Issue #12, opened 2019-02-11
ICS 10: IBC chain naming convention


cwgoes:
Will cover:

- IBC naming system requirements
- Suggested naming convention(s)

cwgoes:
Desiderata:
- Consideration of packet replay prevention across a set of chains (including meta-packets for connection negotiation, channel opening/closing, etc)
- Avoid requiring a central nameservice (although may be utilized for human-readable mappings)
- Avoid requiring new chains needing to look up all previous chain IDs (randomized identifiers)

ebuchman:
See some prior thoughts on "chain descriptions" in [ADR-017](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-017-chain-versions.md).

cwgoes:
I am not sure if this spec needs to be in this repository.

Chain naming did not end up being an important part of the client definitions.

cwgoes:
For now I think we can leave this to [ADR 17](https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-017-chain-versions.md).

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #19, opened 2019-02-26
WIP: Identifier Format

mossid:
Closes: #12


cwgoes:
I think we should develop this specification after the basic protocol semantics, since it will be easier to figure out what properties we need the chain identifier to have in order to ensure safety.

PR #32, opened 2019-03-07
ICS 3: Connection Semantics

cwgoes:
Closes https://github.com/cosmos/ics/issues/3.

Specify datagrams & semantics for bidirectional connections.

Unidirectional connections (one-to-one, or one-to-many, or even many-to-one) are possible but will be specified separately.


spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,143 @@
+---
+ics: 3
```
alexanderbez: I think this oversteps ICS 3 - Connection Semantics?

cwgoes: It's the same spec, we'll merge that PR into this one or vice-versa.

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+(detailed technical specification: syntax, semantics, sub-protocols, algorithms, data structures, etc)
+
+#### Definitions
+
+- Chain `A` is the source blockchain from which the IBC packet is sent
+- Chain `B` is the destination blockchain on which the IBC packet is received
+- `H_h` is the signed header of chain `A` at height `h` 
+- `C_h` is a subset of the consensus ruleset of chain `A` at height `h` 
+- `V_kh` is the value stored on chain `A` under key `k` at height `h` 
+- `P` is the unbonding period of chain `P`, in units of time
```
ebuchman: chain `P`?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+Note that of all these, only `H_h` defines a signature and is thus attributable.
+
+#### Requirements
+
+(reference ICS 2: consensus primitives)
+
+To facilitate an IBC connection, the two blockchains must provide the following proofs:
+
+1. Given a trusted `H_h` and `C_h` and an attributable update message `U_h`,
```
ebuchman: Should this include a statement that `H_h'` is later than `H_h` ?

ebuchman: Also what is an update message and how is it relevant here? 

Same below, what is a change message and how is it relevant?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+(reference ICS 2: consensus primitives)
+
+To facilitate an IBC connection, the two blockchains must provide the following proofs:
+
+1. Given a trusted `H_h` and `C_h` and an attributable update message `U_h`,  
+   it is possible to prove `H_h'` where `C_h' == C_h` and `dt(now, H_h) < P`
+2. Given a trusted `H_h` and `C_h` and an attributable change message `X_h`,  
+   it is possible to prove `H_h'` where `C_h' /= C_h` and `dt(now, H_h) < P`
+3. Given a trusted `H_h` and a Merkle proof `M_kvh` it is possible to prove `V_kh`
```
ebuchman: `Merkle proof` probably needs to be defined prior to use here .

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+It is possible to make use of the structure of BFT consensus to construct extremely lightweight and provable messages `U_h'` and `X_h'`. The implementation of these requirements with Tendermint consensus is defined in [Appendix E](appendices.md#appendix-e-tendermint-header-proofs). Another algorithm able to provide equally strong guarantees (such as Casper) is also compatible with IBC but must define its own set of update and change messages.
```
ebuchman: Paragraph probably doesn't belong here.

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+The Merkle proof `M_kvh` is a well-defined concept in the blockchain space, and provides a compact proof that the key value pair `(k, v)` is consistent with a Merkle root stored in `H_h`. Handling the case where `k` is not in the store requires a separate proof of non-existence, which is not supported by all Merkle stores. Thus, we define the proof only as a proof of existence. There is no valid proof for missing keys, and we design the algorithm to work without it.
```
ebuchman: Seems important. Maybe we need an independent ICS to define the properties of our accumulators? Then we can refer back to that everywhere

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+`valid(H_h, M_kvh) ⇒ true | false`
+
+#### Lifecycle
+
+##### Opening a connection
+
+All proofs require an initial `H_h` and `C_h` for some `h`, where `dt(now, H_h) < P`.
+
+Establishing a bidirectional initial root-of-trust between the two blockchains (`A` to `B` and `B` to `A`) — `H_ah` and `C_ah` stored on chain `B`, and `H_bh` and `C_bh` stored on chain `A` — is necessary before any IBC packets can be sent.
```
ebuchman: We should probably define this notation, ie. `H_ah` means `H_h` for chain `A` ...

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+Provided input:
+- Commitment to root-of-trust for B
+
+On A:
+- Reserve connection slot
+- Prove reserved connection slot (is this necessary?)
+- Prove commitment to root-of-trust for B
+
+Data:
```
ebuchman: What is this ? Data that is part of the relayed message? 

Come to think of it, I wonder if we need a separate ICS for the relay process itself?!

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+- Root-of-trust for A
+- Connection slot identifier on A
+- Desired connection slot identifier on B
```
ebuchman: Why is this necessary? And if its part of the relay data shouldnt it be committed on ChainA already ?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+- Proof of reserved connection slot identifier on A
+
+On B:
+- Check slot identifier
+- Ensure no existing root-of-trust for A
+- Store root-of-trust for A ~= connection identifier
+
+Timeout (committing on B) = stored RoT deleted, connection slot unreserved.
```
ebuchman: Probably need to define root-of-trust somewhere

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+On A:
+- Store root-of-trust for B ~= connection identifier
+- Connection now "open" on A
+
+Timeout (committing on A)
+- On B: Slot unreserved, root-of-trust for A deleted
+- On A: Connection timeout frees slot *OR* second datagram to prove unreserved slot, termined handshake can then free slot on A (eventually)
+
+Forgery: Any B' could commit ACK packet on A (unless RoT-hash is included; alternative version). Only guarantee: non-duplication.
```
ebuchman: How? Didn't we already commit to B's root-of-trust?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,245 @@
...
+- On A: Pending connection deleted.
+- On B: Connection timeout frees slot *OR* second datagram to prove deleted pending connection, slot on B then freed (eventually)
+
+Forgery: Only A could commit `OPENCONFIRM` since root-of-trust for A is now stored on B.
+
+*TWO WAY HANDSHAKE OUTLINE*
+
+The same thing, except `OPENTRY` provides a proof that B's root-of-trust has been stored, and the connection on B is opened after `OPENACK` is sent from B to A.
+
+##### Following headers
```
ebuchman: Looks like this is the lite client logic that belongs in ICS-2 or ICS-7. Tagging https://github.com/cosmos/ics/issues/13

spec/ics-3-connection-semantics/README.md
ebuchman: Oh good https://github.com/cosmos/ics/issues/35

Maybe we want this to come a bit earlier in the stack since it's kind of like the base layer. Before you can even establish a connection, you need a medium for communication, which the relay provides

cwgoes:
Further whiteboarding on ICS 3:

overview

connopen

spec/ics-3-connection-semantics/README.md
cwgoes: Discussion of unbonding periods and other consensus requirements has been moved to ICS 2.

spec/ics-3-connection-semantics/README.md
cwgoes: Will be defined in ICS 23, and now just a generic `AccumulatorProof`.

spec/ics-3-connection-semantics/README.md
cwgoes: This was a leftover from the old spec, deprecated in favor of definitions in ICS 2.

spec/ics-3-connection-semantics/README.md
cwgoes: All related discussion and more specific requirements moved to ICS 2.

spec/ics-3-connection-semantics/README.md
cwgoes: Yes. Will be defined in ICS 23.

spec/ics-3-connection-semantics/README.md
cwgoes: The handshake protocol has now been thoroughly defined, including datagram types.

There are a lot of additional properties it requires that the old spec never enumerated (primarily MitM-prevention) - these are now discussed.

spec/ics-3-connection-semantics/README.md
cwgoes: I think it's important to understand that some relay process will be responsible for actually committing transactions on the two chains, but less important to specify exactly how that process works, as IBC shouldn't rely on it having any particular functionality (other than some degree of liveness in order for the protocol to be live).

spec/ics-3-connection-semantics/README.md
cwgoes: Now defined in ICS 2.

spec/ics-3-connection-semantics/README.md
cwgoes: The handshake protocol has been improved and this is no longer possible.

spec/ics-3-connection-semantics/README.md
cwgoes: The definition belongs in ICS 2 and the implementation in ICS 7, but I think we can include the connection state update logic (which is pretty trivial) here, so we don't have to define state update logic for the connection in more than one ICS.

spec/ics-3-connection-semantics/README.md
cwgoes: The two chains need not use the same identifier (this is *not* necessarily an identifier like "Cosmos Hub", it's just a key for the connection, it could be an unsigned integer for example). The handshake protocol allows the two chains to verify the identifier the other is using, in case they do care about what it is or want a common reference.

cwgoes:
I want to define some timeout semantics here, but those are relatively trivial. Otherwise I think the subprotocols and spec are in a solid enough state that review would be useful - so marking R4R.

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+Accumulator-related primitives are as defined in [ICS 23: Cryptographic Accumulator](../spec/ics-23-cryptographic-accumulator).
+
+Versioning-related primitives are as defined in [ICS 6: Connection & Channel Versioning](../spec/ics-6-connection-channel-versioning).
+
+`Identifier` is an opaque value used as the key for a connection object; it must serialize to a bytestring. The identifier is not necessarily intended to be a human-readable name (and likely should not be, to discourage squatting or racing for identifiers). The opening handshake protocol allows each chain to verify the identifier used to reference the connection on the other chain, so the chains could choose to come to agreemnt on a common identifier. Further discussion is deferred to [ICS 10: Chain Naming Convention](../spec/ics-10-chain-naming-convention).
+
+#### Requirements
+
+Connection handlers and subsequent protocols make use of a simple key-value store interface provided by the underlying state machine. This store must provide three functions, which behave in the way you would expect:
```
cwgoes: I don't think this is worth a separate ICS yet, but I could be convinced otherwise.

spec/ics-3-connection-semantics/closing_equivocation.tex
```diff
@@ -0,0 +1,29 @@
...
+\Field{Identifier}{identifier}
+\Field{Header}{firstHeader}
+\Field{Header}{secondHeader}
+\EndStruct
+\end{algorithmic}
+
+\begin{algorithmic}
+\Function{handleDatagram}{datagram}
+  \Switch{$datagram$}
+    \Case{$CONNCLOSEEQUIVOCATION$}
```
Liamsi: Nit (and not important yet): maybe using SNAKE_CASE or CamelCase can help make this a bit more readable? e.g. `$CONN_CLOSE_EQUIVOCATION`

spec/ics-3-connection-semantics/opening_handshake.tex
```diff
@@ -0,0 +1,82 @@
...
+\Field{AccumulatorProof}{proofAck}
+\EndStruct
+\end{algorithmic}
+
+\begin{algorithmic}
+\Function{handleDatagram}{datagram}
+  \Switch{$datagram$}
+    \Case{$CONNOPENINIT$}
+      \State $state \gets INIT$
+      \State \Call{Set}{identifier, (state, desiredVersion, desiredCounterpartyIdentifier, rootOfTrust)}
```
Liamsi: Does this need to check there is no open, init, or any (connection) state known?

spec/ics-3-connection-semantics/header_tracking.tex
```diff
@@ -0,0 +1,33 @@
...
+
+\begin{algorithmic}
+\Struct[Datagram]{CONNTRACK}
+\Field{Identifier}{identifier}
+\Field{Header}{header}
+\EndStruct
+\end{algorithmic}
+
+\begin{algorithmic}
+\Function{handleDatagram}{datagram}
```
Liamsi: Nit: We probably want to make this consistent across ICSs (compare: #37, e.g., https://github.com/cosmos/ics/pull/37/files#diff-c443c461a2596896f59401c4087cc983 function vs procedure)

cwgoes: Yes, let's standardize on "Function" - updated ICS 18.

spec/ics-3-connection-semantics/closing_equivocation.tex
cwgoes: Switched to CamelCase.

spec/ics-3-connection-semantics/opening_handshake.tex
cwgoes: Appropriately permissioned to who called `ConnOpenInit`, but that seems like a sensible default.

Updated to assert that no current connection exists.

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Specification
+
+### Motivation
+
+The core IBC protocol provides *authorization* and *ordering* semantics for packets: guarantees, respectively, that packets have been committed on the sending blockchain (and according state transitions executed, such as escrowing tokens), and that they have been committed exactly once in a particular order and can be delivered exactly once in that same order. The *connection* abstraction, specified in this standard, defines the *authorization* semantics of IBC (ordering semantics are left to [ICS 4: Channel Semantics](../spec/ics-4-channel-sematics)).
+
+### Desired Properties
+
+- Implementing blockchains should be able to safely allow untrusted users to open and update connections.
```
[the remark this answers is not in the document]
cwgoes: Generally, it's referring to a user who is paying for computation / storage (via gas or something) but is otherwise untrusted. Example users:
- End-users (humans) with keys
- Smart contracts
- Other modules (e.g. in the SDK)

Thinking about where best to explain this...

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,145 @@
...
+![Datatypes](datatypes.png)
+
+Consensus-related primitives are as defined in [ICS 2: Consensus Requirements](../spec/ics-2-consensus-requirements).
+
+Accumulator-related primitives are as defined in [ICS 23: Cryptographic Accumulator](../spec/ics-23-cryptographic-accumulator).
+
+Versioning-related primitives are as defined in [ICS 6: Connection & Channel Versioning](../spec/ics-6-connection-channel-versioning).
+
+`Identifier` is an opaque value used as the key for a connection object; it must serialize to a bytestring. The identifier is not necessarily intended to be a human-readable name (and likely should not be, to discourage squatting or racing for identifiers). The opening handshake protocol allows each chain to verify the identifier used to reference the connection on the other chain, so the chains could choose to come to agreemnt on a common identifier. Further discussion is deferred to [ICS 10: Chain Naming Convention](../spec/ics-10-chain-naming-convention).
+
```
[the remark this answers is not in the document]
cwgoes: Clarified which type/function is defined in which ICS.

spec/ics-3-connection-semantics/README.md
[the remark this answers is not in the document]
cwgoes: I've added essentially the above under "Definitions". That seems OK for now, but maybe this should transition to a separate "common definitions" location in the future.

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,149 @@
...
+`AccumulatorProof` and `verify` are as defined in [ICS 23: Cryptographic Accumulator](../spec/ics-23-cryptographic-accumulator).
+
+`Version` and `checkVersion` are as defined in [ICS 6: Connection & Channel Versioning](../spec/ics-6-connection-channel-versioning).
+
+`Identifier` is an opaque value used as the key for a connection object; it must serialize to a bytestring. The identifier is not necessarily intended to be a human-readable name (and likely should not be, to discourage squatting or racing for identifiers). The opening handshake protocol allows each chain to verify the identifier used to reference the connection on the other chain, so the chains could choose to come to agreement on a common identifier (via `chooseIdentifier` and `checkIdentifier`). Further discussion is deferred to [ICS 10: Chain Naming Convention](../spec/ics-10-chain-naming-convention).
+
+A *user*, as referred to in this specification, is an entity capable of executing datagrams who is paying for computation / storage (via gas or a similar mechanism) but is otherwise untrusted. Possible users include:
```
mossid: ditto

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+This ICS defines four subprotocols: opening handshake, header tracking, closing handshake, and closing by equivocation.
+
+##### Opening Handshake
+
+The opening handshake subprotocol serves to initialize roots of trust for two chains on each other and negotiate an agreeable connection version.
+
+This subprotocol need not be permissioned, modulo anti-spam measures.
+
+![Opening Handshake](opening_handshake.png)
```
mossid: Can we have a simple description of the fields of each datagram, either in or out of the latex block?

mossid: Also might be better if we describe the direction
* `ConnOpenInit`: User -> chain A, 
* `ConnOpenTry`: chain A -> chain B
* `ConnOpenAck`: chain B -> chain A
* `ConnOpenConfirm` chain A -> chain B

cwgoes:
> "Entities", "procedures", "actors" might be some alternative phrasings

I think "actor" is at least more obviously a defined term than "user", I switched to that.

Does it read more clearly?

spec/ics-3-connection-semantics/README.md
cwgoes: https://github.com/cosmos/ics/pull/32#issuecomment-478949434

spec/ics-3-connection-semantics/README.md
cwgoes: I've added a table describing the execution flow, great idea.

Also adding field descriptions.

cwgoes: Added field descriptions in the LaTeX. Are they clear?

cwgoes:
Things that could be further done:
- Turn protocol flow table into LaTeX diagram
- Add `pendingDatagrams` function fragment

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,158 @@
...
+This subprotocol need not be permissioned, modulo anti-spam measures.
+
+![Header Tracking](header_tracking.png)
+
+##### Closing Handshake
+
+The closing handshake protocol serves to cleanly close a connection on two chains.
+
+This subprotocol will likely need to be permissioned to an entity who "owns" the connection on the initiating chain, such as a particular end user, smart contract, or governance mechanism.
+
```
ancazamfir: Similar question as above, can closing be triggered in the same time?
How are header tracking messages handled after the closing? (I see an assert state==open in the handle function but hopefully it doesn't mean a crash :)
In general it maybe it could help if we add a state switch statement on top of handle functions. And mention valid messages in each state....or a state diagram :)

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,158 @@
...
+### Technical Specification
+
+#### Definitions
+
+This ICS defines the `Connection` type:
+
+![Datatypes](datatypes.png)
+
+`RootOfTrust`, `Header, and `updateRootOfTrust` are as defined in [ICS 2: Consensus Requirements](../spec/ics-2-consensus-requirements).
```
ancazamfir:
```suggestion
`RootOfTrust`, `Header`, and `updateRootOfTrust` are as defined in [ICS 2: Consensus Requirements](../spec/ics-2-consensus-requirements).
```
(missing `)
Also, the Header and updateRootOfTrust are not known to the reader at this point in the doc. Header is part of ConnTrack struct and updateRootOfTrust() is hidden further down. Maybe add the other data structures here or move the reference where they are introduced.

nit: is the link broken?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+A correct protocol execution flows as follows:
+
+| Initiator | Datagram          | Chain |
+| --------- | ----------------- | ----- |
+| Actor     | `ConnOpenInit`    | A     |
+| Relayer   | `ConnOpenTry`     | B     |
+| Relayer   | `ConnOpenAck`     | A     |
+| Relayer   | `ConnOpenConfirm` | B     |
+
```
ancazamfir: The table helps a lot! A state machine diagram would also be nice :) A few questions:
What happens if both sides start the connection in the same time?
What happens if version negotiation fails?
Are or can any of these messages be used after the 4-way handshake is done? How are they handled?
Can the connection be made in a single direction?

cwgoes:
> it is all very interesting and I can't wait to go through them all!

:smile: 

> I find it very hard to navigate between the different file types and I wish that all was in a single doc.

In the diff review mode?

> all referenced links seem to be broken and I have a hard time finding the referenced docs when looking across pulls and issue in the repo.

Yes, the links are based on expected future folders. Once the other ICSs are merged as drafts the links will work.

ancazamfir:
>>I find it very hard to navigate between the different file types and I wish that all was in a single doc.

>In the diff review mode?

yes
>Once the other ICSs are merged as drafts the links will work.

ok

cwgoes:
> > > I find it very hard to navigate between the different file types and I wish that all was in a single doc.
> 
> > In the diff review mode?
> 
> yes

One option would be to write the whole specification in LaTeX (could still be rendered on GitHub).

spec/ics-3-connection-semantics/README.md
cwgoes: Can you give an example of the kind of state machine diagram you mean? That sounds useful.

The `asserts` in the protocol description guard against datagram mis-ordering.

I still need to write out timeout logic.

I think we will also have unidirectional connections but I was saving that spec for later since it should be easy to construct once we're agreed on the bidirectional connection semantics.

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,158 @@
...
+Implementing chains may want to allow applications to register handlers to take action upon discovery of an equivocation. Further discussion is deferred to [ICS 12: Byzantine Recovery Strategies](../ics-12-byzantine-recovery-strategies).
+
+### Backwards Compatibility
+
+Not applicable.
+
+### Forwards Compatibility
+
+Once a connection has been established and a version negotiated, future version updates can be negotiated per [ICS 6: Connection & Channel Versioning](../spec/ics-6-connection-channel-versioning). The root of trust can only be updated as allowed by the `updateRootOfTrust` function defined by the consensus protocol chosen when the connection is established.
```
jaekwon: With a open structure (e.g. protobuf structures are open in the sense that new fields can be defined later), additional functionality can be added at a later time, so it may not be necessary to define how version upgrading works initially.

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+The closing handshake protocol serves to cleanly close a connection on two chains.
+
+This subprotocol will likely need to be permissioned to an entity who "owns" the connection on the initiating chain, such as a particular end user, smart contract, or governance mechanism.
+
+![Closing Handshake](closing_handshake.png)
+
+##### Closing by Equivocation
+
+The equivocation closing subprotocol serves to immediately close a connection if a consensus equivocation is discovered and thus prevent further packet transmission.
```
jaekwon: Requires unbonding period in state.

spec/ics-3-connection-semantics/README.md
ancazamfir: > Can you give an example of the kind of state machine diagram you mean? That sounds useful.

Something like that...not sure I got it right though.

image

Can be ASCII art :)
Nit: OPENTRY and TRYOPEN are both used in the spec

ancazamfir: Same for the other (sub)protocols. I can make them if you tell me what to use. And then you can correct them.
BTW, one other thing the spec should probably mention are timeouts, i.e. how long we wait for a datagram and also specify the action taken, e.g. retrying, closing etc.

cwgoes:
To-do: reformat per https://gist.github.com/jaekwon/7d03062b555102a2c41d3d799eeafd65

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,346 @@
...
+- Only the appropriate handshake datagrams can be executed in order.
+- No chain can masquerade as one of the handshaking chains (formalize...)
+
+#### Post-Establishment
+
+Once a negotiation handshake has completed:
+
+- The created connection objects on both chains contain the roots of trust specified by the initiating actor.
+- No other connection objects can be maliciously created on other chains by replaying datagrams.
+- The connection should be able to be voluntarily & cleanly closed by both blockchains.
```
melekes: both - either of two, right?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,346 @@
...
+```
+
+`RootOfTrust`, `Header, and `updateRootOfTrust` are as defined in [ICS 2: Consensus Requirements](../spec/ics-2-consensus-requirements).
+
+`AccumulatorProof` and `verify` are as defined in [ICS 23: Cryptographic Accumulator](../spec/ics-23-cryptographic-accumulator).
+
+`Version` and `checkVersion` are as defined in [ICS 6: Connection & Channel Versioning](../spec/ics-6-connection-channel-versioning).
+
+`Identifier` is an opaque value used as the key for a connection object; it must serialize to a bytestring. The identifier is not necessarily intended to be a human-readable name (and likely should not be, to discourage squatting or racing for identifiers). The opening handshake protocol allows each chain to verify the identifier used to reference the connection on the other chain, so the chains could choose to come to agreement on a common identifier (via `chooseIdentifier` and `checkIdentifier`). Further discussion is deferred to [ICS 10: Chain Naming Convention](../spec/ics-10-chain-naming-convention).
+
```
ancazamfir: Can't find this spec.

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,346 @@
...
+```golang
+type ConnOpenConfirm struct {
+  // Identifier for connection on chain B
+  Identifier        identifier
+  // Proof of stored OPEN state on chain A
+  AccumulatorProof  proofAck
+}
+```
+
```
ancazamfir: Read through all messages a few times an it is not clear to me: a connection is between A and B. Shouldn't the connection identifier be the combination of identifiers on A and B? Also, why is one side proposing an identifier for the other side? Are we trying to solve cases where two chains have the same name?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,346 @@
...
+```coffeescript
+function handleConnOpenInit(identifier, desiredVersion, desiredCounterpartyIdentifier, rootOfTrust)
+  assert(Get(identifier) == null)
+  state <- INIT
+  Set(identifier, (state, desiredVersion, desiredCounterpartyIdentifier, rootOfTrust))
+```
```
ancazamfir: Is this handle function executed by some entity on chain B? To process a ConnOpenInit message from A? 
The "Set(identifier,..." line is hard to understand. What does it mean exactly? We (re)set "identifier" to be the tuple "(state, desiredVersion, desiredCounterpartyIdentifier, rootOfTrust)"? The ConnOpenInit message includes a field "identifier" which I assume is passed in this handle. Do we expect it to be null (assert line)? Then why include it in the message?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,346 @@
...
+
+```coffeescript
+function handleConnOpenTry()
```
ancazamfir: Why aren't there any parameters here and all other handles below?
Could you move maybe the handle immediately after the struct/ message definition?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,346 @@
...
+  expectedRootOfTrust <- getRootOfTrust()
+  assert(verify(rootOfTrust, proofInit,
+    (counterpartyIdentifier, (INIT, desiredVersion, desiredIdentifier, expectedRootOfTrust))))
+  identifier <- chooseIdentifier(desiredIdentifier, counterpartyIdentifier)
```
ancazamfir: What does the chooseIdentifier() do? Selects one of the two? Is this the same "identifier" that is Set further down? 
To be honest I am a bit lost at this point as these two IDs seem to be for two different chains. Maybe an example could help. Also I am not sure how ID collisions are prevented.

cwgoes:
Clarify that multiple headers can be included in the root of trust.

spec/ics-3-connection-semantics/README.md
cwgoes: Yes; clarified

spec/ics-3-connection-semantics/README.md
cwgoes: Unfortunately it does not exist yet. I will clarify the identifier purpose a bit

spec/ics-3-connection-semantics/README.md
cwgoes: Added parameters, thanks.

cwgoes:
- Simplify / clarify identifier usage
- Add timeout logic
- Add state diagram
- Verify subsets of state only

cwgoes:
Still working on pretty diagrams but this can see some more review.

spec/ics-3-connection-semantics/README.md
cwgoes: I have simplified the identifier usage here, now we just assume they are meaningless and prevent collisions.

spec/ics-3-connection-semantics/README.md
cwgoes: Those are now defined in a separate ICS (referenced). Is it clearer?

spec/ics-3-connection-semantics/README.md
cwgoes: We have only local identifiers, not global ones. I think we want chains to know each other's identifier for themselves, but I don't think we need to require that they have the same identifier (although that could be recommended by convention).

spec/ics-3-bidirectional-connection-semantics/README.md
```diff
@@ -0,0 +1,358 @@
...
+  Header firstHeader
+  Header secondHeader
+}
+```
+
+```coffeescript
+function handleConnCloseEquivocation(identifier, firstHeader, secondHeader)
+  (state, version, counterpartyIdentifier, rootOfTrust) = Get(identifier)
+  assert(state == OPEN)
+  assert(checkEquivocation(rootOfTrust, firstHeader, secondHeader))
```
zmanian: Shouldn't this take the current trust state not the root of trust?

spec/ics-3-bidirectional-connection-semantics/README.md
```diff
@@ -0,0 +1,358 @@
...
+  (state, version, counterpartyIdentifier, rootOfTrust) = Get(identifier)
+  assert(state == TRYCLOSE)
+  assert(verify(rootOfTrust, proofTry, (counterpartyIdentifier, CLOSED)))
+  state = CLOSED
+  Set(identifier, (state, version, counterpartyIdentifier, rootOfTrust))
+```
+
+##### Closing by Equivocation
+
+The equivocation closing subprotocol serves to immediately close a connection if a consensus equivocation is discovered and thus prevent further packet transmission.
```
zmanian: Equivocation should be defined as two valid block headers with different block ids at the same height.

spec/ics-3-bidirectional-connection-semantics/README.md
```diff
@@ -0,0 +1,358 @@
...
+  Identifier        counterpartyLightClientIdentifier
+  // Light client identifier for chain A on B
+  Identifier        lightClientIdentifier
+  // Proof of stored INIT state on chain A
+  AccumulatorProof  proofInit
+}
+```
+
+```coffeescript
+function handleConnOpenTry(desiredIdentifier, counterpartyIdentifier, desiredVersion, counterpartyLightClientIdentifier, lightClientIdentifier, proofInit)
```
zmanian: Should all of these functions provide a specify a timeout handler if the next step of protocol isn't executed within N blocks?

spec/ics-3-bidirectional-connection-semantics/README.md
cwgoes: It is that, I guess the nomenclature is confusing. I think we will change it to `ConsensusState`.

edit: renamed

spec/ics-3-bidirectional-connection-semantics/README.md
cwgoes: I think we actually want to move this to ICS 2, which now handles client logic.

cc @mossid

cwgoes: https://github.com/cosmos/ics/pull/20#issuecomment-490851160

spec/ics-3-bidirectional-connection-semantics/README.md
cwgoes: Yes, it should, and for safety I think the datagrams need to contain the block number as well. 

I want to work on channel definitions quickly first but then I'll add this.

cwgoes: See https://github.com/cosmos/ics/issues/86

cwgoes:
Please reformat re: https://github.com/cosmos/ics/issues/84

cwgoes:
To-do:
- [ ] Use updated identifier formats per ICS 24
- [ ] Verify subsets of state only
- [ ] Add timeout logic per https://github.com/cosmos/ics/issues/86
- [ ] Add diagram per https://github.com/cosmos/ics/issues/88

cwgoes:
Going to make the diagram fancier and refer to two chains, but the substance is here.

spec/ics-3-bidirectional-connection-semantics/README.md
cwgoes: Timeout handler specified, also see the state machine diagram.

spec/ics-3-bidirectional-connection-semantics/state.tex
```diff
@@ -0,0 +1,34 @@
...
+  \draw[->, red](open) edge node {\tiny{\textsc{ConnOpenTimeout}}} (empty);
+  \draw[->, red, bend right](tryclose) edge node {\tiny{\textsc{ConnCloseTimeout}}} (open);
+  \draw[->, red](closed) edge node {\tiny{\textsc{ConnCloseTimeout}}} (open);
+
+	\draw[color=black,thick](-0.5,-14) rectangle (15.2,2.8);
+
+	\node at (-0.5,1.5) [above=0.5mm, right=5mm] {\textsc{Connection State Machine}};
+
+\end{tikzpicture}
+\end{document}
```
Liamsi: referencing the current resulting image: https://media.githubusercontent.com/media/cosmos/ics/1b5746682891904a4b982723f5cfff36c1501c30/spec/ics-3-bidirectional-connection-semantics/state.png

cwgoes:
- Also needs timeout recovery on the timing-out chain

cwgoes:
Problem: what to do if we start initializing a channel, then final connection handshake times out?

Maybe we can just delete the channel, the channel handshake will not complete.

Concerns about name collisions?

cwgoes:
- [ ] Reread for consistency
- [ ] Simplify object notation

cwgoes:
- [ ] Ensure correct description of module interface

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,395 @@
...
+  nextTimeoutHeight             uint64
+}
+```
+
+### Subprotocols
+
+This ICS defines three subprotocols: opening handshake, header tracking, closing handshake. Datagrams defined herein are handled as external messages by the IBC relayer module defined in [ICS 26](../spec/ics-26-relayer-module).
+
+![State Machine Diagram](state.png)
+
```
ancazamfir: Nice diagram :) Could you make the text on transitions a bit bigger? It doesn't read easy currently.

ancazamfir: Also, IMHO it would be clearer if you'd separate A and B FSMs.

cwgoes: Thanks for the feedback, makes sense - will do.

cwgoes: Altered the font sizing. I'm not sure separating the FSMs for A and B is the right direction, though - the goal of the protocol construction is to allow reasoning about the combined state of the IBC handlers on both chains at once, so it seems like we want to combine them...

ancazamfir: ...I agree we want to look at chains' FSMs at once. But the current diagram seem to show some transitions upon events that are not valid one one side. Let me try to read it more carefully the way it is together with the code.

ancazamfir: I think we should list the events in the FSM (CONNINIT, CONNOPENTRY, etc) and describe what triggers them. Maybe add also a column in the table below?

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,401 @@
...
+function connOpenTry(desiredIdentifier, counterpartyIdentifier, counterpartyClientIdentifier, clientIdentifier, proofInit, timeoutHeight, nextTimeoutHeight)
+  assert(getConsensusState().getHeight() <= timeoutHeight)
+  consensusState = get("clients/{clientIdentifier}")
+  expectedConsensusState = getConsensusState()
+  expected = Connection{INIT, desiredIdentifier, counterpartyClientIdentifier, clientIdentifier, timeoutHeight}
+  assert(verifyMembership(consensusState.getRoot(), proofInit, "connections/{counterpartyIdentifier}", expected))
+  assert(verifyMembership(consensusState.getRoot(), proofInit, "clients/{counterpartyClientIdentifier}", expectedConsensusState))
+  assert(get("connections/{desiredIdentifier}") == nil)
+  identifier = desiredIdentifier
+  state = OPENTRY
```
ancazamfir:
```suggestion
  state = TRYOPEN
```
Also in other places

spec/ics-3-connection-semantics/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+*ConnOpenInit* initializes a connection attempt on chain A.
+
+```golang
+type ConnOpenInit struct {
+  identifier                    Identifier
+  desiredCounterpartyIdentifier Identifier
+  clientIdentifier              Identifier
+  counterpartyClientIdentifier  Identifier
+  nextTimeoutHeight             uint64
```
ancazamfir: Maybe add a few words in the definition section about timeout heights.

spec/ics-3-connection-semantics/README.md
cwgoes: Unified, thanks.

spec/ics-3-connection-semantics/README.md
cwgoes: Added two columns in each table, for prior state and posterior state - is that what you meant?

spec/ics-3-connection-semantics/README.md
cwgoes: I wonder if the explanation in #86 should actually become an ICS...

cwgoes:
Merging as a draft, future comments are always welcome.

Issue #15, opened 2019-02-12
ICS 15: Cosmos Signed Messages


cwgoes:
Ported from [the spec in the SDK repository](https://github.com/cosmos/cosmos-sdk/blob/develop/docs/spec/_ics/ics-030-signed-messages.md).

Suggested changes/alterations to that spec:
- Organize in [ICS format](https://github.com/cosmos/ics/pull/17)
- More clearly describe requirements of hashing algorithm & signing algorithm (and maybe parameterize the specification over them)
- Investigate distinction between custom messages (more easily defined in the SDK than e.g. Ethereum) and signed messages, and what use we expect for each (block explorers? DEXs? validator authentication mechanisms?)
- Include specification for signing structured data using [JSON schema](https://json-schema.org/)
- Include options for domain specification / replay prevention (nonce, block height, timestamp)

alexanderbez:
Thanks @cwgoes, I'll take care of this shortly.

jackzampolin:
Hell yeah @alexanderbez!

alexanderbez:
@cwgoes  can you elaborate more on the following or possibly reword:

> Investigate distinction between custom messages (more easily defined in the SDK than e.g. Ethereum) and signed messages, and what use we expect for each (block explorers? DEXs? validator authentication mechanisms?)

alexanderbez:
Also, do we want to support messages that are both ASCII text and structured human-readable data?

cwgoes:
> Also, do we want to support messages that are both ASCII text and structured human-readable data? Can we leave this up to implementation?

Can we have a structured data standard which encapsulated ASCII-text (as a datatype option)?

> @cwgoes can you please elaborate more on the following or possibly reword:
> 
> > Investigate distinction between custom messages (more easily defined in the SDK than e.g. Ethereum) and signed messages, and what use we expect for each (block explorers? DEXs? validator authentication mechanisms?)

Let's think about - and write down - which use cases signed messages should be used for (maybe validator / delegator authentication, communication within wallets) and which use cases we expect Cosmos blockchains to specify custom transactions for (most actual state transitions). We can change this later, but it's useful to brainstorm.

cwgoes:
Also [this](https://github.com/maciejhirsz/uos) is interesting.

cwgoes:
I now think it may be prudent to instead put this specification in the Cosmos SDK repository.

Ref https://github.com/cosmos/ics/issues/185

cwgoes:
[...]

PR #33, opened 2019-03-07
ICS 15: Cosmos Signed Messages

alexanderbez:
A (modified) port of the Cosmos Signed Message spec from the SDK. I initialized this spec as `draft` as it already had some form of approval in the SDK.

closes: #15


cwgoes:
Do you mind using ICS 15? Joon & I assigned numbers to all of the issues - https://github.com/cosmos/ics/issues - so that PRs won't conflict.

cwgoes:
Can we also note which functions we require from the signature algorithm (`sign` and `verify`), explain how they are used in the procedures, reference the JSON schema validation function (presumably there are libraries which do this)? Also ref https://github.com/cosmos/ics/issues/15#issuecomment-471541533 - ideally we should be able to encapsulate ASCII text in the JSON schema `@type` field.

cwgoes:
For now, I think we can choose how to construct the "domain separator" based on accounts on Cosmos SDK-esque chains signing messages - so it should include account number / address, some kind of nonce, timestamp at minimum.

cwgoes:
Please reformat re: https://github.com/cosmos/ics/issues/84 (in addition to prior comments)

cwgoes:
Per https://github.com/cosmos/ics/issues/185#issuecomment-522240197, and the observation that this is the only non-IBC specification being considered for inclusion, I wonder if we should put this in the SDK repository (which I believe has other specifications for SDK modules) instead.

cwgoes:
Per the above I'm going to close this. Feel free to reopen if you have objections.

I would be glad to review if this turns into an SDK-side PR.

alexanderbez:
@cwgoes does this belong in ICS or the SDK? How does arbitrary signing of messages relate to IBC?

cwgoes:
> @cwgoes does this belong in ICS or the SDK? How does arbitrary signing of messages relate to IBC?

It does not, and everything else in this repo is just for IBC, so I'm thinking this might be better off in the SDK.

spec/ics-15-cosmos-signed-messages/README.md
```diff
@@ -0,0 +1,147 @@
...
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$id": "cosmos/signing/typeData/schema",
+  "title": "The Cosmos signed message typed data schema.",
+  "type": "object",
+  "properties": {
+    "@chain_id": {
+      "type": "string",
+      "description": "The corresponding Cosmos chain identifier.",
+      "minLength": 1
```
jleni: Is it possible to specify `maxLength` for all (most?) fields?

jleni:
- It would be good to restrict fields to ASCII or even 32..127 when possible to ensure that they can be represented in HW wallets.

- With respect to backwards compatibility, is there any specific requirement? It would be good to make that explicit.

- Canonical JSON: Make some restrictions explicit. Field order for instance (alphabetical or similar). Otherwise, identical content where field order changes results in different hashes / signatures. Similarly, spaces, separators, serialized objects (sets, etc.) that can result in multiple valid representations should be restricted to a unique representation.

- Field names: As a general recommendation, it would be good if field names, in particular in embedded objects:
1) could be user friendly so final users can understand them or...
2) there is some additional field with a short description that could be used for representational purposes

All in all, while this may require some changes in the HW wallet app but I don't see any significant impact that is not solvable. Actually, some improvements would result in a friendlier user interface.

PR #37, opened 2019-03-07
ICS 18: Relayer algorithms

### LINKED ISSUE #35: ICS 18: Off-chain relayer algorithm
Partially a port of the [original spec](https://github.com/cosmos/cosmos-sdk/blob/develop/docs/spec/ibc/channels-and-packets.md#37-packet-relayer).

Will cover:
- Off-chain process relay algorithm
- Permissioning
- Fee payment mechanisms

cwgoes:
Closes #35.

This is a pretty minimal specification for now, since the details of `pendingDatagrams` will be defined in individual ICSs and I think it's too early to spend much time working out relayer fee reimbursement.


cwgoes:
Ready for review as a draft. Specific functions for calculating valid datagrams will need to be defined in individual ICSs as mentioned. Possibly this ICS could include more detail about relayer incentivization later, but I think that design discussion would be too early now (it doesn't seem particularly difficult, and is unlikely to require low-level protocol changes).

spec/ics-18-relayer-algorithms/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+In the IBC protocol, a blockchain can only record the *intention* to send particular data to another chain. Physical datagram relay must be performed by off-chain infrastructure. This standard defines the concept of a *relayer* algorithm, executable by an off-chain process with the ability to query chain state, to perform this relay.
+
+### Desired Properties
+
+- No safety properties of IBC should depend on relayer behavor (assume Byzantine relayers).
+- Liveness properties of IBC should depend only on the existence of at least one correct, live relayer.
+- Relaying should be permissionless, all requisite verification should be performed on-chain.
+- Requisite communication between the IBC user and the relayer should be minimized.
+- Provision for relayer incentivization at the application layer should be considered.
```
Liamsi: This reads more as an instruction/recommendation to chains that implement IBS and will use relayers than a desired property (?). Maybe rephrase to 

```suggestion
- Provision for relayer incentivization should be possible at at the application layer.
```

cwgoes:
> My understanding is that these will be part of separate ICSs.

Yes, that's right. I think that makes more sense because chains are likely to implement subsets of the IBC protocol - that way relayers can compose `pendingDatagrams` based on which ICSs the chain implements.

spec/ics-18-relayer-algorithms/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+`pendingDatagrams` calculates the set of all valid datagrams to be relayed from one chain to another based on the state of both chains. Subcomponents of this function are defined in individual ICSs. The relayer must possess prior knowledge of what subset of the IBC protocol is implemented by the blockchains in the set for which they are relaying (e.g. by reading the source code).
+
+`submitDatagram` is a procedure defined per-chain (submitting a transaction of some sort).
+
+![Relayer](relayer.png)
+
+#### Incentivization
+
+The relay process must have access to accounts on both chains with sufficient balance to pay for transaction fees. Relayers may employ application-level methods to recoup these fees.
```
melekes: So, in order for the relayer (who sent the transaction) to get fees, it will need to give the source chain a proof? how will it look like?

spec/ics-18-relayer-algorithms/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Any number of relayer processes may be safely run in parallel (and indeed, it is expected that separate relayers will serve separate subsets of the interchain). However, they may consume unnecessary fees if they submit the same proof multiple times, so some minimal coordination may be ideal.
```
melekes: > consume unnecessary fees if they submit the same proof multiple times,

What if a relayer sends the same proof 1000 times?

> so some minimal coordination may be ideal.

does this mean that for someone who's initiating IBC transaction, he/she should have enough fees for all existing and future relayers?? Wouldn't it make more sense to just give fees to the first relayer?

spec/ics-18-relayer-algorithms/README.md
cwgoes: Possibly the relayer could include a small payment to itself in the packet data.

I think we will want to specify an exact protocol for this but it is too early to do so (for a later version of this ICS or another ICS).

spec/ics-18-relayer-algorithms/README.md
cwgoes: > What if a relayer sends the same proof 1000 times?

Nothing (it's just like sending a transaction which tries to spend the same funds 1000 times, the first transaction will spend the funds and the remaining 999 will fail).

> does this mean that for someone who's initiating IBC transaction, he/she should have enough fees for all existing and future relayers?? Wouldn't it make more sense to just give fees to the first relayer?

Hmm, no, this has nothing to do with what a user of IBC would do - they would just include a small payment for whichever relayer submits their packet.

zmanian:
Question: Do we need to creation some notion of an `ID` for each datagram.

The relayer algorithm would then be expected to insert any detected datagram if the id wasn't already in the mempool.

cwgoes:
> Question: Do we need to creation some notion of an `ID` for each datagram.
> 
> The relayer algorithm would then be expected to insert any detected datagram if the id wasn't already in the mempool.

This is expected to be included in `pendingDatagrams` (which can read the state of the ledger).

I think usually we will (certainly for packets), but occasionally other state will be sufficient (connection state, for example, in the connection handshake).

spec/ics-18-relayer-algorithms/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Synopsis
+
+Relayer algorithms are the "physical" connection layer of IBC — off-chain processes responsible for relaying data between two chains running the IBC protocol by scanning the state of each chain, constructing appropriate datagrams, and executing them on the opposite chain as allowed by the protocol.
+
+## Specification
+
+### Motivation
+
+In the IBC protocol, a blockchain can only record the *intention* to send particular data to another chain. Physical datagram relay must be performed by off-chain infrastructure. This standard defines the concept of a *relayer* algorithm, executable by an off-chain process with the ability to query chain state, to perform this relay.
```
milosevic: Should we add short explanation why sending of datagrams must happen by off chain infrastructure?

spec/ics-18-relayer-algorithms/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Definitions
+
+A *relayer* is an off-chain process with the ability to read the state of and submit transactions to some set of ledgers utilizing the IBC protocol.
+
+### Desired Properties
+
+- No safety properties of IBC should depend on relayer behavor (assume Byzantine relayers).
```
milosevic: behavor -> behavior

spec/ics-18-relayer-algorithms/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Motivation
+
+In the IBC protocol, a blockchain can only record the *intention* to send particular data to another chain. Physical datagram relay must be performed by off-chain infrastructure. This standard defines the concept of a *relayer* algorithm, executable by an off-chain process with the ability to query chain state, to perform this relay.
+
+### Definitions
+
+A *relayer* is an off-chain process with the ability to read the state of and submit transactions to some set of ledgers utilizing the IBC protocol.
+
+### Desired Properties
```
milosevic: These properties are not necessarily telling us what we expect from relayer. It's more telling us what other part don't expect from relayers.

spec/ics-18-relayer-algorithms/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Technical Specification
+
+#### Relayer Algorithm
+
+The relayer algorithm is defined over a set `C` of chains implementing the IBC protocol.
+
+`pendingDatagrams` calculates the set of all valid datagrams to be relayed from one chain to another based on the state of both chains. Subcomponents of this function are defined in individual ICSs. The relayer must possess prior knowledge of what subset of the IBC protocol is implemented by the blockchains in the set for which they are relaying (e.g. by reading the source code).
+
+`submitDatagram` is a procedure defined per-chain (submitting a transaction of some sort).
```
milosevic: Can we imagine relayers organised in some kind of network so not all relayer processes have access to all chains?If yes, I guess that this is outside the IBC scope?

melekes:
Questions I have so far:

- do relayers programmed to use a Hub?
- if yes, how do they choose what Hub to use?
- do they first send IBC transaction to a Hub? What if a Hub detects a double spend? How it can signal the relayer that IBC transaction is malicious?

spec/ics-18-relayer-algorithms/README.md
cwgoes: Added more explanation.

spec/ics-18-relayer-algorithms/README.md
cwgoes: Thanks, fixed.

spec/ics-18-relayer-algorithms/README.md
cwgoes: Yes - that's what this is intended to convey, since it seems likely. Clarified a bit.

I think the actual physical topology is outside of IBC scope, but we should expect that not all relayers can relay between all chains (perhaps none can).

cwgoes:
@melekes I don't think any of the current IBC specifications talk about "Hubs" yet - that's a distinction at the level of expected network dynamics (for now), not a distinction at the protocol layer.

There's some discussion in progress at https://github.com/cosmos/ics/issues/59.

spec/ics-18-relayer-algorithms/README.md
cwgoes: Hmm, I suppose - these are desired properties of how we construct the relayer algorithm (e.g. permissionless - all datagrams are verified by the blockchain state machine itself). Some are a bit vague but I think still worth listing.

What properties do you think are in scope here?

cwgoes:
Merging to the "draft" stage, further comments always welcome.

PR #40, opened 2019-03-07
ICS 19: Standard queue

### LINKED ISSUE #39: ICS 19: Standard queue
Partially ported from [the original spec](https://github.com/cosmos/cosmos-sdk/blob/develop/docs/spec/ibc/channels-and-packets.md#323-queue).

cwgoes:
Closes #39.

Work in progress.

### DIFF
--- a/spec/ics-19-standard-queue/README.md
+++ b/spec/ics-19-standard-queue/README.md
@@ -1,0 +1,117 @@
+---
+ics: 19
+title: Standard Queue
+stage: proposal
+category: ibc-misc
+author: Christopher Goes <cwgoes@tendermint.com>, Juwoon Yun <joon@tendermint.com>
+created: 2019-03-07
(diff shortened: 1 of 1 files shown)


zmanian:
I think we should specify that the queue must incorporate a cryptographic accumulator against which membership and non-membership proofs can be efficiently checked for a queue.

mossid:
A possible encoding scheme for the indices and `q_head q_tail`:
- `encode :: uint64 | Head | Tail -> bytes`
- `encode(index uint64) -> bigendian64(index)`
- `encode(Head) -> bigendian64(int64_max) ++ 0x00`
- `encode(Tail) -> bigendian64(int64_max) + 0x01`

cwgoes:
ref https://github.com/cosmos/ics/issues/73#issuecomment-491010787

PR #70, opened 2019-04-11
IBC architecture document

cwgoes:
Update per suggestions from @ancazamfir and @milosevic:

- Write motivation section
- Clarify protocol purpose, interfaces, and dataflow
- Take some inspiration from the TCP specification ([here](https://tools.ietf.org/pdf/rfc793.pdf))

Also fix a few links.

Feedback welcome, both on clarity of this writing and any conceptual explanations you think are missing.


docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+> For a broad set of protocol design principles, see [here](./2_IBC_DESIGN_PRINCIPLES.md).
+
+> For definitions of terms used in IBC specifications, see [here](./3_IBC_TERMINOLOGY.md).
+
+This document outlines the architecture of the authentication, transport, and ordering layers of the IBC protocol stack. This document does not describe specific protocol details — those are contained in individual ICSs.
+
+## What is IBC?
+
+The *inter-blockchain communication protocol* is designed for use as a reliable, authenticated module-to-module communication protocol between modules running on independent distributed ledgers across an untrusted network layer.
```
milosevic: Maybe you can say: "The IBC is reliable and secure inter-module communication protocol, where modules run on independent distributed ledgers on top of untrusted network later".

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+## What is IBC not?
```
milosevic: Maybe these use cases can be moved in the previous section to illustrate potential use cases IBC should provide.

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+IBC is an end-to-end, connection-oriented, stateful protocol for reliable, ordered, authenticated communication between modules on separate distributed ledgers. IBC implementations are expected to be co-resident with higher-level modules and protocols on the host ledger. Ledgers hosting IBC must provide a certain set of functions for consensus transcript verification and accumulator proof generation, and IBC packet relayers (off-chain processes) are expected to have access to network protocols and physical datalinks as required to read the state of one ledger and submit data to another.
+
+### Dataflow layers
```
milosevic: I really like the picture, just not really clear how it fits in dataflow layers and what is the purpose of this section.

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+|                          |                           |                          |
+| +----------+     +-----+ |        +---------+        | +-----+     +----------+ |
+| | Module A | <-> | IBC | | <----> | Relayer | <----> | | IBC | <-> | Module B | |
+| +----------+     +-----+ |        +---------+        | +-----+     +----------+ |
++--------------------------+                           +--------------------------+
+```
+
+## Scope
+
+IBC handles authentication, transport, and ordering of structured data packets relayed between modules on separate ledgers. The protocol is intended to be in simultaneous use between any number of modules on any number of ledgers over arbitrarily structured underlying networks.
```
milosevic: The second sentence is not very clear. Does it make sense to initially constrain IBC on inter-module communication between two ledgers?However, we can have several modules on ledger A communicating over IBC with modules on ledger B.

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+IBC is not (only) an atomic-swap protocol: arbitrary cross-chain data transfer and computation is supported.
+
+IBC is not (only) a token transfer protocol: token transfer is a possible application-layer use of the IBC protocol.
+
+IBC is not (only) a sharding protocol: there is no single state machine being split across chains, but rather a diverse set of different state machines on different chains which share some common interfaces.
+
+IBC is not (only) a layer-two scaling protocol: all chains implementing IBC exist on the same "layer", although they may occupy different points in the network topology, and there is no single root chain or single validator set.
+
+## Motivation
```
gamarin2: Good paragraph, but it takes the angle of scalability as primary motivation factor. I agree it is an important motivation factor, but isn't interoperability an other important motivation factor in itself (rather than a simple consequence of solving the scalability problem)? 

Even between two non-scalable ledgers interoperability is a strong desiderata. The ability for two ledgers to exchanges messages in a (somewhat) trustless way is a missing feature of the current blockchian ecosystem, and I think we should also mention it in this section. I would add a paragraph about interoperability before "The *interblockchain communication protocol* takes an orthogonal approach to a differently formulated version of the problem".

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+The *interblockchain communication protocol* takes an orthogonal approach to a differently formulated version of the problem: enabling safe, reliable interoperation of a network of heterogeneous distributed ledgers, arranged in an unknown topology, which can diversify, develop, and rearrange independently of each other or of a particular imposed topology or state machine design. In a wide, dynamic network of interoperating chains, sporadic Byzantine faults are expected, so the protocol must also detect, mitigate, and contain the potential damage of Byzantine faults in accordance with the requirements of the applications & blockchains involved. For a longer list of design principles, see [here](./2_IBC_DESIGN_PRINCIPLES.md).
```
gamarin2: Before this paragraph,I'd suggest to add a paragraph about interoperability as a motivation factor, and change the intro phrase to ", and reformulate the previous sentence to "The *interblockchain communication protocol* takes an orthogonal approach to a differently formulated version of **these** problem"

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Interfaces
+
+IBC sits between modules — smart contracts, state machine components, or otherwise independent pieces of application logic on ledgers — on one side, and underlying consensus protocols, ledgers, and network infrastructure (e.g. TCP/IP), on the other side.
+
+To modules IBC provides a set of functions much like the functions which might be provided to a module for interacting with another module on the same ledger: sending data packets and receiving data packets on an established connection & channel — in addition to calls to manage the protocol state: opening and closing connections and channels, choosing connection, channel, and packet delivery options. Considerable flexibility is provided to ledger developers as to which of these functions to expose to which modules, and how to restrict parameter choices — if at all — the protocol generally assumes the most permissionless setting possible, and implementers can choose to restrict usage according to their application's requirements.
```
milosevic: Can you just list here operations that are provided to modules (opening/closing connections and channels(what is a channel?, and sending and receiving data)? The last sentence is maybe not needed at this point. This can be explained in part where you discuss interface in more details.

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+Of the underlying consensus protocols and ledgers IBC requires a set of primitive functions and properties as defined in [ICS 2](../../spec/ics-2-consensus-requirements), primarily finality, cheaply-verifiable consensus transcripts, and simple key-value store functionality. Of the network infrastructure protocol layer (and physical network layer) IBC requires only eventual data delivery — no authentication, synchrony, or ordering properties are assumed.
```
milosevic: Maybe say "IBC assumes functionalities and properties(?) of ledgers as defined in ...". On the networking side, IBC assumes only...

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Motivation
+
+The two predominant blockchains, Bitcoin and Ethereum, support about seven and about twenty transactions per second respectively. Both have been operating at capacity in recent past despite still being utilized primarily by a userbase of early-adopter enthusiasts, far from the desired mainstream adoption. Throughput limitations are a fundamental limitation of distributed state machines, since every (validating) node in the network must process every transaction and store all state. Faster consensus algorithms, such as [Tendermint](https://github.com/tendermint/tendermint), may increase throughput by a large constant factor but will be unable to scale indefinitely for this reason. In order to support the transaction throughput, application diversity, and cost efficiency required to facilitate wide deployment of distributed ledger applications, execution and storage must be split across many independent consensus instances which can run concurrently.
```
gamarin2: Several comments on this first paragraph:

- two predominant -> two of the most predominant
- support -> currently support
- "far from desired mainstream adoption" -> I would reformulate, because for Bitcoin some argue it's not the case. A lot of people in the Bitcoin community do not see throughput as a fundamental limitation for bitcoin's use case (store of value). I would say that throughput is a limitation for most blockchain use cases, but not necessarily all of them. 
- "since every (validating) node in the network must process every transaction and store all state" -> is this really the limiting factor, or is it that nodes must agree on sets of transactions sequentially? I would say it's the latter

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+IBC is an end-to-end, connection-oriented, stateful protocol for reliable, ordered, authenticated communication between modules on separate distributed ledgers. IBC implementations are expected to be co-resident with higher-level modules and protocols on the host ledger. Ledgers hosting IBC must provide a certain set of functions for consensus transcript verification and accumulator proof generation, and IBC packet relayers (off-chain processes) are expected to have access to network protocols and physical datalinks as required to read the state of one ledger and submit data to another.
+
+### Dataflow layers
+
+```
```
gamarin2: Can we encapsulate Module and IBC in a State machine rectangle?

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
++--------------------------+                           +--------------------------+
+```
+
+## Scope
+
+IBC handles authentication, transport, and ordering of structured data packets relayed between modules on separate ledgers. The protocol is intended to be in simultaneous use between any number of modules on any number of ledgers over arbitrarily structured underlying networks.
+
+## Interfaces
+
+IBC sits between modules — smart contracts, state machine components, or otherwise independent pieces of application logic on ledgers — on one side, and underlying consensus protocols, ledgers, and network infrastructure (e.g. TCP/IP), on the other side.
```
gamarin2: Isn't it part of the state machine though?

docs/ibc/1_IBC_ARCHITECTURE.md
gamarin2: Link to channel and connection definition

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+- Flow control
+- Authentication
+- Connections
+- Multiplexing
+
+The following paragraphs outline the protocol logic within IBC for each area.
+
+### Data relay
+
+IBC assumes the existence of a set of relayer processes with access to an underlying network protocol stack (likely TCP/IP, UDP/IP, or QUIC/IP) and physical interconnect infrastructure. These relayer processes monitor a set of ledgers implementing the IBC protocol, continuously scanning the state of each ledger and executing transactions on another ledger when outgoing datagrams have been committed. For correct operation and progress in a connection between two ledgers, IBC requires only that at least one correct and live relayer process exists which can relay between the ledgers.
```
milosevic: You can maybe just add explanation in the first sentence that in proposed architecture, modules are not directly sending messages to each other over networking infrastructure, but they just create messages to be sent that are then effectively sent with the help of relayer processes. In a second sentence, maybe it's better to explain WHAT is the role of relayer processes, and not HOW they do it. For example, relayer processes are responsible for effectively sending transactions from one module(on sending ledger) to another module(on receiving ledger).   But depending on how relayer processes are introduced, second sentence might not be needed.

gamarin2: Should we say that the access is generally provided by the non state machine part of the distributed ledger?

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Reliability
+
+The network layer and relayer processes may behave in arbitrary ways, dropping, reordering, or duplicating packets, purposely attempting to send invalid transactions, or otherwise acting Byzantine. This must not compromise the safety or liveness of IBC. This is achieved by assigning a sequence number to each packet sent over an IBC connection, which is checked by the IBC handler on the receiving ledger, and providing a method for the sending ledger to check that the receiving ledger has in fact received and handled a packet before sending more packets or taking further action. Cryptographic commitments are used to prevent datagram forgery: the sending ledger commits to outgoing datagrams, and the receiving ledger checks these commitments, so datagrams altered in transit by a relayer will be rejected.
```
gamarin2: Define safety/liveness in the context of IBC

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Authentication
+
+All datagrams in IBC are authenticated: a block finalized by the consensus algorithm of the sending blockchain must commit to the outgoing datagram via a cryptographic accumulator, and the receiving chain must verify both the consensus transcript and the accumulator proof that the datagram was sent (and associated actions executed) before acting upon it. For correct execution semantics, both chains must implement the correct IBC protocol — of course, since IBC is permissionless, any chain could connect to another chain and provide a correct accumulator proof but incorrectly implement IBC such that its IBC handler did not behave according to protocol, so modules utilizing IBC connections (or end-users using those modules) must inspect the state of the connections and ledgers they plan to use to ensure correct implementation.
```
gamarin2: talking about blockchain here, where we talk about "distributed ledger" everywhere else. I think we should stick to distributed ledger, but maybe on the first occurence of the word distributed ledger we should add (sometimes referred to as "blockchain")

milosevic: Blockchain and ledger are being used interchangeably. If these two terms are assuming the same thing, maybe it make sense to state this in vocabulary section.

gamarin2: Define accumulator proof

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Connections
+
+Reliability, flow control, and authentication as described above require that IBC initializes and maintains certain status information for each datastream. This information is called a connection. Each connection object contains information about the consensus state of the connected ledger, negotiated encoding & multiplexing options, and state & sequence numbers. When two modules wish to communicate, they must locate an existing connection between their two ledgers, or initialize a new connection if none yet exists. Initializing a connection requires a multi-step handshake which, once complete, ensures that only the two intended ledgers are connected and future datagrams relayed will be authenticated, encoded, and sequenced as desired.
+
+### Multiplexing
```
gamarin2: Rename "channels"?

docs/ibc/1_IBC_ARCHITECTURE.md
milosevic: This paragraph is not at the same level of abstraction/details as previous part of the document. Can we explain what is meant by "communication is authenticated" without going too much into details?The sentence about IBC handler not behaving according to the protocol is confusing. It seems that there is an implicit assumptions that IBC guarantees are guaranteed to hold only if IBC modules are correct?This might require some clarification. Do we assume the same fault model for IBC module and application module that are collocated?

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+| --> Consensus --> Connection --> Channel --> Packet --> Handler --> Module       |
++----------------------------------------------------------------------------------+
+```
+
+### Steps
+
+1. On chain *A*
+    1. Actor (application-specific)
+    1. Module (application-specific)
+    1. Handler (parts defined in different ICSs)
```
milosevic: What is the handler in this context?

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+| Chain B                                                                          |
+|                                                                                  |
+| --> Consensus --> Connection --> Channel --> Packet --> Handler --> Module       |
++----------------------------------------------------------------------------------+
+```
+
+### Steps
+
+1. On chain *A*
+    1. Actor (application-specific)
```
milosevic: Do we need actor in this scenario? Why not starting from module?

docs/ibc/1_IBC_ARCHITECTURE.md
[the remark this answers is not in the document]
milosevic: I agree here that module should be defined (probably in vocabulary section). It also probably make sense to introduce new terms and not reuse for example process as blockchain modules are very specific processes.

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
+# 1: IBC Architecture
+
+> This is an overview of the high-level architecture & dataflow of the IBC protocol.
+
+> For a broad set of protocol design principles, see [here](./2_IBC_DESIGN_PRINCIPLES.md).
+
+> For definitions of terms used in IBC specifications, see [here](./3_IBC_TERMINOLOGY.md).
+
+This document outlines the architecture of the authentication, transport, and ordering layers of the IBC protocol stack. This document does not describe specific protocol details — those are contained in individual ICSs.
```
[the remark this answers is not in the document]
jackzampolin: I don't think, especially on this repo we need to `Inter-Blockchain Communication` everything.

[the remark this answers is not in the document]
gamarin2: just the first one

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Updated.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Moved to "interfaces", maybe it fits better there.

docs/ibc/1_IBC_ARCHITECTURE.md
zmanian: I'm having trouble actually defining safety in IBC independently of safety in the underlying consensus engines.

Liveness in IBC seems identical to data availability.

Maybe the only parameter of IBC that matters is data availability.

docs/ibc/1_IBC_ARCHITECTURE.md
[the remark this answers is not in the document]
cwgoes: So updated.

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
...
+
+## What is IBC?
+
+The *inter-blockchain communication protocol* is designed for use as a reliable, authenticated module-to-module communication protocol between modules running on independent distributed ledgers across an untrusted network layer.
+
+## What is IBC not?
+
```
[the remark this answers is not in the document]
cwgoes: OK, added a line about that.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: - they are predominant by market capitalization, noted that
- updated
- Satoshi's paper refers to "peer-to-peer digital cash", and it is that use case which I am referring to here; but the point is not to argue with Bitcoin adherents, just to explain the problem. What do you suggest? I just cut the last part of the sentence for now.
- It's the limiting factor on the number of transactions in a single block (or round of consensus), yes. What do you mean by "nodes must agree on sets of transactions sequentially" - that they must run consensus algorithms?

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: I agree that interoperability is a desideratum, but I think it is derived from the primary goal of a scalable, versatile ledger or set of interacting ledgers - if a single ledger could handle a diversity of use cases and scale indefinitely, we wouldn't need to write an interoperability protocol. Maybe the "diversity of use cases" part could be clearer.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: It is, modules are components of the state machine just like IBC is.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: What do you mean? That the validators have network connections?

I don't think the method of access matters for the protocol construction.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Added a note that blockchain and distributed ledger are equivalent, switched to ledger in a few instances.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: OK, done.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: (did you read a contradiction here?)

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,141 @@
+# 1: IBC Architecture
```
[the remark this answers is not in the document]
cwgoes: OK; updated.

docs/ibc/1_IBC_ARCHITECTURE.md
gamarin2: > What do you suggest?

Just say that "throughput is a limitation for most blockchain use cases"

> What do you mean by "nodes must agree on sets of transactions sequentially" - that they must run consensus algorithms?

I'm saying that without IBC we're essentially in a purely sequential world. IBC enables concurrent designs. A validator can scale their ressources to have more storage/bandwidth, but on a single ledger the validator set will always be limited because of the fact that there is a single global state that must be updated all at once.

docs/ibc/1_IBC_ARCHITECTURE.md
gamarin2: well then saying that "IBC sits between state machine components and underlying consensus protocol" when it is a part of the state machine is confusing

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: I think there is a notion of "safety, assuming correct blockchain operation, but possible non-liveness", and a separate notion of "safety, assuming Byzantine blockchain operation" - IBC needs to provide the former and cannot provide the latter in a strong sense but can expose options to applications to mitigate / contain damage from Byzantine blockchains.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Channels provide multiplexing.

docs/ibc/1_IBC_ARCHITECTURE.md
gamarin2: > but I think it is derived from the primary goal of a scalable

Hmm, my point is that I think it deserves its own paragraph. 

> if a single ledger could handle a diversity of use cases and scale indefinitely, we wouldn't need to write an interoperability protocol

We would, e.g. for sovereignty. There is also the idea of optimizing each ledger for their use cases. 

As I said, if two ledgers do not scale, it is still interesting for them to interoperate. Even if it does not help with scalability. Also we have been pitching IBC for itneroeprability for the longest time. I do think it deserves a small paragraph.

docs/ibc/1_IBC_ARCHITECTURE.md
gamarin2: Yes, fair

docs/ibc/1_IBC_ARCHITECTURE.md
gamarin2: Yes, but I feel it's less explicit. Channels are a core part of the architecture, I think they deserve to appear as section title

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: I agree with sovereignty as a separate goal.

cwgoes: Added a paragraph - https://github.com/cosmos/ics/pull/70/commits/ba133b0ccd07870a3a185db7a8bc78554a5d0962

Thoughts?

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: > Just say that "throughput is a limitation for most blockchain use cases"

Updated.

> I'm saying that without IBC we're essentially in a purely sequential world. IBC enables concurrent designs. A validator can scale their ressources to have more storage/bandwidth, but on a single ledger the validator set will always be limited because of the fact that there is a single global state that must be updated all at once.

I would agree with all of this - does something in the text contradict it?

The last sentence in this paragraph is:

```
In order to support the transaction throughput, application diversity, and cost efficiency required to facilitate wide deployment of distributed ledger applications, execution and storage must be split across many independent consensus instances which can run concurrently
```

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: See https://github.com/cosmos/ics/pull/70#discussion_r276290743

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Changed to "other state machine components"

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: The section titles are supposed to be about provided properties, I think that makes more sense to a reader since they won't know what channels are because we haven't defined them.

"Connections" also violates this, I changed it to "Statefulness"

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: There will usually be an actor involved, unless the module is being called by a scheduler or something.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: The part of the state machine implementing the IBC protocol, which routes calls from modules to channels. Added a line in the section synopsis.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Called it "cryptographic accumulator proof", I think we can expect prior knowledge of that.

docs/ibc/1_IBC_ARCHITECTURE.md
melekes: > options to applications to mitigate / contain damage

is there a list somewhere?

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Statefulness
+
+Reliability, flow control, and authentication as described above require that IBC initializes and maintains certain status information for each datastream. This information is called a connection. Each connection object contains information about the consensus state of the connected ledger, negotiated encoding & multiplexing options, and state & sequence numbers. When two modules wish to communicate, they must locate an existing connection between their two ledgers, or initialize a new connection if none yet exists. Initializing a connection requires a multi-step handshake which, once complete, ensures that only the two intended ledgers are connected and future datagrams relayed will be authenticated, encoded, and sequenced as desired.
```
melekes: By connection you mean records on both ledgers (sequence numbers, etc.)? It's not like they maintain a stable network connection, although they had to create one for the handshake. Relayers then utilize their own network connections to relay datagrams.

docs/ibc/1_IBC_ARCHITECTURE.md
gamarin2: Spot on :+1:

docs/ibc/1_IBC_ARCHITECTURE.md
gamarin2: The last sentence is great!

I guess the confusing sentence was "since every (validating) node in the network must process every transaction and store all state". To me it sounded like the limitation was storage/bandwidth, and not the fact that the state must be updated globally. 

In a concurrent model nodes would still have to store all state, but subsets of the state would be update-able concurrently. Maybe rephrasing the sentence slightly would help clarify

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Flow control
+
+IBC does not require specific provision for computation-level flow control since the underlying ledgers will have throughput limitations and flow control mechanisms of their own (such as "gas" markets). Application-level flow control — limiting the rate of particular packets according to their content — may be useful to ensure security properties (limiting the value on a single ledger) and contain damage from Byzantine faults (allowing a challenge period to prove an equivocation, then closing a connection). IBC provides facilities for modules to reject packets and leaves particulars up to the higher-level application protocols.
```
vshvsh: Pushing responsibility for logical security on application is the right way to build IBC but I feel we need a best practice guide soon after the start. Without a proper framework for reasoning about IBC application-level security IBC-relying applications won't be secure in practice.

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Authentication
+
+All datagrams in IBC are authenticated: a block finalized by the consensus algorithm of the sending ledger must commit to the outgoing datagram via a cryptographic accumulator, and the receiving chain must verify both the consensus transcript and the cryptographic accumulator proof that the datagram was sent (and associated actions executed) before acting upon it. For correct execution semantics, both chains must implement the correct IBC protocol — of course, since IBC is permissionless, any chain could connect to another chain and provide a correct cryptographic accumulator proof but incorrectly implement IBC such that its IBC handler did not behave according to protocol, so modules utilizing IBC connections (or end-users using those modules) must inspect the state of the connections and ledgers they plan to use to ensure correct implementation.
```
vshvsh: "receiving chain must verify" means "receiving chain's connection handler must verify"? I think it's better to be more concrete

vshvsh:
The one thing that's not easily understandable from this set of documents is why you presume future IBC topology to be complex and not, say, complete graph: reading how permissionless it is you certainly don't understand where that hub and spoke model in examples comes from.

It's be nice to have list of reasons to limit connections to only well-known well-behaving networks or a discriminating hub - namely expensive on-chain storage, reducing counterparty risk and so on. I'm not sure that speculations on how and why IBC network will shape itself should be in this PR though.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Yes, that's right. I suppose "connection" is terminologically confusing - but that's also what it's called in TCP (where it also isn't a physical wire), so I think it's OK.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Definitely - by "application" I just mean that it isn't part of the "core" protocol - we'll want to specify and analyze concrete constructions for providing additional security in different ways.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Indeed, clarified.

cwgoes:
> It's be nice to have list of reasons to limit connections to only well-known well-behaving networks or a discriminating hub - namely expensive on-chain storage, reducing counterparty risk and so on. I'm not sure that speculations on how and why IBC network will shape itself should be in this PR though.

I agree that it's worthwhile to reason about future network topologies, but I think it is out of scope of this document and the core protocol specification generally - but it could well be in scope for "example topology" documents and analysis in the future, in this repo if you like - some discussion has started at https://github.com/cosmos/ics/issues/59.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: > is there a list somewhere?

No, but there should be - that's a topic for https://github.com/cosmos/ics/issues/6.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Added a line about use cases. I still want to keep the "What is IBC not?" section since these misconceptions frequently come up.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Right, that's confusing. I'm trying to emphasize that it's safe to run multiple instances of the IBC protocol at the same time between many different ledgers which are all connected to each other.

Tried to clarify.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Fair, removed the last sentence.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Thanks, updated.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Added, updated.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: I think we can get rid of that sentence for now, I agree it's too low-level. Done.

Clarified blockchain / ledger at the top of the document.

docs/ibc/1_IBC_ARCHITECTURE.md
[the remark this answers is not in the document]
cwgoes: Added a definition in this sentence.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Added "... and communicate with other validating nodes."

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: ... but I guess we still don't need it for the protocol definition, removed.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Linked to definitions.

cwgoes:
I believe all comments have been addressed.

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,154 @@
...
+Consider the path of an IBC packet between two chains — call them *A* and *B*:
+
+### Diagram
+
+```
++------------------------------------------------------------------------------+
+| Chain A                                                                      |
+|                                                                              |
+| Module --> Handler --> Packet --> Channel --> Connection --> Consensus       |
++------------------------------------------------------------------------------+
```
Liamsi: I think this deserves slightly more structure. In combination with the diagram under "Protocol relations" (which gives a really good high level overview IMO), this is slightly confusing: If you look at both diagrams together, it looks like "Consensus" here is part of the IBC box (under "Protocol relations"). @cwgoes clarified to me (thanks!) that this is the actual Consensus though (as in tendermint Consensus and not the Consensus state as per ICS2).

docs/ibc/1_IBC_ARCHITECTURE.md
```diff
@@ -0,0 +1,154 @@
...
+|                              |                           |                              |
+| +--------------------------+ |                           | +--------------------------+ |
+| | State Machine            | |                           | | State Machine            | |
+| |                          | |                           | |                          | |
+| | +----------+     +-----+ | |        +---------+        | | +-----+     +----------+ | |
+| | | Module A | <-> | IBC | | | <----> | Relayer | <----> | | | IBC | <-> | Module B | | |
+| | +----------+     +-----+ | |        +---------+        | | +-----+     +----------+ | |
+| +--------------------------+ |                           | +--------------------------+ |
++------------------------------+                           +------------------------------+
+```
```
Liamsi: 👍

docs/ibc/README.md
```diff
@@ -0,0 +1,5 @@
+If you're looking for an overview of the IBC protocol, read the documents in numerical order.
```
gamarin2: Add `# Inter-Blockchain Communication Protocol (IBC)` before. 

+ maybe a short intro paragraph (like an abstract) before the following sentence "For an overview of the IBC protocol, read the following documents in numerical order.", as this will probably be the first point of contact with IBC

gamarin2:
```suggestion
For an overview of the IBC protocol, read the following documents in numerical order.
```

cwgoes: So added, so updated.

docs/ibc/1_IBC_ARCHITECTURE.md
cwgoes: Added more structure.

cwgoes:
Merging now, but more comments / suggestions welcome.

PR #74, opened 2019-04-16
ICS 23: Vector commitments

cwgoes:
Closes https://github.com/cosmos/ics/issues/57.

Defines a general vector commitment schema, including types and functions that must be defined.


spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+stage: draft
+category: ibc-core
+author: Christopher Goes <cwgoes@tendermint.com>
+created: 2019-04-16
+modified: 2019-04-25
+---
+
+# Synopsis
+
+An *accumulator*, or *cryptographic accumulator*, is a construction that produces a succinct, binding commitment to a set or indexed vector of elements and short membership and/or non-membership proofs for any element in the set or vector.
```
gamarin2: Is "commitment" a widely understood term in the literature or do we need to define it?

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+An `AccumulatorState` is the full state of the accumulator, which will be stored by the manager.
+
+```golang
+type AccumulatorState struct
+```
+
+#### Root
+
+An `AccumulatorRoot` commits to a particular accumulator state and should be succinct.
```
gamarin2: Same question, I'm wondering if "commitment/commit" is the most explicit word we can use for the root of the accumulator

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+The `remove` function removes an element from an accumulator.
+
+```coffeescript
+remove(AccumulatorState state, Element elem) -> AccumulatorState
+```
+
+#### Proof generation
+
+The `createMembershipWitness` function generates a proof that an element has been added to an accumulator.
```
gamarin2: Why `createMembershipWitness` instead of `createMembershipProof`?

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+```coffeescript
+createMembershipWitness(AccumulatorState state, Element elem) -> AccumulatorProof
+```
+
+The `createNonMembershipWitness` function generates a proof that an element has not been added to an accumulator.
```
gamarin2: Same question as aove

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+### Properties
+
+Accumulators must be *correct* and *sound*. In practice, violations of these properties by computationally-bounded adversaries may be negligible in some security parameter, which is sufficient for use in IBC.
+
+#### Correctness
+
+Accumulator proofs must be *correct*: elements which have been added to the accumulator can always be proved to have been included, but cannot be proved to be excluded.
+
+For an element `elem` in the accumulator `acc`,
```
gamarin2: "For an" -> "For all"?

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+```coffeescript
+verifyNonMembership(root, proof, elem) == false
+```
+
+#### Soundness
+
+Accumulator proofs must be *sound*: elements which have not been added to the accumulator can never be proved to have been included, but can always be proved to have been excluded.
+
+For an element `elem` not in the accumulator `acc`, for all values of `proof`,
```
gamarin2: "For an" -> "For all"

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+Accumulators must be *correct* and *sound*. In practice, violations of these properties by computationally-bounded adversaries may be negligible in some security parameter, which is sufficient for use in IBC.
+
+#### Correctness
+
+Accumulator proofs must be *correct*: elements which have been added to the accumulator can always be proved to have been included, but cannot be proved to be excluded.
+
+For an element `elem` in the accumulator `acc`,
+
+```coffeescript
```
gamarin2: I wonder if this is the best way to present a property that must be always true. Maybe @milosevic can chime in on the best way to formally present a property that must always be verified.

cwgoes: The formal definition is something like on https://eprint.iacr.org/2017/043.pdf, page 7.

We can include that, I don't know if it's necessary but maybe better to be verbose.

cwgoes:
@ValarDragon has convinced me that we should standardize on vector commitments. I'll modify this accordingly.

spec/ics-23-standard-accumulator/README.md
ValarDragon: Commitment is widely understood

spec/ics-23-standard-accumulator/README.md
gamarin2: Sorry maybe I was not clear enough. My point is more about the format of the property. I don't think the way it is currently formatted is the best way to do it (or at least that it explains unequivocally and explicitly that the function must ALWAYS return true if `elem` is in `acc`)

cwgoes: I see - agreed, that is what this sentence is intended to convey, so if it didn't it should be more explicit.

spec/ics-23-standard-accumulator/README.md
cwgoes: "For any"

spec/ics-23-standard-accumulator/README.md
cwgoes: "For any"

spec/ics-23-standard-accumulator/README.md
cwgoes: `witness` is used in the literature but I guess we don't need to call it that

spec/ics-23-standard-accumulator/README.md
cwgoes: I think it's standard - but open to other suggestions.

spec/ics-23-standard-accumulator/README.md
gamarin2: If we use `proof` everywhere I think we should stick to it in the function's name

spec/ics-23-standard-accumulator/README.md
gamarin2: Ok, wasn't sure it was standard. Would be good to define it somewhere still imo

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+  verifyNonMembership(root, proof, elems[1]) && ...
+```
+
+If batch verification is possible and more efficient than individual verification of one proof per element, an accumulator construction SHOULD define batch verification functions.
+
+### Properties
+
+Accumulators must be *correct* and *sound*. In practice, violations of these properties by computationally-bounded adversaries may be negligible in some security parameter, which is sufficient for use in IBC.
+
+#### Correctness
```
ValarDragon: This is typically defined as Completeness

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+and, likewise, for all values of `proof`,
+
+```coffeescript
+verifyNonMembership(root, proof, elem) == false
+```
+
+#### Soundness
+
+Accumulator proofs must be *sound*: elements which have not been added to the accumulator can never be proved to have been included, but can always be proved to have been excluded.
```
ValarDragon: Its impossible to construct an accumulator with perfect soundness :P

spec/ics-23-standard-accumulator/README.md
cwgoes: Oh really? https://eprint.iacr.org/2017/043.pdf uses "correctness"

I like completeness better though

spec/ics-23-standard-accumulator/README.md
cwgoes: I mention the security parameter above. Do you think it's likely that different chains will want to use the same type of accumulator with different security parameters? (otherwise the security parameter can just be part of the definition of the accumulator algorithm, for the purposes of IBC implementers)

spec/ics-23-standard-accumulator/README.md
ValarDragon: oh interesting. I imagine it varies per author then, completeness is definitely the one I've seen in more locations.

spec/ics-23-standard-accumulator/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+If batch verification is possible and more efficient than individual verification of one proof per element, an accumulator construction SHOULD define batch verification functions.
+
+### Properties
+
+Accumulators must be *correct* and *sound*. In practice, violations of these properties by computationally-bounded adversaries may be negligible in some security parameter, which is sufficient for use in IBC.
+
+#### Correctness
+
+Accumulator proofs must be *correct*: elements which have been added to the accumulator can always be proved to have been included, but cannot be proved to be excluded.
```
ValarDragon: Don't require perfect completeness! Having negligible error is extremely useful. This is used in efficient RSA accumulator constructions for instance.

spec/ics-23-standard-accumulator/README.md
ValarDragon: I think most accumulatosr should have a fixed security parameter, as you want probably want to do group-specific optimizations in your code. The only accumulator I know of which you can switch security parameters trivially is merkle trees with a hash like blake2b, since the length is just another parameter, and this seems likely to occur imo. (Remember when Cosmos wanted the MT hash size to be 20 bytes)

spec/ics-23-standard-accumulator/README.md
cwgoes: OK, agreed, updated.

spec/ics-23-standard-accumulator/README.md
cwgoes: Agreed, updated.

spec/ics-23-standard-accumulator/README.md
cwgoes: Updated.

spec/ics-23-standard-accumulator/README.md
```diff
@@ -10,14 +10,15 @@ modified: 2019-04-25
 
 # Synopsis
 
-An *accumulator*, or *cryptographic accumulator*, is a construction that produces a succinct, binding commitment to a set or indexed vector of elements and short membership and/or non-membership proofs for any element in the set or vector.
-This specification enumerates the functions and properties required of accumulator constructions used in the IBC protocol.
+An *accumulator*, or *cryptographic accumulator*, is a construction that produces a succinct, binding commitment to an indexed vector of elements and short membership and/or non-membership proofs for any indicies & elements in the vector.
```
ValarDragon: why refer to accumulators at all, couldn't everything just be described around vector commitments?

cwgoes: I suppose but "accumulator" is more likely to be a known term - vector commitments are considered to be a kind of accumulator, right?

ValarDragon: I'm not an expert by any means on accumulators or vector commitments, but I typically see them refferred to distinctly, not as vector commitments being a subset of the other. Formally one is not the subset of the other, they do different things. (For instance, look at page 1 of https://eprint.iacr.org/2018/1188.pdf)

I don't think accumulators are necessarily a more known term. I am confused as to the title of this ICS. Perhaps the title should be something more general (proof system API requirements?), and therein in can define vector commitments.

cwgoes: "State Commitments" perhaps?

spec/ics-23-standard-accumulator/README.md
cwgoes: OK :+1: Said as much.

spec/ics-23-standard-accumulator/README.md
cwgoes: Aye aye, updated.

spec/ics-23-standard-accumulator/README.md
cwgoes: We can just call them vector commitments if that's all we'll use, it seems so for now.

spec/ics-23-standard-accumulator/README.md
cwgoes: For now, we'll get rid of "accumulator" and just call it "commitment".

spec/ics-23-standard-accumulator/README.md
ValarDragon: Sounds great to use vector commitments for now!

spec/ics-23-standard-accumulator/README.md
cwgoes: I cleaned this section up a bit. We could still explicitly define the computational power of our adversary.

docs/ibc/3_IBC_TERMINOLOGY.md
```diff
@@ -38,9 +38,9 @@ A *header* is an update to the consensus state of a particular blockchain that c
 
 *Finality* is the guarantee provided by a consensus algorithm that a particular block will not be reverted, subject to certain assumptions about the behavior of the validator set. The IBC protocol requires finality.
 
-### Accumulator
+### Commitment 
 
-A cryptographic *accumulator* is a way to cheaply verify membership of a value in a set, where the set can be commited to with a short witness string. An *accumulator proof* refers to the proof structure which proves whether a particular value is or is not a member of a particular committed-to set.
+A cryptographic *commitment* is a way to cheaply verify membership of a key => value pair in a mapping, where the mapping can be commited to with a short witness string. An *vector commitment proof* refers to the proof structure which proves whether a particular key maps to a particular value in a committed-to set or not.
```
ValarDragon: Perhaps reword to state that we are using `commitment` as an alias for `vector commitment`.  Colloquially this is right, though precisely a cryptographic commitment is something akin to a hash and reveal scheme. (You commit to a value in a way that it is computationally binding, and then can reveal the bound value later)

https://en.wikipedia.org/wiki/Commitment_scheme

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+In order to provide a guarantee of a particular state transition having occurred on one chain which can be verified on another chain, IBC requires an efficient cryptographic construction to prove inclusion or non-inclusion of particular values at particular keys in state.
+
+## Definitions
+
+The *manager* of a vector commitment is the actor with the ability and responsibility to add or remove items from the commitment. Generally this will be the state machine of a blockchain.
+
+The *prover* is the actor responsible for generating proofs of inclusion or non-inclusion of particular elements. Generally this will be a relayer (see [ICS 18](../ics-18-relayer-algorithms)).
+
+The *verifier* is the actor who checks proofs in order to verify that the manager of the commitment did or did not add a particular element. Generally this will be an IBC handler running on another chain.
```
Liamsi: Maybe link to the ICS defining what an IBC handler is?

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+Commitments are instantiated with particular *key* and *value* types, which is assumed to be arbitrary serializable data.
```
Liamsi: which are assumed to be (?)

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Desired Properties
+
+This document only defines desired properties, not a concrete implementation — see "Properties" below.
+
+## Technical Specification
+
+### Datatypes
+
+An commitment construction MUST specify the following datatypes, which are otherwise opaque (need not be introspected) but must be serializable:
```
Liamsi: Not sure but should this be consistently capiatlized?
```suggestion
An commitment construction MUST specify the following datatypes, which are otherwise opaque (need not be introspected) but MUST be serializable:
```

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+If batch verification is possible and more efficient than individual verification of one proof per element, an commitment construction SHOULD define batch verification functions.
+
+### Properties
+
+Commitments must be *complete*, *sound*, and *position binding*. These properties are defined with respect to a security parameter `λ`, which must be agreed upon by the manager, prover, and verifier (and often will be constant for the commitment algorithm).
+
+#### Completeness
+
+Commitment proofs must be *complete*: key => value mappings which have been added to the commitment can always be proved to have been included, and keys which have not been included can always be proved to have been excluded, except with probability negligible in `λ`.
```
Liamsi: "except with probability negligible in `λ`." Should this be slightly more formal once in this document or in the glossary document. E.g. stating more formally that the probability of the opposite can be expressed as a function which is negligible in λ. Or point to some definition somewhere?

spec/ics-23-vector-commitments/README.md
cwgoes: It's defined in the top-level docs - added a note

spec/ics-23-vector-commitments/README.md
cwgoes: thanks, fixed

spec/ics-23-vector-commitments/README.md
cwgoes: aye, updated throughout

spec/ics-23-vector-commitments/README.md
cwgoes: Defined more clearly and linked to Wikipedia's definition (which is clear)

spec/ics-23-standard-accumulator/README.md
cwgoes: Noted that success probabilities must be negligible

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,252 @@
...
+stage: draft
+category: ibc-core
+author: Christopher Goes <cwgoes@tendermint.com>
+created: 2019-04-16
+modified: 2019-04-25
+---
+
+# Synopsis
+
+A *vector commitment* is a construction that produces a succinct, binding commitment to an indexed vector of elements and short membership and/or non-membership proofs for any indicies & elements in the vector.
```
cwgoes:
```suggestion
A *vector commitment* is a construction that produces a succinct, binding commitment to an indexed vector of elements and short membership and/or non-membership proofs for any indices & elements in the vector.
```

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,252 @@
...
+The `createMembershipProof` function generates a proof that a particular key has been set to a particular value in an commitment.
+
+```coffeescript
+createMembershipProof(CommitmentState state, Key key, Value value) -> CommitmentProof
+```
+
+The `createNonMembershipProof` function generates a proof that a key has not been set to any value in an commitment.
+
+```coffeescript
+createNonMembersipProof(CommitmentState state, Key key) -> CommitmentProof
```
cwgoes:
```suggestion
createNonMembershipProof(CommitmentState state, Key key) -> CommitmentProof
```

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,252 @@
...
+If defined, these functions MUST be computationally equivalent to the conjunctive union of `verifyMembership` and `verifyNonMembership` respectively (`proof` may vary):
+
+```coffeescript
+batchVerifyMembership(root, proof, items) ==
+  verifyMembership(root, proof, items[0].Key, items[0].Value) &&
+  verifyMembership(root, proof, items[1].Key, items[1].Value) && ...
+```
+
+```coffeescript
+batchVerifyMembership(root, proof, keys) ==
```
cwgoes:
```suggestion
batchVerifyNonMembership(root, proof, keys) ==
```

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,252 @@
...
+#### Position binding
+
+Commitment proofs MUST be *position binding*: a given key can only map to one value, and an commitment proof cannot prove that the same key opens to a different value except with probability negligible in λ.
+
+For any key `key` set in the commitment `acc`, there is one `value` for which:
+
+```coffeescript
+root = getRoot(acc)
+proof = createMembershipProof(acc, key, value)
+verifyMembership(root, proof, key, value) == true
```
cwgoes:
```suggestion
P(verifyMembership(root, proof, key, value) == false) negligible in λ
```

spec/ics-23-vector-commitments/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Other Implementations
+
+Coming soon.
+
+# History
+
+Security definitions are mostly sourced from these papers (and simplified somewhat):
+- [Vector Commitments and their Applications](https://eprint.iacr.org/2011/495.pdf)
+- [commitments with Applications to Anonymity-Preserving Revocation](https://eprint.iacr.org/2017/043.pdf)
```
cwgoes:
```suggestion
- [Commitments with Applications to Anonymity-Preserving Revocation](https://eprint.iacr.org/2017/043.pdf)
```

spec/ics-23-vector-commitments/README.md
ValarDragon: Odd notation. Perhaps use `Pr` to be more suggestive. 
(Still seems odd to me, but I don't have any other suggestions for this)

cwgoes: yeah, that's fair, I'm not sure how best to do it in Markdown

any ideas?

cwgoes:
Merging as a draft, future comments always welcome.

PR #75, opened 2019-04-16
ICS 24: Host state machine requirements

cwgoes:
Closes https://github.com/cosmos/ics/issues/64.

Define requirements of the host state machine.


spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+`RootOfTrust` is as defined in [ICS 2](../ics-2-consensus-requirements).
+
+## Desired Properties
+
+IBC should require as simple an interface from the underlying state machine as possible to maximize the ease of correct implementation.
+
+## Technical Specification
+
+### Keys, identifiers, separators
```
Liamsi: Nit: consistent capitalization of titles (should this be "Keys, Identifiers, Separators)

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+requires: 2
+required-by: 2, 3, 4, 5, 18
+author: Christopher Goes <cwgoes@tendermint.com>
+created: 2019-04-16
+modified: 2019-04-29
+---
+
+# Synopsis
+
+This specification defines the minimal set of interfaces which must be provided and properties which must be fulfilled by a blockchain & state machine hosting an IBC handler (implementation of the interblockchain communication protocol).
```
Liamsi: Should this link to the architecture document here to have quick access to where the IBC handler fits in the larger architecture (and vice versa?).

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Desired Properties
+
+IBC should require as simple an interface from the underlying state machine as possible to maximize the ease of correct implementation.
+
+## Technical Specification
+
+### Keys, identifiers, separators
+
+A `Key` is a bytestring used as the key for an object stored in state. Keys MUST contain only alphanumeric characters and the separator `/`.
```
mossid: Why do the keys & identifiers have to contain only the alphanumeric characters and slashes? In this case, it will be not able to set the arbitrary binary encoded identifiers, for example, lightclient identifier generated from the ROT hash.

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+An `Identifier` is a bytestring used as a key for an object stored in state, such as a connection, channel, or light client. Identifiers MUST consist of alphanumeric characters only.
+
+Identifiers are not intended to be valuable resources — to prevent name squatting, minimum length requirements or pseudorandom generation MAY be implemented.
+
+The separator `/` is used to separate and concatenate two identifiers or an identifier and a constant bytestring. Identifiers MUST NOT contain the `/` character, which prevents ambiguity.
```
mossid: Using slash character as the separator will make it harder to separate each segment of the key harder, as it requires the full iteration on the key bytestring to find the separators. My opinion is that we have to stick with either fixed length or length prefixed key format.

spec/ics-24-host-requirements/README.md
cwgoes: It was consistent at this header level :P - fixed

spec/ics-24-host-requirements/README.md
cwgoes: aye, added.

spec/ics-24-host-requirements/README.md
cwgoes: Either all keys need to be fixed-length, with fixed-length components, or we need a specific separator which cannot be included in identifiers

spec/ics-24-host-requirements/README.md
cwgoes: Hmm, when do we need to separate key segments?

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+stage: draft
+category: ibc-core
+requires: 2
+required-by: 2, 3, 4, 5, 18
+author: Christopher Goes <cwgoes@tendermint.com>
+created: 2019-04-16
+modified: 2019-04-29
+---
+
+# Synopsis
```
gamarin2: I will link to this comment I made on ICS2 which is relevant for this one too (and prob others) https://github.com/cosmos/ics/pull/20/files#r279727245.

Also in github formatting `#` is used for the main title only usually. It might look weird once merged if you use it for sections

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Desired Properties
+
+IBC should require as simple an interface from the underlying state machine as possible to maximize the ease of correct implementation.
+
+## Technical Specification
+
+### Keys, Identifiers, Separators
+
+A `Key` is a bytestring used as the key for an object stored in state. Keys MUST contain only alphanumeric characters and the separator `/`.
```
gamarin2:
```suggestion
A `Key` is a bytestring used as the key for an object stored in state. Keys **must** contain only alphanumeric characters and the separator `/`.
```

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+An `Identifier` is a bytestring used as a key for an object stored in state, such as a connection, channel, or light client. Identifiers MUST consist of alphanumeric characters only.
```
gamarin2:
```suggestion
An `Identifier` is a bytestring used as a key for an object stored in state, such as a connection, channel, or light client. Identifiers **must** consist of alphanumeric characters only.
```

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Identifiers are not intended to be valuable resources — to prevent name squatting, minimum length requirements or pseudorandom generation MAY be implemented.
```
gamarin2:
```suggestion
Identifiers are not intended to be valuable resources — to prevent name squatting, minimum length requirements or pseudorandom generation **may** be implemented.
```

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+The separator `/` is used to separate and concatenate two identifiers or an identifier and a constant bytestring. Identifiers MUST NOT contain the `/` character, which prevents ambiguity.
```
gamarin2:
```suggestion
The separator `/` is used to separate and concatenate two identifiers or an identifier and a constant bytestring. Identifiers **must not** contain the `/` character, which prevents ambiguity.
```

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Variable interpolation, denoted by curly braces, MAY be used as shorthand to define key formats, e.g. `client/{clientIdentifier}/rootOfTrust`.
```
gamarin2:
```suggestion
Variable interpolation, denoted by curly braces, **may** be used as shorthand to define key formats, e.g. `client/{clientIdentifier}/rootOfTrust`.
```

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Key/value Store
+
+Host chains MUST provide a simple key-value store interface, with two functions which behave in the way you would expect:
```
gamarin2:
```suggestion
Host chains **must** provide a simple key-value store interface, with two functions which behave in the standard way:
```

gamarin2: avoid "you", "we", etc. on spec

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```coffeescript
+function set(Key key, Value value)
+```
+
+`Key` is as defined above. `Value` is an arbitrary bytestring encoding of a particular data structure. Encoding details are left to separate ICSs.
+
+### Root-of-trust Introspection
+
+Host chains MUST provide the ability to introspect their own root-of-trust, with `getRootOfTrust`:
```
gamarin2:
```suggestion
Host chains **must** provide the ability to introspect their own root-of-trust, with `getRootOfTrust`:
```

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```coffeescript
+function getRootOfTrust() -> RootOfTrust
+```
+
+`getRootOfTrust` MUST return the current root-of-trust for the consensus algorithm of the host chain.
```
gamarin2:
```suggestion
`getRootOfTrust` **must** return the current root-of-trust for the consensus algorithm of the host chain.
```

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Datagram Submission
+
+Host chains MUST define a unique `submitDatagram` function:
```
gamarin2:
```suggestion
Host chains **must** define a unique `submitDatagram` function:
```

spec/ics-24-host-requirements/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```coffeescript
+function submitDatagram(Datagram datagram)
+```
+
+`submitDatagram` allows relayers to relay IBC datagrams to the host chain. Host chains MAY require that the relayer submitting the datagram has an account to pay transaction fees, signs over the datagram in a larger transaction structure, etc - `submitDatagram` MUST define any such packaging required.
```
gamarin2:
```suggestion
`submitDatagram` allows relayers to relay IBC datagrams to the host chain. Host chains **may** require that the relayer submitting the datagram has an account to pay transaction fees, signs over the datagram in a larger transaction structure, etc - `submitDatagram` **must** define any such packaging required.
```

gamarin2: Link to IBC definition of `datagram`

spec/ics-24-host-requirements/README.md
mossid: Most of my concern with the variable key is that we are no longer able to use arbitrary bytestring for the identifiers... variable length can be easily simulated with the length prefix method I think.

spec/ics-24-host-requirements/README.md
cwgoes: We want to use the capitalized words - ref https://www.ietf.org/rfc/rfc2119.txt

spec/ics-24-host-requirements/README.md
cwgoes: We want to use the capitalized words - ref https://www.ietf.org/rfc/rfc2119.txt

spec/ics-24-host-requirements/README.md
cwgoes: We want to use the capitalized words - ref https://www.ietf.org/rfc/rfc2119.txt

spec/ics-24-host-requirements/README.md
cwgoes: We want to use the capitalized words - ref https://www.ietf.org/rfc/rfc2119.txt

spec/ics-24-host-requirements/README.md
cwgoes: We want to use the capitalized words - ref https://www.ietf.org/rfc/rfc2119.txt

spec/ics-24-host-requirements/README.md
cwgoes: We want to use the capitalized words - ref https://www.ietf.org/rfc/rfc2119.txt

spec/ics-24-host-requirements/README.md
cwgoes: (see above comments)

spec/ics-24-host-requirements/README.md
cwgoes: (see above comments)

spec/ics-24-host-requirements/README.md
cwgoes: we're actually going to eliminate this requirement, ref https://github.com/cosmos/ics/issues/82

spec/ics-24-host-requirements/README.md
cwgoes: > avoid "you", "we", etc. on spec

agreed - fixed

spec/ics-24-host-requirements/README.md
gamarin2: Yep yep, nvm on this one, see https://github.com/cosmos/ics/pull/20/files#r280047888

spec/ics-24-host-requirements/README.md
cwgoes: Ok... we need a method to prevent ambiguity between identifiers with a separator in them. That would require specifying multiple lengths at the beginning of a key, one for each separator. Is that what you're suggesting?

Alternatively, we could standardize identifier lengths, which I think is not that bad for connections / channels, but I wonder if it might be bad in other cases...

Why is requiring that identifiers use a restricted subset of ASCII a problem?

spec/ics-24-host-requirements/README.md
cwgoes: We've used `#` for top-level titles so far - do you think we should use `##` instead?

There is no main title (the header has special formatting).

gamarin2: Yes, I do. I also think we should change the sectioning a bit, as per my comment in the other PR. 

Having both `# Specification` and `# Technical Specification` is confusing

cwgoes: :+1: , https://github.com/cosmos/ics/issues/84

cwgoes:
@mossid This is mostly waiting on your thoughts re: identifier selection. Can we discuss that?

cwgoes:
Synced with @mossid, merging as a draft, further comments always welcome.

PR #79, opened 2019-04-23
ICS 25: Handler interface

### LINKED ISSUE #78: ICS 25: IBC handler interface
(as exposed to modules)

cwgoes:
Closes #78 
Closes https://github.com/cosmos/ics/issues/82

This isn't done (haven't yet described implementation by the IBC handler), but an initial review pass for agreement on the interface would be helpful.


cwgoes:
This isn't done (haven't yet described implementation by the IBC handler), but an initial review pass for agreement on the interface would be helpful.

cwgoes:
ref https://github.com/cosmos/ics/issues/82#issuecomment-488629070

This document will now just reference definitions in other ICSs.

cwgoes:
.... better split this into "all modules" and "IBC relayer module" lest it become too unwieldy

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+```coffeescript
+function updateClient(string id, Header header) -> Maybe<Err>
+```
+
+Implementations of `createClient`, `queryClient`, and `updateClient` are defined in ICS 2.
+
+
+### Connections
+
+By default, connections are unowned, but closure is permissioned to the module which created the connection.
```
ancazamfir: As we discussed offline:
If module A opens a connection and then closes it, it is possible that between open and close, another module starts using the connection (via the module channel). The connection may be closed only if there are no more channels and maybe after a timeout.

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+
+`ConnectionKind` enumerates the connection types supported by the handler implementation.
+
+```golang
+type ConnectionKind enum {
+  Transit
```
ancazamfir: What is "Transit"..."Transmit" maybe? Are "Transit" and "Receive" unidirectional connections?

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+`initConnection` tries to create a new connection with the provided options, failing if the client is not found or the options are invalid.
+
+```coffeescript
+function initConnection(identifier, desiredVersion, desiredCounterpartyIdentifier, lightClientIdentifier) -> Maybe<err>
+```
+
+`tryConnection` tries to initialize a connection based on an initialization attempt on another chain (part one of the three-way connection handshake).
+
```
ancazamfir: When does it return error?

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+`confirmConnection` finalizes a connection (part three of the three-way connection handshake).
+
+```coffeescript
+function confirmConnection(identifier, proofAck)
+```
+
+`queryConnection` queries an existing connection by known identifier, returning the associated metadata if found.
+
+```coffeescript
+function queryConnection(string id) -> Maybe<ConnectionInfo>
```
ancazamfir:
```suggestion
function queryConnection(identifier) -> Maybe<ConnectionInfo>
```

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+`initCloseConnection` initiates the graceful connection closing process. It will fail if there are any open channels using the connection.
+
```
ancazamfir: Should closing fail if there are open channels? Especially if closed from a module?
Similarly, should open fail if the connection already exist?

ancazamfir: initCloseConnection should fail with wrong identifier

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+```coffeescript
+function initCloseConnection(string id) -> Maybe<Err>
+```
```
ancazamfir: I think there is a difference between '(string id)' and '(identifier)', not sure I understand it well, could you please clarify, maybe in some earlier section.

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+
+`tryCloseConnection` continues the graceful connection closing process. It will fail if there are any open channels using the connection.
+
```
ancazamfir: Same comment as above wrt failure.

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+```coffeescript
+function tryCloseConnection(identifier, proofInit) -> Maybe<Err>
+```
+
+`ackCloseConnection` finalizes the graceful connection closing process.
+
```
ancazamfir: When would this fail?

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+### Channels
+
+By default, channels are owned by the creating module, meaning only the creating module can inspect, close, or send on the channel.
+
+`ChannelOptions` contains all the parameter choices required to create a new channel.
+
+```golang
+type ChannelOptions struct {
+  string          connectionIdentifier
+  bool            ordered
```
ancazamfir: What is the meaning of "ordered"?

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+```coffeescript
+function equivocationCloseConnection(identifier) -> Maybe<Err>
+```
+
+Implementations of `initConnection`, `tryConnection`, `ackConnection`, `confirmConnection`, `queryConnection`, `initCloseConnection`, `tryCloseConnection`, `ackCloseConnection`, and `equivocationCloseConnection` are defined in ICS 3.
+
+### Channels
+
+By default, channels are owned by the creating module, meaning only the creating module can inspect, close, or send on the channel.
+
```
ancazamfir: Can a module create more than one channel?

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+}
+```
+
+`ChannelInfo` contains metadata about an existing channel.
+
+```golang
+type ChannelInfo struct {
+  ChannelOptions options
+}
+```
```
ancazamfir: There is no need for a `channelIdentifier`?

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+
+`createChannel` tries to create a new channel with the provided options, failing if the connection is not found or the options are invalid, returning a unique allocated identifier if successful.
+
+```coffeescript
+function createChannel(ChannelOptions options) -> string
+```
```
ancazamfir: should it return err also?

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,314 @@
...
+
+Implementations of `createChannel`, `queryChannel`, and `closeChannel` are defined in ICS 4.
+
+### Packets
+
+Packets are permissioned by channel (only a module which owns a channel can send on it).
+
+`sendPacket` attempts to send a packet, returning an error if the packet cannot be sent (perhaps because the sending module does not own the channel in question), and returning a unique identifier if successful.
+
+The returned identifier will be the same as that sent by the timeout handler, so it can be used by the sending module to associate a specific action with a specific packet timeout.
```
ancazamfir: What timeout handler is this? This sentence needs a bit of clarification IMO.

spec/ics-25-handler-interface/README.md
cwgoes: Agreed - I think we can actually allow any module to close a connection if and only if all associated channels are closed. I think we need a timeout after the connection was opened, to prevent immediate closure, but I'm not sure we need one after the last channel is closed - is there a particular case you are thinking of? If a module wants to keep the connection open it just needs to have at least one open channel.

spec/ics-25-handler-interface/README.md
cwgoes: Typo - "Transmit". Yes, they are - those haven't been specified yet, since bidirectional connections are more pressing, but they should be possible.

Fixed the typo.

spec/ics-25-handler-interface/README.md
cwgoes: Updated.

spec/ics-25-handler-interface/README.md
cwgoes: Yes, closing will fail if there are open channels (I think the sentence says that) - added a bit about wrong identifier.

spec/ics-25-handler-interface/README.md
cwgoes: No, sorry for the confusion. Changed to `string identifier` everywhere.

Identifiers must follow the ruleset as defined in ICS 24, added a link.

spec/ics-25-handler-interface/README.md
cwgoes: Updated.

spec/ics-25-handler-interface/README.md
cwgoes: If the proof or identifier is invalid; updated.

spec/ics-25-handler-interface/README.md
cwgoes: Whether or not the channel should enforce sequential packet delivery. This will be defined in detail in ICS 4.

spec/ics-25-handler-interface/README.md
cwgoes: Yes. Noted that.

spec/ics-25-handler-interface/README.md
cwgoes: Ah yes, indeed there is; added.

spec/ics-25-handler-interface/README.md
cwgoes: Yes, now that the identifier might already be taken.

spec/ics-25-handler-interface/README.md
cwgoes: `timeoutPacket`. The idea is that the module could associate a specific action with a timeout of a packet once it has sent the packet (without explicitly passing function pointers as callbacks).

Clarified a bit.

cwgoes: Also see https://github.com/cosmos/ics/issues/86

cwgoes:
Please reformat re: https://github.com/cosmos/ics/issues/84

cwgoes:
This is blocked on merging ICS 2, 3, and 4.

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,366 @@
...
+```golang
+type ClientKind enum {
+  Tendermint
+}
+```
+
+`ClientOptions` contains all the parameter choices required to create a client.
+
+```golang
+type ClientOptions struct {
```
mossid: Is it needed to differentiate `ClientOptions` and `ClientInfo`?

cwgoes: Ah not really, this part still needs a bit of cleanup to match ICS 4

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,366 @@
...
+```coffeescript
+function updateClient(string identifier, Header header) -> Maybe<Err>
+```
+
+`freezeClient` freezes an existing client by providing proof-of-equivocation, automatically freezing any associated connections & channels.
+
+The default IBC relayer module will allow external calls to `freezeClient`.
+
+```coffeescript
+function freezeClient(string identifier, Header headerOne, Header headerTwo) -> Maybe<Err>
```
mossid: Is the equivocation proof verification should be done inside the `freezeClient`? As far as I understand the handler interface is exposed to the internal logic only, not the external accounts, so the application logic can decide when to call this function. By separating proving and freezing we can handle other cases to freeze the client in the future easily. 

In the implementation side, it will be a bit tricky to verify the equivocation inside this function, as it requires to read past validator set data, which I think better to be separated from the client logic.

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,366 @@
...
+### Packets
+
+Packets are permissioned by channel (only a module which owns a channel can send on it).
+
+`sendPacket` attempts to send a packet, returning an error if the packet cannot be sent (perhaps because the sending module does not own the channel in question or because the channel is frozen), and returning a unique identifier if successful.
+
+The returned identifier will be the same as that sent by the timeout handler `timeoutPacket`, so it can be used by the sending module to associate a specific action with a specific packet timeout.
+
+```coffeescript
+function sendPacket(Packet packet) -> string | Err
```
mossid: Aren't the `xxxPacket` functions should take conn/chans as its argument?

spec/ics-25-handler-interface/README.md
```diff
@@ -0,0 +1,365 @@
...
+`initChannel` tries to start the handshake to create a new channel with the provided options, failing if the connection is not found or the options are invalid.
+
+```coffeescript
+function initChannel(ChannelOptions options) -> Maybe<err>
+```
+
+`tryChannel` tries to initialize a channel based on proof of an initialization attempt on the counterparty chain, failing if the channel identifier is unavailable, the proof is invalid, or the calling module is not authorized.
+
+```coffeescript
+function tryChannel() -> Maybe<Err>
```
Chris-Hibbert: Doesn't try channel (and others below) need a parameter (ChannelOptions or ChannelInfo) identifying the channel the request covers?

cwgoes:
- [ ] Unify with ICS 2 / 3 / 4 as written.

spec/ics-25-handler-interface/README.md
cwgoes: Fixed, thanks.

cwgoes:
Merging as a draft prior to workshop - follow-up: https://github.com/cosmos/ics/issues/122

PR #81, opened 2019-04-26
Some example IBC usecases

cwgoes:
Closes https://github.com/cosmos/ics/issues/83

This is a list of potential application layer use-cases for IBC.

Each is broadly schematized but not specified in detail.

Several of these might turn into concrete protocol ICS specs in the future; additions welcome.


cwgoes:
Merging as a start, future ideas / comments welcome.

PR #105, opened 2019-05-27
Compile the typescript

cwgoes:
Ref https://github.com/cosmos/ics/issues/104

Nothing to see here yet.


cwgoes:
... unfortunately I can't find a clean way to forward declare values.

cwgoes:
Adding a `.d.ts` file in each directory may be a not unreasonable approach.

cwgoes:
I think a bit more verbosity will be worth the typechecking.

cwgoes:
Still to-do; going to merge as-is so this doesn't get out of date.

Lower priority than consensus on the protocol semantics.

PR #115, opened 2019-06-06
Document some design patterns

### LINKED ISSUE #108: Document rationale for receiver vs dispatch call pattern
This isn't clear as evidenced by discussion.

Explain (& debate):
- Verification requirements of module names
- Requirements of the state machine w.r.t. module routing
- Why it makes the IBC core handler a bit simpler (no dispatch table, no return codes)
- What relayers might have to do
- Example patterns on top

Also:
- Should the IBC relayer module have a dispatch table?

### LINKED ISSUE #112: Note on verification vs computation
The IBC handler logic is written to always perform verification instead of computation where possible, since the computation can happen off-chain, and minimizing on-chain operations is preferable.

I think this can be confusing in some places - https://github.com/cosmos/ics/pull/34#discussion_r288037070 - and should be explained.

cwgoes:
Closes #112 
Closes #108


cwgoes:
cc @ancazamfir

cwgoes:
Merging prior to working group next week, further comments always welcome.

PR #116, opened 2019-06-06
Basic data availability requirements

cwgoes:
Closes https://github.com/cosmos/ics/issues/110

### DIFF
--- a/spec/ics-24-host-requirements/README.md
+++ b/spec/ics-24-host-requirements/README.md
@@ -106,7 +106,7 @@ For safety (e.g. exactly-once packet delivery), host chains MUST have eventual d
 
 For liveness (relaying packets, which will have a timeout), host chains MUST have partially synchronous data availability (e.g. within a wall clock or block height bound), such that any key-value pairs in state can be retrieved by relayers within the bound.
 
-Data computable from a subset of state and knowledge of the state machine (e.g. IBC packet data, which is not directly stored) are also assumed to be available to relayers.
+Data computable from a subset of state and knowledge of the state machine (e.g. IBC packet data, which is not directly stored) are also assumed to be available to and efficiently computable by relayers.
 
 Light clients of particular consensus algorithms may have different and/or more strict data availability requirements.
 
--- a/spec/ics-24-host-requirements/README.md
+++ b/spec/ics-24-host-requirements/README.md
@@ -90,7 +90,7 @@ type getCallingModule = () => string
 
 Modules which wish to make use of particular IBC features MAY implement certain handler functions, e.g. to add additional logic to a channel handshake with an associated module on another chain.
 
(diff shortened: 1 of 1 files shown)


spec/ics-24-host-requirements/README.md
```diff
@@ -100,6 +100,16 @@ type submitDatagram = (datagram: Datagram) => void
 
 `submitDatagram` allows relayers to relay IBC datagrams directly to the host chain. Host chains MAY require that the relayer submitting the datagram has an account to pay transaction fees, signs over the datagram in a larger transaction structure, etc - `submitDatagram` MUST define any such packaging required.
 
+### Data availability
+
+For safety (e.g. exactly-once packet delivery), host chains MUST have eventual data availability, such that any key-value pairs in state can be eventually retrieved by relayers.
+
+For liveness (relaying packets, which will have a timeout), host chains MUST have partially synchronous data availability (e.g. within a wall clock or block height bound), such that any key-value pairs in state can be retrieved by relayers within the bound.
+
+Data computable from a subset of state and knowledge of the state machine (e.g. IBC packet data, which is not directly stored) are also assumed to be available to relayers.
```
ValarDragon: It feels to me like there is some notion of efficiency needed here. 

I can't think of any examples in practice that require this since full nodes are required to compute it in the case where the subset is the entire state (which by definition is a subset of state), but in the case where the relayer is restricted to a proper subset of state, efficiency seems relevant.

spec/ics-24-host-requirements/README.md
```diff
@@ -100,6 +100,16 @@ type submitDatagram = (datagram: Datagram) => void
 
 `submitDatagram` allows relayers to relay IBC datagrams directly to the host chain. Host chains MAY require that the relayer submitting the datagram has an account to pay transaction fees, signs over the datagram in a larger transaction structure, etc - `submitDatagram` MUST define any such packaging required.
 
+### Data availability
+
+For safety (e.g. exactly-once packet delivery), host chains MUST have eventual data availability, such that any key-value pairs in state can be eventually retrieved by relayers.
```
ValarDragon: No time bound is needed?

cwgoes: For exactly-once packet delivery, no (the packet will either be delivered once or timeout)

spec/ics-24-host-requirements/README.md
cwgoes: Added a few words about it, not sure how to formalize more concretely though.

Interestingly, there could be models where the packet data actually *is* private, I think, generated off-chain and proved in zero-knowledge to satisfy some invariants, then submitted by
the original transaction author to the destination chain (maybe even also in zero-knowledge...)

cwgoes:
Merging, further comments always welcome.

PR #129, opened 2019-06-20
Rename "module" to "port", clarify usage

cwgoes:
Closes https://github.com/cosmos/ics/issues/125


cwgoes:
Port binding left to https://github.com/cosmos/ics/issues/126 to avoid conflicts.

PR #140, opened 2019-06-28
Linktypo

Desiderius42:
I've found a typo while reading the spec.


cwgoes:
Why did you delete `state.png`?

Desiderius42:
> Why did you delete `state.png`?

it was duplicate from ics-003-connection-semantics/state.png 
 and kept a folder  ics-3-connection-semantics in the git for nothing.

PR #143, opened 2019-06-29
ICS 005: Port Allocation

cwgoes:
Closes https://github.com/cosmos/ics/issues/130


spec/ics-005-port-allocation/README.md
```diff
@@ -0,0 +1,178 @@
...
+}
+```
+
+#### Transferring ownership of a port
+
+If the host state machine supports object-capability keys, no additional protocol is necessary, since the port reference is a bearer capability. If it does not, the IBC handler MAY implement the following `transferPort` function.
+
+```typescript
+function transferPort(id: Identifier) {
+  assert(authenticate(get(portKey(id))))
```
melekes: should it also release the port before assigning a new value to it `del(portKey(id))`?

ancazamfir:
Looks good! Should ICS 25 and ICS 26 also be updated?

cwgoes:
> Looks good! Should ICS 25 and ICS 26 also be updated?

Done.

PR #149, opened 2019-07-15
ICS 020: Fungible Token Transfer

cwgoes:
Ref #9.

First application-level spec. Work in progress. Also should cover denom tracking (https://github.com/cosmos/ics/issues/9#issuecomment-511355528).


cwgoes:
- [ ] proto3 schema files (ref https://github.com/cosmos/ics/issues/146)

cwgoes:
Denomination management split to a follow-up PR.

PR #187, opened 2019-08-08
Fix 24/getConsensusState

mossid:
Under `getConsensusState`,

> `getConsensusState` MUST return the consensus state for the consensus algorithm of the host chain at the specified height, for all heights greater than zero and less than or equal to the current height.

For some cases `getConsensusState` will not be able to return the consensus state with height of `0 < _ < current`. For efficiency, the chains might choose to prune the states and (if the pruning happens for past enough states) the impl will still work, so the restriction should be RECOMMENDED.

### DIFF
--- a/spec/ics-024-host-requirements/README.md
+++ b/spec/ics-024-host-requirements/README.md
@@ -83,7 +83,7 @@ Host chains MUST provide the ability to introspect their own consensus state, wi
 type getConsensusState = (height: uint64) => ConsensusState
 ```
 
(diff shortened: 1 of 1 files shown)


cwgoes:
> For some cases getConsensusState will not be able to return the consensus state with height of 0 < _ < current. For efficiency, the chains might choose to prune the states and (if the pruning happens for past enough states) the impl will still work, so the restriction should be RECOMMENDED.

I agree; this will impose some synchrony constraints beyond what might already be imposed, though, so the chain will need to specify the number of past consensus states which it keeps accessible (some constant `n`).

PR #189, opened 2019-08-12
Add CommitmentPrefix

mossid:


cwgoes:
Is this ready for review?

cwgoes:
I am open to including this but still prefer that it become part of parameterized clients.

Added to `1.0.0-rc4`.

cwgoes:
Bump, @mossid I think we discussed and I ended up agreeing with you, so we should get this in rc3.

cwgoes:
@mossid Is this R4R? Please mark it as such if so, and fix the CI issues.

cwgoes:
Still failing CI.

cwgoes:
Still fails `check_syntax` :(. Reviewing now.

spec/ics-003-connection-semantics/README.md
```diff
@@ -134,6 +137,33 @@ function removeConnectionFromClient(
 }
 ```
 
+Two helper functions are defined to provide automatic application of `CommitmentPrefix`. These functions are recommended to be used
```
cwgoes: Let's make them mandatory if we do it this way, otherwise the spec is ambiguous.

`CommitmentPrefix` can just be empty I presume.

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -889,8 +878,7 @@ function cleanupPacketUnordered(
     abortTransactionUnless(packet.destPort === channel.counterpartyPortIdentifier)
 
     // abortTransactionUnless acknowledgement on the other end
-    client = queryClient(connection.clientIdentifier)
-    abortTransactionUnless(client.verifyMembership(
+    abortTransactionUnless(connection.verifyMembership(
```
cwgoes: These look good, check if we missed any cases.

spec/ics-023-vector-commitments/README.md
```diff
@@ -59,6 +59,20 @@ In certain commitment constructions with constant-size states, `CommitmentState`
 type CommitmentRoot = object
 ```
 
+#### Prefix
+
+A `CommitmentPrefix` defines a substate of the commitment. It is passed to the proof verification functions with the path to provide the substate context information.
```
cwgoes: defines a substate? I think you just mean "defines a store prefix", right?

spec/ics-023-vector-commitments/README.md
```diff
@@ -59,6 +59,20 @@ In certain commitment constructions with constant-size states, `CommitmentState`
...
+A `CommitmentPrefix` defines a substate of the commitment. It is passed to the proof verification functions with the path to provide the substate context information. 
+
+```typescript
+type CommitmentPrefix = object
+```
+
+The function `applyPrefix` constructs a new path from the arguments. It interprets the path argument in the context of the prefix argument. For two `(prefix, path)` tuples, `applyPrefix(prefix, path)` MUST return the same key only if the tuple elements are equal.
+
+```typescript
+type applyPrefix = (prefix: CommitmentPrefix, path: Path) => Path
```
cwgoes: What is this defined by - a commitment type, a state machine, something else?

spec/ics-023-vector-commitments/README.md
cwgoes: I don't really know what "substate" means in this sentence.

spec/ics-024-host-requirements/README.md
```diff
@@ -149,6 +150,32 @@ Host state machines MUST provide the ability to introspect this stored recent co
...
+  prefixedPath = applyPrefix(getCommitmentPrefix(), path)
+  if value != nil {
+    assert(verifyMembership(root, createMembershipProof(state, prefixedPath, value), prefixedPath, value))
+  } else {
+    assert(verifyNonMembership(root, createNonMembershipProof(state, prefixedPath), prefixedPath))
+  }
+}
+```
+
+For a key-value store, the `getCommitmentPrefix` RECOMMENDED to be constant.
```
cwgoes: How could it not be constant - what would that mean? Wouldn't the handshake fail then?

spec/ics-024-host-requirements/README.md
```diff
@@ -149,6 +150,32 @@ Host state machines MUST provide the ability to introspect this stored recent co
...
 
+### Commitment Path Introspection
+
+Host chains MUST provide the ability to inspect their commitment path, with `getCommitmentPrefix`:
+
+```
+type getCommitmentPrefix = () => CommitmentPrefix
+```
+
+If there is no substate used within the host chain, `getCommitmentPrefix` can be defined as identity function.
```
cwgoes: Again, unsure what substate means here

spec/ics-024-host-requirements/README.md
```diff
@@ -149,6 +150,32 @@ Host state machines MUST provide the ability to introspect this stored recent co
...
+
+The result `CommitmentPrefix` is the definition of the key-value store's substate definition.
```
cwgoes: Awkward phrasing "is the definition (...) of the definition"

spec/ics-024-host-requirements/README.md
mossid: For example, the host state machine is running in a consensus machine with prefix of `"ibc/"`. During the lifetime of the state machine, it should not be changed, which means a value set at height `1000` on the (sub)key `"connection/connectionid"` should stay under the full key of `"ibc/connection/connectionid"`. If the `getCommitmentPrefix` is not constant and can be changed, this means the prefix can be modified over time, which modifies the full key to something else; `"ibc-module/connection/connectionid"`.

cwgoes:
Also needs a rebase now (sorry).

spec/ics-024-host-requirements/README.md
mossid: Put `RECOMMENDED` considering the forward compatibility of state migration during the chain upgrade / module version upgrade(which may change the commitment prefix).

cwgoes: If it changes the commitment prefixes, previous connections will no longer work.

cwgoes: If you're talking about upgrades we should discuss that explicitly.

spec/ics-003-connection-semantics/README.md
mossid: Yes `applyPrefix` can be a constant function.

spec/ics-003-connection-semantics/README.md
```diff
@@ -134,6 +137,33 @@ function removeConnectionFromClient(
 }
 ```
 
+Two helper functions are defined to provide automatic application of `CommitmentPrefix`. In the other parts of the specifications,
+these functions should be used for introspecting other chains' state, instead of directly calling the `verifyMembership` with the clients.
```
cwgoes:
```suggestion
these functions MUST be used for introspecting other chains' state, instead of directly calling the `verifyMembership` or `verifyNonMembership` function on the client.
```

mossid:
Also modified `type Path = string` => `type Path = object`

spec/ics-024-host-requirements/README.md
```diff
@@ -149,6 +150,32 @@ Host state machines MUST provide the ability to introspect this stored recent co
...
+    proof = createMembershipProof(state, prefixedPath, value)
+    assert(verifyMembership(root, proof, prefixedPath, value))
+  } else {
+    proof = createNonMembershipProof(state, prefixedPath)
+    assert(verifyNonMembership(root, proof, prefixedPath))
+  }
+}
+```
+
+For a host state machine, the `getCommitmentPrefix` MUST be constant.
```
cwgoes:
```suggestion
For a host state machine, the return value of `getCommitmentPrefix` MUST be constant.
```

mossid:
Per discussion: `Path` type should be separated into two types, one for local state machine accessing on the key-value store, which is a byte string, and one for remote state machine verifying on the key-value store, which is an object.

PR #219, opened 2019-08-23
Support loopback; related changes

cwgoes:
Closes https://github.com/cosmos/ics/issues/151


cwgoes:
- revise subsequent ICSs to use it
- add loopback example (as a client)
- ensure performance penalties are not substantial

PR #220, opened 2019-08-24
Update 2_IBC_DESIGN_PRINCIPLES.md

en:
fix typo


cwgoes:
What is the typo? "Byzantine fraud" is intentional (though maybe we should define that term).

en:
Gotcha, closing ...

PR #224, opened 2019-08-24
Add/update various diagrams

cwgoes:
Closes https://github.com/cosmos/ics/issues/184
Closes https://github.com/cosmos/ics/issues/168

cc @ancazamfir


cwgoes:
...not perfect, but hopefully these will do for now.

PR #271, opened 2019-09-18
ICS 28: Direct Interchain Exchange protocol

zmanian:
Here is a concept for Direct Interchain Exchange that Tendermint/All in Bits is pursuing. 

It was co-developed by @zmanian and @jaekwon.

### DIFF
--- a/spec/ics-027-direct-interchain-exchange/ReadMe.md
+++ b/spec/ics-027-direct-interchain-exchange/ReadMe.md
@@ -11,7 +11,7 @@ Exchange consists of three core components.
 2. Order matching
 3. Execution and delivery
 
-A DEX chain like Binance combines these 3 functions into a single synchronous environment. This comes with risks including asset peg risks, and trusting the validator set of faithfully execute your order.	
+A DEX chain like Binance combines these 3 functions into a single synchronous environment. This comes with risks including asset peg risks, and trusting the validator set of faithfully executing your order.	
 
 We propose an alternative model.  In this model, each of the 3 phases is asynchronously executed on independent blockchains.
 
--- a/spec/ics-027-direct-interchain-exchange/ReadMe.md
+++ b/spec/ics-027-direct-interchain-exchange/ReadMe.md
@@ -1,0 +1,74 @@
+# Cosmos Direct Interchain Exchange Protocol (DIEP)
+
+
+#### Core Idea:
+
+Asynchronous liquidity available with IBC that is resistant to theft by misbehaving validators. This eliminates the need to concentrate liquidity (limit orders) in a third party blockchain/venue. Price discovery and execution becomes native between Cosmos zones (e.g. N^2 IBC connections for N zones). There is a cost in terms of higher latency from this architecture.
+
+Exchange consists of three core components.
+
+1. Commitment of Funds for an order
+2. Order matching
+3. Execution and delivery
+
+A DEX chain like Binance combines these 3 functions into a single synchronous environment. This comes with risks including asset peg risks, and trusting the validator set of faithfully execute your order.	
+
+We propose an alternative model.  In this model, each of the 3 phases is asynchronously executed on independent blockchains.
+
+#### Cast of Characters
+
+Alice has Foo tokens on FooChain.
+
+Bob has Bar tokens on BarChain.
+
+FooChain and BarChain are connected via IBC.
+
+There exists a Hub chain.  FooChain and BarChain are also connected via IBC to the HubChain.
+
+Alice perceives the security of FooChain and Hub chain to be *strong*. 
+Alice perceive the security of BarChain to be /moderate/.
+Or, Alice perceives the security of both FooChain and BarChain to be *strong*.
+
+Bob perceives the security of Bar Chain and Hub chain to be *strong*.
+Bob perceive the security of Foo Chain to be /moderate/.
+Or, Bob perceives the security of both FooChain and BarChain to be *strong*.
+
+#### Protocol
+
+For simplicity, we will describe only one scenario, where Alice wishes to submit a limit order to purchase Bar tokens with her Foo tokens, where the desired price is lower than market (e.g. the limit order will not be matched immediately).
+
+Alice submits to FooChain a limit order transaction on FooChain: {type:"limit", pair:"bar/foo", sell:"20foo", price:"0.9"}
+
+FooChain sees that the price of "0.9" is lower than the current price of "1", so it locks Alice's 20 Foo tokens, and sends this order to BarChain via IBC.
+
+This order persists on FooChain (in the form of locked tokens) and BarChain (in the form of an open limit order) until one of two things happen.
+
+Case A: Alice cancels the order by submitting a cancellation order on BarChain: {type:"limit-cancel", pair:"bar/foo", sell:"20foo", account:"Alice"}.
+Case B: Bob (say) partially or fully matches the order by submitting a limit order on BarChain: {type:"limit", pair:"bar/foo", sell:"100bar", price:"0.8"}.
+
+There are ways to design the protocol such that Alice can cancel the order without having to submit an order to BarChain, but due to the asynchrony of IBC it cannot be immediate for both Alice and Bob simultaneously.  Such extensions to the interchain exchange protocol are not described here.
+
+For both case A and case B, a corresponding opposite IBC message is created from BarChain to FooChain.  In case A, the 20 Foo tokens are released back to Alice.  In case B, the 20 Foo tokens are released back to Bob, and in this example (since Bob's order volume was greater than Alice's), a new partial limit order would be created on FooChain on behalf of Bob.
+
+#### Delivery
+
+While the delivery of matched orders may be routed anywhere, by default we propose that the delivery happens on the Hub as IBC pegged tokens.
+It simplifies custody for Alice and Bob because of common custodial protocols on the Hub.
+It creates a substantial social pressure toward the economic finality of the exchange of tokens by having them delivered to Hub accounts.
+Only delivered tokens need to be on the Hub, not the tokens behind pending limit orders, so as long as users move their funds out of the Hub soon after execution, the security requirement on the hub is minimal.
+
+#### Hub Collateralization
+
+With N^2 IBC connections between N zones for this Direct Interchain Exchange Protocol (DIEP), it becomes difficult for traders and market makers to make sense of the complexity arising from chain security.
+Additionally, by forcing a chain to fail with a double spend attack (a consensus fork), it may become profitable to induce failure, especially with more complex scenarios that require consistency across the interchain.  For example, interchain pair trading may become profitable to attack.
+
+To mitigate such attacks and guarantee more quantifiable skin in the game for the validators, the Hub should play the role of interchain collateralization.  All zones in the DIEP network should connect via IBC to the Hub, and the Hub should be modified to allow staking on behalf of these zones.  ATOM tokens staked on behalf of other zones should not be subject to the full inflationary ATOM tax (e.g. these staked zones may earn a proportionate amount of inflationary ATOMs, less any fees).  And the Hub becomes the arbiter for deciding whether a zone is consistent, or whether it had been double-spend attacked.  Any slashed ATOMs may be used to compensate the victims of an attack, although some ATOMs should probably always be burned.
+
+#### Disadvantages
+
+Alice wants to simultaneously place LimitOrders on FooChain to be taken on BarChain and take orders on Foochain. This requires 2x the value to be locked up by Alice because of asynchrony.
+
+
+#### Prior Art
+
+Agoric’s electronic rights transfer protocol and ZOE exchange protocol bear a pretty strong resemblance to this protocol based on informal descriptions. The Zoe protocol has been informally described by the Agoric team but there isn’t any public documentation I am aware of.


spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
+# Cosmos Direct Interchain Exchange Protocol (DIEP)
+
+
+#### Core Idea:
+
+Asynchronous liquidity available with IBC that is resistant to theft by misbehaving validators. This eliminates the need to concentrate liquidity (limit orders) in a third party blockchain/venue. Price discovery and execution becomes native between Cosmos zones (e.g. N^2 IBC connections for N zones). There is a cost in terms of higher latency from this architecture.
+
+Exchange consists of three core components.
+
+1. Commitment of Funds for an order
```
rigelrozanski:
```suggestion
1. Commitment of funds for an order
```

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
+# Cosmos Direct Interchain Exchange Protocol (DIEP)
```
rigelrozanski: if only there was a way we could make this acronym DIPER... `Direct Interchain Protocol for Exchange and Relay` 👍 👶

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+#### Core Idea:
+
+Asynchronous liquidity available with IBC that is resistant to theft by misbehaving validators. This eliminates the need to concentrate liquidity (limit orders) in a third party blockchain/venue. Price discovery and execution becomes native between Cosmos zones (e.g. N^2 IBC connections for N zones). There is a cost in terms of higher latency from this architecture.
```
rigelrozanski: Okay so is this _also_ meant to support shared staking cross chain? Aka validator-A on chain-A can use funds locked for staking on Chain-A to act as staking tokens to also validate on Chain-B? If so we should more explicitly mention this in this initial paragraph

_________________________

edit: it seems as though this is not specifically intended, but I think there may be a way to have share some fundamental infrastructure of this `DIPER` protocol and the shared-collateralization.

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+
+For both case A and case B, a corresponding opposite IBC message is created from BarChain to FooChain.  In case A, the 20 Foo tokens are released back to Alice.  In case B, the 20 Foo tokens are released back to Bob, and in this example (since Bob's order volume was greater than Alice's), a new partial limit order would be created on FooChain on behalf of Bob.
+
+#### Delivery
+
+While the delivery of matched orders may be routed anywhere, by default we propose that the delivery happens on the Hub as IBC pegged tokens.
+It simplifies custody for Alice and Bob because of common custodial protocols on the Hub.
+It creates a substantial social pressure toward the economic finality of the exchange of tokens by having them delivered to Hub accounts.
+Only delivered tokens need to be on the Hub, not the tokens behind pending limit orders, so as long as users move their funds out of the Hub soon after execution, the security requirement on the hub is minimal.
```
rigelrozanski:
```suggestion
Only delivered tokens need to be on the Hub, not the tokens behind pending limit orders, so as long as users move their funds out of the Hub soon after execution, the security requirement on the Hub is minimal.
```

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+Additionally, by forcing a chain to fail with a double spend attack (a consensus fork), it may become profitable to induce failure, especially with more complex scenarios that require consistency across the interchain.  For example, interchain pair trading may become profitable to attack.
+
+To mitigate such attacks and guarantee more quantifiable skin in the game for the validators, the Hub should play the role of interchain collateralization.  All zones in the DIEP network should connect via IBC to the Hub, and the Hub should be modified to allow staking on behalf of these zones.  ATOM tokens staked on behalf of other zones should not be subject to the full inflationary ATOM tax (e.g. these staked zones may earn a proportionate amount of inflationary ATOMs, less any fees).  And the Hub becomes the arbiter for deciding whether a zone is consistent, or whether it had been double-spend attacked.  Any slashed ATOMs may be used to compensate the victims of an attack, although some ATOMs should probably always be burned.
```
rigelrozanski: I'm a bit confused, so in this model an _entire zone_ can actually be staked on the Hub? Or simply a validator _on_ another zone can also act individually as a sort of "intermediary" validator? (this should be more explicitly worded within this paragraph). 

If you were referring to an entire zone being able to stake with the hub it introduces some potentially unnecessary complexity because now those inflationary rewards provided to the "zone-validator" must be redistributed somehow within the zone (base on whoever fronted the original atoms there).

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+
+While the delivery of matched orders may be routed anywhere, by default we propose that the delivery happens on the Hub as IBC pegged tokens.
+It simplifies custody for Alice and Bob because of common custodial protocols on the Hub.
+It creates a substantial social pressure toward the economic finality of the exchange of tokens by having them delivered to Hub accounts.
+Only delivered tokens need to be on the Hub, not the tokens behind pending limit orders, so as long as users move their funds out of the Hub soon after execution, the security requirement on the hub is minimal.
+
+#### Hub Collateralization
+
+With N^2 IBC connections between N zones for this Direct Interchain Exchange Protocol (DIEP), it becomes difficult for traders and market makers to make sense of the complexity arising from chain security.
+Additionally, by forcing a chain to fail with a double spend attack (a consensus fork), it may become profitable to induce failure, especially with more complex scenarios that require consistency across the interchain.  For example, interchain pair trading may become profitable to attack.
```
rigelrozanski: I'd like to see a bullet point example of how a double spend attack would happen (and affect the hub-collateral) in this cross-chain context.

spec/ics-027-direct-interchain-exchange/ReadMe.md
melekes: entire zone can actually be staked on the Hub? Or simply a validator on another zone can also act individually as a sort of "intermediary" validator?

too much risk for 1 validator imo. probably some percentage of validators (say 2/3 of all stake) is needed.

spec/ics-027-direct-interchain-exchange/ReadMe.md
cwgoes: What does theft of liquidity mean - theft of funds, or just withholding liquidity from some other chain?

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+
+To mitigate such attacks and guarantee more quantifiable skin in the game for the validators, the Hub should play the role of interchain collateralization.  All zones in the DIEP network should connect via IBC to the Hub, and the Hub should be modified to allow staking on behalf of these zones.  ATOM tokens staked on behalf of other zones should not be subject to the full inflationary ATOM tax (e.g. these staked zones may earn a proportionate amount of inflationary ATOMs, less any fees).  And the Hub becomes the arbiter for deciding whether a zone is consistent, or whether it had been double-spend attacked.  Any slashed ATOMs may be used to compensate the victims of an attack, although some ATOMs should probably always be burned.
+
+#### Disadvantages
+
+Alice wants to simultaneously place LimitOrders on FooChain to be taken on BarChain and take orders on Foochain. This requires 2x the value to be locked up by Alice because of asynchrony.
+
```
marbar3778: Another disadvantage is high latency, I think this is the biggest disadvantage to a trader. If we aren't talking about traders but daily app users then they may not care.

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+Bob perceive the security of Foo Chain to be /moderate/.
+Or, Bob perceives the security of both FooChain and BarChain to be *strong*.
+
+#### Protocol
+
+For simplicity, we will describe only one scenario, where Alice wishes to submit a limit order to purchase Bar tokens with her Foo tokens, where the desired price is lower than market (e.g. the limit order will not be matched immediately).
+
+Alice submits to FooChain a limit order transaction on FooChain: {type:"limit", pair:"bar/foo", sell:"20foo", price:"0.9"}
+
+FooChain sees that the price of "0.9" is lower than the current price of "1", so it locks Alice's 20 Foo tokens, and sends this order to BarChain via IBC.
```
cwgoes: The price on BarChain? Where is this price stored, and what if FooChain's data is out of date?

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+
+This order persists on FooChain (in the form of locked tokens) and BarChain (in the form of an open limit order) until one of two things happen.
+
+Case A: Alice cancels the order by submitting a cancellation order on BarChain: {type:"limit-cancel", pair:"bar/foo", sell:"20foo", account:"Alice"}.
+Case B: Bob (say) partially or fully matches the order by submitting a limit order on BarChain: {type:"limit", pair:"bar/foo", sell:"100bar", price:"0.8"}.
+
+There are ways to design the protocol such that Alice can cancel the order without having to submit an order to BarChain, but due to the asynchrony of IBC it cannot be immediate for both Alice and Bob simultaneously.  Such extensions to the interchain exchange protocol are not described here.
```
cwgoes: Without having to submit a cancellation?

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+
+For both case A and case B, a corresponding opposite IBC message is created from BarChain to FooChain.  In case A, the 20 Foo tokens are released back to Alice.  In case B, the 20 Foo tokens are released back to Bob, and in this example (since Bob's order volume was greater than Alice's), a new partial limit order would be created on FooChain on behalf of Bob.
+
+#### Delivery
+
+While the delivery of matched orders may be routed anywhere, by default we propose that the delivery happens on the Hub as IBC pegged tokens.
```
cwgoes: Hmm I don't totally understand this section. Neither Alice nor Bob transferred tokens to the Hub - does the order being settled cause tokens to be transferred to Alice's account on the Hub? Why would she prefer that over her account on FooChain where she had the tokens in the first place?

I guess which chain/account tokens were sent to could just be an option in the order, since Bob doesn't care.

spec/ics-027-direct-interchain-exchange/ReadMe.md
cwgoes: Is there a necessary connection between interchain collateralization and this exchange protocol, or is this just a suggestion for increased security?

cwgoes:
Note - if this is to be an application-level specification, it should contain (eventually):
- A list of all packet types utilized
- Specific packet data structures, including field names & types
- Handler pseudocode that FooChain, BarChain, the Hub, etc. (all roles) must implement

spec/ics-027-direct-interchain-exchange/ReadMe.md
```diff
@@ -0,0 +1,74 @@
...
+
+Asynchronous liquidity available with IBC that is resistant to theft by misbehaving validators. This eliminates the need to concentrate liquidity (limit orders) in a third party blockchain/venue. Price discovery and execution becomes native between Cosmos zones (e.g. N^2 IBC connections for N zones). There is a cost in terms of higher latency from this architecture.
+
+Exchange consists of three core components.
+
+1. Commitment of Funds for an order
+2. Order matching
+3. Execution and delivery
+
+A DEX chain like Binance combines these 3 functions into a single synchronous environment. This comes with risks including asset peg risks, and trusting the validator set of faithfully execute your order.
```
okwme:
```suggestion
A DEX chain like Binance combines these 3 functions into a single synchronous environment. This comes with risks including asset peg risks, and trusting the validator set of faithfully executing your order.	
```

spec/ics-027-direct-interchain-exchange/ReadMe.md
rigelrozanski: Interesting, this is semantics, but we should probably design the protocol in a versatile enough way such that directly or indirectly (through the hub) tokens may be exchanged using the protocol.

spec/ics-027-direct-interchain-exchange/ReadMe.md
rigelrozanski: AFAICT interchain collateralization would be required in order to maintain trustlessness?

cwgoes:
@zmanian Is this still something you and/or @jaekwon are interested in pursuing?

zmanian:
I feel like at very least this should be artifact in the main repo somewhere.

cwgoes:
I would like the requests in the above comment (https://github.com/cosmos/ics/pull/271#issuecomment-533058119) addressed before merging this as a specification draft. @brentxu are you interested in working on those? Otherwise I can merge this in a "ideation" folder for now (or we can wait & leave it as an open PR).

zmanian:
I support merging to `ideation`

cwgoes:
Done in https://github.com/cosmos/ics/pull/347.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: > Two cents: is it fare to call a protocol **direct** in the presence of a Hub (- 3rd party)?

I agree. It is relatively indirect than original solution via direct IBC connection between those two chains.

In this context, it is more like "Cosmos Centralized Interchain Exchange Protocol". People might hate the word "centralized" but it is what this doc proposing I think.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: I think limit orders are not the only component of "liquidity". Custody, matching, settlement are necessary components to handle liquidity. And I think this document suggests huge roles on CosmosHub to provide such important components. These entire set can be called "Exchange".

So I think this doc is about concentrate important roles on route of "liquidity handling, aka exchange" into CosmosHub.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: > If you were referring to an entire zone being able to stake with the hub it introduces some potentially unnecessary complexity because now those inflationary rewards provided to the "zone-validator" must be redistributed somehow within the zone (base on whoever fronted the original atoms there).

yes. I call it "IBC cross-chain distribution protocol". The interchain collaterization will definitely need this additional protocol to transfer rewards from zone to the hub. Definitely "a validator" case is not intended in this context I guess. As melekes said, it is not reasonable to do so from security-wise perspective.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: daily app users also will care because the high latency will lead to deserted liquidity.

algorithm traders always search for any exchange to exploit slow response on price finding. slow exchanges are exposed to be attacked by these entities, making market maker lose a lot of money because of these attackers. 

this extremely effective risk on market makers will result in almost no market maker present in the exchange, which will definitely block the bootstrapping of the liquidity growth.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: i agree that this analogy is inaccurate. there exists no unique "price" in multi-exchange environment. 

also, matching on blockchains are not continues but descrete process. therefore, whether the order is immediately matched or not cannot be decided by recent price.

so, IBC should always assume "no-immediate-matching" condition from order step. the protocol does not know whether it is immediately matched or not until actual matching happens.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: I guess something like "timeout", as an example.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: I think what this doc is suggesting is CosmosHub as a `middleman` to execute "atomic swap" of a confirmed matching. 

I guess technically atomic swap between different chains cannot be syncronous because each transfer are not on one chain. 

By moving this process into one chain, CosmosHub, the atomic swap can be syncronous and also much more trustable, considering provided security on the CosmosHub.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: Every solution is better with higher security by interchain collateralization. So, I think interchain collateralization is an orthogonal topic in this context.

spec/ics-027-direct-interchain-exchange/ReadMe.md
dlguddus: I think the writer is explaining that there might exist high probability of IBC malfunctioning when asset control is too decentralized into so many zones. 

if there exists a bullet point example, it means there exists a bug in IBC. Point is that we will have more risk to experience IBC bug in more complicated design.

Issue #145, opened 2019-06-30
Implementation side Feedback


mossid:
This issue is for documenting the difference between the spec and the current implementation. Discussion will be followed to determine which part we want to merge back to the spec and which part we don't want.

NOTE: the current implementation code is based on pre-1.0.0 spec

## ICS 23

The implementation can be found here: https://github.com/cosmos/cosmos-sdk/pull/4515

1. `verifyMembership` / `verifyNonmembership` distinction

In the spec, the functions are defined separately and takes value only if it is `verifyMembership`. In the implementation, there is one method `Verify` and it takes `nil` as its value if it is an absent proof, which means there is no type distinction between membership proofs and non membership proofs. This approach is easier on implementation(as `nil` means empty value on `KVStore.Get()` too), but also they are semantically equivalent and type distinction in spec is better for its purpose, I suggest to not merge back to spec.

2. `verify*` takes key as argument

`verify*` functions takes both key and value as argument in order to prove the proof. In the implementation, in order to store the proofs in a map, the proofs already contains its key and the `Verify()` function takes only the value. Like as above, this is also equivalent to the spec. 

3. Add `Path` to the types

`Path`s are a prefix distinguisher for the keys. In many cases IBC logic will not be able to occupy the key format that is declared in the specification. For example in Cosmos-SDK case the merkle proof will be prefixed with a `RootMultistore` proof. 

`Path` can be understood as a prefix on keys. For a single KVStore with multiple IBC modules, each module can be stored under prefix `"IBC{index}"`, preventing collision of the keys. The counterparties of the modules will store `Path` with `"IBC{index}"`, which will be prepended to the specified keys.

The `verify*` functions now also have to take `Path` in addition to the `Root`, key, value. The "kind" of `Proof` and `Path` should match(`merkle.Proof` should only be used with `merkle.Path`).

The `Path` should be, I think, unique per connection. If we store `Path` per client, this will make the light clients have more than one purpose(verifying the counterparty consensus).  ICS 02 Client should be only working as a light client for other chains, without storing any other information.

## ICS 02

The implementation can be found here: https://github.com/cosmos/cosmos-sdk/pull/4516

1. May rename client* => verifier*?

The name "client" sometimes conflicts with the client side command line tools, not an importand suggestion.

## ICS 03

The implementation can be found here: https://github.com/cosmos/cosmos-sdk/pull/4517

1. Separation between connection elements

type `ConnectionEnd` is defined as

```typescript
interface ConnectionEnd {
[...]

PR #273, opened 2019-09-19
Remaining rc3 feedback in preparation for rc4

### LINKED ISSUE #258: Seperate Path for accessing and verifying
Per discussion: `Path` type should be separated into two types, one for local state machine accessing on the key-value store, which is a byte string, and one for remote state machine verifying on the key-value store, which is an object.

_Originally posted by @mossid in https://github.com/cosmos/ics/pull/189#issuecomment-530109574_

### LINKED ISSUE #269: Clarify, structure versioning mechanism
- Set of acceptable versions provided by handshake initiator
- Specific version to use for connection/channel provided by handshake receiver
- Version confirmed by handshake initiator

Open question: should channel versions be protocol-level, application-level, or both?

cwgoes:
Ref #145 
Closes #258 
Closes #269

### DIFF
--- a/misc/aspell_dict
+++ b/misc/aspell_dict
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 448 
+personal_ws-1.1 en 449 
 unescrow
 onChanCloseConfirm
 clientKey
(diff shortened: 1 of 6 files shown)


cwgoes:
Remaining bit: https://github.com/cosmos/ics/issues/145#issuecomment-533500547.

Issue #251, opened 2019-09-05
ICS ?: Interchain account


Thunnini:
# Interchain account (draft)

I made some revisions to the proposal to include more specificities regarding the proposal #211 .

In summary, here are the core changes:

- Revised the naming from "IBC managed account" to "Interchain account".
- Added a rough concept for determining interchain account's address.
- Revised so that the IBC packet includes the transaction raw bytes for the receiving chain. The receiving chain must decode these bytes to a transaction, then check that the sending chain has correct permission. If the sending chain has the correct permission, the receiving chain will execute the transaction. Finally, the receiving chain will return result packet to the sending chain.
    - This process will allow interchain accounts to be able to execute arbitrary transactions without additional consideration for the transaction's corresponding module.
    - After writing some [pseudocode for cosmos-sdk](https://github.com/Thunnini/interchain-account), I think that this can be achieved fairly easily without making the Cosmos-sdk's code more complex thanks to Cosmos-SDK's module and router system.
- Some use cases I can see with the implementation of the interchain account are as follows:
    - The use of interchain staking functions
    - Creation of The DAO-like zone that is able to participate and manages multiple hub/zone's governance and staking position.
    - Possibility of a decentralized CDP position (i.e. Kava) where the CDP position is managed by another zone to create a new derivative DeFi product.
    - A protocol that controls arbitrage in DeFi zones such as Terra / Kava and distributes profits to stakeholders in exchange for their capital.
    - And a host of other decentralized financial use-cases.

    ---

ics: (?)
title: (Interchain account)
stage: (Strawman)
category: (ibc-app)
author: ([yunjh1994@ellipti.io](mailto:yunjh1994@ellipti.io))
created: (2019-08-21)
modified: (2019-08-21)

## **Synopsis**

This standard document specifies packet data structure, state machine handling logic, and encoding details for the account management system over an IBC channel between separate chains.

### **Motivation**

On Ethereum, there are two types of accounts: externally owned accounts, controlled by private keys, and contract accounts, controlled by their contract code [[ref](https://github.com/ethereum/wiki/wiki/White-Paper)]. Similar to Ethereum's CA(contract accounts), Interchain accounts are managed by another chain(zone) while retaining all the capabilities of a normal account (i.e. stake, send, vote, etc). While an Ethereum CA's contract logic is performed within Ethereum's EVM, Interchain accounts are managed by another chain via IBC in a trustless way.

### **Definitions**

[...]

PR #278, opened 2019-09-26
ICS 027: Interchain account

Thunnini:
From #251 

## Changes
 - Use acknowledge packet instead of result packet.
 - Module saves the interchain account's addresses in its state.


spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+On Ethereum, there are two types of accounts: externally owned accounts, controlled by private keys, and contract accounts, controlled by their contract code [[ref](https://github.com/ethereum/wiki/wiki/White-Paper)]. Similar to Ethereum's CA(contract accounts), Interchain accounts are managed by another chain(zone) while retaining all the capabilities of a normal account (i.e. stake, send, vote, etc). While an Ethereum CA's contract logic is performed within Ethereum's EVM, Interchain accounts are managed by another chain via IBC in a trustless way.
+
+### **Definitions**
+
+The IBC handler interface & IBC relayer module interface are as defined in [ICS 25](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-025-handler-interface) and [ICS 26](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-026-relayer-module), respectively.
+
+### **Desired Properties**
+
+- Permissionless
+- Fault containment: Interchain account must follow rules of its dwelling chain, even in times of byzantine behavior by the counterparty chain (the chain that manages the account)
```
mossid: :+1:

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+    RegisterIBCAccountPacketData data = packet.data
+      
+    path = "{packet.sourcePort}/{packet.sourceChannel}"
+    address = sha256(path + packet.salt)
+      
+    // Should not block even if there is normal account,
+    // because attackers can distrupt to create an ibc managed account
+    // by sending some assets to estimated address in advance.
+    // And IBC managed account has no public key, but its sequence is 1.
+    // It can be mark for Interchain account, becuase normal account can't be sequence 1 without publish public key.
```
mossid: This assumption seems a bit Cosmos specific. Can we change `address = sha256(path + packet.salt)` to arbitrary address generation logic, so the receiving chain can generate a non-conflicting address? In Cosmos, the receiving chain can generate a new address with that hashing logic, with different bech32 prefix, and returns it, and in other chains, for example in Ethereum, the receiving chain can generate a new contract that works as a proxy for the origin chain and returns its address.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+    account = accountKeeper.getAccount(account)
+    if (account != null) {
+      abortTransactionUnless(account.sequence === 1 && account.pubKey == null)
```
mossid: The `sequence` an `pubkey` field of the account is Cosmos specific

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+    } else {
+      accountKeeper.newAccount(address)
+    }
+      
+    addressesRegisteredChannel[signer] = path
+        
+    // set account's sequence to 1
+    accountKeeper.setAccount(address, {
```
mossid: Also `accountKeeper` seems Cosmos specific

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+      
+    path = "{packet.sourcePort}/{packet.sourceChannel}"
+    address = sha256(path + packet.salt)
+      
+    // Should not block even if there is normal account,
+    // because attackers can distrupt to create an ibc managed account
+    // by sending some assets to estimated address in advance.
+    // And IBC managed account has no public key, but its sequence is 1.
+    // It can be mark for Interchain account, becuase normal account can't be sequence 1 without publish public key.
+    account = accountKeeper.getAccount(account)
```
mossid: Also `accountKeeper` seems Cosmos specific. As an alternative you declare an interface named `accountKeeper` with a set of functions with some constraints(which will be inspired from Cosmos' `accountKeeper` design, but works chain agnostic)

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+    addressesRegisteredChannel[signer] = path
+        
+    // set account's sequence to 1
+    accountKeeper.setAccount(address, {
+      ...,
+      sequence: 1,
+      ...,
+    })
+      
+    // Return acknowledgement byte as 0x0 if account creation succeeds.
```
mossid: If the address generation works arbitrary(as commented above) the return ack should include generated address

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+  } 
+    
+  return 0x
+}
+```
+```typescript
+function onAcknowledgePacket(
+  packet: Packet,
+  acknowledgement: bytes) {
+  // Receiving chain should handle this event as if the tx in packet has failed if ack bytes is not 0x0.
```
mossid: If the address generation works arbitrary(as commented above) the ack receiving should store the controlling interchain accounts(or send the address information to the requested entity for the account generation)

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+      
+    path = "{packet.sourcePort}/{packet.sourceChannel}"
+      
+    // Check tx has right permissions
+    for (const signer of signers){
+      abortTransactionUnless(path === addressesRegisteredChannel[signer])
+    }
+    
+    // Get handler from router
+    handler = router(msg.route())
```
mossid: `router` and `type msg` should be defined somewhere, however I am in favor of using router/handler here

spec/ics-027-interchain-account/README.md
cwgoes: Swap "dwelling" for "host"

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+### **Definitions**
+
+The IBC handler interface & IBC relayer module interface are as defined in [ICS 25](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-025-handler-interface) and [ICS 26](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-026-relayer-module), respectively.
+
+### **Desired Properties**
+
+- Permissionless
+- Fault containment: Interchain account must follow rules of its dwelling chain, even in times of byzantine behavior by the counterparty chain (the chain that manages the account)
+- The chain(that controls the account) requesting a specific transaction from must process the results asynchronously and according to the chain's logic.
```
cwgoes: requesting a specific transaction from where or what?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+`RegisterIBCAccountPacketData` is used for counterparty chain to register an account. Interchain account's address is defined deterministically with channel path and salt. Process of determining address is influenced by [EIP-1014](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1014.md). It allows interactions to be made with addresses that do not exist yet on-chain. These specificities will allow more offchain services to build on top of the data structure.
+
+```typescript
+interface RegisterIBCAccountPacketData {
+  salt: Uint8Array
+}
+```
+
+`RunTxPacketData` is used to run tx for interchain account. Tx bytes is dependent on app, and it is formed with minimal data set except data for validating signatures.
```
cwgoes: You mean without the signatures, right?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+}
+```
+
+The interchain acccount bridge module tracks the addresses that have been registered on particular ports/channels in state. Path is a string which is {path_identifier/channel_identifier}. Fields of the ModuleState are assumed to be in scope.
+
+```typescript
+type Path = string
+
+interface ModuleState {
+    addressesRegisteredChannel: Map<string, Path>
```
cwgoes: Should `Address` be a declared type (instead of a string)?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+### **Packet relay**
+
+In plain English, between chains `A` and `B`. It will describe only the case that chain A wants to register an Interchain account on chain B and control it. Moreover, this system can also be applied the other way around. In these set of examples, we follow Cosmos-SDK's modularity rule and its `msg` system.
+```typescript
+interface ChainAccountTx {
+  msgs: Msg[]
+}
+```
+```typescript    
+function onSendPacket(packet: Packet) {
```
cwgoes: This function no longer exists, see the latest version of ICS 20.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+```typescript
+function onChanCloseConfirm(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier) {
+  // no action necessary
+}
+```
+### **Packet relay**
+
+In plain English, between chains `A` and `B`. It will describe only the case that chain A wants to register an Interchain account on chain B and control it. Moreover, this system can also be applied the other way around. In these set of examples, we follow Cosmos-SDK's modularity rule and its `msg` system.
```
cwgoes: I am not a huge fan of using Cosmos SDK specific abstractions. Can we just leave it at the level of "tx bytes"?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+function onSendPacket(packet: Packet) {  
+  // Sending chain can do preliminary work (i.e. create a log of transactions sent) if needed.
+}
+```
+```typescript
+function onRecvPacket(packet: Packet): bytes {
+  if (packet.data is RunTxPacketData) {
+    ChainAccountTx data = {}
+    // Decode transaction request
+    codec.unmarshal(packet.data.txBytes, &data)
```
cwgoes: You don't need to bother with this in pseudocode

spec/ics-027-interchain-account/README.md
cwgoes: Again, I think this is too Cosmos SDK specific. I think we should abstract what it means to "run a transaction", at least so that the Cosmos SDK specific bits can be an example implementation of the `runTx` function.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+        
+    // Execute handler with msg
+    result = handler(msg)
+        
+    if (result.code === 0) {
+        // Return acknowledgement byte as 0x0 if tx succeeds.
+        return 0x0
+    } else {
+        // Return acknowledgement byte as error code if tx fails.
+        return binary.littleEndian.encode(result.code)
```
cwgoes: No need to bother with encoding in pseudocode, just specify the packet formats.

spec/ics-027-interchain-account/README.md
cwgoes: Agreed, we should avoid Cosmos SDK specific assumptions.

spec/ics-027-interchain-account/README.md
cwgoes: Here is where I would like to see a `runTx` function.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+```
+```typescript
+function onTimeoutPacketClose(packet: Packet) {
+  // nothing is necessary
+}
+```
+
+## **Backwards Compatibility**
+
+(discussion of compatibility or lack thereof with previous standards)
```
cwgoes: If there is nothing here, just put `N/A`.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+## **Forwards Compatibility**
+
+(discussion of compatibility or lack thereof with expected future standards)
```
cwgoes: If there is nothing here, just put `N/A`.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+## **Example Implementation**
+
+(link to or description of concrete example implementation)
```
cwgoes: If there is nothing here, just put `N/A`.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+## **Other Implementations**
+
+(links to or descriptions of other implementations)
```
cwgoes: If there is nothing here, just put `N/A`.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+## **History**
+
+(changelog and notable inspirations / references)
```
cwgoes: Put a date and the first draft version.

spec/ics-027-interchain-account/README.md
Thunnini: Ok

spec/ics-027-interchain-account/README.md
Thunnini: Right. I think that application logic can know what chain requests transaction and whether that chain has the right permission. So, there is no need to include a signature to the interchain account.

spec/ics-027-interchain-account/README.md
Thunnini: Ok, I will recheck.

spec/ics-027-interchain-account/README.md
Thunnini: Ok

spec/ics-027-interchain-account/README.md
Thunnini: I agree. So, is it better to change the determining address login to function interface? For example, make function type`type CreateAddress = funtion(packet:Packet):Uint8Array` and use it for determining address. But, I think there is no need to consider bech32, because the address is just byte address and we can interact by byte address in the codebase.

spec/ics-027-interchain-account/README.md
Thunnini: I think that sequence (or nonce) and a public key are very general in the blockchain. So, I think using sequence and public key is not Cosmos specific. But, in the above discussion, I agree it is better to make account keeper and account as the interface so, I think, it is good to define the account interface that has the sequence (or nonce) and public key and use those variables from the interface.

spec/ics-027-interchain-account/README.md
Thunnini: I agree. I will try to abstract running tx and account keeper as interface.

spec/ics-027-interchain-account/README.md
Thunnini: I agree.

spec/ics-027-interchain-account/README.md
Thunnini: Ok, I'm trying to define interface for the host chain to handle counterparty chain's interchain account.

cwgoes:
> I'm not sure what the interchain multisignature accounts mean. But, I guess it means that the interchain multisignature accounts are handled by multiple chains. It will be interesting feature, however, I don't know how it can be used in a real environment. I think it is better to discuss that feature in post-launch.

Fair enough, probably better to keep the first version of this standard simple.

cwgoes:
Bump. If there's anything we can do on our end to help out here let me know. I really like this proposal.

Thunnini:
@cwgoes @mossid I removed the cosmos-sdk specific part in the specification and rewrote as much as possible based on a generalized interface and requirements.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+ics: 27  
+title: Interchain Account  
+stage: Draft  
+category: IBC/TAO  
+author: Tony Yun <yunjh1994@everett.zone>, Dogemos <josh@tendermint.com>  
+created: 2019-08-01  
+modified: 2019-11-08  
+---  
+
+## **Synopsis**
```
cwgoes: I don't think you need both the header & bold, it will format correctly in a Markdown render with just '##'. (this comment also applies to other headers in this document)

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+modified: 2019-11-08  
+---  
+
+## **Synopsis**  
+
+This standard document specifies packet data structure, state machine handling logic, and encoding details for the account management system over an IBC channel between separate chains.  
+
+### **Motivation**  
+
+On Ethereum, there are two types of accounts: externally owned accounts, controlled by private keys, and contract accounts, controlled by their contract code [[ref](https://github.com/ethereum/wiki/wiki/White-Paper)]. Similar to Ethereum's CA(contract accounts), Interchain accounts are managed by another chain while retaining all the capabilities of a normal account (i.e. stake, send, vote, etc). While an Ethereum CA's contract logic is performed within Ethereum's EVM, Interchain accounts are managed by another chain via IBC in a trustless way.
```
cwgoes:
```suggestion
On Ethereum, there are two types of accounts: externally owned accounts, controlled by private keys, and contract accounts, controlled by their contract code [[ref](https://github.com/ethereum/wiki/wiki/White-Paper)]. Similar to Ethereum's CA (contract accounts), interchain accounts are managed by another chain while retaining all the capabilities of a normal account (i.e. stake, send, vote, etc). While an Ethereum CA's contract logic is performed within Ethereum's EVM, interchain accounts are managed by another chain via IBC in a trustless way.  
```

cwgoes: Unfortunately, I think 'trustless' is so overutilised as to render the term meaningless - can we state something more specific, e.g. 'in a way such that the owner of the account retains full control over how it behaves'?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+On Ethereum, there are two types of accounts: externally owned accounts, controlled by private keys, and contract accounts, controlled by their contract code [[ref](https://github.com/ethereum/wiki/wiki/White-Paper)]. Similar to Ethereum's CA(contract accounts), Interchain accounts are managed by another chain while retaining all the capabilities of a normal account (i.e. stake, send, vote, etc). While an Ethereum CA's contract logic is performed within Ethereum's EVM, Interchain accounts are managed by another chain via IBC in a trustless way.  
+
+### **Definitions**  
+
+The IBC handler interface & IBC relayer module interface are as defined in [ICS 25](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-025-handler-interface) and [ICS 26](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-026-relayer-module), respectively.  
+
+### **Desired Properties**  
+
+- Permissionless  
+- Fault containment: Interchain account must follow rules of its host chain, even in times of byzantine behavior by the counterparty chain (the chain that manages the account)
```
cwgoes:
```suggestion
- Fault containment: Interchain account must follow rules of its host chain, even in times of Byzantine behavior by the counterparty chain (the chain that manages the account)  
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+### **Definitions**  
+
+The IBC handler interface & IBC relayer module interface are as defined in [ICS 25](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-025-handler-interface) and [ICS 26](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-026-relayer-module), respectively.  
+
+### **Desired Properties**  
+
+- Permissionless  
+- Fault containment: Interchain account must follow rules of its host chain, even in times of byzantine behavior by the counterparty chain (the chain that manages the account)  
+- The chain that controls the account must process the results asynchronously and according to the chain's logic.
```
cwgoes: Which results? Results of executing transactions?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+### **Desired Properties**  
+
+- Permissionless  
+- Fault containment: Interchain account must follow rules of its host chain, even in times of byzantine behavior by the counterparty chain (the chain that manages the account)  
+- The chain that controls the account must process the results asynchronously and according to the chain's logic.  
+- Sending and receiving transactions will be processed in an ordered manner.  
+
+## **Technical Specification**  
+
+The implementation of interchain account is non symmetric. This means that each chain must define the way to create a interchain account for counterparty chain deterministically.
```
cwgoes:
```suggestion
The implementation of interchain account is non-symmetric. This means that each chain must define the way to create a interchain account for counterparty chain deterministically.
```

cwgoes: Also I'm a little confused. Does different logic have to be defined for each counterparty chain? Do you mean that the transaction bytes will be different?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+Each chain must satisfy following features to create a interchain account:
+
+- New interchain account must not conflict with existing one.
```
cwgoes:
```suggestion
- New interchain accounts must not conflict with existing ones.
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+- Each chain must keep track of which counterparty chain created the new interchain account.
```
cwgoes:
```suggestion
- Each chain must keep track of which counterparty chain created each new interchain account.
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+Also, each chain must know how the counterparty chains serialize/deserialize transaction bytes in order to send transactions via IBC. And the counterparty chain must implement the process of safely exececuting IBC transactions by verifying the authority of the transaction's signers.
+
+The chain must reject the transaction and must not make a state transition in the following cases:
+
+- The IBC transaction fails to be deserialized.
+
```
cwgoes: (remove empty line)

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+- The IBC transaction requests for signers that was not made by counterparty chain.
```
cwgoes: for signatures, do you mean?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Data Structures
+
+Each chain must implement below interfaces to support interchain account. ```createOutgoingPacket``` method in ```IBCAccountModule``` interface defines the way how to create an outgoing packet for a specific type. Type indicates how IBC account transaction should be constructed and serialized for the host chain. Generally, type indicates what framework the host chain was built from.  
+```generateAddress``` defines the way how to determine the account's address by using identifier and salt. ```createAccount``` is used to create account with genereated address. New interchain account must not conflict with existing one, and chains should keep track of which counterparty chain created the new interchain account in order to verify the authority of transaction's singers in ```authenticateTx```.
+```authenticateTx``` validates a transaction and check the signers in the transaction have right permissions. ```runTx``` executes tx after the transaction was authenticated successfully.
+
+```typescript
+interface IBCAccountModule {
```
cwgoes: Nice abstraction.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+  acknowledgement: bytes) {  
+  if (packet.data is RegisterIBCAccountPacketData) {
+    if (acknowledgement !== 0x) {
+      identifier = "{packet/sourcePort}/{packet.sourceChannel}"
+      onAccountCreated(identifier, acknowledgement)
+    }
+  }
+  if (packet.data is RunTxPacketData) {
+    identifier = "{packet/destPort}/{packet.destChannel}"  
+    onTxSucceeded(identifier: Identifier, packet.data.tyBytes)
```
cwgoes:
```suggestion
    onTxSucceeded(identifier: Identifier, packet.data.txBytes)
```

cwgoes: Shouldn't you pass the acknowledgement? It might be an error.

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+  }
+}  
+```
+
+```typescript  
+function onTimeoutPacket(packet: Packet) {  
+  // Receiving chain should handle this event as if the tx in packet has failed  
+  if (packet.data is RunTxPacketData) {
+    identifier = "{packet/destPort}/{packet.destChannel}"  
+    onTxFailed(identifier: Identifier, packet.data.tyBytes)
```
cwgoes:
```suggestion
    onTxFailed(identifier: Identifier, packet.data.txBytes)
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+Not applicable.
+
+## **Forwards Compatibility**  
+
+Not applicable.
+
+## **Example Implementation**  
+
+Pseudocode for cosmos-sdk: https://github.com/everett-protocol/everett-hackathon/tree/master/x/interchain-account
```
cwgoes: It would also be helpful to reason about how this would work on Ethereum, which has a sufficiently programmable system that it should be able to implement this spec.

spec/ics-027-interchain-account/README.md
Thunnini: Yes, executed transactions will return the result code such as cosmos-sdk's `sdk.Result` type. I think `runTx` should return 0x if interchain tx is succeeded, or return an error code which is not 0x if interchain tx is failed.

spec/ics-027-interchain-account/README.md
Thunnini: It means that each chain can have a different way to deserialize transaction bytes and a different set of transactions that they can execute. For example, chains that use cosmos-sdk will deserialize tx bytes by amino, but if counterparty chain is a smart contract on Ethereum, it may deserialize tx bytes by ABI that is minimal serialize algorithm for smart contract. And a smart contract on Ethereum can't execute the staking, governance transaction if it doesn't support such feature. So, I think interchain account defines the general way to register interchain account and transfer tx bytes, and counterparty chain is responsible to deserialize and execute the tx bytes, and sending chain should know how counterparty chain handle tx bytes in advance.

spec/ics-027-interchain-account/README.md
cwgoes: Sounds good to me.

spec/ics-027-interchain-account/README.md
cwgoes: Makes sense, just include the above paragraph in the spec so that it's clear.

spec/ics-027-interchain-account/README.md
Thunnini: The reason why passes tx bytes is because it let a chain sending interchain transaction to know which transaction is succeeded or failed. But, I think it is needed to pass the acknowledgement if interchain transaction is failed because sending chain should know why transaction is failed. It was my mistake.

spec/ics-027-interchain-account/README.md
Thunnini: I think that it can be implemented in a smart contract level with using CALL, DELEGATECALL, CALLCODE opcode that are used to delegate call to some function in callee. I will research more and try to implement pseudocode in Solidity.

Thunnini:
I fixed the misspelling and added some more specific description.
And, I added the rough  way to implement the interchain account on Ethereum's smart contract. See the link for more details. https://github.com/everett-protocol/ethereum-interchain-account
My implementation of interchain account on Ethereum is a very early stage of POC. It just implements the minimal set of the interface needed to implement interchain account specification because we don't have the implementation of IBC on Ethereum. But it indicates that the interchain account can operate on the Ethereum smart contract level.

Thunnini:
@cwgoes @mossid

spec/ics-027-interchain-account/README.md
cwgoes: (your response makes sense but can you clarify this in the document?)

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+### Definitions  
+
+The IBC handler interface & IBC relayer module interface are as defined in [ICS 25](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-025-handler-interface) and [ICS 26](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-026-relayer-module), respectively.  
+
+### Desired Properties  
+
+- Permissionless  
+- Fault containment: Interchain account must follow rules of its host chain, even in times of Byzantine behavior by the counterparty chain (the chain that manages the account)  
+- The chain that controls the account must process the results asynchronously and according to the chain's logic.  
+- Sending and receiving transactions will be processed in an ordered manner.
```
cwgoes: As in, first in first out? Be specific (there could be many kinds of orderings).

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+### Desired Properties  
+
+- Permissionless  
+- Fault containment: Interchain account must follow rules of its host chain, even in times of Byzantine behavior by the counterparty chain (the chain that manages the account)  
+- The chain that controls the account must process the results asynchronously and according to the chain's logic.  
+- Sending and receiving transactions will be processed in an ordered manner.  
+
+## Technical Specification  
+
+The implementation of interchain account is non-symmetric. This means that each chain can have a different way to generate an interchain account and deserialize the transaction bytes and a different set of transactions that they can execute. For example, chains that use cosmos-sdk will deserialize tx bytes by Amino, but if counterparty chain is a smart contract on Ethereum, it may deserialize tx bytes by ABI that is minimal serializing algorithm for smart contract.
```
cwgoes:
```suggestion
The implementation of interchain account is non-symmetric. This means that each chain can have a different way to generate an interchain account and deserialize the transaction bytes and a different set of transactions that they can execute. For example, chains that use the Cosmos SDK will deserialise tx bytes using Amino, but if the counterparty chain is a smart contract on Ethereum, it may deserialise tx bytes by an ABI that is a minimal serialisation algorithm for the smart contract.  
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+- Permissionless  
+- Fault containment: Interchain account must follow rules of its host chain, even in times of Byzantine behavior by the counterparty chain (the chain that manages the account)  
+- The chain that controls the account must process the results asynchronously and according to the chain's logic.  
+- Sending and receiving transactions will be processed in an ordered manner.  
+
+## Technical Specification  
+
+The implementation of interchain account is non-symmetric. This means that each chain can have a different way to generate an interchain account and deserialize the transaction bytes and a different set of transactions that they can execute. For example, chains that use cosmos-sdk will deserialize tx bytes by Amino, but if counterparty chain is a smart contract on Ethereum, it may deserialize tx bytes by ABI that is minimal serializing algorithm for smart contract.  
+Interchain account specification defines the general way to register interchain account and transfer tx bytes, and counterparty chain is responsible to deserialize and execute the tx bytes, and sending chain should know how counterparty chain handle tx bytes in advance.
```
cwgoes:
```suggestion
The interchain account specification defines the general way to register an interchain account and transfer tx bytes. The counterparty chain is responsible for deserialising and executing the tx bytes, and the sending chain should know how counterparty chain will handle the tx bytes in advance.
```

spec/ics-027-interchain-account/README.md
cwgoes: (please clarify this in the text)

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+Also, each chain must know how the counterparty chains serialize/deserialize transaction bytes in order to send transactions via IBC. And the counterparty chain must implement the process of safely exececuting IBC transactions by verifying the authority of the transaction's signers.
+
+The chain must reject the transaction and must not make a state transition in the following cases:
+
+- The IBC transaction fails to be deserialized.
+- The IBC transaction requests for signers that was not made by counterparty chain.
+
+### Data Structures
+
+Each chain must implement below interfaces to support interchain account. ```createOutgoingPacket``` method in ```IBCAccountModule``` interface defines the way how to create an outgoing packet for a specific type. Type indicates how IBC account transaction should be constructed and serialized for the host chain. Generally, type indicates what framework the host chain was built from.
```
cwgoes:
```suggestion
Each chain must implement the below interfaces to support interchain account. ```createOutgoingPacket``` method in ```IBCAccountModule``` interface defines the way to create an outgoing packet for a specific type. Type indicates how IBC account transaction should be constructed and serialized for the host chain. Generally, type indicates what framework the host chain was built from.  
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+- The IBC transaction fails to be deserialized.
+- The IBC transaction requests for signers that was not made by counterparty chain.
+
+### Data Structures
+
+Each chain must implement below interfaces to support interchain account. ```createOutgoingPacket``` method in ```IBCAccountModule``` interface defines the way how to create an outgoing packet for a specific type. Type indicates how IBC account transaction should be constructed and serialized for the host chain. Generally, type indicates what framework the host chain was built from.  
+```generateAddress``` defines the way how to determine the account's address by using identifier and salt. Using the salt to generate an address is recommended, but not required. If the chain don't support the way to generate an address with salt deterministically, it can be generated by its own way. ```createAccount``` is used to create account with genereated address. New interchain account must not conflict with existing one, and chains should keep track of which counterparty chain created the new interchain account in order to verify the authority of transaction's singers in ```authenticateTx```.
```
cwgoes:
```suggestion
```generateAddress``` defines the way how to determine the account's address by using identifier and salt. Using the salt to generate an address is recommended, but not required. If the chain doesn't support a deterministic way to generate an address with a salt, it can be generated by its own way. ```createAccount``` is used to create account with generated address. New interchains account must not conflict with existing ones, and chains should keep track of which counterparty chain created each new interchain account in order to verify the authority of transaction's signers in ```authenticateTx```.
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+```authenticateTx``` validates a transaction and check the signers in the transaction have right permissions. ```runTx``` executes tx after the transaction was authenticated successfully.
```
cwgoes:
```suggestion
```authenticateTx``` validates a transaction and checks that the signers in the transaction have the right permissions. ```runTx``` executes a transaction after it was authenticated successfully.
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+  createOutgoingPacket(chainType: Uint8Array, data: any)
+  createAccount(address: Uint8Array)
+  generateAddress(identifier: Identifier, salt: Uint8Array): Uint8Array
+  deserializeTx(txBytes: Uint8Array): Tx
+  authenticateTx(tx: Tx): boolean
+  runTx(tx: Tx): uint32
+}
+```
+
+`RegisterIBCAccountPacketData` is used for counterparty chain to register an account. Interchain account's address is defined deterministically with channel identifier and salt. ```generateAccount``` method in module is used to generate interchain account's address. It is recommended to generate address by ```hash(identifier+salt)```, but it doesn't impose a particular process. It must generate unique and deterministic address by utilizing identifier and salt.
```
cwgoes:
```suggestion
`RegisterIBCAccountPacketData` is used by the counterparty chain to register an account. An interchain account's address is defined deterministically with the channel identifier and salt. The ```generateAccount``` method is used to generate a new interchain account's address. It is recommended to generate address by ```hash(identifier+salt)```, but other methods may be used. This function must generate a unique and deterministic address by utilizing identifier and salt.
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+```typescript  
+interface RegisterIBCAccountPacketData {  
+  salt: Uint8Array  
+}  
+```
+
+`RunTxPacketData` is used to run tx for interchain account. Tx bytes is dependent on app, and it is formed with minimal data set except data for validating signatures.
```
cwgoes:
```suggestion
`RunTxPacketData` is used to execute a transaction on an interchain account. The transaction bytes contain the transaction itself & the appropriate signatures, and are serialised in a manner appropriate for the destination chain.
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+`RunTxPacketData` is used to run tx for interchain account. Tx bytes is dependent on app, and it is formed with minimal data set except data for validating signatures.  
+
+```typescript  
+interface RunTxPacketData {  
+  txBytes: Uint8Array  
+}  
+```
+
+```IBCAccountHandler``` interface helps the developers to get results from interactions with interchain account.
```
cwgoes:
```suggestion
The ```IBCAccountHandler``` interface allows the source chain to receive results of executing transactions on an interchain account.
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,292 @@
...
+interface InterchainTxHandler {
+  onAccountCreated(identifier: Identifier, address: Address)
+  onTxSucceeded(identifier: Identifier, txBytes: Uint8Array)
+  onTxFailed(identifier: Identifier, txBytes: Uint8Array, errorCode: Uint8Array)
+}
+```
+
+### Subprotocols  
+
+The subprotocols described herein should be implemented in a "interchain-account-bridge" module with access to a router and codec (decoder or unmarshaller) for app and to the IBC relayer module.
```
cwgoes:
```suggestion
The subprotocols described herein should be implemented in a "interchain-account-bridge" module with access to a router and codec (decoder or unmarshaller) for the application and access to the IBC relayer module.  
```

Thunnini:
@cwgoes Thank you for your advice and for fixing the grammar and misspelling. I fixed it. Please check again.

cwgoes:
> @cwgoes Thank you for your advice and for fixing the grammar and misspelling. I fixed it. Please check again.

I think there are still a few outstanding comments, please check. I can't commit to your branch.

Thunnini:
@cwgoes I handled missed comments. Do I have anything left?

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+This standard document specifies packet data structure, state machine handling logic, and encoding details for the account management system over an IBC channel between separate chains.  
+
+### Motivation  
+
+On Ethereum, there are two types of accounts: externally owned accounts, controlled by private keys, and contract accounts, controlled by their contract code [[ref](https://github.com/ethereum/wiki/wiki/White-Paper)]. Similar to Ethereum's CA (contract accounts), interchain accounts are managed by another chain while retaining all the capabilities of a normal account (i.e. stake, send, vote, etc). While an Ethereum CA's contract logic is performed within Ethereum's EVM, interchain accounts are managed by another chain via IBC in a way such that the owner of the account retains full control over how it behaves.  
+
+### Definitions  
+
+The IBC handler interface & IBC relayer module interface are as defined in [ICS 25](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-025-handler-interface) and [ICS 26](https://github.com/cosmos/ics/blob/bez/15-ics-cosmos-signed-messages/spec/ics-026-relayer-module), respectively.
```
mossid:
```suggestion
The IBC handler interface & IBC relayer module interface are as defined in [ICS 25](https://github.com/cosmos/ics/tree/master/spec/ics-025-handler-interface) and [ICS 26](https://github.com/cosmos/ics/tree/master/spec/ics-026-routing-module), respectively.  
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Data Structures
+
+Each chain must implement the below interfaces to support interchain account. ```createOutgoingPacket``` method in ```IBCAccountModule``` interface defines the way to create an outgoing packet for a specific type. Type indicates how IBC account transaction should be constructed and serialized for the host chain. Generally, type indicates what framework the host chain was built from.  
+```generateAddress``` defines the way how to determine the account's address by using identifier and salt. Using the salt to generate an address is recommended, but not required. If the chain doesn't support a deterministic way to generate an address with a salt, it can be generated by its own way. ```createAccount``` is used to create account with generated address. New interchains account must not conflict with existing ones, and chains should keep track of which counterparty chain created each new interchain account in order to verify the authority of transaction's signers in ```authenticateTx```.
```
mossid: Just a typo

```suggestion
```generateAddress``` defines the way how to determine the account's address by using identifier and salt. Using the salt to generate an address is recommended, but not required. If the chain doesn't support a deterministic way to generate an address with a salt, it can be generated by its own way. ```createAccount``` is used to create account with generated address. New interchain account must not conflict with existing ones, and chains should keep track of which counterparty chain created each new interchain account in order to verify the authority of transaction's signers in ```authenticateTx```.
```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,296 @@
...
+```authenticateTx``` validates a transaction and checks that the signers in the transaction have the right permissions. ```runTx``` executes a transaction after it was authenticated successfully.
+
```
mossid: Add the definition for `Tx`, simply `type Tx = object`

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,296 @@
...
+  channelIdentifier: Identifier,  
+  counterpartyPortIdentifier: Identifier,  
+  counterpartyChannelIdentifier: Identifier,  
+  version: string) {  
+  // only ordered channels allowed  
+  abortTransactionUnless(order === ORDERED)  
+  // only allow channels to "interchain-account" port on counterparty chain  
+  abortTransactionUnless(counterpartyPortIdentifier === "interchain-account")  
+  // version not used at present  
+  abortTransactionUnless(version === "")
```
mossid: For future compatibility, I think it is better to remove version checking logic if we are not actually using version
```suggestion

```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,296 @@
...
+  channelIdentifier: Identifier,  
+  counterpartyPortIdentifier: Identifier,  
+  counterpartyChannelIdentifier: Identifier,  
+  version: string,  
+  counterpartyVersion: string) {  
+  // only ordered channels allowed  
+  abortTransactionUnless(order === ORDERED)  
+  // version not used at present  
+  abortTransactionUnless(version === "")  
+  abortTransactionUnless(counterpartyVersion === "")
```
mossid: For future compatibility, I think it is better to remove version checking logic if we are not actually using version
```suggestion

```

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,296 @@
...
+}  
+```
+
+```typescript  
+function onChanOpenAck(  
+  portIdentifier: Identifier,  
+  channelIdentifier: Identifier,  
+  version: string) {  
+  // version not used at present  
+  abortTransactionUnless(version === "")
```
mossid: For future compatibility, I think it is better to remove version checking logic if we are not actually using version cc @cwgoes 
```suggestion

```

spec/ics-027-interchain-account/README.md
Thunnini: I'll take it

spec/ics-027-interchain-account/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+- New interchain accounts must not conflict with existing ones.
+- Each chain must keep track of which counterparty chain created each new interchain account.
+
+Also, each chain must know how the counterparty chains serialize/deserialize transaction bytes in order to send transactions via IBC. And the counterparty chain must implement the process of safely exececuting IBC transactions by verifying the authority of the transaction's signers.
+
+The chain must reject the transaction and must not make a state transition in the following cases:
+
+- The IBC transaction fails to be deserialized.
+- The IBC transaction requests for signers that was not made by counterparty chain.
```
cwgoes:
```suggestion
- The IBC transaction expects signatures which the counterparty chain did not provide.
```

cwgoes: @Thunnini Can you apply this?

spec/ics-027-interchain-account/README.md
cwgoes: Sure.

spec/ics-027-interchain-account/README.md
Thunnini: I think that signature is not needed because this is not communication between the user and the chain, but between the chain and the chain. It is possible to determine if this account is the interchain account of the counterparty chain. Signatures are not required for communication between contracts on Ethereum. I think this is the same principle.

cwgoes: Oh, do you mean "requests a signer other than the interchain account" ?

Thunnini: I mean that "requests for signers that was not interchain account made by counterparty chain."

cwgoes: Makes sense. Updated my suggestion to clarify.

PR #314, opened 2019-11-26
Alter spec template

cwgoes:
Ref https://github.com/cosmos/ics/issues/286


cwgoes:
Never mind, these should be separate ICSs, the process in the future will be important.

PR #318, opened 2019-11-27
Define & use log emitting & reading interface

cwgoes:
Closes https://github.com/cosmos/ics/issues/313
Closes https://github.com/cosmos/ics/issues/217

### DIFF
--- a/spec/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/ics-004-channel-and-packet-semantics/README.md
@@ -522,6 +522,9 @@ function sendPacket(packet: Packet) {
     nextSequenceSend = nextSequenceSend + 1
     provableStore.set(nextSequenceSendPath(packet.sourcePort, packet.sourceChannel), nextSequenceSend)
     provableStore.set(packetCommitmentPath(packet.sourcePort, packet.sourceChannel, packet.sequence), hash(packet.data, packet.timeout))
+
+    // log that a packet has been sent
+    emitLogEntry("sendPacket", {sequence: packet.sequence, data: packet.data, timeout: packet.timeout})
 }
 ```
 
@@ -585,6 +588,9 @@ function recvPacket(
       nextSequenceRecv = nextSequenceRecv + 1
       provableStore.set(nextSequenceRecvPath(packet.destPort, packet.destChannel), nextSequenceRecv)
     }
+
+    // log that a packet has been received & acknowledged
+    emitLogEntry("recvPacket", {sequence: packet.sequence, timeout: packet.timeout, data: packet.data, acknowledgement})
 
     // return transparent packet
     return packet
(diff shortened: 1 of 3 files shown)


ancazamfir:
Since `pendingDatagrams()` now returns local and counterparty datagrams, should the `relay()` function be something like: 
```
function relay(C: Set<Chain>) {
  for (const chain of C)
    for (const counterparty of C)
      if (counterparty !== chain) {
        const datagrams = pendingDatagrams(chain, counterparty)
        for (const localDatagram of datagrams[0])
          chain.submitDatagram(localDatagram)
        for (const counterpartyDatagram of datagrams[1])
          counterparty.submitDatagram(counterpartyDatagram)
      }
}
```
Then, in `pendingDatagrams()` for the checks `if client.height < height {..` isn't the relayer off-chain process supposed to submit intermediate headers (i.e. do bisection and multiple push ClientUpdate-s) if there was too much change (<1/3rd common val power)?

cwgoes:
> Since `pendingDatagrams()` now returns local and counterparty datagrams, should the `relay()` function be something like:
> 
> ```
> function relay(C: Set<Chain>) {
>   for (const chain of C)
>     for (const counterparty of C)
>       if (counterparty !== chain) {
>         const datagrams = pendingDatagrams(chain, counterparty)
>         for (const localDatagram of datagrams[0])
>           chain.submitDatagram(localDatagram)
>         for (const counterpartyDatagram of datagrams[1])
>           counterparty.submitDatagram(counterpartyDatagram)
>       }
> }
> ```

Yes. Updated in https://github.com/cosmos/ics/commit/8d03d7ada51ad056a9355698fe8c668b1d133dad.

> Then, in `pendingDatagrams()` for the checks `if client.height < height {..` isn't the relayer off-chain process supposed to submit intermediate headers (i.e. do bisection and multiple push ClientUpdate-s) if there was too much change (<1/3rd common val power)?

Hmm, with Tendermint light clients, yes, but bisection is a client-type-specific feature. This is just intended to be a general example. Still we could add it I suppose.

ancazamfir:
> Hmm, with Tendermint light clients, yes, but bisection is a client-type-specific feature. This is just intended to be a general example. Still we could add it I suppose.

True, it depends on the client capabilities. `ClientUpdate`-s are also needed for sequential, for all headers in `(client.height..chain.latestHeight()]`. 
Maybe we can have a comment that a number of updateClient datagrams may be required?

Also, since the chain is continuously making blocks, is it possible that as we build the connection/ channel/ packet datagrams, some of their `proofHeight`-s become higher than the result of the `chain.latestHeight()` call before? Should the call to `chain.latestHeight()` be done last and `ClientUpdate` datagrams pre-pended to the other datagrams?

ancazamfir:
> Also, since the chain is continuously making blocks, is it possible that as we build the connection/ channel/ packet datagrams, some of their `proofHeight`-s become higher than the result of the `chain.latestHeight()` call before? Should the call to `chain.latestHeight()` be done last and ClientUpdate datagrams pre-pended to the other datagrams?

hmm...ok, I guess not :)

PR #320, opened 2019-12-02
Add alternative handshake path

cwgoes:
Closes https://github.com/cosmos/ics/issues/300


cwgoes:
@mossid Does this cover all the cases you were concerned about?

Also see https://github.com/cosmos/ics/issues/300#issuecomment-558597567, which this PR does not address.

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -86,14 +86,14 @@ Channel ends have a *state*:
 ```typescript
 enum ChannelState {
   INIT,
-  OPENTRY,
+  TRYOPEN,
```
fedekunze: ++

PR #322, opened 2019-12-03
Fix ics-027

Thunnini:
I read carefully after merging. And I found some things needed to be fixed. 🙏
I removed  "the appropriate signatures" in the sentence "The transaction bytes contain the transaction itself & the appropriate signatures". And I fixed a typo that I found.
@cwgoes And, It seems that you use "serialise/deserialise". But, my English dictionary says it as "serialize/deserialize". I'm not sure that which one is correct or both are correct. But, I used "serialize/deserialize". So, there are two types of "serializ(s)e/deserializ(s)e" in this document. Do you think that it is better to use only "serialise/deserialise"?

### DIFF
--- a/spec/ics-027-interchain-account/README.md
+++ b/spec/ics-027-interchain-account/README.md
@@ -44,7 +44,7 @@ Also, each chain must know how the counterparty chains serialize/deserialize tra
 The chain must reject the transaction and must not make a state transition in the following cases:
 
 - The IBC transaction fails to be deserialized.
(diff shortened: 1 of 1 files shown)


cwgoes:
`serialize` / `deserialize` is the American English spelling, `serialise` / `deserialise` is British English; by convention this repository uses the latter.

cwgoes:
I've updated the spelling in https://github.com/cosmos/ics/commit/0aebe25aab0c0253001fc77c719bc0399cebf469.

PR #331, opened 2019-12-09
ICS 6: Solo Machine Client

cwgoes:
Ref https://github.com/cosmos/ics/issues/286.

Solo machine client using new store layout abstraction.


cwgoes:
... so now the handshake desugars to key exchange, interesting, I wonder if we should compare it with existing public-key exchange protocols.

cwgoes:
"Big table of signatures in the universe"

cwgoes:
> ACK. Do we want to implement this on the SDK as a separate pkg?

Hmm, that's a good question - probably, I expect. The ICS 2 interface should be modular so new client types can easily be added as upgrades (ideally with the upgrade module if we're adding that).

cwgoes:
Merging; further reviews are welcome at any time.

PR #348, opened 2020-01-09
Update paths for consistency; revise ICS 2 for client type sub-specs

cwgoes:
Closes https://github.com/cosmos/ics/issues/345
Closes https://github.com/cosmos/ics/issues/346

### DIFF
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -253,10 +253,10 @@ but they must expose this common set of query functions to the IBC handler.
 type ClientState = bytes
 ```
 
-Client types MUST define a method to initialize a client state with a provided consensus state, writing to state as appropriate.
+Client types MUST define a method to initialise a client state with a provided consensus state, writing to state as appropriate.
 
 ```typescript
-type initialize = (state: ConsensusState) => ()
+type initialise = (consensusState: ConsensusState) => ClientState
 ```
 
 Client types MUST define a method to fetch the current height (height of the most recent validated header).
@@ -544,7 +544,7 @@ function commit(
 }
 
 // initialisation function defined by the client type
-function initialize(consensusState: ConsensusState): () {
+function initialise(consensusState: ConsensusState): () {
   clientState = {
     frozen: false,
     pastPublicKeys: Set.singleton(consensusState.publicKey),
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -253,10 +253,18 @@ but they must expose this common set of query functions to the IBC handler.
 type ClientState = bytes
 ```
 
-Client types must also define a method to initialize a client state with a provided consensus state, writing to state as appropriate.
+Client types MUST define a method to initialize a client state with a provided consensus state, writing to state as appropriate.
 
 ```typescript
 type initialize = (state: ConsensusState) => ()
+```
+
+Client types MUST define a method to fetch the current height (height of the most recent validated header).
+
+```typescript
+type latestHeight = (
+  clientState: ClientState)
+  => uint64
 ```
 
 #### CommitmentProof
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -573,7 +573,7 @@ function verifyClientConsensusState(
   proof: CommitmentProof,
   clientIdentifier: Identifier,
   consensusState: ConsensusState) {
-    path = applyPrefix(prefix, "consensusStates/{clientIdentifier}/" + height)
+    path = applyPrefix(prefix, "clients/{clientIdentifier}/consensusStates/{height}")
     abortTransactionUnless(!clientState.frozen)
     return clientState.verifiedRoots[sequence].verifyMembership(path, consensusState, proof)
 }
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -253,10 +253,10 @@ but they must expose this common set of query functions to the IBC handler.
 type ClientState = bytes
 ```
 
-Client types must also define a method to initialize a client state with a provided consensus state:
+Client types must also define a method to initialize a client state with a provided consensus state, writing to state as appropriate.
 
 ```typescript
-type initialize = (state: ConsensusState) => ClientState
+type initialize = (state: ConsensusState) => ()
 ```
 
 #### CommitmentProof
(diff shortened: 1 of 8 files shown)


fedekunze:
@cwgoes any chance we could switch the channel paths from `"ports/{portID}/channels/{channelID}"` to `"channels/{channelID}/ports/{portID}"`

cwgoes:
> @cwgoes any chance we could switch the channel paths from `"ports/{portID}/channels/{channelID}"` to `"channels/{channelID}/ports/{portID}"`

Hmm, we could, doesn't it seem a bit confusing though (since channel identifiers are unique "per-port")? What query do you want to run that requires contiguous keys?

cwgoes:
For consensus state heights, we need to pick a height at which to check that the counterparty chain has the correct consensus state. I think we should say that this must be the height at which we started the handshake, so that means we'll need to keep it in state (temporarily). That seems safest though.

cwgoes:
I wonder if client types should also define an `initialise` function, we want to avoid paths in ICS 2.

cwgoes:
Also, the spec should define what a loopback client implementation looks like.

cwgoes:
> Also, the spec should define what a loopback client implementation looks like.

Split to https://github.com/cosmos/ics/issues/352.

cwgoes:
How to correctly sub-space with non-uniformly-prefixed keys?

@fedekunze Do you know if the SDK has a substore wrapper which can limit read/write to *specific* keys (instead of just a prefix)? We want to limit client type instances to reading/writing only keys belonging to them, but we altered the layout like you mentioned so that keys for a given client don't have a uniform prefix (this makes for easier iteration, I suppose).

cwgoes:
- [ ] Fix ICS 4, thanks Fede.

cwgoes:
> > @cwgoes any chance we could switch the channel paths from `"ports/{portID}/channels/{channelID}"` to `"channels/{channelID}/ports/{portID}"`
> 
> Hmm, we could, doesn't it seem a bit confusing though (since channel identifiers are unique "per-port")? What query do you want to run that requires contiguous keys?

Confirmed per discussion that we don't need this.

cwgoes:
For now I am going to update the paths so that everything is prefixed under `clients/{identifier}` - this means that we can use the prefix store for permissioning. IMO that is worth possibly having to skip items while iterating, since that only ought to be happening out of transaction processing.

cwgoes:
This PR really makes me want a type-checked spec...

spec/ics-024-host-requirements/README.md
```diff
@@ -99,11 +99,14 @@ Parts of the private store MAY safely be used for other purposes as long as the
...
 | -------------- | ------------------------------------------------------------------------------ | ----------------- | ---------------------- |
-| privateStore   | "clients/{identifier}"                                                         | ClientState       | [ICS 2](../ics-002-client-semantics) |
-| provableStore  | "clients/{identifier}/consensusState"                                          | ConsensusState    | [ICS 2](../ics-002-client-semantics) |
 | provableStore  | "clients/{identifier}/type"                                                    | ClientType        | [ICS 2](../ics-002-client-semantics) |
+| privateStore   | "clients/{identifier}"                                                         | ClientState       | [ICS 2](../ics-007-tendermint-client) |
+| provableStore  | "clients/{identifier}/consensusStates/{height}"                                | ConsensusState    | [ICS 7](../ics-007-tendermint-client) |
+| privateStore   | "clients/{identifier}/connections                                              | []Identifier      | [ICS 3](../ics-003-connection-semantics) |
```
fedekunze: what is this path for?

spec/ics-002-client-semantics/README.md
```diff
@@ -576,12 +544,13 @@ function commit(
 }
 
 // initialisation function defined by the client type
-function initialize(consensusState: ConsensusState): ClientState {
```
fedekunze: I actually prefer to return the `ClientState` and set it to the store on `createClient`

fedekunze: at least as I see it, `initialize` is just the constructor of the `ClientState`

cwgoes: Hmm. We can do that, I suppose, since this store location does need to be known outside the client-type-specific spec.

spec/ics-024-host-requirements/README.md
cwgoes: It's used for reverse lookup of what connections are using a client.

spec/ics-002-client-semantics/README.md
```diff
@@ -468,28 +449,15 @@ function createClient(
...
-}
-```
-
-```typescript
-function queryClient(id: Identifier): ClientState {
-    return privateStore.get(clientStatePath(id))
-}
-```
+Client consensus state and client internal state can be queried by identifier, but
+the specific paths which must be queried are defined by each client type.
```
fedekunze: > the specific paths which must be queried are defined by each client type.

why is this now defined per client? All clients need to still be able to query the consensus and client states

spec/ics-002-client-semantics/README.md
```diff
@@ -253,10 +253,18 @@ but they must expose this common set of query functions to the IBC handler.
...
 
 ```typescript
-type initialize = (state: ConsensusState) => ClientState
+type initialize = (state: ConsensusState) => ()
+```
+
+Client types MUST define a method to fetch the current height (height of the most recent validated header).
+
+```typescript
+type latestHeight = (
```
fedekunze: this was confusing for me at least in the beginning. The naming of the third-party client consensus state vs the **self** consensus state was not clear. So whenever it said `getConsensusState(height)` it meant self consensus state (i.e block header and validator info), while the `getConsensusState(clientID, height)` meant the counterparty client.

I'd suggest adding the `client` or `self` prefix to the `ConsensusState` to clarify this item.

fedekunze:
```suggestion
type latestClientHeight = (
```

fedekunze: or, alternatively `clientState.latestHeight()`.

spec/ics-002-client-semantics/README.md
cwgoes: The client state, yes, but the consensus state, not necessarily, e.g. for a solo machine there is not a consensus state stored for each height. Client types can be required to expose `queryConsensusState` I suppose - is this required? The relayer should use a client-type-agnostic method.

fedekunze: > The client state, yes, but the consensus state, not necessarily, e.g. for a solo machine there is not a consensus state stored for each height. Client types can be required to expose queryConsensusState I suppose...

👍 It'd be awesome if we could define a standard interface for all clients with the functions that they need to implement.

cwgoes: Aye, agreed. I'll add that.

spec/ics-002-client-semantics/README.md
cwgoes: Aye, that makes sense, I'll change the client function name to `getClientConsensusState`.

cwgoes: Actually, I can't find any uses of `getConsensusState` now which take a `clientID` and `height`. Can you point to a specific location which is confusing?

spec/ics-002-client-semantics/README.md
cwgoes: With regards to a standard interface, I completely agree, but the only parts of that I can reliably guess are those which will be required by relayers (querying clients), so if you want additional functions for convenience or CLI UX, let me know.

fedekunze: I defined a few interfaces for clients in cosmos/cosmos-sdk#5485

spec/ics-002-client-semantics/README.md
fedekunze: > Hmm. We can do that, I suppose, since this store location does need to be known outside the client-type-specific spec.

Should we keep the return value then?

spec/ics-002-client-semantics/README.md
cwgoes: Yeah. I think I'll split that out to a separate issue & we can discuss first. https://github.com/cosmos/ics/issues/356

spec/ics-002-client-semantics/README.md
cwgoes: Yes. Updated.

PR #355, opened 2020-01-17
ICS 9 (loopback client)

cwgoes:
Loop-back client for interacting through the IBC interface with modules on the local ledger.

Closes https://github.com/cosmos/ics/issues/352.

### DIFF
--- a/spec/ics-009-loopback-client/README.md
+++ b/spec/ics-009-loopback-client/README.md
@@ -1,0 +1,69 @@
+---
+ics: 9
+title: Loopback Client
+stage: draft
+category: IBC/TAO
+kind: instantiation
+author: Christopher Goes <cwgoes@tendermint.com>
+created: 2020-01-17
+modified: 2020-01-17
+implements: 2
+---
+
+## Synopsis
+
+(high-level description of and rationale for specification)
+
+### Motivation
+
+(rationale for existence of standard)
+
+### Definitions
+
+(definitions of any new terms not defined in common documentation)
+
+### Desired Properties
+
+(desired characteristics / properties of protocol, effects if properties are violated)
+
+## Technical Specification
+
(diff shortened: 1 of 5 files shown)


spec/ics-009-loopback-client/README.md
```diff
@@ -0,0 +1,196 @@
...
+implements: 2
+---
+
+## Synopsis
+
+This specification describes a loop-back client, designed to be used for interaction over the IBC interface with modules present on the same ledger.
+
+### Motivation
+
+Loop-back clients may be useful in cases where the calling module does not have prior knowledge of where precisely the destination module lives and would like to use the uniform IBC message-passing interface.
```
fedekunze: Maybe provide a more concrete use case? I understand the rationale, but Idk why would one use a lookback client in the first place

cwgoes: Because the interface is standardised. Say that you want to use the same (asynchronous) logic for handling token transfers to/from other modules on the same chain and to/from other modules on remote chains - this makes that easy to do without any overhead costs.

cwgoes: As a direct analogy, people use TCP/IP loopback clients all the time (`127.0.0.1`) for precisely this reason - so that they can use all the existing tooling which supports the protocol.

cwgoes: I added a quick note that `127.0.0.1` is similar.

fedekunze:
can we rename this ICS to localhost client? I think the term is more widely used than loopback

cwgoes:
> can we rename this ICS to localhost client? I think the term is more widely used than loopback

I scanned the [RFC](https://tools.ietf.org/html/rfc6761) and I think "loopback" is what we want, "localhost" is just an alias.

PR #367, opened 2020-01-24
Fixes to ICS 7 from review

cwgoes:
Closes https://github.com/cosmos/ics/issues/366
Closes https://github.com/cosmos/ics/issues/365
Closes https://github.com/cosmos/ics/issues/364

Thanks @AdityaSripal & @fedekunze.

### DIFF
--- a/spec/ics-007-tendermint-client/README.md
+++ b/spec/ics-007-tendermint-client/README.md
@@ -168,7 +168,7 @@ function checkMisbehaviourAndUpdateState(
       verify(consensusState.validatorSet, evidence.fromHeight, evidence.h2)
       )
     // set the frozen height
-    clientState.frozenHeight = evidence.h1.height // which is same as h2.height
+    clientState.frozenHeight = min(clientState.frozenHeight, evidence.h1.height) // which is same as h2.height
     // save the client
     set("clients/{identifier}", clientState)
 }
--- a/spec/ics-007-tendermint-client/README.md
+++ b/spec/ics-007-tendermint-client/README.md
@@ -22,6 +22,8 @@ State machines of various sorts replicated using the Tendermint consensus algori
 
 Functions & terms are as defined in [ICS 2](../ics-002-client-semantics).
 
+`currentTimestamp` is as defined in [ICS 24](../ics-024-host-requirements).
+
 The Tendermint light client uses the generalised Merkle proof format as defined in ICS 8.
 
 `hash` is a generic collision-resistant hash function, and can easily be configured.
@@ -125,9 +127,9 @@ function checkValidityAndUpdateState(
   clientState: ClientState,
   header: Header) {
     // assert trusting period has not yet passed
-    assert(now - clientState.latestTimestamp < trustingPeriod)
+    assert(currentTimestamp() - clientState.latestTimestamp < trustingPeriod)
     // assert header timestamp is not in the future (& transitively that is not past the trusting period)
-    assert(header.timestamp <= now)
+    assert(header.timestamp <= currentTimestamp())
     // assert header timestamp is past current timestamp
     assert(header.timestamp > clientState.latestTimestamp)
     // assert header height is newer than any we know
@@ -159,7 +161,7 @@ function checkMisbehaviourAndUpdateState(
     // fetch the previously verified commitment root & validator set
     consensusState = get("clients/{identifier}/consensusStates/{evidence.fromHeight}")
     // assert that the timestamp is not from more than an unbonding period ago
-    assert(now - consensusState.timestamp < clientState.unbondingPeriod)
+    assert(currentTimestamp() - consensusState.timestamp < clientState.unbondingPeriod)
     // check if the light client "would have been fooled"
     assert(
       verify(consensusState.validatorSet, evidence.fromHeight, evidence.h1) &&
--- a/spec/ics-007-tendermint-client/README.md
+++ b/spec/ics-007-tendermint-client/README.md
@@ -51,7 +51,7 @@ interface ClientState {
 
 ### Consensus state
 
-The Tendermint client tracks the timestamp (block time), validator set, and commitment root for all previously verified consensus states (these can be pruned after awhile).
+The Tendermint client tracks the timestamp (block time), validator set, and commitment root for all previously verified consensus states (these can be pruned after the unbonding period has passed, but should not be pruned beforehand).
 
 ```typescript
 interface ConsensusState {
--- a/spec/ics-007-tendermint-client/README.md
+++ b/spec/ics-007-tendermint-client/README.md
@@ -41,7 +41,10 @@ The Tendermint client state tracks the current validator set, latest height, and
 ```typescript
 interface ClientState {
   validatorSet: List<Pair<Address, uint64>>
+  trustingPeriod: uint64
+  unbondingPeriod: uint64
   latestHeight: uint64
+  latestTimestamp: uint64
   frozenHeight: Maybe<uint64>
 }
 ```
@@ -52,6 +55,7 @@ The Tendermint client tracks the validator set & commitment root for all previou
 
 ```typescript
 interface ConsensusState {
+  timestamp: uint64
   validatorSet: List<Pair<Address, uint64>>
   commitmentRoot: []byte
 }
@@ -64,6 +68,7 @@ The Tendermint client headers include a height, the commitment root, the complet
 ```typescript
 interface Header {
   height: uint64
+  timestamp: uint64
   commitmentRoot: []byte
   validatorSet: List<Pair<Address, uint64>>
   signatures: []Signature
@@ -88,10 +93,13 @@ interface Evidence {
 Tendermint client initialisation requires a (subjectively chosen) latest consensus state, including the full validator set.
 
 ```typescript
-function initialise(consensusState: ConsensusState, validatorSet: List<Pair<Address, uint64>>, height: uint64): ClientState {
+function initialise(consensusState: ConsensusState, validatorSet: List<Pair<Address, uint64>>, height: uint64, trustingPeriod: uint64, unbondingPeriod: uint64): ClientState {
   return ClientState{
     validatorSet,
     latestHeight: height,
+    latestTimestamp: consensusState.timestamp,
+    trustingPeriod,
+    unbondingPeriod,
     pastHeaders: Map.singleton(latestHeight, consensusState)
   }
 }
@@ -113,14 +121,20 @@ Tendermint client validity checking uses the bisection algorithm described in th
 function checkValidityAndUpdateState(
   clientState: ClientState,
   header: Header) {
-    // assert that header is newer than any we know
-    assert(header.height < clientState.latestHeight)
+    // assert trusting period has not yet passed
+    assert(now - clientState.latestTimestamp < trustingPeriod)
+    // assert header timestamp is not in the future (& transitively that is not past the trusting period)
+    assert(header.timestamp <= now)
+    // assert header timestamp is past current timestamp
+    assert(header.timestamp > clientState.latestTimestamp)
+    // assert header height is newer than any we know
+    assert(header.height > clientState.latestHeight)
     // call the `verify` function
     assert(verify(clientState.validatorSet, clientState.latestHeight, header))
     // update latest height
     clientState.latestHeight = header.height
     // create recorded consensus state, save it
-    consensusState = ConsensusState{validatorSet, header.commitmentRoot}
+    consensusState = ConsensusState{validatorSet, header.commitmentRoot, header.timestamp}
     set("clients/{identifier}/consensusStates/{header.height}", consensusState)
     // save the client
     set("clients/{identifier}", clientState)
@@ -141,6 +155,8 @@ function checkMisbehaviourAndUpdateState(
     assert(evidence.h1.commitmentRoot !== evidence.h2.commitmentRoot)
     // fetch the previously verified commitment root & validator set
     consensusState = get("clients/{identifier}/consensusStates/{evidence.fromHeight}")
+    // assert that the timestamp is not from more than an unbonding period ago
+    assert(now - consensusState.timestamp < clientState.unbondingPeriod)
     // check if the light client "would have been fooled"
     assert(
       verify(consensusState.validatorSet, evidence.fromHeight, evidence.h1) &&
--- a/spec/ics-007-tendermint-client/README.md
+++ b/spec/ics-007-tendermint-client/README.md
@@ -36,7 +36,7 @@ This specification depends on correct instantiation of the [Tendermint consensus
 
 ### Client state
 
-The Tendermint client state tracks the current validator set, latest height, and a possible frozen height.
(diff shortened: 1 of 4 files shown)


spec/ics-007-tendermint-client/README.md
```diff
@@ -41,18 +41,22 @@ The Tendermint client state tracks the current validator set, latest height, and
...
 
 ### Consensus state
 
-The Tendermint client tracks the validator set hash & commitment root for all previously verified consensus states (these can be pruned after awhile).
+The Tendermint client tracks the validator set & commitment root for all previously verified consensus states (these can be pruned after awhile).
 
 ```typescript
 interface ConsensusState {
-  validatorSetHash: []byte
+  timestamp: uint64
```
fedekunze: is this the block time?

spec/ics-007-tendermint-client/README.md
```diff
@@ -41,18 +41,22 @@ The Tendermint client state tracks the current validator set, latest height, and
 ```typescript
 interface ClientState {
   validatorSet: List<Pair<Address, uint64>>
+  trustingPeriod: uint64
+  unbondingPeriod: uint64
   latestHeight: uint64
+  latestTimestamp: uint64
```
fedekunze: ditto?

spec/ics-007-tendermint-client/README.md
```diff
@@ -41,18 +41,22 @@ The Tendermint client state tracks the current validator set, latest height, and
 ```typescript
 interface ClientState {
   validatorSet: List<Pair<Address, uint64>>
+  trustingPeriod: uint64
+  unbondingPeriod: uint64
```
fedekunze: do we need to check that ubd > trusting period?

spec/ics-007-tendermint-client/README.md
```diff
@@ -64,6 +68,7 @@ The Tendermint client headers include a height, the commitment root, the complet
 ```typescript
 interface Header {
   height: uint64
+  timestamp: uint64
```
fedekunze: ditto block time

spec/ics-007-tendermint-client/README.md
cwgoes: Yes.

spec/ics-007-tendermint-client/README.md
cwgoes: Yes.

spec/ics-007-tendermint-client/README.md
cwgoes: Yes. Do you want it to be renamed?

spec/ics-007-tendermint-client/README.md
cwgoes: Sure, we can.

spec/ics-007-tendermint-client/README.md
```diff
@@ -114,14 +124,20 @@ Tendermint client validity checking uses the bisection algorithm described in th
 function checkValidityAndUpdateState(
   clientState: ClientState,
   header: Header) {
-    // assert that header is newer than any we know
-    assert(header.height < clientState.latestHeight)
+    // assert trusting period has not yet passed
+    assert(now - clientState.latestTimestamp < trustingPeriod)
```
AdityaSripal: I'm confused as to what `now` would mean in this context, could you clarify @cwgoes

spec/ics-007-tendermint-client/README.md
```diff
@@ -140,14 +156,14 @@ function checkMisbehaviourAndUpdateState(
...
+    // assert that the timestamp is not from more than an unbonding period ago
+    assert(now - consensusState.timestamp < clientState.unbondingPeriod)
     // check if the light client "would have been fooled"
     assert(
-      verify(evidence.fromValidatorSet, evidence.fromHeight, evidence.h1) &&
-      verify(evidence.fromValidatorSet, evidence.fromHeight, evidence.h2)
+      verify(consensusState.validatorSet, evidence.fromHeight, evidence.h1) &&
+      verify(consensusState.validatorSet, evidence.fromHeight, evidence.h2)
       )
     // set the frozen height
```
AdityaSripal: We must also check that frozen height either doesn't already exist or exists for a later height before setting

spec/ics-007-tendermint-client/README.md
```diff
@@ -89,12 +93,18 @@ interface Evidence {
...
-function initialise(consensusState: ConsensusState, validatorSet: List<Pair<Address, uint64>>, height: uint64): ClientState {
-  return ClientState{
-    validatorSet,
-    latestHeight: height,
-    pastHeaders: Map.singleton(latestHeight, consensusState)
-  }
+function initialise(
+  consensusState: ConsensusState, validatorSet: List<Pair<Address, uint64>>,
+  height: uint64, trustingPeriod: uint64, unbondingPeriod: uint64): ClientState {
+  assert(trustingPeriod < unbondingPeriod)
```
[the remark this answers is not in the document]
cwgoes: Hmm. For safety, it should be strictly less than, as if they are equal at the very end of the trusting period there won't be enough time to submit evidence (since blocks aren't instant).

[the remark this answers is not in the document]
fedekunze: @melekes is there any restriction to this check? ^^

spec/ics-007-tendermint-client/README.md
```diff
@@ -36,34 +36,39 @@ This specification depends on correct instantiation of the [Tendermint consensus
...
   latestHeight: uint64
+  latestTimestamp: uint64
   frozenHeight: Maybe<uint64>
 }
 ```
 
 ### Consensus state
 
-The Tendermint client tracks the validator set hash & commitment root for all previously verified consensus states (these can be pruned after awhile).
+The Tendermint client tracks the timestamp (block time), validator set, and commitment root for all previously verified consensus states (these can be pruned after awhile).
```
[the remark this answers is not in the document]
cwgoes: Ah, yes, indeed, noted both.

spec/ics-007-tendermint-client/README.md
fedekunze: time.Now()

spec/ics-007-tendermint-client/README.md
```diff
@@ -36,34 +36,39 @@ This specification depends on correct instantiation of the [Tendermint consensus
...
 
 ```typescript
 interface ConsensusState {
-  validatorSetHash: []byte
+  timestamp: uint64
+  validatorSet: List<Pair<Address, uint64>>
```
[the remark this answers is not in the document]
cwgoes: This update is per request by @fedekunze. I don't have a strong opinion on the tradeoff. Note that it is a lot of state to store if header updates are frequent, but as long as pruning happens after 3 weeks, the state usage should be constant.

spec/ics-007-tendermint-client/README.md
cwgoes: The current timestamp. Clarified & added as an explicit requirement to ICS 24.

spec/ics-007-tendermint-client/README.md
cwgoes: Indeed. Updated!

spec/ics-007-tendermint-client/README.md
AdityaSripal: Hmm so `currentTimeStamp` in an SDK chain would be `ctx.BlockHeader().Time()`?

spec/ics-007-tendermint-client/README.md
```diff
@@ -114,14 +126,20 @@ Tendermint client validity checking uses the bisection algorithm described in th
 function checkValidityAndUpdateState(
   clientState: ClientState,
   header: Header) {
-    // assert that header is newer than any we know
-    assert(header.height < clientState.latestHeight)
+    // assert trusting period has not yet passed
+    assert(currentTimestamp() - clientState.latestTimestamp < trustingPeriod)
+    // assert header timestamp is not in the future (& transitively that is not past the trusting period)
+    assert(header.timestamp <= currentTimestamp())
```
AdityaSripal: Don't have an intuitive sense of why these 2 checks are necessary.

The first check is ensuring that we don't allow updates to happen if the last update to the connection chain happened before our chain's current time minus the trustingPeriod. What is the purpose of this check? Seems like it will force connections to update within a specific frequency, however i don't see the danger of allowing connections to update as slow as they wish (perhaps a connection with very rare messages sent through it, may choose to update lazily?). It also seems like if this assertion ever fails, then this connection is no longer updatable at all. Is this the desired behavior?

The second check prevents a connecting chain from updating with a header if its timestamp happens to be greater than the current timestamp. This doesn't seem to be an issue to have this check, but i don't understand the attack it is preventing.

Don't think these are necessarily issues to be resolved, but would like an explanation of why these checks are necessary

spec/ics-007-tendermint-client/README.md
```diff
@@ -114,14 +126,20 @@ Tendermint client validity checking uses the bisection algorithm described in th
...
+    // assert header timestamp is past current timestamp
+    assert(header.timestamp > clientState.latestTimestamp)
```
AdityaSripal: There still needs to be this check, correct?

```go
assert(header.timestamp - clientState.latestTimeStamp < trustingPeriod)
```

spec/ics-007-tendermint-client/README.md
cwgoes: Yes, precisely.

spec/ics-007-tendermint-client/README.md
cwgoes: > The first check is ensuring that we don't allow updates to happen if the last update to the connection chain happened before our chain's current time minus the trustingPeriod. What is the purpose of this check? Seems like it will force connections to update within a specific frequency, however i don't see the danger of allowing connections to update as slow as they wish (perhaps a connection with very rare messages sent through it, may choose to update lazily?). It also seems like if this assertion ever fails, then this connection is no longer updatable at all. Is this the desired behavior?

This is the client update logic - the "trusting period" is just the "unbonding period", less some buffer of time intended to allow for the submission of evidence if misbehaviour happens.

> The second check prevents a connecting chain from updating with a header if its timestamp happens to be greater than the current timestamp. This doesn't seem to be an issue to have this check, but i don't understand the attack it is preventing.

Because it might no longer fall within the trusting period (unbonding period).

spec/ics-007-tendermint-client/README.md
cwgoes: That check is provided by lines 130 & 132. We need to check that the *local* time has not passed the trusting period because the header time could be falsified, checking the header time alone is not sufficient.

cwgoes:
I think we need to come to a clearer consensus about what should be in this spec / code & what should be in the Tendermint light client spec / code.

In particular, the [Tendermint spec](https://github.com/tendermint/spec/blob/master/spec/consensus/light-client/verification.md) and this spec duplicate some checks, we should ensure a clean separation. We also need to ensure that `Verify` (imported by the ICS 7 implementation) is deterministic (no local clock reading, instead "currentTime" should be provided as a parameter).

spec/ics-007-tendermint-client/README.md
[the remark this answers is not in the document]
melekes: @cwgoes is right (it must be strictly less) minus some delta for evidence propagation (network delay + how often blocks are created). e.g. if blocks are created once a day, then delta should be 2 days.

cwgoes:
Bump @AdityaSripal, did you see my responses above / do you have any follow-up questions?

spec/ics-007-tendermint-client/README.md
```diff
@@ -140,17 +158,17 @@ function checkMisbehaviourAndUpdateState(
...
     // check if the light client "would have been fooled"
     assert(
-      verify(evidence.fromValidatorSet, evidence.fromHeight, evidence.h1) &&
-      verify(evidence.fromValidatorSet, evidence.fromHeight, evidence.h2)
+      verify(consensusState.validatorSet, evidence.fromHeight, evidence.h1) &&
+      verify(consensusState.validatorSet, evidence.fromHeight, evidence.h2)
       )
     // set the frozen height
-    clientState.frozenHeight = evidence.h1.height // which is same as h2.height
+    clientState.frozenHeight = min(clientState.frozenHeight, evidence.h1.height) // which is same as h2.height
```
fedekunze: shouldn't this be `max`? `clientState` always starts with `frozenHeight = 0`

cwgoes: No, it should be `min`, since we want to freeze at an earlier height if misbehaviour is discovered. We should check if it is `0` though first, you're right.

fedekunze: see? the `IsFrozen()` function is handy 😜

cwgoes: indeed!

spec/ics-007-tendermint-client/README.md
```diff
@@ -114,14 +126,20 @@ Tendermint client validity checking uses the bisection algorithm described in th
 function checkValidityAndUpdateState(
   clientState: ClientState,
   header: Header) {
-    // assert that header is newer than any we know
-    assert(header.height < clientState.latestHeight)
+    // assert trusting period has not yet passed
+    assert(currentTimestamp() - clientState.latestTimestamp < trustingPeriod)
```
fedekunze:
```suggestion
    assert(currentTimestamp() - clientState.latestTimestamp < clientState.trustingPeriod)
```

spec/ics-007-tendermint-client/README.md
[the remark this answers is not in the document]
fedekunze: @AdityaSripal @cwgoes I thought by storing the full valset, it wouldn't be necessary to have the former `Committer` struct on the SDK. Please correct me if I'm wrong here

spec/ics-007-tendermint-client/README.md
fedekunze: I updated the implementation to take this into account

spec/ics-007-tendermint-client/README.md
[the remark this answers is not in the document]
AdityaSripal: Don't believe Committer is necessary regardless

cwgoes:
> I understand that the unbonding period needs to be part of ClientState to verify misbehaviour. Can trusting period simply be a default fraction of the client-specific unbonding period?

Hmm, a fraction might not be right, you generally want at least some flat amount of time to submit evidence - is there a reason not to leave them variable? We can still have defaults at a higher layer.

spec/ics-007-tendermint-client/README.md
zmanian: Debugging issues in the relayer with @jackzampolin and @AdityaSripal.

132 introduces a fine clocked clock sync between the chains on each side of the connection.

You can replace `  assert(header.timestamp <= currentTimestamp())` with what Aditya proposed above `assert(header.timestamp - clientState.latestTimeStamp < trustingPeriod)`.

This will achieve the same clock sync with the same guarantees.

PR #368, opened 2020-01-26
Standardised query & proof construction interface

cwgoes:
List the set of queries & proof construction functions which must be defined by each client type.

This is a bit high-level right now, I think it makes sense to keep it as such until we add another substantive client type (solo machines don't count) - then we can figure out the cleanest set of common abstractions (e.g. returning proofs with state queries vs. separate construction functions).

This interface is what should be used by the relayer process (e.g. [here](https://github.com/cosmos/relayer/blob/master/relayer/strategies.go)).

Closes https://github.com/cosmos/ics/issues/356

@fedekunze @jackzampolin Let me know if you think any query functionality is missing.


spec/ics-002-client-semantics/README.md
```diff
@@ -378,6 +378,110 @@ type verifyNextSequenceRecv = (
...
+type queryHeader = (
+  height: uint64
+) => Header
+```
+
+`queryChainConsensusState` MAY be defined by the chain which is validated by a particular client, to allow for the retrieval of the current consensus state which can be used to construct a new client.
+When used in this fashion, the returned `ConsensusState` MUST be manually confirmed by the querying entity, since it is subjective. This endpoint is assumed to be untrusted.
+
+```typescript
+type queryChainConsensusState = (height: uint64) => ConsensusState
```
jackzampolin: Might be helpful to define what is in the `ConsensusState` struct

spec/ics-002-client-semantics/README.md
```diff
@@ -378,6 +378,110 @@ type verifyNextSequenceRecv = (
...
+
+The `ClientState` type SHOULD expose its latest verified height (from which the consensus state can then be retrieved using `queryConsensusState` if desired).
+
+```typescript
+type latestHeight = (state: ClientState) => uint64
+```
+
+Client types SHOULD define the following standardised query functions in order to allow relayers & other off-chain entities to interface with on-chain state in a standard API.
+
+`queryConsensusState` allows stored consensus states to be retrieved by height.
```
jackzampolin:
```suggestion
`queryStoredConsensusState` allows stored consensus states to be retrieved by height.
```

spec/ics-002-client-semantics/README.md
```diff
@@ -378,6 +378,110 @@ type verifyNextSequenceRecv = (
...
+Note that retrieval of past consensus states by height (as opposed to just the current consensus state) is convenient but not required.
+
+`queryChainConsensusState` MAY also return other data necessary to create clients, such as the "unbonding period" for certain proof-of-stake security models. This data MUST also be verified by the querying entity.
+
+##### On-chain state queries
+
+This specification defines a single function to query the state of a client by-identifier.
+
+```typescript
+function queryClientState(identifier: Identifier): ClientState {
```
jackzampolin: What is in the `ClientState` struct?

spec/ics-002-client-semantics/README.md
```diff
@@ -378,6 +378,110 @@ type verifyNextSequenceRecv = (
...
+```
+
+##### Proof construction
+
+Each client type SHOULD define functions to allow relayers to construct the proofs required by the client's state verification algorithms. These may take different forms depending on the client type.
+For example, Tendermint client proofs may be returned along with key-value data from store queries, and solo client proofs may need to be constructed interactively on the solo machine in question (since the user will need to sign the message).
+These functions may constitute external queries over RPC to a full node as well as local computation or verification.
+
+```typescript
+type queryAndProveClientConsensusState = (
```
jackzampolin: Noting the transactions that each of these proofs are used in would be very helpful here.

cwgoes:
> I would like to see a tad more detail (and/or examples) on when/where these queries and proofs are used, also a bit more detail in what some of the mentioned abstractions contain, but otherwise this is a significant improvement over the existing documentation and should be merged.

Aye. I'll update the example relayer algorithm to use these query functions.

spec/ics-002-client-semantics/README.md
cwgoes: Why?

spec/ics-002-client-semantics/README.md
cwgoes: That's defined per-client-type. I'll add a note.

spec/ics-002-client-semantics/README.md
cwgoes: Client-type-specific.

spec/ics-002-client-semantics/README.md
cwgoes: That will be covered in the relayer example.

cwgoes:
- [ ] Update relayer example
- [ ] Add timeout / acknowledgement handling to relayer example

PR #375, opened 2020-02-03
Miscellaneous spec updates from review (mostly ICS 20)

cwgoes:
- [x] Standardise version string (closes https://github.com/cosmos/ics/issues/381)
- [x] Fix initialisation logic (closes https://github.com/cosmos/ics/issues/378)
- [x] Correctly handle acknowledgements of success & failure (closes https://github.com/cosmos/ics/issues/373)
- [x] Alter interface to more closely mirror implementation (or vice versa)
- [x] Clarify abort / failure distinction
- [x] Discuss multi-hop case (closes https://github.com/cosmos/ics/issues/232)


cwgoes:
Refactoring ICS 20 (spec & impl) to correctly handle acknowledgements, my current thought is that application-level specs should always use acknowledgement data to indicate failure instead of aborting the transaction (which is also possible). The reason I think this is preferable is that the acknowledgement data can contain detailed information about the failure, so that the original sender of the packet can handle it appropriately (this may mean different things based on the kind of packet & kind of failure - e.g. if the destination address was invalid, refund the tokens, but if the denomination was invalid, something went wrong with the version negotiation process, if there aren't enough tokens, there has been some sort of Byzantine behaviour somewhere in the system that needs to be handled, etc.)

Just aborting the transaction can't provide this because it's only 1 bit of data (success/failure) and can't communicate the difference between different kinds of failures which might need to be handled differently. Writing acknowledgements will require a little bit more state usage, but not much, and it seems worthwhile. Relayers can still elect not to relay acknowledgements of success (which are no-ops on the sending chain) in the optimistic all-ok path.

Does this make sense as a design pattern? Anything I missed?

cwgoes:
- [ ] Also deal with source/dest channel/port arguments required in `createOutgoingPacket` [here](https://github.com/cosmos/ics/blob/master/spec/ics-020-fungible-token-transfer/README.md#packet-relay)

cwgoes:
Per discussion with @fedekunze:

- [ ] Alter spec so that whether or not we are the source chain is deduced from the denom
- [ ] Note a bit about future registries of "root chain" for each denom & supply accounting

spec/ics-020-fungible-token-transfer/README.md
```diff
@@ -189,52 +199,63 @@ function createOutgoingPacket(
...
   FungibleTokenPacketData data = packet.data
-  if data.source {
-    // sender was source chain: mint vouchers
-    // construct receiving denomination, check correctness
-    prefix = "{packet/destPort}/{packet.destChannel}"
-    abortTransactionUnless(data.denomination.slice(0, len(prefix)) === prefix)
-    // mint vouchers to receiver (assumed to fail if balance insufficient)
-    bank.MintCoins(data.receiver, data.denomination, data.amount)
+  // inspect the denomination to determine whether or not we are the source chain
+  prefix = "{packet/destPort}/{packet.destChannel}"
```
fedekunze:
```suggestion
  prefix = "{packet.destPort}/{packet.destChannel}"
```

spec/ics-020-fungible-token-transfer/README.md
```diff
@@ -244,28 +265,40 @@ function onRecvPacket(packet: Packet): bytes {
...
+}
+```
+
+`refundTokens` is called by both `onAcknowledgePacket`, on failure, and `onTimeoutPacket`, to refund escrowed tokens to the original sender.
+
+```typescript
+function refundTokens(packet: Packet) {
   FungibleTokenPacketData data = packet.data
-  if data.source {
+  prefix = "{packet/sourcePort}/{packet.sourceChannel}"
```
fedekunze:
```suggestion
  prefix = "{packet.sourcePort}/{packet.sourceChannel}"
```

spec/ics-020-fungible-token-transfer/README.md
```diff
@@ -244,28 +265,40 @@ function onRecvPacket(packet: Packet): bytes {
...
     escrowAccount = channelEscrowAddresses[packet.destChannel]
     // construct receiving denomination, check correctness
-    prefix = "{packet/sourcePort}/{packet.sourceChannel}"
-    abortTransactionUnless(data.denomination.slice(0, len(prefix)) === prefix)
     // unescrow tokens back to sender
     bank.TransferCoins(escrowAccount, data.sender, data.denomination.slice(len(prefix)), data.amount)
   } else {
     // receiver was source chain, mint vouchers
     // construct receiving denomination, check correctness
     prefix = "{packet/sourcePort}/{packet.sourceChannel}"
```
fedekunze:
```suggestion
    prefix = "{packet.sourcePort}/{packet.sourceChannel}"
```

PR #384, opened 2020-02-25
Fix ICS 4

junha1:
`ConnectionState` can not be `CLOSED`
https://github.com/cosmos/ics/tree/master/spec/ics-003-connection-semantics#data-structures

### DIFF
--- a/spec/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/ics-004-channel-and-packet-semantics/README.md
@@ -278,7 +278,6 @@ function chanOpenInit(
 
     // optimistic channel handshakes are allowed
     abortTransactionUnless(connection !== null)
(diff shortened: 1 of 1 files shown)


cwgoes:
Ah, right, however, what we want to test is just that the connection exists in some state.

junha1:
@cwgoes Then it is ok to start the channel opening handshake while the connection is not fully establish? (`INIT` OR `TRYOPEN`)

cwgoes:
> @cwgoes Then it is ok to start the channel opening handshake while the connection is not fully established? (means being just `INIT` OR `TRYOPEN`)

Yes, this is safe because fields are immutable once `INIT` or `TRYOPEN` has happened.

junha1:
Ok, then just removing the line seems to be enough, since there is already `abortTransactionUnless(connection !== null)`.

PR #386, opened 2020-02-28
Fix missing hash() in ICS 02

junha1:
I think here `hash()` is needed.

### DIFF
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -615,7 +615,7 @@ function verifyPacketData(
   data: bytes) {
     path = applyPrefix(prefix, "ports/{portIdentifier}/channels/{channelIdentifier}/packets/{sequence}")
     abortTransactionUnless(!clientState.frozen)
-    return clientState.verifiedRoots[sequence].verifyMembership(path, data, proof)
+    return clientState.verifiedRoots[sequence].verifyMembership(path, hash(data), proof)
 }
 
 function verifyPacketAcknowledgement(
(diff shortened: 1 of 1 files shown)


junha1:
@cwgoes I don't know why, but strangely 'request review' button doesn't appear.

cwgoes:
Hmm, odd.

The lack of `hash` here is intentional; that should instead happen inside the client-specific verification function definitions, that way the implementation of `hash` can vary per-client-type.

junha1:
@cwgoes 
That really make sense. Then `verify_...()` in ICS02 has to be updated with `hash()` I think.

cwgoes:
> @cwgoes
> That really make sense. Then `verify_...()` in ICS02 has to be updated with `hash()` I think.

Yes, it probably does, I suspect I forgot to do that.

cwgoes:
Do you see anywhere `hash()` is missing? ICS 7 has it, e.g. [here](https://github.com/cosmos/ics/tree/master/spec/ics-007-tendermint-client#state-verification-functions).

junha1:
@cwgoes I found something. Regarding `verify_packet_data()` in ICS07, it hashes only `data` of packet, not together with `timeout`. I think we need to add a new parameter `timeout` in `verify_packet_data()` to fix this.

cwgoes:
> @cwgoes I found something. Regarding `verify_packet_data()` in ICS07, it hashes only `data` of packet, not together with `timeout`. I think we need to add a new parameter `timeout` in `verify_packet_data()` to fix this.

The concatenation of data & timeout occurs prior to `verifyPacketCommitment` - see `recvPacket` [here](https://github.com/cosmos/ics/blob/master/spec/ics-004-channel-and-packet-semantics/README.md). Maybe the concatenation should be performed in the client-type-specific code though, did you have a use-case in mind?

junha1:
@cwgoes, Not really, but seems to be worth it to just update.
Btw ICS 02 has to be updated because `hash()` is actually missing there, unlike ICS 07. (Regardless of concat responsibility)

cwgoes:
> @cwgoes, Not really.
> Btw ICS 02 has to be updated because `hash()` is actually missing there, unlike ICS 07. (Regardless of concat responsibility)

Aye, you're right, and thanks for the review!

junha1:
@cwgoes Thanks!

PR #390, opened 2020-03-08
Remove an unnecessary fine grained timing dependency

zmanian:
Remove an unnecessary timing dependency in validity checks.

Add some normative guidance what validity checks are unrecoverable.

### DIFF
--- a/spec/ics-007-tendermint-client/README.md
+++ b/spec/ics-007-tendermint-client/README.md
@@ -128,10 +128,10 @@ Tendermint client validity checking uses the bisection algorithm described in th
 function checkValidityAndUpdateState(
   clientState: ClientState,
   header: Header) {
-    // assert trusting period has not yet passed
+    // assert trusting period has not yet passed. This should fatally terminate a connection.
     assert(currentTimestamp() - clientState.latestTimestamp < clientState.trustingPeriod)
-    // assert header timestamp is not in the future (& transitively that is not past the trusting period)
-    assert(header.timestamp <= currentTimestamp())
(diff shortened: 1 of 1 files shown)


zmanian:
Related: https://github.com/cosmos/ics/pull/367#discussion_r389323458

spec/ics-007-tendermint-client/README.md
```diff
@@ -128,10 +128,10 @@ Tendermint client validity checking uses the bisection algorithm described in th
 function checkValidityAndUpdateState(
   clientState: ClientState,
   header: Header) {
-    // assert trusting period has not yet passed
+    // assert trusting period has not yet passed. This should fatally terminate a connection.
```
cwgoes: Right now, if the trusting period has passed, it will be impossible to update a client, so therefore impossible to use any connections associated with that client. Some sort of governance-based state recovery would be necessary.

spec/ics-007-tendermint-client/README.md
```diff
@@ -128,10 +128,10 @@ Tendermint client validity checking uses the bisection algorithm described in th
...
     assert(currentTimestamp() - clientState.latestTimestamp < clientState.trustingPeriod)
-    // assert header timestamp is not in the future (& transitively that is not past the trusting period)
-    assert(header.timestamp <= currentTimestamp())
+    // assert header timestamp is less than trust period in the future. This should be resolved with an intermediate header.
```
cwgoes: Right, although it's worth noting that the only case when this check would fail when the first succeeds is when the chain which we're running a light client of is running more quickly than the local chain, so it should be an unusual case (this is not the same as bisection, where we need intermediate headers for a different reason).

PR #392, opened 2020-03-11
Fix ICS04 to set right connection in verification

junha1:
`channelEnd.connectionHops` is supposed to mean its own connection. (not counterparty's)

It is possible to put a notable further work of 'reversed' hops semantics only if we query multiple connection first. (Doing so doesn't seem clear now I think)


junha1:
@cwgoes Sorry, but requesting review still doesn't appear for me :cry:

cwgoes:
Right, right, this needs to be revised in the future for multiple connections.

PR #398, opened 2020-03-19
Add timeout timestamp to IBC packets

cwgoes:
Closes https://github.com/cosmos/ics/issues/353


cwgoes:
- [ ] Add sentinel `0` height to disable height-based timeouts

PR #438, opened 2020-06-23
Update Chinese translation for recent changes.

en:


en:
Currently, we use [GitLocalize](https://gitlocalize.com/) to track changes in the original text and complete the translation.
Although it has some minor problems, it is still very helpful.

PR #447, opened 2020-07-08
Abstract height type

cwgoes:
Closes https://github.com/cosmos/ics/issues/439

### DIFF
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -277,6 +277,10 @@ type compare = (h1: Height, h2: Height) => Ord
 
 A height is either `LT` (less than), `EQ` (equal to), or `GT` (greater than) another height.
 
+`>=`, `>`, `===`, `<`, `<=` are defined through the rest of this specification as aliases to `compare`.
+
+There must also be a zero-element for a height type, referred to as `0`, which is less than all non-zero heights.
+
 #### ClientState
 
 `ClientState` is an opaque data structure defined by a client type.
@@ -728,7 +732,7 @@ function checkValidityAndUpdateState(
 
 function verifyClientConsensusState(
   clientState: ClientState,
-  height: uint64,
+  height: Height,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   clientIdentifier: Identifier,
@@ -740,7 +744,7 @@ function verifyClientConsensusState(
 
 function verifyConnectionState(
   clientState: ClientState,
-  height: uint64,
+  height: Height,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   connectionIdentifier: Identifier,
@@ -752,7 +756,7 @@ function verifyConnectionState(
 
 function verifyChannelState(
   clientState: ClientState,
-  height: uint64,
+  height: Height,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -765,7 +769,7 @@ function verifyChannelState(
 
 function verifyPacketData(
   clientState: ClientState,
-  height: uint64,
+  height: Height,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -779,7 +783,7 @@ function verifyPacketData(
 
 function verifyPacketAcknowledgement(
   clientState: ClientState,
-  height: uint64,
+  height: Height,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -793,7 +797,7 @@ function verifyPacketAcknowledgement(
 
 function verifyPacketAcknowledgementAbsence(
   clientState: ClientState,
-  height: uint64,
+  height: Height,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -806,7 +810,7 @@ function verifyPacketAcknowledgementAbsence(
 
 function verifyNextSequenceRecv(
   clientState: ClientState,
-  height: uint64,
+  height: Height,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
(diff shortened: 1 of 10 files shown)


spec/ics-002-client-semantics/README.md
```diff
@@ -256,6 +256,27 @@ Once misbehaviour is detected, clients SHOULD be frozen so that no future update
...
+
+`Height` is an opaque data structure defined by a client type.
+It must form a partially ordered set & provide operations for comparison.
+
+```typescript
+type Height
+```
+
+```typescript
+enum Ord {
```
fedekunze, line 269: can we rename it to  `Cmp` instead? it's more widely used by various libraries in Go (eg big.Int)

fedekunze: actually, I think Cmp should be a function from `Height`

```go
func (h Height) Cmp(b Height): int {
  if h.epochNumber < b.epochNumber {
    return -1
  } else if h.epochNumber == b.epochNumber {
    if a.epochHeight < b.epochHeight {
      return -1
    } else if h.epochHeight == b.epochHeight {
      return 0
    }
  }
  return 1
}
```

something in those lines

fedekunze: https://golang.org/pkg/math/big/#Int.Cmp

cwgoes: Hmm, I'm not a huge fan of returning an `int` since there are more valid values for `int` than valid return values for this function. It's fine if we use `Cmp` to implement this in the Cosmos SDK code, though.

cwgoes:
Probably makes sense to add https://github.com/cosmos/ics/issues/445#issuecomment-655603487 here as well.

cwgoes:
We'll have to fix a key & commitment prefix for upgrade handling.

cwgoes:
This is not yet ready to be merged, I need to check that the fields all match up to the ICS 7 implementation, but want to check that the upgrade method proposed here makes sense, cc @adityasripal.

spec/ics-007-tendermint-client/README.md
```diff
@@ -200,18 +249,48 @@ function checkMisbehaviourAndUpdateState(
...
+    // check proof of updated client state in state at predetermined commitment prefix and key
+    path = applyPrefix(clientState.upgradeCommitmentPrefix, clientState.upgradeKey)
+     // check that the client is at a sufficient height
+    assert(clientState.latestHeight >= height)
+    // check that the client is unfrozen or frozen at a higher height
+    assert(clientState.frozenHeight === null || clientState.frozenHeight > height)
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that the provided consensus state has been stored
+    assert(root.verifyMembership(path, clientState, proof))
```
AdityaSripal, line 273: Should this be `newClientState`?

```suggestion
    assert(root.verifyMembership(path, newClientState, proof))
```

spec/ics-007-tendermint-client/README.md
```diff
@@ -200,18 +249,48 @@ function checkMisbehaviourAndUpdateState(
...
+
+As the client state change will be performed immediately, once the new client state information is written to the predetermined key, the client will no longer be able to follow blocks on the old chain, so it must upgrade promptly.
+
+```typescript
+function upgradeClientState(
+  clientState: ClientState,
+  newClientState: ClientState,
+  height: Height,
+  proof: CommitmentPrefix) {
+    // check proof of updated client state in state at predetermined commitment prefix and key
```
AdityaSripal, line 264: How would we know what the updated client state is going to be before the upgrade, since we don't know the first header.

This could be fixed with the comment here: https://github.com/cosmos/cosmos-sdk/issues/6736#issuecomment-659374828

Tho we still wouldn't know something like `LatestTimestamp uint64` beforehand

spec/ics-007-tendermint-client/README.md
```diff
@@ -200,18 +249,48 @@ function checkMisbehaviourAndUpdateState(
...
+    path = applyPrefix(clientState.upgradeCommitmentPrefix, clientState.upgradeKey)
+     // check that the client is at a sufficient height
+    assert(clientState.latestHeight >= height)
```
AdityaSripal, line 267: Should we make this check more specific in implementation. (i.e. EpochNumber must be incremented)

AdityaSripal:
The Tendermint ClientState contains things that we would only know after the upgrade happened
Currently it contains a Header
But as pointed out in https://github.com/cosmos/cosmos-sdk/issues/6736 this is unnecessary, but we would still need a field like LatestTimestamp  which we don't know ahead of time
So we cannot store it exactly on the old chain before upgrade
However, we could just store with that field zeroed out, and when we go to verify we set the newClientState to zero for verification purposes
and maybe just check that `newClientState.Timestamp > clientstate.TimeStamp`

spec/ics-007-tendermint-client/README.md
cwgoes: Ah, yes, thanks.

cwgoes:
> but we would still need a field like LatestTimestamp which we don't know ahead of time

Hmm, the problem here is that `latestTimestamp` is required to determine whether a header is or is not within the trusting period. In the case of an upgrade, to preserve the same security model, we still want client updates to fail if they aren't submitted within the trusting period *from when the new chain starts*. Given that `newClientState` is expected to be written just when the old version halts, I think it is reasonable to expect an estimate to be put in for `latestTimestamp` (e.g. an hour into the future, depending on how long the restart is expected to take).

spec/ics-007-tendermint-client/README.md
cwgoes: Yes, added that check, thanks.

spec/ics-007-tendermint-client/README.md
cwgoes: See below comment.

cwgoes:
Just this one remaining question on client state - https://github.com/cosmos/cosmos-sdk/issues/6736#issuecomment-660868766.

PR #449, opened 2020-07-16
Update ics-027

Thunnini:
I added the packet's proto definition.
And, I updated the packet 
Register and runtx packet has been merged as one packet.
Salt is no longer required to create an Interchain account, and allows the destination chain to define how the account is created. The address of the account is returned as an acknowledgement packet.


spec/ics-027-interchain-accounts/packets.proto
```diff
@@ -0,0 +1,19 @@
+syntax = 'proto3';
+
+enum Type {
```
cwgoes, line 3: You can use `oneof` in proto, it is probably cleaner

spec/ics-027-interchain-accounts/README.md
```diff
@@ -50,88 +50,111 @@ It does not restrict how you can distinguish signers that was not made by the co
...
-`RegisterIBCAccountPacketData` is used by the counterparty chain to register an account. An interchain account's address is defined deterministically with the channel identifier and salt. The ```generateAccount``` method is used to generate a new interchain account's address. It is recommended to generate address by ```hash(identifier+salt)```, but other methods may be used. This function must generate a unique and deterministic address by utilising identifier and salt.
+`IBCAccountPacketData` specifies the type of packet. The `IBCAccountPacketData.data` is arbitrary and can be used for different purposes depending on the value of `IBCAccountPacketData.type`.
+
+When the value of `type` is `REGISTER`, `IBCAccountPacketData.data` can be used as information by which the account is created on the host chain. When the value of `type` is `RUNTX`, `IBCAccountPacketData.data` contains the tx bytes.
 
 ```typescript
-interface RegisterIBCAccountPacketData {
-  salt: Uint8Array
+// `Type` enumerator defines the packet type.
+enum Type {
```
cwgoes, line 84: You can probably use `oneof`, see below comment

spec/ics-027-interchain-accounts/README.md
```diff
@@ -210,54 +233,94 @@ function onChanCloseConfirm(
...
+      return IBCAccountPacketAcknowledgement{
+        type: Type.REGISTER,
+        chainID,
+        code: 1,
+        data: [],
+        error: e.message,
+      }
+    }
+    // Return ack with generated address.
+    return IBCAccountPacketAcknowledgement{
```
cwgoes, line 259: Put this in the `try` block, it's cleaner flow-wise

spec/ics-027-interchain-accounts/README.md
```diff
@@ -210,54 +233,94 @@ function onChanCloseConfirm(
...
+    }
 
-  if (packet.data is RegisterIBCAccountPacketData) {
-    RegisterIBCAccountPacketData data = packet.data
-    identifier = "{packet/sourcePort}/{packet.sourceChannel}"
-    const address = generateAddress(identifier, packet.salt)
-    createAccount(address)
-    // Return generated address.
-    return address
+    return IBCAccountPacketAcknowledgement{
```
cwgoes, line 283: Ditto

spec/ics-027-interchain-accounts/packets.proto
Thunnini: I think that the all fields in this definition can be used mutually, so I think the enumerator is enough.

spec/ics-027-interchain-accounts/README.md
Thunnini: I think that the all fields in this definition can be used mutually, so I think the enumerator is enough.

spec/ics-027-interchain-accounts/README.md
Thunnini: Ok

spec/ics-027-interchain-accounts/README.md
Thunnini: Ok

spec/ics-027-interchain-accounts/packets.proto
cwgoes: Hmm, alright

spec/ics-027-interchain-accounts/README.md
Thunnini: I fixed it.

spec/ics-027-interchain-accounts/README.md
Thunnini: I fixed it.

PR #464, opened 2020-08-05
Add definition for Acknowledgement Envelope

### LINKED ISSUE #461: Provide standard Result<success, error> acknowlegement packet envelope
**Problem**

When I send a packet, there are 2 possible results in the eyes of IBC:

* Timeout - OnTimeout
* Return Packet (success or error) - OnAcknowledgePacket

For my application, there are also 2 cases I need to handle (and these are what most clients care about):

* Packet Succeeded (with some result data)
* Packet failed (application error or timeout)

These are not the same cases and we need to do some protocol-specific packet inspection to map the IBC `OnAcknowledgePacket` call to success or failure.

If I build a IBC packet visualizer, that can handle dynamic protocols, which makes most sense to display to the user: "delivered / timed out" or "succeeded / failed"? Sure "Succeed / failed / timed out" may be the most desirable, but clearly the key information is whether the packet was accepted (and transfer made).

**Proposal**

I do not want to change the IBC protocol, but rather add some standard conventions on top of it that will enable better tooling for all users. Currently, [`acknowledgePacket`](https://github.com/cosmos/ics/tree/master/spec/ics-004-channel-and-packet-semantics#acknowledgements) receives `acknowledgement: bytes` for both success and failure cases. Where `timeoutPacket` receives no bytes. While `timeoutPacket` can easily be mapped to "error", we would need a standard envelope for the acknowledgement bytes to do so.

I propose the following standard as a protobuf format:

```proto
message Acknowledgement {
  option (gogoproto.equal) = true;
  option (gogoproto.goproto_stringer) = false;
  oneof response {
    bytes result = 21;
    string error = 22;
  }
}
```

I choose the fields 21 and 22 to not accidentally conflict with some other, unrelated format. If `len(success) > 0 && len(error) == 0` then this is a clear success with `msg.success` representing any app-specific data to be parsed down the line. If `len(success) == 0 && len(error) > 0` then this is a clear error with some error string, which may be displayed. If neither of those are true, then we cannot introspect this packet and leave it as "unknown" (for relayers / visualizers) or use app-specific parsing methods (for the on-chain logic).

An observer can then map `timeoutPacket` to `Acknowledgement { error: "timeout" }` and all packet responses can be marked as success/green with some opaque app-data, or failure/red with some human-readable error message. Furthermore, we could provide a helper function to do this mapping and then directly call success/failure callbacks, which would slightly simplify the logic in the modules, or at least [ensure they do they same thing on error and timeout](https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc-transfer/keeper/relay.go#L186-L201)

**Related Work**

This is similar to the case that there is no enforced format to the [version identifier in ICS3](https://github.com/cosmos/ics/tree/master/spec/ics-003-connection-semantics#data-structures), but it is [recommended to use](https://github.com/cosmos/cosmos-sdk/blob/master/docs/ibc/custom.md#channel-handshake-version-negotiation) the [version/feature format](https://github.com/cosmos/cosmos-sdk/blob/master/x/ibc/03-connection/types/version.go) defined in the cosmos-sdk (https://github.com/cosmos/cosmos-sdk/pull/6640). 

I think this level of "recommended types" really belongs in the ICS repo first, with cosmos-sdk following it (or at least backported to the specs as "recommended formats"). Thus, I make the issue on the ics repo.

**Next Steps**

If this issue is generally accepted, I would write a PR adding this opt-in protocol to ics4 and define ics20 to make use of it. 

After that is finalized, I would prepare a PR to update ics20 in the cosmos-sdk (unless someone else wishes to), so it uses the defined envelope. 

I would also add this logic to the upcoming ibc introspection tools in cosmjs

**History**

The proto format has been updated, the original proto proposal we refer to below was:

```proto
message Acknowledgement {
  option (gogoproto.equal) = true;
  option (gogoproto.goproto_stringer) = false;
  bytes success = 21;
  string error = 22;
}
```

ethanfrey:
Closes #461 

This adds a recommended Acknowledgement envelope to the packet specification along with the definition of the acknowledgement algorithm.

I propose this under the RFC2119 keyword "SHOULD"

> 3. SHOULD   This word, or the adjective "RECOMMENDED", mean that there
>   may exist valid reasons in particular circumstances to ignore a
>   particular item, but the full implications must be understood and
>   carefully weighed before choosing a different course.

If you disagree with that, it could be reduced to "MAY":

> 5. MAY   This word, or the adjective "OPTIONAL", mean that an item is
>    truly optional.  One vendor may choose to include the item because a
>    particular marketplace requires it or because the vendor feels that
>    it enhances the product while another vendor may omit the same item.
>    An implementation which does not include a particular option MUST be
>    prepared to interoperate with another implementation which does
>    include the option, though perhaps with reduced functionality. In the
>    same vein an implementation which does include a particular option
>    MUST be prepared to interoperate with another implementation which
>    does not include the option (except, of course, for the feature the
>    option provides.)

However, I think much of the utility for allowing intelligent client-side code in a dynamic IBC world would be reduced by watering it down from "SHOULD". Likewise I feel it is not so essential to warrant MUST/SHALL.

### DIFF
--- a/spec/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/ics-004-channel-and-packet-semantics/README.md
@@ -681,6 +681,30 @@ function acknowledgePacket(
 }
 ```
 
+##### Acknowledgement Envelope
+
+The acknowledgement returned from the remote chain is defined as arbitrary bytes in the ics protocol. This data
+may either encode a successful execution or a failure (anything besides a timeout). There is no generic way to
+distinguish the two cases, which require any client-side packet visualizer to understand every app-specific protocol
+in order to distinguish the case of successful or failed relay. In order to reduce this issue, we offer additional
+specification for acknowledgement formats, which [SHOULD](https://www.ietf.org/rfc/rfc2119.txt) be used by the 
+app-specific protocols.
+
+```proto
+message Acknowledgement {
+  oneof response {
+    bytes result = 21;
+    string error = 22;
+  }
+}
+```
+
+If an application uses a different format for acknowledge bytes, it MUST not deserialize to a valid protobuf message
+of this format. Note that all packets contain exactly one non-empty field, and it must be result or error.  The field
+numbers 21 and 22 were explicitly chosen to avoid accidental conflicts with other protobuf message formats used
+for acknowledgements. The first byte of any message with this format will be the non-ascii values `0xaa` (result) 
+or `0xb2` (error).
+
 #### Timeouts
 
 Application semantics may require some timeout: an upper limit to how long the chain will wait for a transaction to be processed before considering it an error. Since the two chains have different local clocks, this is an obvious attack vector for a double spend - an attacker may delay the relay of the receipt or wait to send the packet until right after the timeout - so applications cannot safely implement naive timeout logic themselves.
--- a/spec/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/ics-004-channel-and-packet-semantics/README.md
@@ -683,10 +683,10 @@ function acknowledgePacket(
 
 ##### Acknowledgement Envelope
 
-The acknowledgement returned from the remote chain is defined as arbitrary bytes in the ics protocol. This data
+The acknowledgement returned from the remote chain is defined as arbitrary bytes in the IBC protocol. This data
 may either encode a successful execution or a failure (anything besides a timeout). There is no generic way to
-distinguish the two cases, which require any client-side packet visualizer to understand every app-specific protocol
-in order to distinguish the case of successful or failed relay. In order to reduce this issue, we offer additional
+distinguish the two cases, which requires that any client-side packet visualizer understands every app-specific protocol
+in order to distinguish the case of successful or failed relay. In order to reduce this issue, we offer an additional
 specification for acknowledgement formats, which [SHOULD](https://www.ietf.org/rfc/rfc2119.txt) be used by the 
 app-specific protocols.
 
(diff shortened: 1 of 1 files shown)


spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -681,6 +681,30 @@ function acknowledgePacket(
 }
 ```
 
+##### Acknowledgement Envelope
+
+The acknowledgement returned from the remote chain is defined as arbitrary bytes in the ics protocol. This data
```
cwgoes, line 686:
```suggestion
The acknowledgement returned from the remote chain is defined as arbitrary bytes in the IBC protocol. This data
```

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -681,6 +681,30 @@ function acknowledgePacket(
...
+may either encode a successful execution or a failure (anything besides a timeout). There is no generic way to
+distinguish the two cases, which require any client-side packet visualizer to understand every app-specific protocol
```
cwgoes, line 688:
```suggestion
distinguish the two cases, which requires that any client-side packet visualizer understands every app-specific protocol
```

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -681,6 +681,30 @@ function acknowledgePacket(
...
+in order to distinguish the case of successful or failed relay. In order to reduce this issue, we offer additional
```
cwgoes, line 689:
```suggestion
in order to distinguish the case of successful or failed relay. In order to reduce this issue, we offer an additional
```

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -681,6 +681,30 @@ function acknowledgePacket(
...
+```proto
+message Acknowledgement {
+  oneof response {
+    bytes result = 21;
+    string error = 22;
+  }
+}
+```
+
+If an application uses a different format for acknowledge bytes, it MUST not deserialize to a valid protobuf message
```
cwgoes, line 702:
```suggestion
If an application uses a different format for acknowledgement bytes, it MUST not deserialize to a valid protobuf message
```

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -681,6 +681,30 @@ function acknowledgePacket(
...
+of this format. Note that all packets contain exactly one non-empty field, and it must be result or error.  The field
+numbers 21 and 22 were explicitly chosen to avoid accidental conflicts with other protobuf message formats used
+for acknowledgements. The first byte of any message with this format will be the non-ascii values `0xaa` (result)
```
cwgoes, line 705:
```suggestion
for acknowledgements. The first byte of any message with this format will be the non-ASCII values `0xaa` (result) 
```

ethanfrey:
Okay, accepted those changes, nice improvements.

Issue #472, opened 2020-08-26
Adding Japanese translation


3100:
Hi,
My colleague and I translated several specs into Japanese. Can I send them as a PR?

cwgoes:
Absolutely!

cwgoes:
Closed by #473.

3100:
@cwgoes Thanks!

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #473, opened 2020-08-26
Adding Japanese Translation

3100:
This PR adds Japanese translation for several specs and `/ibc` docs.

cf. #472


cwgoes (approved):
ACK (although I can't review the Japanese, alas)

PR #477, opened 2020-08-27
Rectify ICS 007 between Golang implementation / spec

cwgoes:
Closes https://github.com/cosmos/ics/issues/456

Not yet totally sure what to do about the `SignedHeader` type.


zramsay:
ping @colin-axner

PR #479, opened 2020-09-01
ICS3 protocol fix for version negotiation 

### LINKED ISSUE #459: ICS3 handshake version negotiation problem
# Overview

It may be possible that the connection handshake protocol aborts at a chain during step connOpenTry if this chain previously executed step connOpenInit.

# Execution example

Imagine the following scenario for establishing a connection between chains Alice and Bob, consisting of 4 steps:

1. chain __Alice__ executes __connOpenInit__
    - Alice stores a `ConnectionEnd` object where the field `version <- X` (`X` is a list of versions that Alice supports, obtained from `getCompatibleVersions()`)
    - the other fields are not relevant

2. chain __Bob__ executes __connOpenInit__
    - Bob stores a `ConnectionEnd` object where the field `version <- Y` (`Y` is the list of versions that Bob supports, obtained from `getCompatibleVersions()` on Bob's side)
    - suppose this is for the same connection that Alice executed at step 1 above

3. a correct relayer constructs a connOpenTry datagram to further progress in the connection handshake
     - assume that the destination of this datagram is Bob
     - the field `counterpartyVersions` in this datagram consists of list `X` (again, this is the list of versions that Alice supports)

4. chain __Bob__ delivers and executes __connOpenTry__
     - recall that Bob already stores a `ConnectionEnd` with `version === Y` (step 2 above)
     - Bob will abort execution of connOpenTry handler because the following check (marked with `~~>`) fails:

```typescript
abortTransactionUnless(
      (previous === null) ||
      (previous.state === INIT &&
        previous.counterpartyConnectionIdentifier === counterpartyConnectionIdentifier &&
        previous.clientIdentifier === clientIdentifier &&
        previous.counterpartyClientIdentifier === counterpartyClientIdentifier &&
   ~~>  previous.version === version))
```

### Further clarifications:

Bob aborts execution of connOpenTry only if `X =/= Y`, i.e., only if the two lists of versions `X` and `Y` returned from  `getCompatibleVersions()` are different at the two chains. If we assume that the two chains always return the same list of versions, then this handler should not abort, so there should not be a problem.

Alice would also abort execution if she executes the __connOpenTry__ handler by the same reasoning as above: Alice stores a `ConnectionEnd` object in the variable `previous` with `version === X` and she is comparing this against `Y`.

# Fix

It seems that removing the verification `previous.version === version` would fix this problem.

adizere:
Solution candidate to close #459.

For a full discussion on the underlying problem, see [ibc-rs/disclosure-log](https://github.com/informalsystems/ibc-rs/blob/master/docs/disclosure-log.md#3-ics3-problems-due-to-version-negotiation).

### DIFF
--- a/spec/ics-003-connection-semantics/README.md
+++ b/spec/ics-003-connection-semantics/README.md
@@ -245,7 +245,7 @@ In the future, it might be used to indicate what kinds of channels can utilise t
 what encoding formats channel-related datagrams will use. At present, host state machine MAY utilise the version data
 to negotiate encodings, priorities, or connection-specific metadata related to custom logic on top of IBC.
 
-Host state machines MAY also safely ignore the version data or specify an empty string. It is assumed that the two chains running the opening handshake have at least once compatible version in common (i.e., the compatible versions at the two chains must have a non-empty intersection).
+Host state machines MAY also safely ignore the version data or specify an empty string. It is assumed that the two chains running the opening handshake have at least one compatible version in common (i.e., the compatible versions of the two chains must have a non-empty intersection). If the two chains do not have any mutually acceptable versions, the handshake will fail.
 
 An implementation MUST define a function `getCompatibleVersions` which returns the list of versions it supports, ranked by descending preference order.
 
--- a/spec/ics-003-connection-semantics/README.md
+++ b/spec/ics-003-connection-semantics/README.md
@@ -245,7 +245,7 @@ In the future, it might be used to indicate what kinds of channels can utilise t
 what encoding formats channel-related datagrams will use. At present, host state machine MAY utilise the version data
 to negotiate encodings, priorities, or connection-specific metadata related to custom logic on top of IBC.
 
-Host state machines MAY also safely ignore the version data or specify an empty string.
+Host state machines MAY also safely ignore the version data or specify an empty string. It is assumed that the two chains running the opening handshake have at least once compatible version in common (i.e., the compatible versions at the two chains must have a non-empty intersection).
 
 An implementation MUST define a function `getCompatibleVersions` which returns the list of versions it supports, ranked by descending preference order.
 
@@ -253,7 +253,7 @@ An implementation MUST define a function `getCompatibleVersions` which returns t
 type getCompatibleVersions = () => []string
 ```
 
-An implementation MUST define a function `pickVersion` to choose a version from a list of versions proposed by a counterparty.
+An implementation MUST define a function `pickVersion` to choose a version from a list of versions.
 
 ```typescript
 type pickVersion = ([]string) => string
--- a/spec/ics-003-connection-semantics/README.md
+++ b/spec/ics-003-connection-semantics/README.md
@@ -320,7 +320,8 @@ function connOpenTry(
     expectedConsensusState = getConsensusState(consensusHeight)
     expected = ConnectionEnd{INIT, desiredIdentifier, getCommitmentPrefix(), counterpartyClientIdentifier,
                              clientIdentifier, counterpartyVersions}
-    version = pickVersion(counterpartyVersions)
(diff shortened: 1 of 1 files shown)


spec/ics-003-connection-semantics/README.md
```diff
@@ -245,15 +245,15 @@ In the future, it might be used to indicate what kinds of channels can utilise t
 what encoding formats channel-related datagrams will use. At present, host state machine MAY utilise the version data
 to negotiate encodings, priorities, or connection-specific metadata related to custom logic on top of IBC.
 
-Host state machines MAY also safely ignore the version data or specify an empty string.
+Host state machines MAY also safely ignore the version data or specify an empty string. It is assumed that the two chains running the opening handshake have at least once compatible version in common (i.e., the compatible versions at the two chains must have a non-empty intersection).
```
cwgoes, line 248:
```suggestion
Host state machines MAY also safely ignore the version data or specify an empty string. It is assumed that the two chains running the opening handshake have at least one compatible version in common (i.e., the compatible versions of the two chains must have a non-empty intersection). If the two chains do not have any mutually acceptable versions, the handshake will fail.
```

spec/ics-003-connection-semantics/README.md
```diff
@@ -353,7 +354,9 @@ function connOpenAck(
   consensusHeight: Height) {
     abortTransactionUnless(consensusHeight < getCurrentHeight())
     connection = provableStore.get(connectionPath(identifier))
-    abortTransactionUnless(connection.state === INIT || connection.state === TRYOPEN)
+    abortTransactionUnless(
+        (connection.state === INIT && connection.version.indexOf(version) !== -1)
```
cwgoes, line 358: it can't both be equal and `indexOf`, I think you want `indexOf`

cwgoes (changes requested):

spec/ics-003-connection-semantics/README.md
adizere: Not sure I understand your comment.

This conditions is trying to ensure that: If the (current) state of the connection is `INIT`, then the incoming argument `version` must exist inside the list of versions `connection.version`. Agree this is not the ideal way to code this check.

cwgoes: Oh, never mind, just got confused reading the diff

cwgoes (approved):
ACK

I think it may be worth noting that if the two chains implement different `pickVersion` functions, a (possibly misbehaving) relayer may be able to stall the handshake by doing INIT - OPENTRY on both chains (at which point they will pick different versions and be unable to continue).

spec/ics-003-connection-semantics/README.md
```diff
@@ -245,15 +245,15 @@ In the future, it might be used to indicate what kinds of channels can utilise t
...
+Host state machines MAY also safely ignore the version data or specify an empty string. It is assumed that the two chains running the opening handshake have at least one compatible version in common (i.e., the compatible versions of the two chains must have a non-empty intersection). If the two chains do not have any mutually acceptable versions, the handshake will fail.
 
 An implementation MUST define a function `getCompatibleVersions` which returns the list of versions it supports, ranked by descending preference order.
 
 ```typescript
 type getCompatibleVersions = () => []string
 ```
 
-An implementation MUST define a function `pickVersion` to choose a version from a list of versions proposed by a counterparty.
+An implementation MUST define a function `pickVersion` to choose a version from a list of versions.
```
cwgoes, line 256:
```suggestion
An implementation MUST define a function `pickVersion` to choose a version from a list of versions. Note that if the two chains performing the handshake implement different `pickVersion` functions, a (possibly misbehaving) relayer may be able to stall the handshake by executing `INIT` and `OPENTRY` on both chains, at which point they will pick different versions and be unable to continue.
```

cwgoes (approved):

PR #480, opened 2020-09-04
Asynchronous acknowledgements

cwgoes:
Closes https://github.com/cosmos/ics/issues/478

- [x] Separate keys
- [x] `writeAcknowledgement` function
- [x] Update ICS 18 (relayer logic)
- [x] Search for any lingering references to old functions
- [x] Read over again
- [x] Fix the PDF `make` process (somehow broken)

### DIFF
--- a/spec/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/ics-004-channel-and-packet-semantics/README.md
@@ -561,9 +561,9 @@ function sendPacket(packet: Packet) {
 
 #### Receiving packets
 
-The `recvPacket` function is called by a module in order to receive & process an IBC packet sent on the corresponding channel end on the counterparty chain.
+The `recvPacket` function is called by a module in order to receive an IBC packet sent on the corresponding channel end on the counterparty chain.
 
-Calling modules MUST execute application logic atomically in conjunction with calling `recvPacket`, likely beforehand to calculate the acknowledgement value.
+Atomically in conjunction with calling `recvPacket`, calling modules MUST either execution application logic or queue the packet for future execution.
 
 The IBC handler performs the following steps in order:
 
@@ -573,7 +573,7 @@ The IBC handler performs the following steps in order:
 - Checks that the packet sequence is the next sequence the channel end expects to receive (for ordered channels)
 - Checks that the timeout height has not yet passed
 - Checks the inclusion proof of packet data commitment in the outgoing chain's state
-- Sets the opaque acknowledgement value at a store path unique to the packet (if the acknowledgement is non-empty or the channel is unordered)
+- Sets a store path to indicate that the packet has been received (unordered channels only)
 - Increments the packet receive sequence associated with the channel end (ordered channels only)
 
 ```typescript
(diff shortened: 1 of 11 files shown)


cwgoes:
Still trying to fix PDF creation but the spec changes are ready for review.

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -553,9 +559,9 @@ function sendPacket(packet: Packet) {
 
 #### Receiving packets
 
-The `recvPacket` function is called by a module in order to receive & process an IBC packet sent on the corresponding channel end on the counterparty chain.
+The `recvPacket` function is called by a module in order to receive an IBC packet sent on the corresponding channel end on the counterparty chain.
 
-Calling modules MUST execute application logic atomically in conjunction with calling `recvPacket`, likely beforehand to calculate the acknowledgement value.
+Atomically in conjunction with calling `recvPacket`, calling modules MUST either execution application logic or queue the packet for future execution.
```
colin-axner, line 564:
```suggestion
Atomically in conjunction with calling `recvPacket`, calling modules MUST either execute application logic or queue the packet for future execution.
```

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -599,41 +604,75 @@ function recvPacket(
...
-
     if (channel.order === ORDERED) {
       nextSequenceRecv = provableStore.get(nextSequenceRecvPath(packet.destPort, packet.destChannel))
       abortTransactionUnless(packet.sequence === nextSequenceRecv)
       nextSequenceRecv = nextSequenceRecv + 1
       provableStore.set(nextSequenceRecvPath(packet.destPort, packet.destChannel), nextSequenceRecv)
-    } else
-      abortTransactionUnless(provableStore.get(packetAcknowledgementPath(packet.destPort, packet.destChannel, packet.sequence) === null))
+    } else {
+      // for unordered channels we must set the receipt so it can be verified on the other side
```
colin-axner, line 613: I know this isn't needed for correctness since it is implicit once you understand the spec, but I don't see any references saying the receipt is just a bit and doesn't contain any other information. I could see a new reader thinking the receipt should be provided by the calling module

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -599,41 +604,75 @@ function recvPacket(
...
+      abortTransactionUnless(provableStore.get(packetReceiptPath(packet.destPort, packet.destChannel, packet.sequence) === null))
+      provableStore.set(
+        packetReceiptPath(packet.destPort, packet.destChannel, packet.sequence),
+        ""
```
colin-axner, line 617: the quotes seem strange to me. maybe add a comment why it can be empty?

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -599,41 +604,75 @@ function recvPacket(
...
-#### Acknowledgements
+#### Writing acknowledgements
+
+The `writeAcknowledgement` function is called by a module in order to write data which resulted from processing an IBC packet that the sending chain can then verify, a sort of "execution receipt" or "RPC call response".
+
+Calling modules MUST execute application logic atomically in conjunction with calling `writeAcknowledgement`.
+
+This is an asynchronous acknowledgement, the contents of which do not need to be determined when the packet is received, only when processing is complete. In the synchronous case, `writeAcknowledgement` can be called in the same transaction (atomically) with `recvPacket`.
+
+Acknowledging packets is not required; however, if an ordered channel uses acknowledgements, either all or no packets must be acknowledged (since the acknowledgements are processed in order).
```
colin-axner, line 638: One thing that is kinda annoying here is packet commitments will never be deleted and basic relayers will continually ask chainB if they have written an ack for a packet commitment. It might be useful to someone allow modules to signal if they plan to write acks or not

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -599,41 +604,75 @@ function recvPacket(
...
+    abortTransactionUnless(provableStore.get(packetAcknowledgementPath(packet.destPort, packet.destChannel, packet.sequence) === null))
+
+    // write the acknowledgement
+    provableStore.set(
+      packetAcknowledgementPath(packet.destPort, packet.destChannel, packet.sequence),
+      hash(acknowledgement)
+    )
+
+    // log that a packet has been acknowledged
+    emitLogEntry("writeAcknowledgement", {sequence: packet.sequence, timeoutHeight: packet.timeoutHeight,
```
colin-axner, line 663: shouldn't it emit the channel end port ID and channel ID the ack was written to (ie packet.destPort, packet.destChannel)?

colin-axner (approved):
LGTM, just left some suggestions for clarity

spec/ics-004-channel-and-packet-semantics/README.md
cwgoes: Added a comment.

spec/ics-004-channel-and-packet-semantics/README.md
cwgoes: Aye, added a comment.

spec/ics-004-channel-and-packet-semantics/README.md
cwgoes: Added a comment to this effect; I think this we will leave for future versions of the protocol.

spec/ics-004-channel-and-packet-semantics/README.md
cwgoes: Aye, added.

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -599,41 +604,77 @@ function recvPacket(
...
 ```
 
-#### Acknowledgements
+#### Writing acknowledgements
+
+The `writeAcknowledgement` function is called by a module in order to write data which resulted from processing an IBC packet that the sending chain can then verify, a sort of "execution receipt" or "RPC call response".
+
+Calling modules MUST execute application logic atomically in conjunction with calling `writeAcknowledgement`.
+
+This is an asynchronous acknowledgement, the contents of which do not need to be determined when the packet is received, only when processing is complete. In the synchronous case, `writeAcknowledgement` can be called in the same transaction (atomically) with `recvPacket`.
```
ethanfrey, line 638: Okay, so in the "normal" case we have today, we just call `writeAcknowledgement` in the OnRecv handler rather than return the ack bytes, correct? 

This may cause hard to debug errors for people's code if they don't follow this issue. It would be good to maintain a simple migration guide for all IBC changes that affect people building on IBC. (I asked for a migration guide to the sdk as well. At least IBC is more or less stable since some weeks, and makes sense to keep a diff of what changed since some recent point in time)

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -565,15 +571,14 @@ The IBC handler performs the following steps in order:
 - Checks that the packet sequence is the next sequence the channel end expects to receive (for ordered channels)
 - Checks that the timeout height has not yet passed
 - Checks the inclusion proof of packet data commitment in the outgoing chain's state
-- Sets the opaque acknowledgement value at a store path unique to the packet (if the acknowledgement is non-empty or the channel is unordered)
+- Sets a store path to indicate that the packet has been received (unordered channels only)
```
ethanfrey, line 574: Don't we need to set recvPacket all the time now? As we do not rely on acknowledgements?

If a synchronous acknowledgement obviates the need for `recvPacket` then this should specify that either acknowledgement or receipt must be written atomically with `onRecv`.

Or maybe you can explain why I misunderstand this.

ethanfrey (commented):
A few questions, so I can update our IBC modules properly.

spec/ics-004-channel-and-packet-semantics/README.md
cwgoes: That's correct; to replicate the current behaviour simply call `writeAcknowledgement` synchronously.

spec/ics-004-channel-and-packet-semantics/README.md
cwgoes: `recvPacket` always needs to be called, as usual, regardless of whether or not an acknowledgement will be written.

cwgoes:
This build issue seems to be related to https://github.com/sergiocorreia/panflute/issues/142; downgrading my `pandoc` for now.

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -207,6 +207,14 @@ function packetCommitmentPath(portIdentifier: Identifier, channelIdentifier: Ide
 
 Absence of the path in the store is equivalent to a zero-bit.
 
+Packet receipt data are stored under the `packetReceiptPath`
+
+```typescript
+function packetReceiptPath(portIdentifier: Identifier, channelIdentifier: Identifier, sequence: uint64): Path {
+    return "acks/ports/{portIdentifier}/channels/{channelIdentifier}/receipts/" + sequence
```
fedekunze, line 214: the prefix is wrong here, should be `receipts` or smth else

cwgoes: Thanks! Fixed in https://github.com/cosmos/ics/commit/46a40a5910fb9e193dccdac48259cee37b34b904 to `receipts`.

PR #482, opened 2020-09-15
Flexible identifier selection

cwgoes:
Flexible identifier & version selection.

Closes https://github.com/cosmos/ics/issues/462
Closes https://github.com/cosmos/ics/issues/465

### DIFF
--- a/spec/ics-003-connection-semantics/README.md
+++ b/spec/ics-003-connection-semantics/README.md
@@ -306,6 +306,7 @@ function connOpenInit(
 ```typescript
 function connOpenTry(
   desiredIdentifier: Identifier,
+  provedIdentifier: Identifier,
   counterpartyConnectionIdentifier: Identifier,
   counterpartyPrefix: CommitmentPrefix,
   counterpartyClientIdentifier: Identifier,
@@ -318,7 +319,11 @@ function connOpenTry(
     abortTransactionUnless(validateConnectionIdentifier(desiredIdentifier))
     abortTransactionUnless(consensusHeight < getCurrentHeight())
     expectedConsensusState = getConsensusState(consensusHeight)
-    expected = ConnectionEnd{INIT, desiredIdentifier, getCommitmentPrefix(), counterpartyClientIdentifier,
+    abortTransationUnless(
+      provedIdentifier === desiredIdentifier ||
+      provedIdentifier === ""
+      )
+    expected = ConnectionEnd{INIT, provedIdentifier, getCommitmentPrefix(), counterpartyClientIdentifier,
                              clientIdentifier, counterpartyVersions}
     versionsIntersection = intersection(counterpartyVersions, getCompatibleVersions())
     version = pickVersion(versionsIntersection)
@@ -348,6 +353,7 @@ function connOpenTry(
 function connOpenAck(
   identifier: Identifier,
   version: string,
+  counterpartyIdentifier: Identifier,
   proofTry: CommitmentProof,
   proofConsensus: CommitmentProof,
   proofHeight: Height,
@@ -355,17 +361,22 @@ function connOpenAck(
     abortTransactionUnless(consensusHeight < getCurrentHeight())
     connection = provableStore.get(connectionPath(identifier))
     abortTransactionUnless(
+      counterpartyIdentifier === connection.counterpartyConnectionIdentifier ||
+      connection.counterpartyConnectionIdentifier === ""
+      )
+    abortTransactionUnless(
         (connection.state === INIT && connection.version.indexOf(version) !== -1)
         || (connection.state === TRYOPEN && connection.version === version))
     expectedConsensusState = getConsensusState(consensusHeight)
     expected = ConnectionEnd{TRYOPEN, identifier, getCommitmentPrefix(),
                              connection.counterpartyClientIdentifier, connection.clientIdentifier,
                              version}
-    abortTransactionUnless(connection.verifyConnectionState(proofHeight, proofTry, connection.counterpartyConnectionIdentifier, expected))
+    abortTransactionUnless(connection.verifyConnectionState(proofHeight, proofTry, counterpartyConnectionIdentifier, expected))
     abortTransactionUnless(connection.verifyClientConsensusState(
       proofHeight, proofConsensus, connection.counterpartyClientIdentifier, consensusHeight, expectedConsensusState))
     connection.state = OPEN
     connection.version = version
+    connection.counterpartyConnectionIdentifier = counterpartyConnectionIdentifier
     provableStore.set(connectionPath(identifier), connection)
 }
 ```
--- a/spec/ics-003-connection-semantics/README.md
+++ b/spec/ics-003-connection-semantics/README.md
@@ -299,6 +299,8 @@ function connOpenInit(
     abortTransactionUnless(provableStore.get(connectionPath(identifier)) == null)
     state = INIT
     if version != "" {
+      // manually selected version must be one we can support
+      abortTransactionUnless(getCompatibleVersions().indexOf(version) > -1)
       versions = [version]
     } else {
       versions = getCompatibleVersions()
--- a/spec/ics-003-connection-semantics/README.md
+++ b/spec/ics-003-connection-semantics/README.md
@@ -281,6 +281,7 @@ At the end of an opening handshake between two chains implementing the sub-proto
 
 This sub-protocol need not be permissioned, modulo anti-spam measures.
 
+In `connOpenInit`, a sentinel empty-string identifier can be used to allow the recipient chain to choose its own connection identifier.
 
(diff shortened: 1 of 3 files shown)


spec/ics-003-connection-semantics/README.md
```diff
@@ -318,24 +328,27 @@ function connOpenTry(
...
-      proofHeight, proofConsensus, counterpartyClientIdentifier, consensusHeight, expectedConsensusState))
     previous = provableStore.get(connectionPath(desiredIdentifier))
     abortTransactionUnless(
       (previous === null) ||
       (previous.state === INIT &&
         previous.counterpartyConnectionIdentifier === counterpartyConnectionIdentifier &&
         previous.counterpartyPrefix === counterpartyPrefix &&
         previous.clientIdentifier === clientIdentifier &&
-        previous.counterpartyClientIdentifier === counterpartyClientIdentifier &&
-        previous.version === getCompatibleVersions()))
```
colin-axner, line 338: why was this removed? or maybe it is represented somehow by the additional checks?

colin-axner: oh I see because in ConnOpenInit the version might not be the compatible ones but the provided one

cwgoes: yes, that's right

spec/ics-003-connection-semantics/README.md
```diff
@@ -290,12 +293,18 @@ function connOpenInit(
...
   counterpartyPrefix: CommitmentPrefix,
   clientIdentifier: Identifier,
-  counterpartyClientIdentifier: Identifier) {
+  counterpartyClientIdentifier: Identifier,
+  version: string) {
     abortTransactionUnless(validateConnectionIdentifier(identifier))
     abortTransactionUnless(provableStore.get(connectionPath(identifier)) == null)
     state = INIT
+    if version != "" {
+      versions = [version]
```
colin-axner, line 302: should their be a check that the version supplied is contained in the compatible versions?

cwgoes: It was manually provided, but I suppose, sure

cwgoes: Added a check

colin-axner: yea but what if the manually provided version isn't actually supported. Wouldn't the connection handshake be doomed to fail in connection ack or confirm?

spec/ics-003-connection-semantics/README.md
```diff
@@ -306,6 +317,7 @@ function connOpenInit(
 ```typescript
 function connOpenTry(
   desiredIdentifier: Identifier,
+  provedIdentifier: Identifier,
```
colin-axner, line 320: couldn't this just be a boolean indicating if the sending chain decided to choose the identifier or not?

spec/ics-003-connection-semantics/README.md
```diff
@@ -318,24 +330,27 @@ function connOpenTry(
     abortTransactionUnless(validateConnectionIdentifier(desiredIdentifier))
     abortTransactionUnless(consensusHeight < getCurrentHeight())
     expectedConsensusState = getConsensusState(consensusHeight)
-    expected = ConnectionEnd{INIT, desiredIdentifier, getCommitmentPrefix(), counterpartyClientIdentifier,
+    abortTransationUnless(
+      provedIdentifier === desiredIdentifier ||
```
ethanfrey, line 334: This is a bit confusing to me.

proved identifier is empty or comes from the init chain.

Desired Identifier comes from the receiving chain. But where? There doesn't seem to be a way to say "use whatever name they suggest", or "use whatever name they suggest if it starts with 'foo'" for example. 

Rather than comparing here, it would make more sense to have

`function desiredIdentifier(provedIdentifier) -> Identifier {}`

So the receiving module can make this decision and about if desired

spec/ics-003-connection-semantics/README.md
```diff
@@ -318,24 +330,27 @@ function connOpenTry(
...
       (previous === null) ||
       (previous.state === INIT &&
         previous.counterpartyConnectionIdentifier === counterpartyConnectionIdentifier &&
         previous.counterpartyPrefix === counterpartyPrefix &&
         previous.clientIdentifier === clientIdentifier &&
-        previous.counterpartyClientIdentifier === counterpartyClientIdentifier &&
-        previous.version === getCompatibleVersions()))
+        previous.counterpartyClientIdentifier === counterpartyClientIdentifier))
+    versionsIntersection = intersection(counterpartyVersions, previous !== null ? previous.version : getCompatibleVersions())
+    version = pickVersion(versionsIntersection)
```
ethanfrey, line 348: This throws if the intersection is empty, right?

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -311,6 +313,7 @@ function chanOpenTry(
   connectionHops: [Identifier],
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
+  provedIdentifier: Identifier,
```
ethanfrey, line 316: Is this proved or provided?

ethanfrey (commented):
Looks good, a few detail comments

spec/ics-003-connection-semantics/README.md
```diff
@@ -348,24 +363,30 @@ function connOpenTry(
...
+      )
     abortTransactionUnless(
         (connection.state === INIT && connection.version.indexOf(version) !== -1)
         || (connection.state === TRYOPEN && connection.version === version))
     expectedConsensusState = getConsensusState(consensusHeight)
     expected = ConnectionEnd{TRYOPEN, identifier, getCommitmentPrefix(),
                              connection.counterpartyClientIdentifier, connection.clientIdentifier,
                              version}
-    abortTransactionUnless(connection.verifyConnectionState(proofHeight, proofTry, connection.counterpartyConnectionIdentifier, expected))
+    abortTransactionUnless(connection.verifyConnectionState(proofHeight, proofTry, counterpartyConnectionIdentifier, expected))
```
colin-axner, line 384:
```suggestion
    abortTransactionUnless(connection.verifyConnectionState(proofHeight, proofTry, counterpartyIdentifier, expected))
```
?

spec/ics-003-connection-semantics/README.md
```diff
@@ -348,24 +363,30 @@ function connOpenTry(
...
     abortTransactionUnless(connection.verifyClientConsensusState(
       proofHeight, proofConsensus, connection.counterpartyClientIdentifier, consensusHeight, expectedConsensusState))
     connection.state = OPEN
     connection.version = version
+    connection.counterpartyConnectionIdentifier = counterpartyConnectionIdentifier
```
colin-axner, line 389:
```suggestion
    connection.counterpartyConnectionIdentifier = counterpartyIdentifier
```

spec/ics-003-connection-semantics/README.md
colin-axner, line 334: I'm running into issues implementing this as well. 

chainA sets the connection pointing at chainB with `proviedIdentifier`

chainB is verifying the connection set by chainA and setting its own connection pointing at chainB if everything is valid. This connection will use `desiredIdentifier` so why do `desiredIdentifier` and `provedIdentifier` need to be equal? This breaks our existing implementation:

chainA sets the connection on its chain with "connectionAID"

chainB wants to set the connection on its chain with "connectionBID" 

that would fail this if state. Was our implementation wrong? Am I missing something?

cwgoes: Right, that's the intended usage, it just doesn't need to happen "within" the IBC module, per se. I can add a note about it.

cwgoes: `provedIdentifier` is the counterparty identifier chain A selected for chain B, if chain A chose a specific identifier chain B needs to use it or the handshake will fail.

chain A can use whatever identifier it wants on its own side for its own connection

spec/ics-003-connection-semantics/README.md
cwgoes: sure, though we'll still construct a "provedIdentifier" since we need to check it

spec/ics-003-connection-semantics/README.md
cwgoes: Aye. Added comment to that effect.

spec/ics-004-channel-and-packet-semantics/README.md
cwgoes: It's the identifier which was proved on the counterparty side

spec/ics-003-connection-semantics/README.md
cwgoes: Added a comment with your example.

cwgoes:
Merging this as we've implemented it downstream, but let me know if comments were not adequately addressed.

PR #485, opened 2020-10-12
Fix syntax highlighting codeblock of ics-005

nkmr-jp:
When I was reading the spec/isc-005 , I found a code block that was not colored, so I fixed it.
https://github.com/cosmos/ics/tree/master/spec/ics-005-port-allocation


cwgoes (approved):
ACK, thanks!

PR #500, opened 2020-11-24
Spec changes for per-packet delay period

cwgoes:
Ref https://github.com/cosmos/cosmos-sdk/pull/7884

### DIFF
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -318,6 +318,8 @@ at a particular finalised height (necessarily associated with a particular commi
 Client types must define functions to authenticate internal state of the state machine which the client tracks.
 Internal implementation details may differ (for example, a loopback client could simply read directly from the state and require no proofs).
 
+`currentTimestamp` and `delayPeriod` are passed to packet-related verification functions in order to allow packets to specify a period which must pass after a header is verified before the packet is allowed to be processed.
+
 ##### Required functions
 
 `verifyClientConsensusState` verifies a proof of the consensus state of the specified client stored on the target machine.
@@ -366,6 +368,7 @@ type verifyChannelState = (
 type verifyPacketData = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -381,6 +384,7 @@ type verifyPacketData = (
 type verifyPacketAcknowledgement = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -396,6 +400,7 @@ type verifyPacketAcknowledgement = (
 type verifyPacketReceiptAbsence = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -410,6 +415,7 @@ type verifyPacketReceiptAbsence = (
 type verifyNextSequenceRecv = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -770,6 +776,7 @@ function verifyChannelState(
 function verifyPacketData(
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -784,6 +791,7 @@ function verifyPacketData(
 function verifyPacketAcknowledgement(
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -799,6 +807,7 @@ function verifyPacketReceiptAbsence(
   clientState: ClientState,
   height: Height,
   prefix: CommitmentPrefix,
+  delayPeriod: uint64,
   proof: CommitmentProof,
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
@@ -811,6 +820,7 @@ function verifyPacketReceiptAbsence(
 function verifyNextSequenceRecv(
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
--- a/spec/ics-003-connection-semantics/README.md
+++ b/spec/ics-003-connection-semantics/README.md
@@ -167,47 +167,51 @@ function verifyChannelState(
 function verifyPacketData(
   connection: ConnectionEnd,
   height: Height,
+  delayPeriod: uint64,
   proof: CommitmentProof,
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
   sequence: Height,
   data: bytes) {
     client = queryClient(connection.clientIdentifier)
-    return client.verifyPacketData(connection, height, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, data)
+    return client.verifyPacketData(connection, height, delayPeriod, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, data)
 }
 
 function verifyPacketAcknowledgement(
   connection: ConnectionEnd,
   height: Height,
+  delayPeriod: uint64,
   proof: CommitmentProof,
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
   sequence: uint64,
   acknowledgement: bytes) {
     client = queryClient(connection.clientIdentifier)
-    return client.verifyPacketAcknowledgement(connection, height, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, acknowledgement)
+    return client.verifyPacketAcknowledgement(connection, height, delayPeriod, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, acknowledgement)
 }
 
 function verifyPacketReceiptAbsence(
   connection: ConnectionEnd,
   height: Height,
+  delayPeriod: uint64,
   proof: CommitmentProof,
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
   sequence: uint64) {
     client = queryClient(connection.clientIdentifier)
-    return client.verifyPacketReceiptAbsence(connection, height, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier)
+    return client.verifyPacketReceiptAbsence(connection, height, delayPeriod, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier)
 }
 
 function verifyNextSequenceRecv(
   connection: ConnectionEnd,
   height: Height,
+  delayPeriod: uint64,
   proof: CommitmentProof,
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
   nextSequenceRecv: uint64) {
     client = queryClient(connection.clientIdentifier)
-    return client.verifyNextSequenceRecv(connection, height, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, nextSequenceRecv)
+    return client.verifyNextSequenceRecv(connection, height, delayPeriod, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, nextSequenceRecv)
 }
 
 function getTimestampAtHeight(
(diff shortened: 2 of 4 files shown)


spec/ics-002-client-semantics/README.md
```diff
@@ -366,6 +368,7 @@ type verifyChannelState = (
 type verifyPacketData = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
```
AdityaSripal, line 371: These require the `currentTimestamp` as well

spec/ics-003-connection-semantics/README.md
```diff
@@ -167,47 +167,51 @@ function verifyChannelState(
...
   height: Height,
+  delayPeriod: uint64,
   proof: CommitmentProof,
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
   sequence: Height,
   data: bytes) {
     client = queryClient(connection.clientIdentifier)
-    return client.verifyPacketData(connection, height, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, data)
+    return client.verifyPacketData(connection, height, delayPeriod, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, data)
```
AdityaSripal, line 177:
```suggestion
    currentTimestamp = queryCurrentTimestamp()
    return client.verifyPacketData(connection, height, currentTimestamp, delayPeriod, connection.counterpartyPrefix, proof, portIdentifier, channelIdentifier, data)
```

ditto for other connection verify packet methods

spec/ics-007-tendermint-client/README.md
```diff
@@ -84,6 +84,7 @@ The Tendermint client tracks the timestamp (block time), validator set, and comm
 ```typescript
 interface ConsensusState {
   timestamp: uint64
+  processedTime: uint64
```
AdityaSripal, line 87: ConsensusState should not contain `processedTime`

AdityaSripal (changes requested):

spec/ics-002-client-semantics/README.md
cwgoes: This is the SDK interface, but isn't it only that way because we don't pass a `sdk.Context`?

spec/ics-007-tendermint-client/README.md
cwgoes: Aye, updated.

spec/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -116,6 +117,7 @@ interface Packet {
 - The `sequence` number corresponds to the order of sends and receives, where a packet with an earlier sequence number must be sent and received before a packet with a later sequence number.
 - The `timeoutHeight` indicates a consensus height on the destination chain after which the packet will no longer be processed, and will instead count as having timed-out.
 - The `timeoutTimestamp` indicates a timestamp on the destination chain after which the packet will no longer be processed, and will instead count as having timed-out.
+- The `delayPeriod` indicates a period that must elapse after validation of a header before a packet can be processed. It is only safe to allow the user to set this field when using unordered channels.
```
milosevic, line 120: And what about ordered channels? How we ensure the same level of security for ordered channels?

spec/ics-007-tendermint-client/README.md
```diff
@@ -214,8 +214,9 @@ function checkValidityAndUpdateState(
     // update latest timestamp
     clientState.latestTimestamp = header.timestamp
     // create recorded consensus state, save it
-    consensusState = ConsensusState{header.validatorSet, header.commitmentRoot, header.timestamp}
+    consensusState = ConsensusState{header.timestamp, header.validatorSet, header.commitmentRoot}
     set("clients/{identifier}/consensusStates/{header.height}", consensusState)
+    set("clients/{identifier}/processedTimes/{header.height}", currentTimestamp())
```
milosevic, line 219: Does it make sense adding `processedTime` as a field to the ConsensusState? It might be a bit easier to understand the semantic of `delayPeriod`?

spec/ics-007-tendermint-client/README.md
```diff
@@ -378,6 +385,10 @@ function verifyPacketAcknowledgement(
     assert(clientState.latestHeight >= height)
     // check that the client is unfrozen or frozen at a higher height
     assert(clientState.frozenHeight === null || clientState.frozenHeight > height)
+    // fetch the processed time
+    processedTime = get("clients/{identifier}/processedTimes/{height}")
```
milosevic, line 389: I am puzzled here. Let's say packet p with delay d was sent from chain A to chain B at height h. Packet is correctly received at chain B if there is a header for height h+1 installed at time t and packet is received not before time t+d. Verifying packet acknowledgment for p is happening on the chain A, and if we want to verify that packet was indeed not processed before t+d, we will need to have access to the consensusState at chain B at the header reception time and processingTime of the packet p at chain B. So both these information need to be carried to the chain A with the corresponding proofs so chain A can verify that indeed packet was correctly processed on B. Does this make sense?

spec/ics-007-tendermint-client/README.md
```diff
@@ -397,6 +409,10 @@ function verifyPacketReceiptAbsence(
...
```
milosevic, line 413: Why do we need this check in packet receipt absence? I guess we only need the proof that the packet hasn't been received by the counterparty?

spec/ics-007-tendermint-client/README.md
cwgoes: No, because we don't want to *store* it in the consensus state, which needs to be verified e.g. during the handshake

spec/ics-004-channel-and-packet-semantics/README.md
cwgoes: Ordered channels would need to apply some bound to the `delayPeriod` which their packets are allowed to set, or even fix a particular value (possibly `0` or a low value).

spec/ics-007-tendermint-client/README.md
cwgoes: This is a good question, I'm not sure. In the security model as I understand it, we want delays on all verification which could result in movement of funds, in order to allow time for evidence to be submitted, yes? Packet receipt absence does fall into this category, since it is used for proof-of-timeout which will e.g. refund funds to the sender of a fungible token transfer.

spec/ics-007-tendermint-client/README.md
cwgoes: Hmm. Maybe I'm misunderstanding the intended model here. I thought we were trying to ensure that after a given header was submitted to a chain - call it *C* - a period of time specified per-packet must pass before any proof verification involving that packet or related data (such as the acknowledgement) which uses that header is allowed to complete, so that there would be time for misbehaviour to be submitted to the chain. I don't see what relevance the timestamp on the other chain would have to this since it's the light client on *this* chain which we want time to submit misbehaviour to - also, the timestamp on the other chain is not a local clock, so it's not clear how we would compare it. Am I misunderstanding something fundamental here?

milosevic: So high level invariant we want is the following: packet verification on the reception side should fail if delayPeriod hasn't passed since the corresponding header verification/update. This allows correct relayer to detect faulty header and shutdown light client. Now, as packet reception is a precondition for valid acknowledgment processing on the sender side, acknowledgment verification depends on it. In the model where we trust state transitions to be valid if we see valid light client proof, nothing would need to be changed on the sender side (ack/timeout verification). Note that in the model without delay this is the case as we only verify light client proofs, and not if the corresponding transition was executed (if I am not wrong). In this concrete case we have a way to verify if time has passed, so we could indeed validate this on the sender side, but I think we will need more data in ack/timeout packets to be able to do this. The question is if that is needed.

AdityaSripal: > Let's say packet p with delay d was sent from chain A to chain B at height h. Packet is correctly received at chain B if there is a header for height h+1 installed at time t and packet is received not before time t+d. 

Correct, I think we are all agreed on this.

>  Now, as packet reception is a precondition for valid acknowledgment processing on the sender side, acknowledgment verification depends on it. In the model where we trust state transitions to be valid if we see valid light client proof, nothing would need to be changed on the sender side (ack/timeout verification). Note that in the model without delay this is the case as we only verify light client proofs, and not if the corresponding transition was executed (if I am not wrong). In this concrete case we have a way to verify if time has passed, so we could indeed validate this on the sender side, but I think we will need more data in ack/timeout packets to be able to do this. 

So from my understanding you're talking about the acknowledgement logic verifying that the packet delay period had indeed passed on the counterparty before packet receipt was processed. ie ChainA verifies that chainB waited for delay period before receiving, and only then will successfully ACK. (and similarly with timeouts)

I'm not sure this is what we want to do, compared with Chris's comment which is to enforce the packet-specified delay period for any packet-related verification that occurs on our own chain. And not check that packet-delay is correctly enforced on the counterparty chain.

Going back to your example, a user sends a packet p with delay period d from chainA to chainB.

Assuming a correct implementation on chainB, the consensus state used to verify chainA's packet commitment must at least exist for `d` time before a relayer is allowed to use it to receive the packet on chainB.
**Note that this feature protects the recipient on chainB from potential malicious behaviour of chainA.**

Now there are only 2 instances in my mind that a chainB will receive a packet before the user-specified delay period.

**1. Incorrect implementation:**

In this case, chainB verifies the packet without enforcing the delay period. One could imagine an alternate IBC implementation that simply ignored the packet's delay period.
Note from above that this is a choice the chain is allowed to make that simply affects its own users. ie the recipient on chainB is vulnerable to misbehaviour on chainA because chainB chose not to enforce delay period.
In this case, I don't believe chainA should care whether chainB follows delay period or not as it does not affect chainA.
Why should chainA enforce that chainB protects users against chainA becoming malicious? I believe it is safe for a chain to assume itself to be honest and avoid unnecessary verification work that way.
Furthermore, suppose chainB does ignore the delay period and receives the packet anyway. It's an open question what should be done in light of this.
Should chainA never acknowledge the packet since chainB incorrectly received before delay period? I believe this is what your proposal would do since if we pass the counterparty's consensus state and processing time back to chainA and the delay period check fails, chainB will have received the packet while chainA can never acknowledge that receipt because it will always fail the delay period check.
This can be dangerous for applications. Take transfer for example, if chainB receives the packet before delay period and for whatever reason errors and returns an `Failed` acknowledgement. chainA will never be able to process that acknowledgement and refund the sender.

I believe it is safe and simpler for chainA to just not check whether chainB enforces the delay period during packet receipt or not. What is important is that each chain enforces the delay period on its own end to protect its own users from misbehaviour by counterparties. It is unnecessary for chainA to check that counterparties are protecting their users from potential misbehaviour by chainA.

**2. Fork**

The other case where chainA may try to ack a packet receipt is if there is a malicious fork on chainB such that the malicous fork has received the packet before delay period. A relayer might then submit a proof against the malicous fork to chainA and try to get the ack processed.

However, this case is already handled by the current spec and implementation since the packet-delay period specified when the packet is sent will apply to **all packet verification logic**, including acks and timeouts. Thus the consensus state on chainA must exist for at least `d` time before a relayer can use it to verify an ACK on chainB.

This offers the same level of misbehaviour protection for acks and timeouts as well, since a relayer will have at least `d` time to submit misbehaviour that could happen at any stage in the packet lifecycle.

AdityaSripal: > So high level invariant we want is the following: packet verification on the reception side should fail if delayPeriod hasn't passed since the corresponding header verification/update.

I believe this invariant should only be enforced on the receiver side and not the sender side for the reasons stated above.

I think the invariant we want to enforce is actually this: Each chain enforces a delay period before accepting any packet-level verification (receipts/acks/timeouts) on its own chain to allow time for misbehaviour to be submitted against a counterparty.

spec/ics-007-tendermint-client/README.md
AdityaSripal: I think we want to enforce the delay period specified in packet send on any packet-related verification. Since each step involves a race between the packet verification and misbehaviour submission that could invalidate the verification.

The original sender of the packet should be able to define a delay period that gives a bounded time for misbehaviour submission before packet-verification at every stage of packet lifecycle.

spec/ics-007-tendermint-client/README.md
milosevic: Thanks a lot @AdityaSripal for clarifying all this. I misunderstood the ack/timeout logic. I thought it's related to packet reception and as you nicely explained it ensures the same guarantee for ack/timeout data. I think that I agree with your proposal regarding 1, but we can synchronously confirm it during the IBC call.

cwgoes: Sounds like we're in agreement here (thanks @AdityaSripal for the clearer explanation) but let's confirm on the call.

cwgoes:
Per discussion decided to revise this to a fixed delay period set per-connection, and later on push value flow control to the app-level.

cwgoes:
Replaced by https://github.com/cosmos/ics/pull/507.

PR #507, opened 2020-12-03
Delay period on a per-connection basis

cwgoes:
Replaces https://github.com/cosmos/ics/pull/500 per discussion.

### DIFF
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -318,7 +318,7 @@ at a particular finalised height (necessarily associated with a particular commi
 Client types must define functions to authenticate internal state of the state machine which the client tracks.
 Internal implementation details may differ (for example, a loopback client could simply read directly from the state and require no proofs).
 
-`currentTimestamp` and `delayPeriod` are passed to packet-related verification functions in order to allow packets to specify a period which must pass after a header is verified before the packet is allowed to be processed.
+- The `delayPeriod` is passed to packet-related verification functions in order to allow packets to specify a period which must pass after a header is verified before the packet is allowed to be processed.
 
 ##### Required functions
 
--- a/spec/ics-002-client-semantics/README.md
+++ b/spec/ics-002-client-semantics/README.md
@@ -318,6 +318,8 @@ at a particular finalised height (necessarily associated with a particular commi
 Client types must define functions to authenticate internal state of the state machine which the client tracks.
 Internal implementation details may differ (for example, a loopback client could simply read directly from the state and require no proofs).
 
+`currentTimestamp` and `delayPeriod` are passed to packet-related verification functions in order to allow packets to specify a period which must pass after a header is verified before the packet is allowed to be processed.
+
 ##### Required functions
 
 `verifyClientConsensusState` verifies a proof of the consensus state of the specified client stored on the target machine.
@@ -366,6 +368,7 @@ type verifyChannelState = (
 type verifyPacketData = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -381,6 +384,7 @@ type verifyPacketData = (
 type verifyPacketAcknowledgement = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -396,6 +400,7 @@ type verifyPacketAcknowledgement = (
 type verifyPacketReceiptAbsence = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -410,6 +415,7 @@ type verifyPacketReceiptAbsence = (
 type verifyNextSequenceRecv = (
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -770,6 +776,7 @@ function verifyChannelState(
 function verifyPacketData(
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -784,6 +791,7 @@ function verifyPacketData(
 function verifyPacketAcknowledgement(
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
@@ -799,6 +807,7 @@ function verifyPacketReceiptAbsence(
   clientState: ClientState,
   height: Height,
   prefix: CommitmentPrefix,
+  delayPeriod: uint64,
   proof: CommitmentProof,
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
@@ -811,6 +820,7 @@ function verifyPacketReceiptAbsence(
 function verifyNextSequenceRecv(
   clientState: ClientState,
   height: Height,
+  delayPeriod: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
--- a/spec/ics-003-connection-semantics/README.md
+++ b/spec/ics-003-connection-semantics/README.md
@@ -95,7 +95,7 @@ interface ConnectionEnd {
 - The `counterpartyClientIdentifier` field identifies the client on the counterparty chain associated with this connection.
 - The `version` field is an opaque string which can be utilised to determine encodings or protocols for channels or packets utilising this connection.
   If not specified, a default `version` of `""` should be used.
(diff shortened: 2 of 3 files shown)


AdityaSripal (approved):
ACK, same approach taken in PR

spec/ics-002-client-semantics/README.md
```diff
@@ -318,6 +318,8 @@ at a particular finalised height (necessarily associated with a particular commi
 Client types must define functions to authenticate internal state of the state machine which the client tracks.
 Internal implementation details may differ (for example, a loopback client could simply read directly from the state and require no proofs).
 
+- The `delayPeriod` is passed to packet-related verification functions in order to allow packets to specify a period which must pass after a header is verified before the packet is allowed to be processed.
```
milosevic, line 321: Is packet defining a `delayPeriod`?

spec/ics-003-connection-semantics/README.md
```diff
@@ -345,11 +350,11 @@ function connOpenTry(
     abortTransactionUnless(consensusHeight < getCurrentHeight())
     expectedConsensusState = getConsensusState(consensusHeight)
     expected = ConnectionEnd{INIT, "", getCommitmentPrefix(), counterpartyClientIdentifier,
-                             clientIdentifier, counterpartyVersions}
+                             clientIdentifier, counterpartyVersions, delayPeriod}
     versionsIntersection = intersection(counterpartyVersions, previous !== null ? previous.version : getCompatibleVersions())
     version = pickVersion(versionsIntersection) // throws if there is no intersection
     connection = ConnectionEnd{TRYOPEN, counterpartyConnectionIdentifier, counterpartyPrefix,
-                               clientIdentifier, counterpartyClientIdentifier, version}
+                               clientIdentifier, counterpartyClientIdentifier, version, delayPeriod}
```
milosevic, line 357: My understanding is that `delayPeriod` is a parameter defined by the chain to provide security to its users. Shouldn't it be reasonable to give each side of the connection possibility to choose `delayPeriod`? Also what happen in crossing hello case when different `delayPeriod` is chosen by counter parties?

spec/ics-002-client-semantics/README.md
colin-axner: `delayPeriod` is being defined upon creation of a connection (In conn open init). Each connection end must have the same `delayPeriod` for a connection to be established. The connection then passes this delayPeriod to the packet verification functions implemented by the on chain light clients

milosevic:
I have few more high level questions: 

1. What are the benefits of having `delayPeriod` as part of `ConnectionEnd` and not `ClientState`? The potential benefits of
having it as part of `ClientState` are the fact that as soon as client is created, we know what delay it will ensure. So from higher level protocols and abstractions as soon as they are using right client id it is safe. Second, at the moment `verifyClientConsensusState` and other `verifyX` functions do not have `packetDelay` as a parameter. Are we sure who is using those functions and aren't those functions also potentially targets for the same attack? Having `packetDelay` as part of `ClientState` could potentially make also these functions secured the same way (if there is a need for this).
2. What happens with optimistic channel handshake and packet sends? I guess from the channel perspective, at the point `channelOpenInit` is executed, user should be aware of what `delayPeriod` it will ensure, and the way to configure this is by choosing `connectionId`. My understanding is that in this proposal we assume that `delayPeriod` will be defined at the connection level, so the fact that connection with the given id is created gives user possibility to inspect the connectionEnd and see if its `delayPeriod` field satisfies its security needs. This seems fine if we assume that `delayPeriod` is agreed out of band, so there is no need to negotiate the value during the handshake. 
3. Does it make sense adding `minDelayPeriod` also as a parameter to a `chanOpenInit` and `chanOpenTry` so IBC can ensure that the chosen connection end will indeed ensure that the `delayPeriod` will be greater than `minDelayPeriod`. Having `minDelayPeriod` as a field in `ChannelEnd` structure might simplify figuring out which channel to use in case application wants to have flexibility in choosing what delay to ensure depending on the packet data. Also in multi-hop scenarios, we might want to ensure that some `minDelayPeriod` is ensured across all path, but it does not need necessarily being the same.

spec/ics-003-connection-semantics/README.md
colin-axner: I don't think chains are defining the delay period in this implementation. As you say, each side of the conenction has the possibility to choose the `delayPeriod`. 

In the case of crossing hello's, the connection handshake will succeed with different delay periods. It can be viewed the same as if the crossing hellos selected different client identifiers (they cannot connect with each other because they have different parameters)

colin-axner:
> I have few more high level questions:

I think I can take a shot at answering some of these questions.

1. The benefit of having the `delayPeriod` in the connection instead of the `clientState` is that you don't have to create new clients just to have different delay periods, you could simply create a new connection. This would result in having to update less clients overall which can be costly and would also encourage maintaining less clients for a chain, but still allow flexibility in the security of each connection. 

I think the delay period is left out of the other verify functions since there is not really any risk involved until a packet is sent (at which point we verify the delay period)

2. 
> I guess from the channel perspective, at the point channelOpenInit is executed, user should be aware of what delayPeriod it will ensure, and the way to configure this is by choosing connectionId. 

correct. 

3. Will leave this to @cwgoes and @AdityaSripal but this sounds to me like a UI improvement as opposed to something necessary for spec correctness. Since delayPeriod is set and finalized upon initialization of a new connection, the delay period to be associated with a channel can always be queried for.

spec/ics-003-connection-semantics/README.md
AdityaSripal: > In the case of crossing hello's, the connection handshake will succeed with different delay periods.

I think you mean to say it will fail.

---- 

We want the initiator of the handshake to define the `delayPeriod` on both ends. Consider that I want to create a connection with ChainA and chainB, and I want to use the `delayPeriod` feature to protect packets going across this coonnection from misbehaviour on either side.

I submit a `ConnOpenInit` to chainA with a `delayPeriod=60`. Now, I intend to also set a `delayPeriod` on chainB, however another relayer has submitted a `ConnOpenTry` before me on chainB with `delayPeriod=0`. This tx passes, and I now have a connection with `delayPeriod` on chainA and no `delayPeriod` on chainB.

This is clearly not what I wanted as the initiator. This is why we fix the delay period at the `INIT` step to ensure that the final connection, regardless of which relayer completes it, has the delay period that was specified by the initiating relayer. In `OpenTry`, we enforce that this connectionEnd also has a delay period specified by the initiating relayer.
As @colin-axner said, in case of crossing hello's that don't agree on delay period, we error just as we do in other cases where parameters do not match.

Now, it is possible that in the `INIT` step, we could define different delay periods for each chain. So `ConnOpenInit` might specify `delayPeriod` AND `counterpartyDelayPeriod`.
I could see this being potentially useful in cases where a highly secure chain is connecting to a more risky chain, the initiating relayer may specify no delay period for the `cosmos-hub` client but a high delay period for the `risky-chain` client.

However, this is a (slight) complication that doesn't seem all that useful imo, and unnecessary for IBC 1.0

AdityaSripal:
> The benefit of having the delayPeriod in the connection instead of the clientState is that you don't have to create new clients just to have different delay periods, you could simply create a new connection. This would result in having to update less clients overall which can be costly and would also encourage maintaining less clients for a chain, but still allow flexibility in the security of each connection.
I think the delay period is left out of the other verify functions since there is not really any risk involved until a packet is sent (at which point we verify the delay period)

Correct. In the non-packet verify function, the order of verification success and misbehaviour submission doesn't really matter. Given it doesn't matter, it makes sense to prioritize speed here.
ie. A channel handshake message may fails because client is frozen, or it may pass and the client is frozen afterwards. The end result is the same. Once packets start getting sent, they will all fail to be received.

However, the order of packet submission and misbehaviour submission really does matter. Which is why delay period gives a chance for misbehaviour to be submitted first.

3. It could make sense to do this. And certainly this check should happen somewhere such that users are always using channels that have underlying connections that suit the user's needs.
However, I would prefer we don't add this check to the core state-machine. Specifying a connection-id in channel handshake, is an implicit acknowledgement of all the connection parameters.
We should implement this using connection queries on the relayer-side and make the relayer responsible for using connections that match user's needs when creating a channel.

4. Yes it is true, we rely on correct relayer. However, this is true of misbehaviour in general. All of the misbehaviour and freezing logic we have are only useful if there is a correct relayer monitoring the client state updates.

spec/ics-002-client-semantics/README.md
milosevic: I understand now what you mean with packet-related verification functions.

spec/ics-003-connection-semantics/README.md
milosevic: It seems that giving possibility for each party to choose a delayPeriod would conflict with optimistic sends, so it seems like good tradeoff to have it the same on both ends.

milosevic:
Thanks a lot for the answers, it makes sense. Looks good to me. Just a note that we will put some effort in adding this change to the IBC tla+ spec and try to run it through model checker so there are no surprises :)

milosevic (approved):

spec/ics-002-client-semantics/README.md
cwgoes: Aye. I didn't include that definition in this particular spec document because a client shouldn't care "where" the `delayPeriod` is defined. Maybe that's not so important as to avoid mentioning this though.

spec/ics-003-connection-semantics/README.md
cwgoes: Aye, we discussed this a bit internally, I guess it should be summarised somewhere public. Great summary @AdityaSripal!

Issue #539, opened 2021-03-18
Hybrid delayPeriod


milosevic:
In the current specification (https://github.com/cosmos/ibc/tree/master/spec/core/ics-003-connection-semantics) delayPeriod parameter is time based, i.e., the protocol enforces that at minimum delayPeriod must pass between the point header (consensusState) is installed and the corresponding packet data is verified. This provides security window during which correct relayer can submit a proof of misbehaviour and freeze client in case consensusState is invalid. 

However, if the destination chain (where invalid consensusState is installed) does not create blocks or is censoring misbehaviour processing during delayPeriod, then security protection is of no use. Potential solution that would increase security guarantees would be having hybrid delayPeriod where it is expressed both in terms of time and number of blocks. In this case verification would be possible only when both some time and number of blocks has passed.

Although it is very hard completely preventing censorship issue (with the existing system), this proposal should still increase security guarantees as it decreases a chance for censorship of misbehaviour by increasing packetDelay.

PR #552, opened 2021-03-25
Update ICS 2, 3, 7 for hybrid delay period

cwgoes:
Closes https://github.com/cosmos/ibc/issues/539

Split delay period into both a minimum time period which must have passed and a minimum number of blocks which must have passed in order to allow packets, acknowledgements, etc. to be processed.

Note that this breaks the connection handshake.

### DIFF
--- a/spec/client/ics-007-tendermint-client/README.md
+++ b/spec/client/ics-007-tendermint-client/README.md
@@ -348,7 +348,8 @@ function verifyChannelState(
 function verifyPacketData(
   clientState: ClientState,
   height: Height,
-  delayPeriod: uint64,
+  delayPeriodTime: uint64,
+  delayPeriodBlocks: uint64,
   prefix: CommitmentPrefix,
   proof: CommitmentProof,
   portIdentifier: Identifier,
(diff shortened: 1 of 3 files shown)


milosevic:
@cwgoes In general looks good. If I am not wrong, we need to remember also processedHeight in ics07, `checkValidityAndUpdateState`, the similar way we set processed time (`set("clients/{identifier}/processedHeight/{header.height}", currentHeight()`) and then use it in verify functions.

cwgoes:
> @cwgoes In general looks good. If I am not wrong, we need to remember also processedHeight in ics07, `checkValidityAndUpdateState`, the similar way we set processed time (`set("clients/{identifier}/processedHeight/{header.height}", currentHeight()`) and then use it in verify functions.

Ah yes, of course, I had the wrong height! Fixed.

milosevic (approved):
LGTM. Thanks Chris!

AdityaSripal (approved):
This looks correct to me.

I guess I'm still uncertain of why this change is needed.

PR #557, opened 2021-04-06
add protobuf files

colin-axner:


cwgoes (commented):
Do we want to put these files in separate directories from the specifications themselves, or include them alongside the specification documents?

colin-axner:
> Do we want to put these files in separate directories from the specifications themselves, or include them alongside the specification documents?

I don't have a strong preference, but it'd be good to clearly document where to find the files if they don't live in `proto/`

cc @AdityaSripal

colin-axner:
I spoke with @AdityaSripal a little while back and we concluded that having the proto files live under `proto/` made the most sense. The specs could link to the protobuf files as necessary. 

I can add a README as well which talks about #568 

I will probably split this pr into smaller ones so each review is more digestable. I'd like the protobuf files to be reviewed since they should contain up to date comments and no references to the SDK if they are to be included here. The only exception is the gogoproto can be left since it'd make our lives very difficult if they were removed

cwgoes:
Thanks for the explanation, your proposal sounds good to me.

AdityaSripal:
Are you splitting these up @colin-axner or should I review as is?

colin-axner:
> Are you splitting these up @colin-axner or should I review as is?

I should split these up. I think it'll be easier for all that way. This has fallen on my backlog (happy to let someone take over)

colin-axner:
I won't be taking this up anytime soon, so closing this pr

Issue #560, opened 2021-04-14
ICS ?: IBFT 2.0 client support


bluele:
Hi, there.

We propose the spec of IBC Client that supports [the IBFT 2.0 consensus protocol](https://arxiv.org/abs/1909.10194).

IBFT 2.0 is one of the major consensus protocols for Hyperledger Besu.

We believe that supporting Hyperledger Besu(IBFT 2.0), a popular blockchain in the enterprise blockchain space, will be valuable in expanding Cosmos IBC ecosystem.

The IBFT 2.0 Light Client spec can be found [here](https://github.com/datachainlab/ibc-solidity/blob/main/docs/ibft2-light-client.md), and an example implementation under development can be found [here](https://github.com/datachainlab/ibc-solidity/blob/main/contracts/core/IBFT2Client.sol).

We will open a PR for the spec soon.

bluele:
Hi @cwgoes,
We've already opened a PR(#561)  for this spec.
Could we get a review from someone? Or please let us know if any other process is needed for ICS proposal.

cwgoes:
Hi @bluele - thanks for the PR! We will try to take a look at it soon.

crodriguezvega:
@bluele Is this issue still relevant?

PR #561, opened 2021-04-14
Add IBFT 2.0 client spec

bluele:
ref. #560

### DIFF
--- a/spec/client/ics-xx-ibft2-client/README.md
+++ b/spec/client/ics-xx-ibft2-client/README.md
@@ -1,0 +1,463 @@
+---
+ics: XX
+title: IBFT 2.0 Client
+stage: draft
+category: IBC/TAO
+kind: instantiation
+author: Jun Kimura <jun.kimura@datachain.jp>
+created: 2021-04-14
+implements: 2
+---
+
+## Synopsis
+
+This specification document describes a client (verification algorithm) for a blockchain using IBFT 2.0 consensus.
+
+### Motivation
+
+State machines of various sorts replicated using the IBFT 2.0 consensus protocol might like to interface with other replicated state machines or solo machines over IBC.
+
+### Definitions
+
+Functions & terms are as defined in [ICS 2](https://github.com/cosmos/ibc/tree/master/spec/core/ics-002-client-semantics).
+
+### Desired Properties
+
+This specification must satisfy the client interface defined in ICS 2.
+
+## Technical Specification
+
+This specification depends on correct instantiation of the [IBFT 2.0 consensus protocol](https://arxiv.org/abs/1909.10194) and [light client protocol](https://github.com/datachainlab/ibc-solidity/blob/main/docs/ibft2-light-client.md).
+
+### Client State
+
+The IBFT 2.0 client state tracks the trusting period, latest height, ibcStoreAddress, and a possible frozen height. The ibcStoreAddress represents the address of the contract account which stores the commitments.
+
+```typescript
+interface ClientState {
+  chainId: uint64
+  trustLevel: Fraction
+  trustingPeriod: uint64
+  ibcStoreAddress: Address
+  latestHeight: uint64
+  frozenHeight: Maybe<uint64>
+}
+```
+
+### Consensus State
+
+The IBFT 2.0 client tracks the timestamp (block time), validator set, and commitment root for all previously verified consensus states (these can be pruned after the trusting period has passed, but should not be pruned beforehand). The commitmentRoot is a storage root of the account corresponding to the ibcStoreAddress in the ClientState.
+
+```typescript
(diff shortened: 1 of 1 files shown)


zmanian:
There is currently no implementation of ICS-23 for the Ethereum Merkle Patricia tree.

Is there a plan to create one?

The reasons are here: https://github.com/confio/ics23#unsupported

> I spent quite some time to wrestle out a well-defined serialization and a validation logic that didn't involve importing too much code from go-ethereum (copying parts and stripping it down to the minimum). At the end, I realized the key is only present from parsing the entire path and is quite a painstaking process, even with go code that imports rlp and has the custom patricia key encodings. After a long time reflecting, I cannot see any way to implement this that doesn't either: (A) allow the provider to forge a different key that cannot be detected by the verifier or (B) include a huge amount of custom code in the client side.
> 
> If anyone has a solution to this, please open an issue in the proofs-ethereum repo.

bluele:
Thanks for your comment, @zmanian.

Unless I've misunderstood ics23, it already provides a solidity implementation. It looks something like the following:

- `CommitmentRoot` is the storage root of the Contract Account and `Path` is represented as the storage slot. `verifyMembership` implementation using them is here:
  - https://github.com/datachainlab/ibc-solidity/blob/28757d2cbfd4d03de1aa9a58eceba35f5a75585e/contracts/core/IBFT2Client.sol#L290

- The Merkle Patricia tree verification is implemented based on [aragon's evm-storage-proofs](https://github.com/aragon/evm-storage-proofs/tree/master/packages/evm/contracts).
  - https://github.com/datachainlab/ibc-solidity/blob/987b179f3331ccebdab3e591dee25fe15e96c4f1/contracts/lib/TrieProofs.sol

- The prover currently uses `eth_getproof` API. It can be used as implemented in geth and Hyperledger Besu.
  - https://github.com/datachainlab/ibc-solidity/blob/56828df34c4a778c0ae44967ab69307048ca6ad9/pkg/client/proof.go

zmanian:
For full IBC compatibility, you need to be able to represent the proof as ICS23 operations so the golang side can implement it.

Instead of using the merkle patricia tree to store packets, I would consider using a in contract merkle tree like optics does.

https://github.com/celo-org/optics-monorepo

ethanfrey:
Note that ics23 *should* be able to represent the SMT which I believe will be used in Ethereum 2.0.

While working on that, it seemed very unlikely that we will have a finality gadget for Eth 1.0 and only from Eth 2.0 would some sort of IBC-compatible light client proof be available. Since we only need ics23 merkle proofs where there is a valid light client proof, I felt this was a fair simplification to make.

Note that no transformer from SMT proofs to ICS23 format has been written, but if this is important to you, I am can discuss the design with you and give some advice.

bluele:
Will the proof and the operation spec defined in confio's ics23 repository be merged into the ics23 spec in the future? Or is it a standard for Go implementation only?

Also, as far as I understand, even if it implements `Commitments` with Incremental Merkle Tree in a chain compatible with eth1.0, we can't avoid verifying Merkle Patricia Tree as long as it does the verification based on the state root of a header.
Of course, it may be possible to avoid header-based verification by optimistically accepting a Tree-root submission and introducing a dispute resolution process, but I'm not sure that approach is appropriate for permissioned chains like Besu.

By the way, we don't currently have a proof implementation that complies with confio's ics23. Does that prevent us from moving forward with this PR review process?

ethanfrey:
I am not maintainer of IBC, so this is just my personal opinion...

I think proposing a new **client** type (equivalent to tendermint) should be independent of the merkle proof. You could remove the sections discussing verifyPacket, etc here and just focus on the consensus algorithms, and this should be unblocked.

The packet verification should work for all clients, and just require an ics23-compatible proof spec.

The ics23 idea is in the ibc spec. The implementation is in https://github.com/confio/ics23. This defines the binary interchange format as a set of protobuf specs and provides verification functions in Go, Rust, and TypeScript. There was an attempt to provide an ics23 **verifier** in solidity, but I believe that was abandoned (look at the PR).

The source blockchain needs to know how to convert the internal merkle tree proof format into ics23 format. The receiving chain doesn't need to make any changes to support this.

cwgoes:
I realise ICS 7 does essentially the same thing, but in general I think we should move _away_ from separating out the light client definitions into a "verify" function and an ICS specification which calls it - this is an awkward interface boundary which splits verification-related code into two places. Rather, I would prefer that the minimal light client be specified as a part of the ICS so that the standard is self-contained.

Perhaps another interface boundary is possible that's less prone to error, open to suggestions. What do you think?

cc @milosevic

Raulnori77 (approved):

mpoke:
Please reopen if it's still relevant.

Issue #550, opened 2019-06-25
Directed acyclic graph of causal dependencies (partial packet ordering)


cwgoes:
Split off from https://github.com/cosmos/ics/issues/126.

Desiderata
- Packets can specify a set of ancestor identifiers which must be received prior to receiving the packet
- Super-set of:
    - Ordered channels: one ancestor per packet that is the direct predecessor
    - Unordered channels: no ancestors
- Ancestor identifiers can refer to packets sent on other channels from other chains / modules

Implementation
- Ancestor identifier represented as hash of
  `(source connection || source channel || packet sequence || packet data)`
- Accumulator in receiving IBC handler adds identifiers of all received packets (and stores forever?) - shared across all channels and connections
- Proof of inclusion of ancestor packets required to receive
- Proof of non-inclusion of particular identifier for timeouts (a bit different than proof of receive sequence number currently), can batch timeouts safely if a packet upon which other packets depended times out

Questions
- How to deal with timeouts? Should they still close channels? (could be an option)
- Does the accumulator have to store received packet identifiers forever?

colin-axner:
Interchain Accounts would benefit greatly from this (it would order based on account sequences). See [discussion](https://github.com/cosmos/interchain-accounts/issues/22#issue-837410534)

I think this would be a great feature to add to TAO v1.1.0

PR #563, opened 2021-04-22
Cross-chain valset update spec

josef-widder:
This is a draft for the specification for cross-chain validator set update by @jovankomatovic. It currently consists of two modules

- a shared queue (causally consistent)
- a module that uses the shared queue to maintain the validator sets

This PR is for discussion. 

The plan so to replace the queue by a new sliding window module where one chains calls `add` to append an item at the top of the list, and the other chain calls `ack` to acknowledge an item.

### DIFF
--- a/spec/app/ics-028-cross-chain-valset-update/README.md
+++ b/spec/app/ics-028-cross-chain-valset-update/README.md
@@ -1,0 +1,257 @@
+# Technical Specification
+
+## Introduction
+
+This document presents a technical specification for the **Cross-Chain Validation** protocol.
+The basic idea of the Cross-Chain Validation protocol is to allow validators that are already securing some existing blockchain (parent blockchain) to also secure a "new" blockchain (baby blockchain).
+The stake bonded at the parent blockchain guarantees that a validator behaves correctly at the baby blockchain.
+Otherwise, the validator is slashed on the parent blockchain.
+
+Therefore, at a high level, we can imagine the Cross-Chain Validation protocol to be concerned with following entities:
+  - Parent blockchain: This is a blockchain that "provides" validators. Namely, "provided" validators have some stake at the parent blockchain. Any misbehavior of a validator is slashed at the parent blockchain. Moreover, parent blockchain manipulates the validator set of a chain that "borrows" validators from it.
+  - Baby blockchain: Baby blockchain is a blockchain that is being secured by the parent blockchain. In other words, validators that secure and operate the baby blockchain are bonded on the parent blockchain. Any misbehavior of a validator at the baby blockchain is punished at the parent blockchain (i.e., the validator is slashed at the parent blockchain).
+  - Causally consistent shared FIFO queue: This queue is used for communication among two blockchains, as we explain in the rest of the document.
+
+Note that the protocol we present is generalized to multiple baby blockchains.
+In other words, a single parent blockchain might have multiple baby blockchains under its "jurisdiction".
+Hence, the Cross-Chain Validation protocol is concerned with more than two entities (one parent blockchain and potentially multiple baby blockchains).
+
+### Properties
+
+This subsection is devoted to defining properties that the Cross-Chain Validation protocol ensures.
+Recall that the parent blockchain has an ability to demand a change to the validator set of the baby chain.
+Moreover, we want to ensure the stake of validators of the baby blockchain are "frozen" at the parent chain.
+
+We present the interface of the protocol:
+  - Request: <ChangeValidatorSet, babyChain, V> - request made by the parent blockchain to change the validator set of babyChain using the validator set updates V.
+  We assume that each validator set update is unique (e.g., each validator set update V could simply have a unique identifier).
+  - Indication: <Unbonded, babyChain, V> - indication to the parent blockchain that the validator set update V (i.e., its "effect") has unbonded on the baby blockchain.
+
+Hence, we aim to achieve the following properties:
+- *Liveness*: If the parent blockchain demands a validator set change V for the baby chain, then the validator set of the baby blockchain reflects this demand.
+- *Validator set change safety*: Suppose that the validator set of the baby blockchain changes from *V* to *V'*. Then, there exists a sequence *seq = V1, V2, ..., Vn*, where *n > 0*, of change validator set demands issued by the parent blockchain such that *apply(V, seq) = V'*.
+- *Validator set change liveness*: If there exist infinitely many <ChangeValidatorSet, babyChain, V'> issued after <ChangeValidatorSet, babyChain, V>, then <Unbonded, babyChain, V> is eventually triggered.
+
+Let us explain the validator set change liveness property.
+First, note that <Unbonded, babyChain, Vlast> is never triggered if Vlast represent the **final** validator set change demand issued by the parent blockchain (the reason is that the validator set of the baby blockchain will never change after Vlast).
+Hence, it might be enough for V not to be the last demanded change.
+Unfortunatelly, that is not the case since multiple validator set change demands might be committed in the same block as the last demand.
+That is why we demand infinitely many demands to be issued after V.
+Note that if B represents the number of transactions committed in each block, then it is sufficient to demand B validator set change demands to be issued in order for V to eventually unbond.
+
+## High-level design of the Cross-Chain Validation protocol
+
+In this section, we provide an intuition behind our protocol.
+
+We use the causally consistent FIFO queue for "communication" among two blockchains (parent and baby).
+Namely, the parent blockchain enqueues new validator set changes and the baby blockchain dequeues these validator set changes once they have unbonded on the baby blockchain (which represents a signal that the stake of those validator could be set free by the parent blockchain).
+
+- Parent blockchain: Once the parent blockchain demands a change of validator set, it enqueues the validator set.
+
+- Baby blockchain: Once the baby blockchain observes that there exist new validator set demands, the baby blockchain applies all these changes.
+Moreover, whenever the validator set of the baby blockchain is modified, the "old" validator set should start unbonding.
+That is the baby blockchain uses the **local** *UnbondingQueue* queue.
+Namely, whenever the validator set of the baby blockchain is modified, we enqueue to *UnbondingQueue* the sequence number of the validator set change demand that was **last** applied in order to obtain the old (i.e., changed) validator set.
+Once the unbonding period elapses for the validator set, the appropriate entry is dequeued from the shared queue.
+
+## Data Structures
+
+We devote this section to defining the data structures used to represent the states of both parent and baby blockchain, as well as the packets exchanged by the two blockchains.
+
+### Application data
+
+#### Parent blockchain
+
+- Q: The causally consistent FIFO shared queue.
+
+#### Baby blockchain
+
+- Q: The causally consistent FIFO shared queue.
+- *UnbondingQueue*: Keeps track of validators that are currently unbonding.
+- validatorSetSeqNum: Number of validator set change demands processed; initialized to 0.
+- dequeueSeqNum: Number of elements dequeued from Q; initialized to 0.
+- lastObserved: Element of Q that was observed during the last Q.read() operation; initializated to nil.
+
+## Transitions
+
+In this section, we informally discuss the state transitions that occur in our protocol.
+We observe state transitions that are driven by a user (i.e., the new staking module on the parent chain), driven by the relayer and driven by elapsed time.
+
+  - User-driven state transitions: These transitions "start" the entire process of changing the validator set of the baby blockchain.
+  We assume that the will to change the validator set of the baby blockchain babyChain is expressed by invoking <ChangeValidatorSet, babyChain, V>.
+
+  - Time-driven state transitions: These transitions are activated since some time has elapsed. As we will present in the rest of the document, time-driven state transitions help us determine when the unbonding period, measured at the baby blockchain, has elapsed for a validator set.
+
+In the rest of the document, we will discuss the aforementioned state transitions in more detail.
+
+## Function Definitions
+
+### Parent blockchain
+
+This subsection will present the functions executed at the parent blockchain.
+
+```golang
+// expresses will to modify the validator set of the baby blockchain;
+func changeValidatorSet(
+  babyChainId: ChainId
+  valSetUpdate: Validator[]
+) {
+  // enqueue to the queue shared with babyChainId
+  Q[babyChainId].enqueue(valSetUpdate)
+}
+```
+
+- Expected precondition
+  - There exists a blockchain with *babyChainId* identifier
+  - All validators from *valSetUpdate* are validators at the parent blockchain
+- Expected postcondition
+  - The valSetUpdate is enqueued to the queue shared with the blockchain with *babyChainId*
+- Error condition
+  - If the precondition is violated
+
+TODO discuss and write a staking logic on endBlock
+Important: By reading the shared queue Q, the parent blockchain is able to conclude which validators could get their stake back.
+
+### Baby blockchain
+
+```golang
+// initialization function
+func init() {
+  UnbondingQueue = new local FIFO queue
+  validatorSetSeqNum = 0
+  dequeueSeqNum = 0
+  lastObserved = nil
+}
+```
+
+
+```golang
+// function used in End-Block method
+// returns all the elements enqueued since the last invocation of this method
+func observeChangesQ() { 
+  // get the content of the shared queue
+  content = Q.read()
+
+  // check whether the last observed element is in content
+  if (!content.has(lastObserved)) {
+    // update the lastObserved to the last element of content if content is not empty
+    if (!content.empty()) lastObserved = content.last()
+
+    // return the entire content
+    return content
+  } else {
+    // set "old" lastObserved
+    oldLastObserved = lastObserved
+
+    // update the lastObserved to the last element of content
+    lastObserved = content.last()
+
+    // return the entire content of content starting from oldLastObserved (excluding oldLastObserved)
+    return content.startFrom(oldLastObserved)
+  }
+}
+```
+
+- Expected precondition
+  - None
+- Expected postcondition
+  - Returns the elements enqueued since the last Q.read() operation was invoked.
+- Error condition
+  - None
+
+```golang
+// End-Block method executed at the end of each block
+func endBlock(block: Block) {
+  // get time
+  time = block.time
+
+  // finish unbonding for mature validator sets
+  while (!UnbondingQueue.isEmpty()) {
+    // peak the first queue entry
+    seqNum, startTime = UnbondingQueue.peak()
+
+    if (startTime + UNBONDING_PERIOD >= time) {
+      // remove from the unbonding queue
+      seqNum, startTime = UnbondingQueue.dequeue()
+
+      // dequeue all elements until seqNum
+      for (i = dequeueSeqNum + 1; i <= seqNum; i++) {
+        Q.dequeue()
+      }
+      dequeueSeqNum = seqNum
+
+    } else {
+      break
+    }
+  }
+
+  // get the new changes
+  changes = observeChangesQ()
+
+  // if there are no changes, return currnt validator set
+  if (changes.empty()) {
+    // validator set remains the same
+    ABCI.updateValidatorSet(block.validatorSet)
+    return
+  }
+
+  // get the old validator set
+  oldValSet = block.validatorSet
+
+  // get the new validator set; init to the old one
+  newValSet = oldValSet
+
+  // start unbonding for the old validator set represented by the validator set update;
+  // "start unbonding" simply means adding to the queue of validator set changes that started unbonding
+  UnbondingQueue.enqueue(validatorSetSeqNum, time)
+
+  // update the validator set
+  while (!changes.isEmpty()) {
+    valSetUpdate = changes.dequeue()
+
+    // update the new validator set
+    newValSet = applyValidatorUpdate(newValSet, valSetUpdate)
+
+    // increment validatorSetSeqNum
+    validatorSetSeqNum++
+
+    // remember which demands participate
+    if (content.isEmpty()) {
+      newSeqNum = validatorSetSeqNum
+    }
+  }
+
+  return newValSet
+}
+```
+
+- Expected precondition
+  - Every transaction from the *block* is executed
+- Expected postcondition
+  - Unbonding starts for the old validator set
+  - Unbonding finishes for all validator set that started unbonding more than *unbondingTime* before *time = block.time*. Moreover, the *UnbondingOver* packet is created for each such validator set
+  - The new validator set *newValSet* is pushed to the Tendermint protocol and *newValSet* reflects all the change validator set demands from the *block*
+- Error condition
+  - If the precondition is violated
+
+
+## Correctness Arguments
+
+Here we provide correctness arguments for the liveness, validator set change safety and liveness properties.
+
+### Liveness
+Suppose that the IBC communication indeed successfully relays the change validator set demand to the baby blockchain.
+Therefore, the validator set of the baby blockchain should reflect this demand.
+This indeed happens at the end of a block, since every observed demand is applied in order for the baby blockchain to calculate the new validator set.
+Hence, the property is satisfied.
+
+### Validator set change safety
+Suppose that the validator set of the baby blockchain changes from *V* to *V'* in two consecutive blocks.
+By construction of the protocol, we conclude that there exists a sequence of change validator set demands issued by the parent blockchain that result in *V'* when applied to *V*.
+Recursively, we conclude that this holds for any two validator sets (irrespectively of the "block distance" between them).
+
+### Validator set change liveness
+
+Since infinitely many demands are issued after the demand V, we conclude that eventually the unbonding period is started for V.
+The unbonding period eventually elapses and the V is dequeued at that moment.
+As soon as the dequeued is "observed" by the parent blockchain, the indication is triggered.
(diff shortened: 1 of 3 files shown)


spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+  - Indication: <Unbonded, babyChain, V> - indication to the parent blockchain that the validator set update V (i.e., its "effect") has unbonded on the baby blockchain.
+
+Hence, we aim to achieve the following properties:
+- *Liveness*: If the parent blockchain demands a validator set change V for the baby chain, then the validator set of the baby blockchain reflects this demand.
+- *Validator set change safety*: Suppose that the validator set of the baby blockchain changes from *V* to *V'*. Then, there exists a sequence *seq = V1, V2, ..., Vn*, where *n > 0*, of change validator set demands issued by the parent blockchain such that *apply(V, seq) = V'*.
+- *Validator set change liveness*: If there exist infinitely many <ChangeValidatorSet, babyChain, V'> issued after <ChangeValidatorSet, babyChain, V>, then <Unbonded, babyChain, V> is eventually triggered.
+
+Let us explain the validator set change liveness property.
+First, note that <Unbonded, babyChain, Vlast> is never triggered if Vlast represent the **final** validator set change demand issued by the parent blockchain (the reason is that the validator set of the baby blockchain will never change after Vlast).
```
AdityaSripal, line 36: Unbonded acknowledgement should be sent for `Vlast` even if this is the final validator set change demanded by parent chain correct?

The baby chain validator set after `VLast` should be:

`ApplyChanges(VLast, PrevValidatorSet)`

immediately after the validator set change is received.

Applying the last validator set change may cause some baby chain validators to unbond. Thus, we still need to send  the acknowledgement so that the stake on parent chain can be unlocked for any validators that unbonded on the final validator set change.

This should happen regardless independent of whether any future validator set changes come in.

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+This subsection will present the functions executed at the parent blockchain.
+
+```golang
+// expresses will to modify the validator set of the baby blockchain;
+func changeValidatorSet(
+  babyChainId: ChainId
+  valSetUpdate: Validator[]
+) {
+  // enqueue to the queue shared with babyChainId
+  Q[babyChainId].enqueue(valSetUpdate)
```
AdityaSripal, line 100: This is `SendPacket` on ORDERED channel

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+  while (!UnbondingQueue.isEmpty()) {
+    // peak the first queue entry
+    seqNum, startTime = UnbondingQueue.peak()
+
+    if (startTime + UNBONDING_PERIOD >= time) {
+      // remove from the unbonding queue
+      seqNum, startTime = UnbondingQueue.dequeue()
+
+      // dequeue all elements until seqNum
+      for (i = dequeueSeqNum + 1; i <= seqNum; i++) {
```
AdityaSripal, line 178: dequeueSeqNum not defined

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+        Q.dequeue()
```
AdityaSripal, line 179: This is `WriteAcknowledgement` on ORDERED channel

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+    // increment validatorSetSeqNum
+    validatorSetSeqNum++
+
+    // remember which demands participate
+    if (content.isEmpty()) {
+      newSeqNum = validatorSetSeqNum
+    }
+  }
+
+  return newValSet
```
AdityaSripal, line 224: Endblock just needs to return the `abci.ValidatorUpdate` not the entire validator set

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+Hence, it might be enough for V not to be the last demanded change.
+Unfortunatelly, that is not the case since multiple validator set change demands might be committed in the same block as the last demand.
+That is why we demand infinitely many demands to be issued after V.
+Note that if B represents the number of transactions committed in each block, then it is sufficient to demand B validator set change demands to be issued in order for V to eventually unbond.
+
+## High-level design of the Cross-Chain Validation protocol
+
+In this section, we provide an intuition behind our protocol.
+
+We use the causally consistent FIFO queue for "communication" among two blockchains (parent and baby).
```
ethanfrey, line 46: Isn't this an ordered channel?
Can't we stick with the existing IBC primitives in all cases we can, and if they are not sufficient, then a clear explanation of why we cannot use them, and proposals for what is needed (do you really need a DAG with causal ordering? it seems FIFO is same as "ordered")

spec/app/ics-070-shared-queue/README.md
```diff
@@ -0,0 +1,485 @@
...
+
+We present two versions of the queue module.
+The versions provide slightly different interface and guarantees, as we describe in the following subsection.
+
+### Shared FIFO Queue Specification
+
+Our shared FIFO queue is a concurrent object.
+We now define both versions of our queue.
+
+#### Sequential Consistent FIFO Queue
```
ethanfrey, line 21: Maybe I am being dense, but this seems a long description of exisiting ordered channels

spec/app/ics-070-shared-queue/README.md
```diff
@@ -0,0 +1,485 @@
...
+- *contains(x)* operation: Operation that returns whether the element *x* belongs to the queue.
+
+We set the following constraints:
+- Enqueue and contains operations are invoked solely by the parent blockchain.
+- Dequeue and peak operations are invoked solely by the baby blockchain.
+
+Lastly, our implementation satisfies the **sequential consistency** correctness criterium:
+The result of any execution is the same as if the operations of all the processes were executed in some sequential order, and the operations of each individual process appear in this sequence in the order specified by its protocol.
+
+#### Causally Consistent FIFO Queue
```
ethanfrey, line 36: And this the proposed causally ordered channels. It would make sense to contribute to https://github.com/cosmos/ibc/issues/550 rather than invent new concepts.

ethanfrey (commented):
It seems there is a lot of work to redefine the existing channel concepts and minimal info on the actual packet flow.

I would suggest removing all that queue stuff, replacing it with channels, and then focusing on the actual packets and handling.

If queues are really important, please define those in a different PR, so we can just focus on actual validator update logic here

spec/app/ics-028-cross-chain-valset-update/README.md
josef-widder: I was thinking about that a lot recently. For me, the write-up of the ICS focuses about send/receive packet (also when we consider how token transfer is designed/implemented), rather than on, e.g., how to read (over iterate over) e.g., all (un)acknowledged packets. I believe that these reads may be an important programming abstractions that are somehow written between the lines in the ICSs. In writing this spec, it turned out that this abstractions are quite useful. 

It might well be that our queues coincide with what can be obtained from reading the channel, as you suggest. The point for me is that we should try to make this explicit. I now believe this can be done with the existing primitives. What is missing for me is an explicit explanation/API what can be done with these primitives in addition to send/receive of individual packet.

For instance, we could introduce a function `pendingpackets()`. It returns a list of all pending packets in a channel (we can define it over the store). The return value of these function might have the same semantics as our queues. Perhaps this function already exists in the implementation? Then it would just be a matter of mentioning it in e.g., the channel ICS, and we can the use it directly for the valset update instead of introducing a new module.

Does that make sense?

spec/app/ics-070-shared-queue/README.md
josef-widder: Thanks a lot for the pointer. As far as I see there are slight differences. #550 talks about explicitly specifying predecessors, while in causal consistency (on the interface) relations are not explicitly stated. Causal consistency is a well-establish correctness criteria for shared data structures in concurrent programming. Actually, from our current understanding in the two chain case, ordered channels coincide with this "Causally Consistent FIFO Queue", while in the multi-chain case, we would need additional indices.

AdityaSripal:
@ethanfrey agreed that the queue can be implemented with ORDERED channels and this is the direction that the implementation will take. Once the implementation uses ORDERED channels, we can come back to this queue spec and see what can be omitted and what can be repurposed (perhaps as documentation on how to use ORDERED channels).

In the implementation we will have ORDERED channel. Parent chain sends `ValidatorSetChangePacket` across the channel.
Baby chain will send an **asynchronous acknowledgement** back after the unbonding period is over. This allows IBC to take care of ordering completely and reduces the need for a separate unbonding message

AdityaSripal:
I've run into two issues related to initialization while implementing. Can we please port over the init spec here so people can give it a review. I believe @gamarin2 had a proposal for how it should work. The exact mechanism by which the intial validator set is finalized and the baby chain can be started can be determined later.

I would like to surface 2 concerns:

**1. How should a baby chain verify that it is talking to the parent chain?**

It is not enough to simply check the chainID as that can be trivially spoofed.
Proposals:

- Set a root of trust consensus state in baby chain's genesis file. In channel handshake, the baby chain verifies that this consensus state exists on the underlying client. Benefit is that we rely on parent chain validator set honesty and we can verify within the channel handshake itself. Con is that we must enforce that someone updates the client on baby chain with this specific consensus state. This requires some manual work on part of a relayer.
- Baby chain validators post signatures on parent chain and require they post to only correct parent chain if they're honest. Parent chain sends packet with signatures over to baby chain who then verifies signatures. Pros: No manual relayer work. Cons: Reliance on honesty of baby chain (not a big deal), expensive to store and verify.

**2. How to handle slashing events on parent chain during channel handshake**

The protocol requires the initial validator set of baby chain to be frozen until the channel handshake is fully complete, since parent chain cannot send validator set changes until channel is OPEN. This prevents validators from unbonding their stake during this time, however validators may still have their stake changed by getting slashed on the parent chain. This information cannot be propogated to baby chain until channel is OPEN.
Proposals:
- Slash immediately and close any in-progress channel and require baby chain to start with new initial validator set. Pros: Simple, secure. Cons: Inconvenient, baby chain must generate new genesis file and start channel handshake process again. They must repeat process until they get an OPEN channel before a slashing event can occur.
- Slash parent stake immediately, but allow channel handshake to continue with outdated baby-chain validator set. Once the channel is OPEN, immediately send over the validator set changes from slashing events. Pros: Convenient, no need to keep restarting chains. Cons: Security implications unclear, implementation may be more complicated.

AdityaSripal:
I've proposed what I believe is a very workable fix for 1 here (see child genesis logic): https://github.com/cosmos/ibc-go/pull/116

Effectively the child chain's genesis state includes the parent client information. The client state for parent is constructed on child chain during chain initialization. The CCV channel must be built on top of this underlying client.

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Introduction
+
+This document presents a technical specification for the **Cross-Chain Validation** protocol.
+The basic idea of the Cross-Chain Validation protocol is to allow validators that are already securing some existing blockchain (parent blockchain) to also secure a "new" blockchain (baby blockchain).
+The stake bonded at the parent blockchain guarantees that a validator behaves correctly at the baby blockchain.
+Otherwise, the validator is slashed on the parent blockchain.
+
+Therefore, at a high level, we can imagine the Cross-Chain Validation protocol to be concerned with following entities:
+  - Parent blockchain: This is a blockchain that "provides" validators. Namely, "provided" validators have some stake at the parent blockchain. Any misbehavior of a validator is slashed at the parent blockchain. Moreover, parent blockchain manipulates the validator set of a chain that "borrows" validators from it.
```
cwgoes, line 11:
```suggestion
  - Parent blockchain: This is a blockchain that "provides" validators. Namely, "provided" validators have some stake at the parent blockchain. Any misbehavior of a validator is slashed on the parent blockchain. Moreover, the parent blockchain manipulates the validator set of a chain that "borrows" validators from it.
```

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+  - Request: <ChangeValidatorSet, babyChain, V> - request made by the parent blockchain to change the validator set of babyChain using the validator set updates V.
+  We assume that each validator set update is unique (e.g., each validator set update V could simply have a unique identifier).
+  - Indication: <Unbonded, babyChain, V> - indication to the parent blockchain that the validator set update V (i.e., its "effect") has unbonded on the baby blockchain.
+
+Hence, we aim to achieve the following properties:
+- *Liveness*: If the parent blockchain demands a validator set change V for the baby chain, then the validator set of the baby blockchain reflects this demand.
+- *Validator set change safety*: Suppose that the validator set of the baby blockchain changes from *V* to *V'*. Then, there exists a sequence *seq = V1, V2, ..., Vn*, where *n > 0*, of change validator set demands issued by the parent blockchain such that *apply(V, seq) = V'*.
+- *Validator set change liveness*: If there exist infinitely many <ChangeValidatorSet, babyChain, V'> issued after <ChangeValidatorSet, babyChain, V>, then <Unbonded, babyChain, V> is eventually triggered.
```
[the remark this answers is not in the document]
cwgoes: shouldn't there only need to be one change that unbonds the validator in question?

spec/app/ics-028-cross-chain-valset-update/README.md
cwgoes: I'm just confused by the semantics intended here. If we want the baby chain to control when unbonding happens - so that we can ensure that the baby chain reduces voting power (safely) before a validator is unbonded - can't we just have it acknowledge the validator set updates in some standard fashion (even using IBC-provided acknowledgements, ideally)? Why do we need these separate semantics and infinite messages?

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+
+```golang
+// function used in End-Block method
+// returns all the elements enqueued since the last invocation of this method
+func observeChangesQ() { 
+  // get the content of the shared queue
+  content = Q.read()
+
+  // check whether the last observed element is in content
```
cwgoes, line 135: Do we need to do all of this? Can't we just iterate over anything in the queue and apply the changes?

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+- Error condition
+  - If the precondition is violated
+
+
+## Correctness Arguments
+
+Here we provide correctness arguments for the liveness, validator set change safety and liveness properties.
+
+### Liveness
+Suppose that the IBC communication indeed successfully relays the change validator set demand to the baby blockchain.
```
cwgoes, line 243: There should be a synchrony bound here, no? What about timeouts?

spec/app/ics-070-shared-queue/README.md
```diff
@@ -0,0 +1,485 @@
...
+3) There exists an operation C such that A --> C and C --> B.  
+Moreover, operations A and B are *concurrent* if neither causally precedes the other.
+
+Finally, we are able to define the causal consistency correctness criterium.
+Operations that are related by the causal precedence relation are observed by all parties (i.e., blockchains) in their causal precedence order.
+Note that the definition allows concurrent operations to be seen in different order by different blockchains.
+
+### Closer Look at the IBC Channels
+
+An IBC channel assumes two parties (the respective blockchains) involved in the communication. However, it also assumes a relayer which handles message transmissions between the two blockchains. The relayer carries a central responsibility in ensuring communication between the two parties through the channel.
```
cwgoes, line 67: Hmm, the relayer plays a role, yes, but how would you avoid that with a different design? I'm confused.

cwgoes (changes requested):
Sorry it took me so long to review this, see comments. I admit I don't understand the rationale behind a separate queue design, can you clarify what specifically IBC ordered channels _do not_ provide in terms of causal ordering semantics which you require here? Could IBC channels be modified somehow to achieve that?

josef-widder:
> Sorry it took me so long to review this, see comments. I admit I don't understand the rationale behind a separate queue design, can you clarify what specifically IBC ordered channels _do not_ provide in terms of causal ordering semantics which you require here? Could IBC channels be modified somehow to achieve that?

Thanks for the comments, @cwgoes. I will take a closer look next week. Just a quick reply on the queue: In the current ICSs, the use case for channels is basically via send/receive packet, that is, via an event-based semantics. For this application (and potentially for other future application) it seems convenient to have a more state-based semantics, where a module can read, e.g., the packets that have been sent in the past. We were initially trying to express this via a queue. After discussions with @AdityaSripal, we realized that in principle we can do this directly on the channel. We are currently waiting for Aditya's initial implementation to see what is the best way to capture this.

For me right now the question is whether it would make sense to define functions that allow us to, e.g., iterate over the sent/received packets (that would correspond to a "read") within the ICSs. @AdityaSripal was more in favor of doing this in form of a manual that explains how to perform a "read". I would be happy to discuss this.

ethanfrey:
> In the current ICSs, the use case for channels is basically via send/receive packet, that is, via an event-based semantics. For this application (and potentially for other future application) it seems convenient to have a more state-based semantics, where a module can read, e.g., the packets that have been sent in the past.

The application could easily write each received packet to a local queue to keep a history (formatted/processed before storing).
What I don't understand is why this is needed here? 

Packet Handler: Receive packet -> locally store valset diff -> return acknowledgement
EndBlocker: Return valset diff to tendermint (if it exists) -> clear local diff

This seems simple enough and would obviate the need for any queue - basically you reduce all pending packets into one value as they arrive, and consume in the and blocker

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
+# Technical Specification
+
+## Introduction
+
+This document presents a technical specification for the **Cross-Chain Validation** protocol.
+The basic idea of the Cross-Chain Validation protocol is to allow validators that are already securing some existing blockchain (parent blockchain) to also secure a "new" blockchain (baby blockchain).
```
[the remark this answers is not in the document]
jovankomatovic: No, they should simply be managed by the parent chain.

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+Note that the protocol we present is generalized to multiple baby blockchains.
+In other words, a single parent blockchain might have multiple baby blockchains under its "jurisdiction".
+Hence, the Cross-Chain Validation protocol is concerned with more than two entities (one parent blockchain and potentially multiple baby blockchains).
+
+### Properties
+
+This subsection is devoted to defining properties that the Cross-Chain Validation protocol ensures.
+Recall that the parent blockchain has an ability to demand a change to the validator set of the baby chain.
+Moreover, we want to ensure the stake of validators of the baby blockchain are "frozen" at the parent chain.
```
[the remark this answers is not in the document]
jovankomatovic: Yes, exactly. The baby blockchain does not make any changes.

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+We present the interface of the protocol:
+  - Request: <ChangeValidatorSet, babyChain, V> - request made by the parent blockchain to change the validator set of babyChain using the validator set updates V.
+  We assume that each validator set update is unique (e.g., each validator set update V could simply have a unique identifier).
+  - Indication: <Unbonded, babyChain, V> - indication to the parent blockchain that the validator set update V (i.e., its "effect") has unbonded on the baby blockchain.
```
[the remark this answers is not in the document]
jovankomatovic: Yes, sure. I will clarify it!

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+
+Hence, we aim to achieve the following properties:
+- *Liveness*: If the parent blockchain demands a validator set change V for the baby chain, then the validator set of the baby blockchain reflects this demand.
```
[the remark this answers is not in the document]
jovankomatovic: No, I will put "eventually".

spec/app/ics-028-cross-chain-valset-update/README.md
jovankomatovic: We need to figure out which updates are "new". Otherwise, we could apply one validator set update more than once, which we want to avoid.

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+The stake bonded at the parent blockchain guarantees that a validator behaves correctly at the baby blockchain.
+Otherwise, the validator is slashed on the parent blockchain.
+
+Therefore, at a high level, we can imagine the Cross-Chain Validation protocol to be concerned with following entities:
+  - Parent blockchain: This is a blockchain that "provides" validators. Namely, "provided" validators have some stake at the parent blockchain. Any misbehavior of a validator is slashed at the parent blockchain. Moreover, parent blockchain manipulates the validator set of a chain that "borrows" validators from it.
+  - Baby blockchain: Baby blockchain is a blockchain that is being secured by the parent blockchain. In other words, validators that secure and operate the baby blockchain are bonded on the parent blockchain. Any misbehavior of a validator at the baby blockchain is punished at the parent blockchain (i.e., the validator is slashed at the parent blockchain).
+  - Causally consistent shared FIFO queue: This queue is used for communication among two blockchains, as we explain in the rest of the document.
```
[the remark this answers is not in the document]
jovankomatovic: I meant that queue is an important "building block" of the protocol. This part is removed now that we are not talking about the queue, but about IBC channels.

spec/app/ics-028-cross-chain-valset-update/README.md
jovankomatovic: It was defined in the init.
Now that we have abandoned the queue semantics, this is removed.

spec/app/ics-028-cross-chain-valset-update/README.md
jovankomatovic: Fixed.

spec/app/ics-028-cross-chain-valset-update/README.md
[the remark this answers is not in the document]
jovankomatovic: This is fixed. Please, take a look at the updated version.

spec/app/ics-028-cross-chain-valset-update/README.md
```diff
@@ -0,0 +1,257 @@
...
+  - None
+
+```golang
+// End-Block method executed at the end of each block
+func endBlock(block: Block) {
+  // get time
+  time = block.time
+
+  // finish unbonding for mature validator sets
+  while (!UnbondingQueue.isEmpty()) {
```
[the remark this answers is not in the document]
jovankomatovic: Not sure I understand.
This simply implements a timer for each of the validator set changes.
I clarified it in the updated version.
Please, let me know if I am missing something.

spec/app/ics-070-shared-queue/README.md
jovankomatovic: Sorry, I do not understand your question.

spec/app/ics-028-cross-chain-valset-update/README.md
jovankomatovic: Yes, the synchrony bound is the IBC timeout.

AdityaSripal:
> The application could easily write each received packet to a local queue to keep a history (formatted/processed before storing).
What I don't understand is why this is needed here

This is already how the implementation works: https://github.com/cosmos/ibc-go/pull/116

 We discussed in one of our first meetings how to use the IBC ORDERED channel and local stores to implement the queue. I believe the next step in the spec is to incorporate those changes.

spec/app/ics-028-cross-chain-valset-update/README.md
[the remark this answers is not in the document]
cwgoes: Aye, just wanted to clarify that.

spec/app/ics-028-cross-chain-valset-update/README.md
cwgoes: I think this question resulted from my confusion in how the queue was being used, since clarified.

spec/app/ics-028-cross-chain-valset-update/README.md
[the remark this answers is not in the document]
cwgoes: Never mind, sorry.

spec/app/ics-028-cross-chain-valset-update/README.md
cwgoes: Can we clarify that?

spec/app/ics-070-shared-queue/README.md
cwgoes: I think we need to be more specific. IBC requires the existence of at least one correct relayer for liveness, but that's a pretty weak requirement, the relayer has nothing to do with correctness and only one correct relayer needs to exist.

AdityaSripal:
We want to make sure that the following properties are enforced during channel initialization (handshake callbacks):

- A correct parent chain will not complete handshake with counterfeit child chain
- A correct child chain will not complete handshake with counterfeit parent chain
- A correct child chain and parent chain will always be able to complete handshake if channel is not yet open.
- There cannot be more than 1 open CCV channel between parent chain and child chain
- The first open CCV channel will remain the only open CCV channel unless a voluntary closing is initiated

It would be good to get some formal understanding of whether the current work in the implementation meets these properties, and a way to encode these properties into the spec itself.

mpoke:
PR is deprecated.

Issue #564, opened 2021-04-22
ICS-27 Revision


seantking:
Continuing the conversation that has been ongoing here -> https://github.com/cosmos/interchain-accounts/issues/22

Issues that have been outlined thus far:

### **Channel Ordering:** 
At present, the specification does not outline how many accounts can be used/registered per channel.
The implementation [here](https://github.com/cosmos/interchain-accounts/blob/master/x/ibc-account/keeper/account.go#L13) interprets the spec to mean that you can multiplex multiple accounts on a single channel namespaced via `{port-id}/{channel-id}/{salt}`.

A potential actor vector with this approach can be described as follows (thanks @ethanfrey):
```
In an unordered channel, if 5 packets come in: 1, 2, 3, 4, 5, a relayer could relay 5 and 2. An hour later relay 4. And never relay 1 or 3 (which would eventually time out).

In an ordered channel, I could not relay packet 2 until I relay 1 or I have sent proof that packet 1 timed out. That is the meaning of an ordered channel.

Now, in this case, assume there is a community pool that controls an account on another chain. I have a group account registered on the same channel. I see there is an upcoming vote from the community pool, so I send 100s of packets with long (2 weeks?) timeouts. I never pay for them. The community pool then writes packet 863 and wants to relay it. But to do so, they must wait until all my packets time out (and relay the timeout proof), or relay packets 1-862.

I see multiple distrusting accounts operating on one ordered channel as a huge attack vector. I think the definition of unordered channels and allowing multiple ordered channels on one connection are the principal advances in the ICS spec since I wrote one in late 2017 and we really should make use of this.
```

@okwme
You asked me earlier why can we not just say that the standard here (maybe defined in developer documentation somewhere) should be that relayers/client/helper module developers should only allow one account per channel. The problem is how can we enforce this? A user will always have to trust the client they are using, unless they are running their own relayer. In the short term, I guess this won't be a problem as we can communicate this to application developers who can enforce this in their own application logic. I guess longer term this would get messy. 

**Short term solution: 1 account per 1 ordered channel**
Quoting @AdityaSripal 
So I'm in favor of a single ORDERED channel dedicated to a single account for now, with partially ordered channels as an ideal upgrade that might happen in the 6+month time frame as ethan suggests.

This simplifies the packet data since we already know who the authorizer is whenever they send a packet through a channel.

[...]

PR #567, opened 2021-04-27
update: major revisions to ics-27 specification

seantking:
Updating the ics-027 specification based on ongoing discussions:

https://github.com/cosmos/ibc/issues/564
https://github.com/cosmos/interchain-accounts/issues/22


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 ```typescript
-interface IBCAccountPacketAcknowledgement {
-    type: Type
-    code: uint32
-    data: Uint8Array
-    error: string
+message Acknowledgement {
+  // response contains either a result or an error and must be non-empty
+  oneof response {
+    bytes  result = 21;
```
seantking, line 121: What do we want to send back to the sending chain on a successful `RunTx`? I specified above that we should send the chain-id and the address of the interchain account but I'm wondering what is actually necessary here.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
-  data: Uint8Array
+`authenticateTx` validates the transaction and checks that the signers in the transaction have the right permissions (are the owners of the account in question).`
+
+`runTx` executes a transaction after the transaction has been successfully authenticated.
+
+### Packet Data
+`InterchainAccountPacketData` contains an array of messages that an interchain account can run and a memo string that is sent to the recieving chain. The example below is defined as a proto encoded message but each chain can encode this differently. This should be clearly defined in the documentation for each chain specific implementation.
+
+```typescript
+message InterchainAccountPacketData  {
```
seantking, line 109: This is basically what our `RunTx` message type was previously. Seeing as we have removed the need for the register message entirely this makes sense to me. The only message type we will be sending will be for running transactions on the receiving chain.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+    repeated google.protobuf.Any messages = 1;
+    string memo = 2;
```
seantking, line 111: @ethanfrey What was your idea for using `memo` here? I'm wondering what data is useful to send here.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
-`runTx` executes a transaction after the transaction has been successfully authenticated.
+The recieving chain must implement authentication with regards to ensuring that the incoming messages are sent by the owner of the targetted interchain account. With regards to authorization, it is up to each chain specific implementation to decide if the hosted interchain accounts have authority to invoke all of the chains message types or only a subset. This configuration may be set up at module initialization.
```
seantking, line 71: The current implementation does not have any concept of `Authorization` on the receiving chain i.e. what messages interchain accounts can execute. I think having this as an optional feature makes sense. Some chains may not want remote chains to be able to execute every message type. Wdyt?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -201,14 +208,15 @@ function onChanOpenAck(
   // port has already been validated
   // assert that version is "ics27-1"
   abortTransactionUnless(version === "ics27-1")
+  // state change to keep track of successfully registered interchain account
+  confirmInterchainAccountRegistration()
```
seantking, line 212: This is where we would invoke a state change to keep a local registry of what accounts are registered on what chains. I'm wondering, can we pass back the address of the interchain account here? 

@AdityaSripal Is this what you had in mind?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
-Each chain must implement the interfaces as defined in this section in order to support interchain accounts.
+##### Partially Ordered Channels 
+Interchain accounts is a perfect use case for partially ordered channels, whereby the order is based on account sequences. However, this has yet to be implemented in IBC. As of the time of writing this specification the current progress can be tracked [here](https://github.com/cosmos/ibc/issues/550). 
 
-`tryRegisterIBCAccount` method in `IBCAccountModule` interface defines the way to request the creation of an IBC account on the host chain (or counterparty chain that the IBC account lives on). The host chain creates an IBC account using its account creation logic, which may use data within the packet (such as destination port, destination channel, etc) and other additional data for the process. The origin chain can receive the address of the account that was created from the acknowledge packet.
+### Architecture Diagram
+![](https://i.imgur.com/HX1h2u2.png)
```
seantking, line 64: I hosted this on imgur just for now. Where should we keep this asset?

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: We can drop it. It just came from the standard `sdk.Tx` protobuf type, and figured it might be interesting for documentation. The same way the memo field on transactions doesn't actually do anything but can be observed by clients.

Fine to remove if you don't want it. Figured an optional field wouldn't hurt

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: When we defined the [subcall standard in CosmWasm](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/results/subcall.rs#L52-L65) (to allow getting results from a call to another contract/module), we pass back the data field and the events returned.

A similar format is used in the tx responsed (data and events for each message).

Maybe events are too variable for returning over IBC, but I would at least include the data field (which is part of the ResultHash)

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: I agree. But you may want to make multiple types for future extensibility? `ics27-2` 

Keeping a type field (Any) here with just one defined, supported concrete form is a good step there.
(Not necessary, but an idea)

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: I think the idea is good in the spec. Allowing the receiver to reject packets.

The wire protocol would remain the same, correct? Just another potential error case

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: I like the idea of passing back account here

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: https://giphy.com/ :trollface: 

But seriously, can't you commit the png to this repo and link it in the same directory?

seantking: Yeah, that's the obvious solution. I just wanted to determine a protocol for this. All the images in the other specs are currently broken.

spec/app/ics-027-interchain-accounts/README.md
seantking: Exactly. In Cosmos SDK land this could just be a list of message types we can pass into the module at init time as an optional field. If the incoming message type is disallowed it will throw an unauthorized error.

emadami3060 (approved):

spec/app/ics-027-interchain-accounts/README.md
seantking: I'm not entirely sure what an `event` would be in this context. Would you mind giving me an example?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+
 The IBC handler interface & IBC relayer module interface are as defined in [ICS 25](../ics-025-handler-interface) and [ICS 26](../ics-026-routing-module), respectively.
 
 ### Desired Properties
 
 - Permissionless
-- Fault containment: Interchain account must follow rules of its host chain, even in times of Byzantine behaviour by the counterparty chain (the chain that manages the account)
-- The chain that controls the account must process the results asynchronously and according to the chain's logic. The result's code should be 0 if the transaction was successful and an error code other than 0 if the transaction failed.
-- Sending and receiving transactions will be processed in an ordered channel where packets are delivered exactly in the order which they were sent.
+- Fault containment: An interchain account must follow rules of its host chain, even in times of Byzantine behavior by the counterparty chain (the chain that manages the account)
```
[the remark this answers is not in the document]
seantking: Great question, I left it in but I wasn't entirely sure either if I'm being honest. I think the point here is that the interchain account is just a regular account, even tho it's being controlled by another chain it has no more power than any other account on the chain. 

I'm okay with removing this if others are.

spec/app/ics-027-interchain-accounts/README.md
seantking: Any thoughts on the actual diagram in general?

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: Every message in the sdk returns optional binary data and a list of events.

Events each have a type string, and a list of key-value attributes (also strings).

Eg.bank send emits "transfer" type with recipient, sender, and amount as attributes

spec/app/ics-027-interchain-accounts/README.md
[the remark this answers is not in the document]
cwgoes: I see - I think that's a good point, but it's not what I assumed the sentence meant, maybe we can just rephrase?

spec/app/ics-027-interchain-accounts/README.md
cwgoes: I don't really understand the arrows, why are arrows going to the relayer? The ordering (numbers) makes sense to me.

seantking: Thanks for the feedback, perhaps we can remove/improve the arrows in some way. 

In this context, we're calling onChanOpenInit and emitting an event, which the relayer picks up and then begins the handshake process on both chains for creating a channel. The arrows are trying to symbolize that the relayer is both listening for events + coordinating the handshake process. I guess this isn't being communicated clearly tho.

cwgoes: I see - yes, that's true, maybe labelling the arrows or something would help.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+- The chain that controls the account must process the results asynchronously and according to the chain's logic. The acknowledgment message should contain a result of an error as described in [ics-4](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#acknowledgement-envelope).
```
colin-axner, line 31: I don't understand `according to the chain's logic`. This seems to be referring to receiving acknowledgements but what logic is done when processing a failed/successful acknowledgement?

Or maybe the wording is incorrect? Do you think it is referring to the OnRecv or OnAcknowledge callbacks?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+- Sending and receiving transactions will be processed in an ordered channel where packets are delivered exactly in the order in which they were sent. 
+- Each ordered channel must only allow one interchain account to use it, otherwise, malicious users may be able to block transactions from being received. Practically, N accounts per N channels can be achieved by creating a new port for each user (owner of the interchain account) and creating a unique channel for each new registration request. Future versions of ics-27 will use partially ordered channels to get around this security issue and make the process more straight forward
```
colin-axner, line 33:
```suggestion
- Each ordered channel must only allow one interchain account to use it, otherwise, malicious users may be able to block transactions from being received. Practically, N accounts per N channels can be achieved by creating a new port for each user (owner of the interchain account) and creating a unique channel for each new registration request. Future versions of ics-27 will use partially ordered channels to allow multiple interchain accounts on the same channel to preserve account sequence ordering without being reliant on one another
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
-The interchain account specification defines the general way to register an interchain account and transfer tx bytes. The counterparty chain is responsible for deserialising and executing the tx bytes, and the sending chain should know how counterparty chain will handle the tx bytes in advance.
+The implementation of interchain accounts is non-symmetric. This means that each sending chain can have a different way to generate an interchain account and each receiving chain a different way to deserialise the transaction bytes and a different set of transactions that they can execute. For example, chains that use the Cosmos SDK will deserialise tx bytes using Protobuf, but if the counterparty chain is a smart contract on Ethereum, it may deserialise tx bytes by an ABI that is a minimal serialisation algorithm for the smart contract.
```
colin-axner, line 37:
```suggestion
The implementation of interchain accounts is non-symmetric. This means that each sending chain can have a different way to generate an interchain account and each receiving chain can have a different set of transactions that may be deserialised in different ways. For example, chains that use the Cosmos SDK will deserialise tx bytes using Protobuf, but if the counterparty chain is a smart contract on Ethereum, it may deserialise tx bytes by an ABI that is a minimal serialisation algorithm for the smart contract.
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+
+A chain can implement one or both parts to the interchain accounts protocol (sending and receiving). A sending chain registering and controlling an account does not necessarily have to allow other chains to register accounts on its own chain, and vice versa. 
 
-Each chain must satisfy following features to create a interchain account:
+The interchain account specification defines the general way to register an interchain account and transfer tx bytes. The counterparty chain is responsible for deserialising and executing the tx bytes, and the sending chain should know how the counterparty chain will handle the tx bytes in advance. Each chain-specific implementation should clearly document how serialization/deserialization of transactions happens and ensure the required packet data format is clearly outlined.
```
colin-axner, line 41:
```suggestion
The interchain account specification defines the general way to register an interchain account and transfer transaction bytes. The counterparty chain is responsible for deserialising and executing the transaction bytes, and the sending chain should know how the counterparty chain will handle the transaction bytes in advance. Each chain-specific implementation should clearly document how serialization/deserialization of transactions happens and ensure the required packet data format is clearly outlined. 
```


`the sending chain should know how the counterparty chain will handle the transaction bytes in advance`


I think maybe the sending chain just needs to know what transaction format are supported? The counterparty chain could do black box magic and the sending chain could still successfully interact so long as it uses the correct transaction format

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+The interchain account specification defines the general way to register an interchain account and transfer tx bytes. The counterparty chain is responsible for deserialising and executing the tx bytes, and the sending chain should know how the counterparty chain will handle the tx bytes in advance. Each chain-specific implementation should clearly document how serialization/deserialization of transactions happens and ensure the required packet data format is clearly outlined. 
 
-- New interchain accounts must not conflict with existing ones.
-- Each chain must keep track of which counterparty chain created each new interchain account.
+Each chain must satisfy the following features to create an interchain account:
 
-Also, each chain must know how the counterparty chains serialise/deserialise transaction bytes in order to send transactions via IBC. And the counterparty chain must implement the process of safely executing IBC transactions by verifying the authority of the transaction's signers.
+- New interchain accounts must not conflict with existing ones
+- Each chain must keep track of which counterparty chain created each new interchain account
```
colin-axner, line 46: Isn't this implicit via IBC channels?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
 The chain must reject the transaction and must not make a state transition in the following cases:
 
-- The IBC transaction fails to be deserialised.
-- The IBC transaction expects signers other than the interchain accounts made by the counterparty chain.
+- The IBC transaction fails to be deserialised
+- The authentication step on the receiving chain (where the interchain account is hosted) fails 
 
-It does not restrict how you can distinguish signers that was not made by the counterparty chain. But the most common way would be to record the account in state when the interchain account is registered and to verify that signers are recorded interchain account.
+#### Known Issues
```
colin-axner, line 53:
```suggestion
### Known Issues
```

This seems like this section should be at the bottom of the spec?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+##### Ordered Channels 
+In an ordered channel, packet 2 cannot be relayed until packet 1 has been relayed or a valid proof that packet 1 has timed out has been received. If multiple distrusting accounts are allowed to use a single ordered channel this creates an attack vector whereby malicious users can block packets from being received in a timely manner. An example of this may be a malicious user sending hundreds of packets that are never paid for. Other users trying to use this channel will have to wait for all of these packets to be fully relayed or receive valid proof of these packets having timed out before their own packets can be relayed. Therefore, multiplexing N interchain accounts on 1 ordered channel is not viable. This specification assumes the following: N interchain accounts will exist on N ordered channels.
```
colin-axner, line 55:
```suggestion
In an ordered channel, the next packet cannot be relayed until the previous packet has been relayed. If the previous packet is timed out, the ordered channel will be closed and the next packet can never be relayed. If multiple distrusting accounts are allowed to use a single ordered channel this creates an attack vector whereby malicious users can block packets from being received in a timely manner. An example of this may be a malicious user sending hundreds of packets that are never paid for. Other users trying to use this channel will have to wait for all of these packets to be fully relayed or receive valid proof of these packets having timed out before their own packets can be relayed. Therefore, multiplexing N interchain accounts on 1 ordered channel is not viable. This specification assumes the following: N interchain accounts will exist on N ordered channels. 
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
-### Data Structures
+##### Unordered Channels 
+An unordered channel lets packets be relayed in any order. A user can send 100 packets and never relay them. Another user submits one packet and can immediately relay just that one without having to relay all other packets. In theory interchain accounts can be implemented using an unordered channel so long as messages that are dependent on one another are packed into a single packet. However, this pattern differs from other standards such as how Tendermint's mempool operates. Rather than deviate from well defined standards for message processing it is recommended to not to use this approach or at a minimum clearly document how messages are processed on the receiving chain.
```
colin-axner, line 58: It feels slightly odd that this section goes from a declaration into an example without a transition

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
-### Data Structures
+##### Unordered Channels 
+An unordered channel lets packets be relayed in any order. A user can send 100 packets and never relay them. Another user submits one packet and can immediately relay just that one without having to relay all other packets. In theory interchain accounts can be implemented using an unordered channel so long as messages that are dependent on one another are packed into a single packet. However, this pattern differs from other standards such as how Tendermint's mempool operates. Rather than deviate from well defined standards for message processing it is recommended to not to use this approach or at a minimum clearly document how messages are processed on the receiving chain. 
 
-Each chain must implement the interfaces as defined in this section in order to support interchain accounts.
+##### Partially Ordered Channels 
+Interchain accounts are a perfect use case for partially ordered channels, whereby the order is based on account sequences. However, this has yet to be implemented in IBC. As of the time of writing this specification the current progress can be tracked [here](https://github.com/cosmos/ibc/issues/550).
```
colin-axner, line 61:
```suggestion
Interchain accounts are a perfect use case for partially ordered channels, whereby the order is based on account sequences. However, this has yet to be specified and implemented in IBC. As of the time of writing this specification the current progress can be tracked [here](https://github.com/cosmos/ibc/issues/550). 
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
-`runTx` executes a transaction after the transaction has been successfully authenticated.
+The receiving chain must implement authentication with regard to ensuring that the incoming messages are sent by the owner of the targeted interchain account. With regards to authorization, it is up to each chain-specific implementation to decide if the hosted interchain accounts have the authority to invoke all of the chain's message types or only a subset. This configuration may be set up at module initialization.
```
colin-axner, line 71: I think @AdityaSripal might have good suggestions here, but I think it needs to be specified that the receiving chain authorization is dependent upon the sending chain authorization. The receiving chain should just be checking that the owners the sending chain authorized to send the messages are also authorized to execute those messages on the receiving chain. 

Ex:
Sending chain authorizes a packet which says addrA will send 100 atoms from addrA to addrB. Receiving chain should check that the owner (addrA) matches the from address (addrA) and that the spendable balance is correct. 

If the scenario was: Sending chain authorizes a packet which says addrA will send 100 atoms from addrC to addrB
Then the receiving chain should throw an error because addrA is not authorized to spend from addrC account. What cannot really happen is that the receiving chain verifies some signature by addrA over the messages being sent. This sort of verification should be done by the sending chain

colin-axner (commented):
part way through, will continue reviewing later - currently on sending interface

spec/app/ics-027-interchain-accounts/README.md
seantking: Thanks for the feedback. 

Just so we're on the same page: Authentication confirms that users are who they say they are. Authorization gives those users permission to access a resource.

Authentication & Authorization on the sending side determines who can create accounts, and what transactions those accounts can send. Every chain that wants to act as a sender will need to implement this logic. 

Authentication on the receiving side is simply checking is the sender of this message, the owner of the targetted account.

An additional, optional step, may be to add authorization on the receiving side,  by disallowing certain messages from being invoked by interchain accounts. This can just be an array of message types that get passed in at init time. Maybe, the creators of a chain do not want external chains to be able to access all message types via an interchain account. This was raised by B-harvest in a conversation.

spec/app/ics-027-interchain-accounts/README.md
seantking: What do you mean by "via IBC channels"? 

I think this is quite implicit in general, you're correct. I think it's okay to leave it explicit though.

spec/app/ics-027-interchain-accounts/README.md
seantking: Very good.

spec/app/ics-027-interchain-accounts/README.md
seantking: Good point, I'll update this.

spec/app/ics-027-interchain-accounts/README.md
colin-axner: Ahh, I think it is supposed to convey that each account needs to be associated with a channel?

spec/app/ics-027-interchain-accounts/README.md
colin-axner: I see, thanks for the clarification. I think I was expecting to read the authentication on the receiving side before the additional, optional step as the authentication on the receiving side is non-optional

spec/app/ics-027-interchain-accounts/README.md
seantking: I think the point here is that the sending chain needs to keep track of which accounts it has registered and on what chain. Likewise, the receiving side needs to do the same. 

Clearly, this should be communicated better. The line in question was added in the previous version of the spec, and I may have interrupted it differently than you have.

spec/app/ics-027-interchain-accounts/README.md
seantking: Cool. I'll think about that and update it to be more clear. Thanks.

colin-axner: I think it is clear and concise as is. It just took me a second pass/more detailed read. Looks good to me :+1:

spec/app/ics-027-interchain-accounts/README.md
colin-axner: Yea, I think given the context of IBC, it is more concise to specify that the sending/receivers must track which channels are used for which accounts. IBC doesn't really have a strong notion of "chain", rather it has clients. Connections built on clients also have security implications. Thus two accounts using the different channels but the same clients might have completely different security guarantees

colin-axner (commented):
Fantastic work!! LGTM. Left mostly typo fixes and a couple suggestions

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+
+
+The `tryRunTx` method in the`InterchainAccountModule` interface defines the way to create an outgoing packet for a specific chain type. The chain type determines how the IBC account transaction should be constructed and serialised for the receiving chain. The sending side should know in advance how the receiving side expects the incoming IBC packet to be structured. 
+
+#### Recieving Interfacce
+
+`createAccount` defines the way to determine the account's address by using the port & channel id. A newly created interchain account must not conflict with an existing account. Therefore, the host chain (the chain that the account will live on) must keep track of which blockchains have created an interchain account in order to verify the transaction signing authority in `authenticateTx`. `createAccount` should be called in the `chanOpenTry` callback as part of the channel creation handshake process defined in [ics-4](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics).
```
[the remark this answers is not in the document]
colin-axner: Thinking on this some more, I don't understand why we still need receiving side authentication if there is only one account per channel. The receiving chain should just check that `msg.GetSigners()` == `getAccount(channel-0, interchain-account)`. But `getAccount` can be stateless

spec/app/ics-027-interchain-accounts/README.md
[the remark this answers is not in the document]
milosevic: My understanding is that we need to clarify what are the semantics we want to provide in case of forks on controlling chain.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
-```typescript
-// `Type` enumerator defines the packet type.
-enum Type {
-  REGISTER,
-  RUNTX
-}
+#### Sending Interface
+The `tryRegisterInterchainAccount` method in the `InterchainAccountModule` interface defines the way to request the creation of an interchain account on a remote chain. The remote chain creates an interchain account using its own account creation logic. Due to the limitation of ordered channels, the recommended way to achieve this when calling `tryRegisterInterchainAccount` is to dynamically bind a new port with the port id set as the address of the owner of the account (if the port is not already bound), invoke `OpenChanInit` via an IBC module which will initiate the handshake process and emit an event signaling to a relayer to generate a new channel between both chains. The remote chain can then create the interchain account in the `ChanOpenTry` callback as part of the channel creation handshake process defined in [ics-4](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics). Once the `chanOpenAck` callback fires the handshake-originating (sending) chain can assume the account registration is successful.
```
milosevic, line 92: Is the reason why the protocol is tied to channel handshake is a requirement that interchain account needs to be mapped to a single channel? What I don't get is how the mapping between interchain account on dst chain, controller on source chain and channel is achieved.

seantking: I'm planning to push some updates to this tomorrow. I don't think the emitting of the event is necessary here or needs to be part of the protocol. 

> Is the reason why the protocol is tied to channel handshake is a requirement that interchain account needs to be mapped to a single channel? 

This was the idea. But perhaps this does not need to be part of the protocol at all. I'm thinking that we begin registering the account on the `onChanOpenInit` callback rather than wait for the event as mentioned here.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
   onTxFailed(sourcePort:string, sourceChannel:string, txBytes: Uint8Array)
 }
 ```
 
-### Subprotocols
-
-The subprotocols described herein should be implemented in an "interchain-account-bridge" module with access to a router and codec (decoder or unmarshaller) for the application and access to the IBC relayer module.
-
 ### Port & channel setup
+Receiving chains (chains that will host interchain accounts) must always bind to a port with the id `interchain_account`. Sending chains will bind to ports dynamically, with each port id being the address of the interchain account owner.
```
milosevic, line 136: Can a single account owner (identified with its address) be owning several interchain accounts on a given receiving chain?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -139,21 +155,14 @@ function setup() {
...
 ### Channel lifecycle management
 
-Both machines `A` and `B` accept new channels from any module on another machine, if and only if:
+An interchain account module will accept new channels from any module on another machine, if and only if:
 
-- The other module is bound to the "ibcaccount" port.
-- The channel being created is ordered.
-- The version string is "ics27-1".
+- The channel being created is ordered
+- The version string is "ics27-1"
```
milosevic, line 165: Is there a need to check if port id is equal to account address and that message is signed with the address owner, or we don't care if someone else create interchain account in name of some owner's address as only owner can control it after?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -201,14 +208,15 @@ function onChanOpenAck(
...
 }
 ```
 
 ```typescript
 function onChanOpenConfirm(
   portIdentifier: Identifier,
   channelIdentifier: Identifier) {
-  // accept channel confirmations, port has already been validated
 }
 ```
```
milosevic, line 222: In case channel is closed, access to the interchain account is lost if I am not wrong. If that's true, access to assets is lost, so I wonder if there is something we can do here.

spec/app/ics-027-interchain-accounts/README.md
colin-axner: I'm also curious if a single account can own several interchain accounts on different receiving chains. I'd like to use a single account on chainA to create an account on chainB, chainC, and chainD and I can use my one authentication setup to control all 3 accounts

spec/app/ics-027-interchain-accounts/README.md
colin-axner: 1. No, channels are really not designed to be passing information. The only mechanism I can think of is using the channel version
2. Correct, although I'm a little confused. Isn't this the controlling side so won't the account be the portID? The counterparty can get this information if needed via the counterparty PortID
3. Packets cannot be received if the channel isn't OPEN. I don't see any reasons for concern. The handshake can always be completed unless there is a bug or the counterparty chain refuses to communicate. OnChanAck is like the account is created and OnChanConfirm is like the account is ready to be used

spec/app/ics-027-interchain-accounts/README.md
okwme, line 30:
```suggestion
- Fault containment: The origin chain of an Interchain Account has complete control of authorization and authentication for the account. The host chain accepts at face value that the origin chain is acting correctly when sending messages on behalf of the Interchain Account.
```

This was my interpretation of the meaning and purpose of the section.

spec/app/ics-027-interchain-accounts/README.md
okwme: I don't see a problem with this beyond that it's maybe irrelevant/distracting information. The origin chain may have custom logic like: once first tx is submitted and successfully processed create a new one and begin the process of submitting it. Basically the origin chain is must process the results async and is _able_ to use that ack within it's own logic.

okwme:
```suggestion
- The chain that controls the account must process the results asynchronously. The acknowledgment message should contain a result of an error as described in [ics-4](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#acknowledgement-envelope) which can be used as part of the chain's logic.
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
-The interchain account specification defines the general way to register an interchain account and transfer tx bytes. The counterparty chain is responsible for deserialising and executing the tx bytes, and the sending chain should know how counterparty chain will handle the tx bytes in advance.
+The implementation of interchain accounts is non-symmetric. This means that each sending chain can have a different way to generate an interchain account and each receiving chain can have a different set of transactions that may be deserialised in different ways. For example, chains that use the Cosmos SDK will deserialise tx bytes using Protobuf, but if the counterparty chain is a smart contract on Ethereum, it may deserialise tx bytes by an ABI that is a minimal serialisation algorithm for the smart contract.
+
+A chain can implement one or both parts to the interchain accounts protocol (sending and receiving). A sending chain registering and controlling an account does not necessarily have to allow other chains to register accounts on its own chain, and vice versa.
```
okwme, line 39:
```suggestion
A chain can implement one or both parts of the interchain accounts protocol (sending and receiving). A sending chain registering and controlling an account does not necessarily have to allow other chains to register accounts on its own chain, and vice versa. 
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+The interchain account specification defines the general way to register an interchain account and transfer tx bytes. The counterparty chain is responsible for deserialising and executing the tx bytes, and the sending chain should know how the counterparty chain will handle the tx bytes in advance. Each chain-specific implementation should clearly document how serialization/deserialization of transactions happens and ensure the required packet data format is clearly outlined. 
 
-- New interchain accounts must not conflict with existing ones.
-- Each chain must keep track of which counterparty chain created each new interchain account.
+Each chain must satisfy the following features to create an interchain account:
 
-Also, each chain must know how the counterparty chains serialise/deserialise transaction bytes in order to send transactions via IBC. And the counterparty chain must implement the process of safely executing IBC transactions by verifying the authority of the transaction's signers.
+- New interchain accounts must not conflict with existing ones
+- Each chain must keep track of which counterparty chain created each new interchain account
```
okwme, line 46:
```suggestion
- Each origin chain and host chain must track which channels are used for which accounts.
```

spec/app/ics-027-interchain-accounts/README.md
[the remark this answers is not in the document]
dtribble: I would also like interchain accounts that have limited off-chain control.  e.g., you can control staking but not selling. In that case, byzantine behavior by the controller could only compromise the allowed subset of functionality.

spec/app/ics-027-interchain-accounts/README.md
dtribble: > result of an error

"result *or* error..."?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+- Fault containment: An interchain account must follow rules of its host chain, even in times of Byzantine behavior by the counterparty chain (the chain that manages the account)
+- The chain that controls the account must process the results asynchronously and according to the chain's logic. The acknowledgment message should contain a result of an error as described in [ics-4](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#acknowledgement-envelope).
+- Sending and receiving transactions will be processed in an ordered channel where packets are delivered exactly in the order in which they were sent. 
+- Each ordered channel must only allow one interchain account to use it, otherwise, malicious users may be able to block transactions from being received. Practically, N accounts per N channels can be achieved by creating a new port for each user (owner of the interchain account) and creating a unique channel for each new registration request. Future versions of ics-27 will use partially ordered channels to allow multiple interchain accounts on the same channel to preserve account sequence ordering without being reliant on one another
```
dtribble, line 33: There are plenty of scenarios in which a module on a remote chain might be controlling multiple accounts on the destination chain, so sharing a channel adds no additional risk.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
+
+A chain can implement one or both parts to the interchain accounts protocol (sending and receiving). A sending chain registering and controlling an account does not necessarily have to allow other chains to register accounts on its own chain, and vice versa. 
 
-Each chain must satisfy following features to create a interchain account:
+The interchain account specification defines the general way to register an interchain account and transfer tx bytes. The counterparty chain is responsible for deserialising and executing the tx bytes, and the sending chain should know how the counterparty chain will handle the tx bytes in advance. Each chain-specific implementation should clearly document how serialization/deserialization of transactions happens and ensure the required packet data format is clearly outlined.
```
dtribble, line 41: Since it's asymmetric, let's use terms like source and destination or some such. "counterparty" is too context sensitive.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
-Each chain must implement the interfaces as defined in this section in order to support interchain accounts.
+##### Partially Ordered Channels 
+Interchain accounts are a perfect use case for partially ordered channels, whereby the order is based on account sequences. However, this has yet to be specified and implemented in IBC. As of the time of writing this specification the current progress can be tracked [here](https://github.com/cosmos/ibc/issues/550). 
 
-`tryRegisterIBCAccount` method in `IBCAccountModule` interface defines the way to request the creation of an IBC account on the host chain (or counterparty chain that the IBC account lives on). The host chain creates an IBC account using its account creation logic, which may use data within the packet (such as destination port, destination channel, etc) and other additional data for the process. The origin chain can receive the address of the account that was created from the acknowledge packet.
+### Architecture Diagram
+![](https://i.imgur.com/HX1h2u2.png)
```
dtribble, line 64: Isn't the diagram just `TryRegisterChainAccount` followed by normal channel setup handshake?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -20,110 +20,126 @@ On Ethereum, there are two types of accounts: externally owned accounts, control
...
 
-`runTx` executes a transaction after the transaction has been successfully authenticated.
+The receiving chain must implement authentication with regard to ensuring that the incoming messages are sent by the owner of the targeted interchain account. With regards to authorization, it is up to each chain-specific implementation to decide if the hosted interchain accounts have the authority to invoke all of the chain's message types or only a subset. This configuration may be set up at module initialization.
```
dtribble, line 71: If it's controlling access, then it's not authentication, it's authorization. Why does something have to be from the owner? Does that mean control of an account is not transferrable?

spec/app/ics-027-interchain-accounts/README.md
seantking: Hey Dean, thanks for the feedback! 

Regarding multiple untrusted users (account owners) using the same channel please take a look at this comment for reference: https://github.com/cosmos/interchain-accounts/issues/22#issuecomment-820431565. 

The issue stems from multiple parties using the same ordered channel.

spec/app/ics-027-interchain-accounts/README.md
seantking: Thanks for the feedback. I think you're right.

spec/app/ics-027-interchain-accounts/README.md
seantking: The diagram, in this case, is signaling that the registration of the account is occurring during the channel creation process, specifically in the handshake callbacks.

spec/app/ics-027-interchain-accounts/README.md
seantking: In this design, the account can only be controlled by a single owner, correct.

Regarding authorization/authentication: I guess I was thinking of this as a validation step on the receiving side for ensuring the incoming IBC packet/instructions for the account are actually from the account owner. However, as Colin pointed out [here](https://github.com/cosmos/ibc/pull/567#discussion_r627449733) this shouldn't be necessary. I will update this section.

spec/app/ics-027-interchain-accounts/README.md
seantking: It is a good point, as technically anyone can create a channel. I think it's okay to allow anyone to create a channel as the account can only be controlled by the owner anyway. 

I'm not sure how we would enforce this otherwise. The relayer will be signing these messages right?

spec/app/ics-027-interchain-accounts/README.md
seantking: I'm thinking there are probably several ways to achieve reclaiming an account. 

A question that comes to mind: Do we want several channels to be active at one time, all of which can send instructions to the account? 

Also, if owner A needs to reclaim an account, what channel are they using to do that? In the case of a DAO the reclaiming process would need to be communicated over IBC.

spec/app/ics-027-interchain-accounts/README.md
seantking: I removed the diagram

spec/app/ics-027-interchain-accounts/README.md
seantking: I removed this in the end.

spec/app/ics-027-interchain-accounts/README.md
seantking: Removed this section

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,114 +16,119 @@ This standard document specifies packet data structure, state machine handling l
...
   onTxFailed(sourcePort:string, sourceChannel:string, txBytes: Uint8Array)
 }
 ```
 
-### Subprotocols
-
-The subprotocols described herein should be implemented in an "interchain-account-bridge" module with access to a router and codec (decoder or unmarshaller) for the application and access to the IBC relayer module.
-
 ### Port & channel setup
+The interchain account module on a host chain must always bind to a port with the id `interchain_account`. Controller chains will bind to ports dynamically, with each port id set as `ics27-1-{owner-address}`.
```
AdityaSripal, line 125: Could we make the port naming a bit more obvious? `interchain_account_host`, `interchain_account_control_{owner-address}`

Or replace `interchain_account` with `ics27-1` in both cases

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: Since `ChanOpenInit` is being initiated by module from the `RegisterAddress` function, we can check address against the signer of the `RegisterAddress` message

AdityaSripal: But this should be checked on controlling chain, NOT host chain as this will only get relayer-signed messages

AdityaSripal: Since the code on controller chain is enforcing this property we don't need a check at all

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -201,14 +197,16 @@ function onChanOpenAck(
   // port has already been validated
   // assert that version is "ics27-1"
   abortTransactionUnless(version === "ics27-1")
+  // state change to keep track of successfully registered interchain account
+  setActiveChannel(SourcePortId)
 }
 ```
 
 ```typescript
 function onChanOpenConfirm(
```
AdityaSripal, line 206: You also need `Close` callbacks since that needs to "unset" active channel. Is the plan that we create a new channel against the same portID to "recover" the account?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -139,21 +144,14 @@ function setup() {
...
+An interchain account module will accept new channels from any module on another machine, if and only if:
 
-- The other module is bound to the "ibcaccount" port.
-- The channel being created is ordered.
-- The version string is "ics27-1".
+- The channel being created is ordered
+- The version string is "ics27-1"
 
 ```typescript
 function onChanOpenInit(
```
AdityaSripal, line 157: We should be erroring here and on all handshake messages if `activeChannel` already set

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: I think we should just maintain a single active channel at a time. Multiple channels at a time means there's no global ordering of messages

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -308,39 +262,9 @@ function onAcknowledgePacket(
...
 ```
 
-```typescript
-function onTimeoutPacket(packet: Packet) {
-  // Receiving chain should handle this event as if the tx in packet has failed
-  switch (ack.type) {
-  case Type.RUNTX:
-    onTxFailed(packet.sourcePort, packet.sourceChannel, packet.data.data)
-    return
-}
```
AdityaSripal, line 318: Why is this no longer true?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -166,8 +164,6 @@ function onChanOpenInit(
   version: string) {
   // only ordered channels allowed
   abortTransactionUnless(order === ORDERED)
-  // only allow channels to "ibcaccount" port on counterparty chain
-  abortTransactionUnless(counterpartyPortIdentifier === "ibcaccount")
```
AdityaSripal, line 170: Don't these handshakes still need to check counterparty port based on:
 
The interchain account module on a host chain must always bind to a port with the id `interchain_account`. Controller chains will bind to ports dynamically, with each port id set as `ics27-1-{owner-address}`.

AdityaSripal (commented):
Have some questions regarding account recovery

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,114 +16,119 @@ This standard document specifies packet data structure, state machine handling l
...
+On Ethereum, there are two types of accounts: externally owned accounts, controlled by private keys, and contract accounts, controlled by their contract code ([ref](https://github.com/ethereum/wiki/wiki/White-Paper)). Similar to Ethereum's contract accounts, interchain accounts are controlled by another chain (not a private key) while retaining all the capabilities of a normal account (i.e. stake, send, vote, etc). While an Ethereum CA's contract logic is performed within Ethereum's EVM, interchain accounts are managed by a seperate chain via IBC in a way such that the owner of the account retains full control over how it behaves. ICS27-1 primarily targets the use cases of DAO investing and staking derivatives over IBC.
```
charleenfei, line 19: 'separate'

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,114 +16,119 @@ This standard document specifies packet data structure, state machine handling l
...
-  createAccount(packet: Packet, data: Uint8Array): Address
-  deserialiseTx(txBytes: Uint8Array): Tx
-  authenticateTx(tx: Tx): boolean
-  runTx(tx: Tx): Result
-}
-```
+`createInterchainAccount`  A newly created interchain account must not conflict with an existing account. `createInterchainAccount` should be called in the `chanOpenTry` callback as part of the channel creation handshake process defined in [ics-4](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics).
 
-`IBCAccountPacketData` specifies the type of packet. The `IBCAccountPacketData.data` is arbitrary and can be used for different purposes depending on the value of `IBCAccountPacketData.type`.
+`runTx` executes a transaction based on the IBC packet recieved from the controller chain.
```
charleenfei, line 92: received

spec/app/ics-027-interchain-accounts/README.md
seantking: yes exactly

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -291,16 +260,10 @@ function onRecvPacket(packet: Packet) {
...
-      onAccountCreated(packet.sourcePort, packet.sourceChannel, ack.data)
-    }
-    return
-  }
-  case Type.RUNTX:
-    if (ack.code === 0) {
+    if (acknowledgement.result) {
       onTxSucceeded(packet.sourcePort, packet.sourceChannel, packet.data.data)
+      // Sets the interchainAccount address on the controller side for a given owner
+      setInterchainAccountAddress(sourcePort, result)
```
AdityaSripal, line 266: This doesn't make sense. The result is not the interchain address anymore from my understanding. Should this be removed?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -201,87 +204,53 @@ function onChanOpenAck(
...
+  setActiveChannel(CounterPartyPortId)
 }
 ```
 
 ```typescript
 function onChanCloseInit(
   portIdentifier: Identifier,
   channelIdentifier: Identifier) {
-  // no action necessary
+  unsetActiveChannel(CounterPartyPortId)
```
AdityaSripal, line 224: delete interchain account as well

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -201,87 +204,53 @@ function onChanOpenAck(
...
 }
 ```
 
 ```typescript
 function onChanCloseConfirm(
   portIdentifier: Identifier,
   channelIdentifier: Identifier) {
-  // no action necessary
+  unsetActiveChannel(SourcePort)
```
AdityaSripal, line 232: ditto

AdityaSripal (changes requested):
Everything looks great. My only blocking concern is whether we're sending address back in the acknowledgment

spec/app/ics-027-interchain-accounts/README.md
colin-axner: I think the idea is to leave the interchain account and allow it to be associated with a different active channel

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,114 +16,121 @@ This standard document specifies packet data structure, state machine handling l
...
-  createAccount(packet: Packet, data: Uint8Array): Address
-  deserialiseTx(txBytes: Uint8Array): Tx
-  authenticateTx(tx: Tx): boolean
-  runTx(tx: Tx): Result
-}
-```
+`createInterchainAccount`  A newly created interchain account must not conflict with an existing account. `createInterchainAccount` should be called in the `chanOpenTry` callback as part of the channel creation handshake process defined in [ics-4](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics).
 
-`IBCAccountPacketData` specifies the type of packet. The `IBCAccountPacketData.data` is arbitrary and can be used for different purposes depending on the value of `IBCAccountPacketData.type`.
+`executeTx` executes a transaction based on the IBC packet received from the controller chain.
```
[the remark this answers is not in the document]
seantking: Agree

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,114 +16,121 @@ This standard document specifies packet data structure, state machine handling l
...
+ CounterPartyPortId: `interchain_account`,
+ CounterPartyChannelId: `channel-2`,
+}
+```
 
 ### Data Structures
 
-Each chain must implement the interfaces as defined in this section in order to support interchain accounts.
+```typescript
+interface InterchainAccountModule {
```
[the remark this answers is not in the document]
seantking: Well, the helper module uses the controller part of the `InterchainAccountModule` API. I wouldn't necessarily say that the helper module is the controller module per say (although in some ways it is). 

I think for now a single interface is fine, but in terms of readability of code, it may be nice to have two separate interfaces in the future.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,114 +16,121 @@ This standard document specifies packet data structure, state machine handling l
...
 
-Each chain must satisfy following features to create a interchain account:
+### Authentication & Authorization
 
-- New interchain accounts must not conflict with existing ones.
-- Each chain must keep track of which counterparty chain created each new interchain account.
+For the controller chain to register an interchain account on a host chain, first the controlling side must bind a new port to the interchain account module with the id `ics27-1-{connection-number}-{owner-address}` where `{owner-address}` is the account address of the interchain account owner & connection number is the number parsed from the connection id (e.g. `connection-1` would have a connection number of `1`). This port will be used to create channels between the controller & host chain for a specific owner/interchain account pair. Only the account with `{owner-address}` matching the bound port will be authorized to send IBC packets over channels created with `SourcePort` `ics27-1-{connection-number}-{owner-address}`. The host chain trusts that each controller chain will enforce this port registration and access.
```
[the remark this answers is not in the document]
AdityaSripal: I'm ok with keeping this restriction permanently I don't see a reason not to

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,114 +16,121 @@ This standard document specifies packet data structure, state machine handling l
...
-interface IBCAccountModule {
-  tryRegisterIBCAccount(data: Uint8Array)
-  tryRunTx(chainType: Uint8Array, data: any)
-  createAccount(packet: Packet, data: Uint8Array): Address
-  deserialiseTx(txBytes: Uint8Array): Tx
-  authenticateTx(tx: Tx): boolean
-  runTx(tx: Tx): Result
-}
-```
+`createInterchainAccount`  A newly created interchain account must not conflict with an existing account. `createInterchainAccount` should be called in the `chanOpenTry` callback as part of the channel creation handshake process defined in [ics-4](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics).
```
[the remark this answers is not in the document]
AdityaSripal: I'm in favor of approach 1

spec/app/ics-027-interchain-accounts/README.md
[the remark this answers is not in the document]
seantking: Updated this

spec/app/ics-027-interchain-accounts/README.md
seantking: Updated this

spec/app/ics-027-interchain-accounts/README.md
seantking: True. We do need a check on `executeTx` (host chain) to ensure that the signer of a message (returned from getSigners()) is in fact the interchain account address associated with the SourcePortId. Impl;cemented like so: https://github.com/cosmos/ibc-go/blob/interchain-accounts/modules/apps/ibc-account/keeper/relay.go#L144

spec/app/ics-027-interchain-accounts/README.md
seantking: Yes updated.

spec/app/ics-027-interchain-accounts/README.md
seantking: As colin suggests here: https://github.com/cosmos/ibc/pull/567/files#r669637554

We want to return the address of the interchain account on packet sequence 1. So in this case result would include the address.

seantking: https://github.com/cosmos/ibc/pull/567/files#diff-56983566230cc9f705648cb67111ac9a42e097e0de832ed165b6da08331cb299R249

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -241,10 +241,14 @@ function onRecvPacket(packet: Packet) {
...
-
+      executeTx(tx)
+      
+      // Gets the interchain account address, and returns it to the controller chain
+      // This step is necessary in order for the controller chain to know the address of the registered interchain account 
+      // We should only return the interchain account on packet sequence 1
+      const interchainAccount = getInterchainAccount(ctx, SourcePortId)
       return Acknowledgement{
-        result: result
+        result: interchainAccount
```
AdityaSripal: Acking since this is a quick change. But this should not just contain the interchain account. Otherwise we don't properly get the result of our first message execution.

We should get the interchain account **in addition** to first message result on ack sequence 1. The way to marshal both in ack, and then unmarshal should be specified in this document.

Will create a separate issue for this to be tackled in subsequent PR

AdityaSripal (approved):
Great work!! Splitting out my final comment into an issue to be tackled in separate PR

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -166,10 +166,12 @@ function onChanOpenInit(
...
   // only ordered channels allowed
   abortTransactionUnless(order === ORDERED)
-  // only allow channels to "ibcaccount" port on counterparty chain
-  abortTransactionUnless(counterpartyPortIdentifier === "ibcaccount")
+  // only allow channels to "interchain_account" port on counterparty chain
+  abortTransactionUnless(counterpartyPortIdentifier === "interchain_account")
   // version not used at present
   abortTransactionUnless(version === "ics27-1")
+  // Only open the channel if there is no active channel already set (with status OPEN)
+  abortTransactionUnless(activeChannel === nil)
```
colin-axner, line 174: This check isn't sufficient. Lets say I create an ICA account. The channel for that account closes, and it becomes inactive. Now there's a race condition between me trying to reopen a channel for that port and someone else sending a `ChanOpenInit` message

There's no check here that prevents any user from submitting a valid `ChanOpenInit` for my port ID except this one. As soon as the channel is inactive, my ICA account is vulnerable 

What is the primary concern with channels closing? From my understanding, the channel can only close if a packet gets timed out. Presumably this should never occur unless a light client attack occurs. If a light client attack occurs, I'd expect the funds in question to have been stolen anyways

seantking: If a random user manages to call OnChanOpenInit all it will do is open a new channel, it won't give that user access to the account. I think the only reason we need this check is so that we don't have random people creating lots of unnecessary channels. 

I think we can potentially remove the unSet active channel step entirely on OnCloseChan

colin-axner: Yes this is correct, after more exploration, the part that was missing for me is that `InitInterchainAccount` should only be called once to create a certain port ID. To recover a ICA after channel closure, the user must send a `MsgChanOpenInit`. I was looking at what would happen if a user didn't initiate a channel handshake via `InitInterchainAccount`

colin-axner:
@milosevic and @cwgoes whenever time arises to review this pr again, I think it'd be useful to reserve any minor fixes for a separate issue/followup pr and significant issues for requested changes. This pr is a major improvement to the existing spec, but is also overflowed with comments. I think it'd be a lot easier to review any remaining changes in separate prs

seantking:
> @milosevic and @cwgoes whenever time arises to review this pr again, I think it'd be useful to reserve any minor fixes for a separate issue/followup pr and significant issues for requested changes. This pr is a major improvement to the existing spec, but is also overflowed with comments. I think it'd be a lot easier to review any remaining changes in separate prs

I pushed a bunch more changes to this, mostly just making everything clearer and using a similar layout to the fee spec which had a nice section for the interface. 

@milosevic If you have time please take a look, it would be nice to get this merged in as 100+ comments make this PR extremely challenging to review. I think future PRs can tackle the minor updates that are necessary. It is evolving as the implementation evolves.

milosevic (approved):

Issue #569, opened 2021-05-03
ICS2: Allow Light Client implementations more flexibility in setting/getting of client states and consensus states


colin-axner:
Edit: see [comment](https://github.com/cosmos/ibc/issues/569#issuecomment-832051686)

### Summary

IBC clients as defined by ICS 02 should contain a version to allow backwards compatible changes to proof key formats and proto definitions. 

### Problem

As we continue to iterate on IBC after 1.0.0 has been released, we have ran into difficulty in making clean backwards compatible changes to proof keys. This is because the first notion of versioning occurs at the second layer of the TAO stack (03 connection). At the first layer we create clients and consensus states, but without versioning information it is impossible to cleanly make changes at the 02-client level in a backwards compatible way. 

It seems very strange to me that the first notion of versioning occurs at the second level. It has become clear to me that versions are primarily useful in providing information for usage at the layer above. For example, the current connection versioning is useful for establishing channel types. The current version in a channel is useful for providing information about the application. A client version would be useful in establishing information about proof keys and proto definitions used in connection handshake verification. 

Currently, if we want to change our client state or consensus state, it is impossible to do so in a backwards compatible fashion. We create our clients and update them before we establish an IBC connection version. There is no flexibility in using different store keys or different proto definitions. 

### Proposed Solution

Add a client version into each IBC client which represents the versions of IBC the counterparty currently supports. 

In ibc-go we would not store this version inside the `ClientState`. This is because we need to know the version of a client before we can even get the client as the version may indicate which store key to use. Instead we store the client version under a version key which should not be verified by counterpartys (this way the version key can change without waiting for the next major IBC version). Properly verifying the counterparty in connection handshakes acts as an implicit verification of the counterparty version. The client version is used to ensure the connection handshake can be completed properly, if the counterparty is using an unsupported version, the connection handshake would fail. 

#### Positives

Proof keys and proto definitions for clients can change in a backwards compatible way. 

#### Negatives

Another layer of versioning, may become confusing

### Unresolved Questions

- Does the client version need to be verified
- How does the counterparty chain indicate it updates its versioning if the version is not stored in the client state

Opening to start discussion as I think this is a useful feature in helping us decrease the amount of times IBC needs to do a synchronous network upgrade. Would love to see someone sketch out this idea in more detail

cwgoes:
[...]

PR #571, opened 2021-05-07
ICS 28: WASM Client

ParthDesai:
Reopened this PR: https://github.com/cosmos/ibc/pull/506

cc: @AdityaSripal @hxrts @cwgoes

### DIFF
--- a/spec/client/ics-028-wasm-client/README.md
+++ b/spec/client/ics-028-wasm-client/README.md
@@ -1,0 +1,451 @@
+---
+ics: 28
+title: WASM Client
+stage: draft
+category: IBC/TAO
+kind: instantiation
+implements: 2
+author: Parth Desai <parth@chorus.one>
+created: 2020-10-13
+modified: 2020-10-13
+---
+
+## Synopsis
+
+This specification document describes an interface to a client (verification algorithm) stored as a wasm bytecode for a blockchain.
+
+### Motivation
+
+WASM based light clients are decoupled from SDK source code, which enables one to upgrade existing light client or add support for other blockchain without modifying SDK source code.
+
+### Definitions
+
+Functions & terms are as defined in [ICS 2](../ics-002-client-semantics).
+
+`currentTimestamp` is as defined in [ICS 24](../ics-024-host-requirements).
+
+`WASM Client Code` refers to wasm bytecode stored in the client store, which provides target blockchain specific implementation of [ICS 2](../ics-002-client-semantics).
+
+`WASM Client` refers to a particular instance of `WASM Client Code` defined as a tuple `(WASM Client Code, ClientID)`.
+
+`WASM VM` refers to a virtual machine capable of executing valid wasm bytecode.
+
+
+### Desired Properties
+
+This specification must satisfy the client interface defined in ICS 2.
+
+## Technical Specification
+
+This specification depends on correctness of the `WASM Client code` in context of consensus algorithm of its target `blockchain`, as well as correct instantiation of `WASM Client`.
+
+### Client state
+
+The wasm client state tracks location of the wasm bytecode via `codeId`. Binary data represented by `data` field is opaque and only interpreted by the WASM Client Code. `type` represents client type.
+`type` and `codeId` both are immutable.
+
+```typescript
+interface ClientState {
+  codeId: []byte
+  data: []byte
+  frozen: boolean
+  frozen_height: Height
+  latest_height: Height
+  type: String
+}
+```
+
+### Consensus state
+
+The WASM consensus state tracks the timestamp (block time), `WASM Client code` specific fields and commitment root for all previously verified consensus states.
+`type` and `codeId` both are immutable.
+
+```typescript
+interface ConsensusState {
+  codeId: []byte
+  data: []byte
+  timestamp: uint64
+  root: MerkleRoot
+  type: String
+}
+```
+
+### Height
+
+The height of a WASM light client instance consists of two `uint64`s: the epoch number, and the height in the epoch.
+
+```typescript
+interface Height {
+  epochNumber: uint64
+  epochHeight: uint64
+}
+```
+
+Comparison between heights is implemented as follows:
+
+```typescript
+function compare(a: Height, b: Height): Ord {
+  if (a.epochNumber < b.epochNumber)
+    return LT
+  else if (a.epochNumber === b.epochNumber)
+    if (a.epochHeight < b.epochHeight)
+      return LT
+    else if (a.epochHeight === b.epochHeight)
+      return EQ
+  return GT
+}
+```
+
+This is designed to allow the height to reset to `0` while the epoch number increases by one in order to preserve timeouts through zero-height upgrades.
+
+### Headers
+
+Contents of wasm client headers depend upon `WASM Client Code`.
+
+```typescript
+interface Header {
+  data: []byte
+}
+```
+
+### Misbehaviour
+
+The `Misbehaviour` type is used for detecting misbehaviour and freezing the client - to prevent further packet flow - if applicable.
+WASM client `Misbehaviour` consists of two headers at the same height both of which the light client would have considered valid.
+
+```typescript
+interface Misbehaviour {
+  fromHeight: Height
+  h1: Header
+  h2: Header
+}
+```
+
+### Client initialisation
+
+WASM client initialisation requires a (subjectively chosen) latest consensus state interpretable by the target WASM Client Code. `wasmCodeId` field is unique identifier for `WASM Client Code`, and `initializationData` refers to opaque data required for initialization of the particular client managed by that WASM Client Code.
+
+```typescript
+function initialise(
+    wasmCodeId: String,
+    initializationData: []byte,
+    consensusState: []byte,
+  ): ClientState {
+    codeHandle = getWASMCode(wasmCodeId)
+    assert(codeHandle.isInitializationDataValid(initializationData, consensusState))
+    set("clients/{identifier}/consensusStates/{height}", consensusState)
+    return codeHandle.initialise(initializationData, consensusState)
+}
+```
+
+The `latestClientHeight` function returns the latest stored height, which is updated every time a new (more recent) header is validated.
+
+```typescript
+function latestClientHeight(clientState: ClientState): Height {
+  codeHandle = clientState.codeHandle();
+  codeHandle.latestClientHeight(clientState)
+}
+```
+
+### Validity predicate
+
+WASM client validity checking uses underlying WASM Client code. If the provided header is valid, the client state is updated & the newly verified commitment written to the store.
+
+```typescript
+function checkValidityAndUpdateState(
+  clientState: ClientState,
+  epoch: uint64,
+  header: Header) {
+    codeHandle = clientState.codeHandle()
+    isValid, consensusState = codeHandle.validateHeaderAndCreateConsensusState(clientState, header, epoch)
+    set("clients/{identifier}/consensusStates/{header.height}", consensusState)
+    // save the client
+    set("clients/{identifier}", clientState)
+}
+```
+
+### Misbehaviour predicate
+
+WASM client misbehaviour checking determines whether or not two conflicting headers at the same height would have convinced the light client.
+
+```typescript
+function checkMisbehaviourAndUpdateState(
+  clientState: ClientState,
+  misbehaviour: Misbehaviour) {
+    codeHandle = clientState.codeHandle()
+    consensusState = get("clients/{identifier}/consensusStates/{misbehaviour.fromHeight}")
+    assert(codeHandle.handleMisbehaviour(clientState, consensusState, misbehaviour))
+    // save the client
+    set("clients/{identifier}", clientState)
+}
+```
+
+### Upgrades
+
+The chain which this light client is tracking can elect to write a special pre-determined key in state to allow the light client to update its client state (e.g. with a new chain ID or epoch) in preparation for an upgrade.
+
+As the client state change will be performed immediately, once the new client state information is written to the predetermined key, the client will no longer be able to follow blocks on the old chain, so it must upgrade promptly.
+
+```typescript
+function upgradeClientState(
+  clientState: ClientState,
+  newClientState: ClientState,
+  height: Height,
+  proof: CommitmentPrefix) {
+    codeHandle = clientState.codeHandle()
+    codeHandle.verifyNewClientState(oldClientState, newClientState, height)
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that the provided consensus state has been stored
+    assert(root.verifyMembership(path, newClientState, proof))
+    // update client state
+    clientState = newClientState
+    set("clients/{identifier}", clientState)
+}
+```
+
+In case of wasm client, upgrade of `WASM Client Code` is also possible via blockchain specific management functionality.
+
+### State verification functions
+
+WASM client state verification functions check a Merkle proof against a previously validated commitment root.
+
+```typescript
+function verifyClientConsensusState(
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  clientIdentifier: Identifier,
+  consensusStateHeight: Height,
+  consensusState: ConsensusState) {
+    path = applyPrefix(prefix, "clients/{clientIdentifier}/consensusState/{consensusStateHeight}")
+    codeHandle = getCodeHandleFromClientID(clientIdentifier)
+    assert(codeHandle.isValidClientState(clientState, height))
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that the provided consensus state has been stored
+    assert(root.verifyMembership(path, consensusState, proof))
+}
+
+function verifyConnectionState(
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  connectionIdentifier: Identifier,
+  connectionEnd: ConnectionEnd) {
+    path = applyPrefix(prefix, "connections/{connectionIdentifier}")
+    codeHandle = clientState.codeHandle()
+    assert(codeHandle.isValidClientState(clientState, height))
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that the provided connection end has been stored
+    assert(root.verifyMembership(path, connectionEnd, proof))
+}
+
+function verifyChannelState(
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  channelEnd: ChannelEnd) {
+    path = applyPrefix(prefix, "ports/{portIdentifier}/channels/{channelIdentifier}")
+    codeHandle = clientState.codeHandle()
+    assert(codeHandle.isValidClientState(clientState, height))
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that the provided channel end has been stored
+    assert(root.verifyMembership(codeHandle.getProofSpec(clientState), path, channelEnd, proof))
+}
+
+function verifyPacketData(
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  sequence: uint64,
+  data: bytes) {
+    path = applyPrefix(prefix, "ports/{portIdentifier}/channels/{channelIdentifier}/packets/{sequence}")
+    codeHandle = clientState.codeHandle()
+    assert(codeHandle.isValidClientState(clientState, height))
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that the provided commitment has been stored
+    assert(root.verifyMembership(codeHandle.getProofSpec(clientState), path, hash(data), proof))
+}
+
+function verifyPacketAcknowledgement(
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  sequence: uint64,
+  acknowledgement: bytes) {
+    path = applyPrefix(prefix, "ports/{portIdentifier}/channels/{channelIdentifier}/acknowledgements/{sequence}")
+    codeHandle = clientState.codeHandle()
+    assert(codeHandle.isValidClientState(clientState, height))
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that the provided acknowledgement has been stored
+    assert(root.verifyMembership(codeHandle.getProofSpec(clientState), path, hash(acknowledgement), proof))
+}
+
+function verifyPacketReceiptAbsence(
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  sequence: uint64) {
+    path = applyPrefix(prefix, "ports/{portIdentifier}/channels/{channelIdentifier}/receipts/{sequence}")
+    codeHandle = clientState.codeHandle()
+    assert(codeHandle.isValidClientState(clientState, height))
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that no acknowledgement has been stored
+    assert(root.verifyNonMembership(codeHandle.getProofSpec(clientState), path, proof))
+}
+
+function verifyNextSequenceRecv(
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  nextSequenceRecv: uint64) {
+    path = applyPrefix(prefix, "ports/{portIdentifier}/channels/{channelIdentifier}/nextSequenceRecv")
+    codeHandle = clientState.codeHandle()
+    assert(codeHandle.isValidClientState(clientState, height))
+    // fetch the previously verified commitment root & verify membership
+    root = get("clients/{identifier}/consensusStates/{height}")
+    // verify that the nextSequenceRecv is as claimed
+    assert(root.verifyMembership(codeHandle.getProofSpec(clientState), path, nextSequenceRecv, proof))
+}
+```
+
+### WASM Client Code Interface
+
+#### What is code handle?
+Code handle is an object that facilitates interaction between WASM code and go code. For example, consider the method `isValidClientState` which could be implemented like this:
+
+```go
+func (c *CodeHandle) isValidClientState(clientState ClientState, height u64) {
+    clientStateData := json.Serialize(clientState)
+    packedData := pack(clientStateData, height)
+    // VM specific code to call WASM contract
+}
+```
+
+#### WASM Client interface
+Every WASM client code need to support ingestion of below messages in order to be used as light client.
+
+```rust
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct MisbehaviourMessage {
+    pub client_state: Vec<byte>,
+    pub consensus_state: Vec<byte>,
+    pub height: u64,
+    pub header1: Vec<byte>,
+    pub header2: Vec<byte>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct CreateConsensusMessage {
+    pub client_state: Vec<byte>,
+    pub epoch: u64,
+    pub height: u64
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct InitializeClientStateMessage {
+    pub initialization_data: Vec<byte>,
+    pub consensus_state: Vec<byte>
+}
+
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum HandleMsg {
+    HandleMisbehaviour(MisbehaviourMessage),
+    TryCreateConsensusState(CreateConsensusMessage),
+    InitializeClientState(InitializeClientStateMessage)
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct ValidateClientStateMessage {
+    client_state: Vec<byte>,
+    height: u64
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct ValidateNewClientStateMessage {
+    client_state: Vec<byte>,
+    new_client_state: Vec<byte>,
+    height: u64
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct ValidateInitializationDataMessage {
+    init_data: Vec<byte>,
+    consensus_state: Vec<byte>
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ValidityPredicate {
+    ClientState(ValidateClientStateMessage),
+    NewClientState(ValidateNewClientStateMessage),
+    InitializationData(ValidateInitializationDataMessage),
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum QueryMsg {
+   IsValid(ValidityPredicate),
+   LatestClientHeight(Vec<byte>),
+   ProofSpec(Vec<byte>)
+}
+
+```
+
+### Properties & Invariants
+
+Correctness guarantees as provided by the underlying algorithm implemented by `WASM Client Code`.
+
+## Backwards Compatibility
+
+Not applicable.
+
+## Forwards Compatibility
+
+As long as `WASM Client Code` keeps interface consistent with `ICS 02` it should be forward compatible
+
+## Example Implementation
+
+None yet.
+
+## Other Implementations
+
+None at present.
+
+## History
+
+
+## Copyright
+
+All content herein is licensed under [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0).


spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+kind: instantiation
+implements: 2
+author: Parth Desai <parth@chorus.one>
+created: 2020-10-13
+modified: 2020-10-13
+---
+
+## Synopsis
+
+This specification document describes an interface to a client (verification algorithm) stored as a wasm bytecode for a blockchain.
```
cwgoes, line 15: Should "wasm" be capitalised?

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+
+### Motivation
+
+WASM based light clients are decoupled from SDK source code, which enables one to upgrade existing light client or add support for other blockchain without modifying SDK source code.
```
cwgoes, line 19:
```suggestion
WASM-based light clients are decoupled from SDK source code, which enables one to upgrade an existing light client or add support for other blockchain without modifying SDK source code.
```

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+
+### Definitions
+
+Functions & terms are as defined in [ICS 2](../ics-002-client-semantics).
+
+`currentTimestamp` is as defined in [ICS 24](../ics-024-host-requirements).
+
+`WASM Client Code` refers to wasm bytecode stored in the client store, which provides target blockchain specific implementation of [ICS 2](../ics-002-client-semantics).
```
cwgoes, line 27:
```suggestion
`WASM Client Code` refers to WASM bytecode stored in the client store, which provides a target blockchain specific implementation of [ICS 2](../ics-002-client-semantics).
```

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+`WASM VM` refers to a virtual machine capable of executing valid wasm bytecode.
+
+
+### Desired Properties
+
+This specification must satisfy the client interface defined in ICS 2.
+
+## Technical Specification
+
+This specification depends on correctness of the `WASM Client code` in context of consensus algorithm of its target `blockchain`, as well as correct instantiation of `WASM Client`.
```
cwgoes, line 40: Does it? The chain hosting the client should not be exposed to any correctness risk, only users who use the client

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+
+### Client state
+
+The wasm client state tracks location of the wasm bytecode via `codeId`. Binary data represented by `data` field is opaque and only interpreted by the WASM Client Code. `type` represents client type.
```
cwgoes, line 44: please capitalise WASM consistently (or not, but pick one)

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+`type` and `codeId` both are immutable.
+
+```typescript
+interface ClientState {
```
cwgoes, line 48: why can't this whole type be up to the WASM client code

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+}
+```
+
+### Consensus state
+
+The WASM consensus state tracks the timestamp (block time), `WASM Client code` specific fields and commitment root for all previously verified consensus states.
+`type` and `codeId` both are immutable.
+
+```typescript
+interface ConsensusState {
```
cwgoes, line 64: ditto, same question, why can't this all be up to the WASM code?

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+}
+```
+
+### Misbehaviour
+
+The `Misbehaviour` type is used for detecting misbehaviour and freezing the client - to prevent further packet flow - if applicable.
+WASM client `Misbehaviour` consists of two headers at the same height both of which the light client would have considered valid.
+
+```typescript
+interface Misbehaviour {
```
cwgoes, line 117: why can't this be up to the WASM code

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+WASM client validity checking uses underlying WASM Client code. If the provided header is valid, the client state is updated & the newly verified commitment written to the store.
+
+```typescript
+function checkValidityAndUpdateState(
+  clientState: ClientState,
+  epoch: uint64,
+  header: Header) {
+    codeHandle = clientState.codeHandle()
+    isValid, consensusState = codeHandle.validateHeaderAndCreateConsensusState(clientState, header, epoch)
+    set("clients/{identifier}/consensusStates/{header.height}", consensusState)
```
cwgoes, line 161: can we give the WASM code limited state read/write access instead, it would be more flexible

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+
+```typescript
+function checkMisbehaviourAndUpdateState(
+  clientState: ClientState,
+  misbehaviour: Misbehaviour) {
+    codeHandle = clientState.codeHandle()
+    consensusState = get("clients/{identifier}/consensusStates/{misbehaviour.fromHeight}")
+    assert(codeHandle.handleMisbehaviour(clientState, consensusState, misbehaviour))
+    // save the client
+    set("clients/{identifier}", clientState)
```
cwgoes, line 179: ditto, can we give the WASM code limited state read/write access instead, it would be more flexible

cwgoes (changes requested):
Apologies for taking awhile to review this. Besides the nits, my substantive comment is that it would be better to allow the WASM code to read/write state and specify all the datatypes rather than partially specifying them here, I think.

spec/client/ics-028-wasm-client/README.md
ethanfrey: I agree completely. Since you are using wasmvm, it would not be hard to give each contract a sandboxed storage area, like we do in `x/wasm`. The contract knows what data it needs and can just store it locally.

However, it seems you actually need to write to some particular ibc keys here. Could you not just make a prefix store of `clients/{identifier}` and pass that to the contract. It could then store client state under the empty key and  consensus state under `/consensusStates/{header.height}`. Meaning full access to the state it manages, but no access to other ibc state

spec/client/ics-028-wasm-client/README.md
ethanfrey: I believe "Wasm" (not WASM or wasm) is [the preferred abbreviation](https://webassembly.org/)

cwgoes:
@ParthDesai Bump on this, any thoughts on my comments? Anything I should clarify?

spec/client/ics-028-wasm-client/README.md
ParthDesai: @cwgoes That was the initial idea. 

`ClientState`, `ConsensusState` and `Misbehaviour` interfaces  have some functions like `GetProofSpec`, `IsFrozen`, `Validate` which do not give access to store and/or are being called during tx creation on client side. In both cases we do not have access to the WASM VM. But, we still have to return valid answer as per semantics of that function and to do that we need to have some structure instead of opaque blob of byte only understood by WASM code.

spec/client/ics-028-wasm-client/README.md
ParthDesai: @cwgoes See my previous comment.

spec/client/ics-028-wasm-client/README.md
ParthDesai: @cwgoes See my previous comment.

spec/client/ics-028-wasm-client/README.md
ParthDesai: @ethanfrey @cwgoes I did not want WASM code read/write  access to storage, because current version of WASM manager does not allow validation of the WASM code thoroughly. So, to err on safe side, I removed storage access to the WASM code.

ParthDesai:
> @ParthDesai Bump on this, any thoughts on my comments? Anything I should clarify?

@cwgoes Replied. Please check and let me know.

spec/client/ics-028-wasm-client/README.md
ethanfrey: If wasm can return results that will be written to some specified storage locations, you could also provide the wasm code direct read/write access to only those storage locations.

I would assume wasm code is potentially malicious (even if this code must have been verified by governance to run), but you should be able to define a sandbox where it can work without impacting other modules (just it's own state). It is trusted to approve packets for a subset of clients/connections/channels, so should be able to write any data relative to those without increasing the attack surface at all

spec/client/ics-028-wasm-client/README.md
ethanfrey: This is a good point - if the spec defines data structures that must be validated without any access to storage of the VM, then these types must be understood by the Go code.

However, this is in the spec level and I wonder if "In both cases we do not have access to the WASM VM." is obvious in the spec or an implementation detail we could work to improve.

Could you link the functions that require inspection of these types, so they cannot be defined in Wasm. It might make sense to revisit those definitions with @cwgoes and see if we can make them more flexible.

spec/client/ics-028-wasm-client/README.md
cwgoes: Then let's use that consistently :)

spec/client/ics-028-wasm-client/README.md
cwgoes: Hmm, that's an excellent point, but I think we may want to alter the structure of IBC or whatever code is involved here so that the VM _can_ be accessed - sounds like we're on the same page @ethanfrey.

spec/client/ics-028-wasm-client/README.md
cwgoes: See my previous response

spec/client/ics-028-wasm-client/README.md
cwgoes: See my previous response

spec/client/ics-028-wasm-client/README.md
cwgoes: What do you mean, "does not allow validation of the WASM code thoroughly"? That sounds dangerous. I would be far more worried about compute issues than storage issues, since storage usage is gas-metered and safely sub-store-constrained by the keeper.

cwgoes:
> > @ParthDesai Bump on this, any thoughts on my comments? Anything I should clarify?
> 
> @cwgoes Replied. Please check and let me know.

Responded!

spec/client/ics-028-wasm-client/README.md
mkaczanowski: I think @ParthDesai is referring to [this piece of code](https://github.com/ChorusOne/ibc-go/blob/ics_28_wasm_client/modules/core/28-wasm/keeper/validation.go#L22)

Right now, Wasm manager validates by instantiating the Wasm VM (via CosmWasm lib). There is no code introspection (unless provided by CosmWasm)

ethanfrey: > Right now, Wasm manager validates by instantiating the Wasm VM (via CosmWasm lib). There is no code introspection (unless provided by CosmWasm)

Sure. But who can define such a wasm code as a light client? This is out of scope for ICS repo maybe, but I feel essential to understand how this integrates in a blockchain, and what security guarantees we need. 

Can I upload some wasm as `light-client-foo` and then make a new client/connection on top of it? Does governance need to approve it? If this code misbehaves what is the worst that can happen? (I assume fake packets in any channels built on top of this client).

spec/client/ics-028-wasm-client/README.md
mkaczanowski: These are methods without access to storage:

### Consensus State ([link](https://github.com/cosmos/ibc-go/blob/main/modules/core/exported/client.go#L188)):
```
ClientType() string
GetRoot() Root
GetTimestamp() uint64
ValidateBasic() error
```

spec/client/ics-028-wasm-client/README.md
mkaczanowski: These are the methods without access to storage
## Misbehaviour ([link](https://github.com/cosmos/ibc-go/blob/main/modules/core/exported/client.go#L201)):

```
ClientType() string
GetClientID() string
ValidateBasic() error
```

spec/client/ics-028-wasm-client/README.md
cwgoes: Maybe @colin-axner or @AdityaSripal can offer some more context here and/or thoughts on whether we should change the interface.

spec/client/ics-028-wasm-client/README.md
mkaczanowski: These are methods without access to the storage:

## Misbehaviour ([link](https://github.com/cosmos/ibc-go/blob/main/modules/core/exported/client.go#L201)):**
```
ClientType() string
GetClientID() string
ValidateBasic() error
```

This seems to be a stale definition:
```
interface Misbehaviour {
  fromHeight: Height  // doesn't seem to be used
  client_id: string, // missing, used by GetClientID()
  h1: Header 
  h2: Header
}
```

More correct version:
```
interface Misbehaviour {
  client_id: string,
  h1: Header 
  h2: Header
}
```

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,451 @@
...
+
+This is designed to allow the height to reset to `0` while the epoch number increases by one in order to preserve timeouts through zero-height upgrades.
+
+### Headers
+
+Contents of Wasm client headers depend upon `Wasm Client Code`.
+
+```typescript
+interface Header {
+  data: []byte
```
mkaczanowski, line 107: Methods that don't have access to storage

### Header ([link](https://github.com/cosmos/ibc-go/blob/main/modules/core/exported/client.go#L210)

```
ClientType() string
GetHeight() Height
ValidateBasic() error
```

This definition looks stale. It should be:
```
interface Header { 
   data: []byte,
   height: Height, // required by GetHeight() 
}
```

spec/client/ics-028-wasm-client/README.md
ethanfrey: Thanks for the detailed list of those methods.

`ProofSpecs` is static over the lifetime of the client and there may be somewhere to include this in code not in state.

`LatestHeight` obviously requires a query to the KV store. How does the current Go code handle returning this without storage access?

spec/client/ics-028-wasm-client/README.md
ethanfrey: With the two headers, we should be able to compare the headers for misbehavior without storage access, just code.

mkaczanowski: yes, that sounds right

spec/client/ics-028-wasm-client/README.md
AdityaSripal: Thanks, @ParthDesai and @mkaczanowski for the explanations. In principle, I'm ok with breaking any Go APIs that are not expected by counterparty chains or relayers if it will help make WASM and future client code more clean.

`ClientType() string`: Agreed that you should just be returning `10-wasm` here.

`Validate() error`: Since this is getting used in messages, it has to be stateless. Any additional checks you want to do with state must be done elsewhere. We could rename this to `ValidateBasic` to emphasize this.

`ZeroCustomFields`: This is just used for upgrades, we can provide store here if necessary.

---

### Interfaces that may be expected by counterparty chains

`GetProofSpecs`: The proof specs get used in the client `Verify` functions. This of course doesn't require an external interface, in face we just use the clientstate field directly. However, it also gets used on connection handshake to ensure that the counterparty client is using the correct proof specs: [here](https://github.com/cosmos/ibc-go/blob/2e95805e2d/modules/core/02-client/keeper/keeper.go#L301). Currently, it is not using the interface itself but rather getting the `ProofSpecs` field directly because we are making an assumption that all counterparty clients for an SDK chain are `07-tendermint` clients. However, once `WASM` client is introduced we need to remove that assumption since we could have a wasm smart contract client tracking the SDK chain (this could be better than 07-tendermint since it may have easier upgrade logic). So once that happens, we will have to rely on the interface itself. We cannot give you access to the store here, since this is getting called by the **counterparty** chain, which does not contain the store for your light client. You will need to find a way to provide this information in the client state struct itself.

`GetLatestHeight() Height`: Similar to the `ProofSpec` case, we check that the latest height of the counterparty client is not greater than our height during connection handshake [here](https://github.com/cosmos/ibc-go/blob/2e95805e2d/modules/core/02-client/keeper/keeper.go#L295). Again we currently use the field directly, but will have to switch to using an interface. 

We don't strictly require the interface, seeing as we could just execute separate `ValidateSelfClient` logic for wasm/go clients, however the wasm client must still have proof specs and latest height directly retrievable from its client state in order to perform all necessary checks on connection handshake.

AdityaSripal: > LatestHeight obviously requires a query to the KV store. How does the current Go code handle returning this without storage access?

We "cache" the latest height directly into the client state on update: https://github.com/cosmos/ibc-go/blob/2e95805e2d/modules/light-clients/07-tendermint/types/update.go#L251

spec/client/ics-028-wasm-client/README.md
AdityaSripal: `GetRoot() Root`: This is only getting used in internal client logic, we could remove it altogether if desired.

`GetTimestamp() uint64`: We need the timestamp of the consensus state to process packet timeouts. We can give you the client store here. The function you posted is merely a convenience, we could use the local time instead.

`ValidateBasic() error`: Again, this must be a stateless check. You can do very basic checks (e.g. codeId not empty), or even a no-op, and perform your stateful checks later on in message execution when you have access to the store.

spec/client/ics-028-wasm-client/README.md
AdityaSripal: ditto on ValidateBasic

I'm a bit hesitant to make `GetHeight` stateful here since it gets used in ValidateBasic checks and without it the ValidateBasic isn't doing much at all. If you could clarify why you need access to storage, and the difficulties with keeping it stateless that would be helpful.

spec/client/ics-028-wasm-client/README.md
AdityaSripal: > With the two headers, we should be able to compare the headers for misbehavior without storage access, just code.

Agreed.

spec/client/ics-028-wasm-client/README.md
AdityaSripal: > If wasm can return results that will be written to some specified storage locations, you could also provide the wasm code direct read/write access to only those storage locations.

Agreed. The WASM code handlers should just be given the client-id prefixed store. This is **already** what is given to the client handlers. Core IBC already prevents the ability for one client to read/write outside of its sandbox. So WASM can follow same security model. Just pass the store that the light client receives directly into WASM, it is already sandboxed by core IBC.

> But who can define such a wasm code as a light client? This is out of scope for ICS repo maybe, but I feel essential to understand how this integrates in a blockchain, and what security guarantees we need.

> Can I upload some wasm as light-client-foo and then make a new client/connection on top of it? Does governance need to approve it? If this code misbehaves what is the worst that can happen? (I assume fake packets in any channels built on top of this client).

We may rollout more conservatively, but I'm of the belief that the spec itself should enable random users uploading wasm light client code. This is a big selling point of WASM clients imo, since you don't need buy in from governance to support your eclectic choice of consensus algorithm.

Since IBC is designed to sandbox any issues, any malicious light client code submitted by user can only affect users of that light client. Though of course, they can do whatever they want to the users of their light client.

Here it will be incumbent on counterparty chains to enforce wasm contract correctness, by saying I will only accept a connection where the client has this smart contract code hash(es). This way we do not allow malicious clients to be submitted against correct chains. By creating the restriction on the destination chain rather than the source chain, we can enable all the flexibility of dynamic light clients while still preventing malicious connections over correctly implemented chains.

AdityaSripal (commented):
Thanks for pinging me on this. I'd like to confirm why storage is needed in these client interfaces from WASM.

From my understanding, this information would be in `data` but you would need to retrieve the smart contract code from WASM store in order to parse it and return the value. If you can't retrieve code, then you will have to store directly in client state so that go handlers can implement the interfaces. Is that correct? If not, please clarify why WASM client needs storage to implement these interfaces.

In general, I'm open to changing the interfaces for `ClientState` and `ConsensusState` since those are mostly created and stored directly on-chain, and the GO apis can change.

I'm less comfortable changing the `Header` and `Misbehaviour` interfaces, since they are very client facing and so having some stateless interfaces is useful here.

I also believe you should just be passing the client-prefixed store into the wasm handlers. I don't see any security risk there. It's already an assumption of core IBC that clients may do arbitrary things to their own client-prefixed store.

spec/client/ics-028-wasm-client/README.md
AdityaSripal: Though we should double check that `02-client` does not make any assumptions that something will be in the store in ways that will affect its overall functioning.

It's ok if the client becomes non-functional or downright malicious imo.

What would be bad, is if deleting something (like say the clientstate itself), causes the client handler to panic and shutdown the node. In case, the wasm contract does something like this, we should just return an error and fail th tx whenever client state is needed

spec/client/ics-028-wasm-client/README.md
mkaczanowski: Thanks, @AdityaSripal for such a detailed response

> however the wasm client must still have proof specs and latest height directly retrievable from its client state in order to perform all necessary checks on connection handshake.

We'll update the ICS with:
```
interface ClientState {
  codeId: []byte
  data: []byte
  
  latestHeight: Height
  proofSpecs: ProofSpec
}
```

> ZeroCustomFields: This is just used for upgrades, we can provide a store here if necessary.

Since the method intends to reset some fields in the ClientState, I don't think we need storage access. However, possibly changing the interface to allow returning the `error` would be helpful. Please see the initial implementation:
https://github.com/ChorusOne/ibc-go/blob/ics_28_wasm_client/modules/light-clients/10-wasm/types/client_state.go#L228

what do you think?

spec/client/ics-028-wasm-client/README.md
mkaczanowski: > GetRoot() Root: This is only getting used in internal client logic, we could remove it altogether if desired.

Yes, if `GetRoot` is removed from the interface, then there is no need for us to require `root: MerkleRoot` field anymore.

> GetTimestamp() uint64: We need the timestamp of the consensus state to process packet timeouts. We can give you the client store here. The function you posted is merely a convenience, we could use the local time instead.

Great, then we can modify the function `modules/apps/transfer/client/cli/tx.go` and the only place where `GetTimeout()` is called is [03-connection/keeper/keeper.go](https://github.com/cosmos/ibc-go/blob/2523fdbfd8fc14554d81dc92ce58b3a590964647/modules/core/03-connection/keeper/keeper.go#L99) and here where we have access to the VM.

Similar question to `ZeroCustomFields`... To pull the timestamp from the Wasm light client we would have to call the VM, which may fail. The current interface `GetTimestamp() uint64` doesn't allow to return error, so would it be alright to change the signature to `GetTimestamp() (uint64, error)`?

mkaczanowski: > GetRoot() Root: This is only getting used in internal client logic, we could remove it altogether if desired.

Yes, if `GetRoot` is removed from the interface, then there is no need for us to require `root: MerkleRoot` field anymore.

> GetTimestamp() uint64: We need the timestamp of the consensus state to process packet timeouts. We can give you the client store here. The function you posted is merely a convenience, we could use the local time instead.

Great, then we can modify the function `modules/apps/transfer/client/cli/tx.go` and the only place where `GetTimeout()` is called is [03-connection/keeper/keeper.go](https://github.com/cosmos/ibc-go/blob/2523fdbfd8fc14554d81dc92ce58b3a590964647/modules/core/03-connection/keeper/keeper.go#L99) and here where we have access to the VM. 

To pull the timestamp from the Wasm light client we would have to:
* call the VM, which may fail. The current interface `GetTimestamp() uint64` doesn't allow to return an error, so would it be alright to change the signature to `GetTimestamp() (uint64, error)`?
* `ConsensusState` doesn't hold `codeId` that is required to call the WasmVM. So far, it was only stored by `ClientState`, so here are the possible options:
  * change the signature `ConsensusState.GetTimestamp() uint64` -> `ConsensusState.GetTimestamp(cs ClientState) uint64` but that looks weird
  * add the `codeId` field to `ConsensusState`
  * other options?

spec/client/ics-028-wasm-client/README.md
mkaczanowski: @AdityaSripal sorry I phrased it wrongly. The storage is not needed here, but I wanted to give some context as to why the `height: Height` is a separate field (not a part of `data`), this is [during the UpdateClient call the `GetHeight()` is required](https://github.com/cosmos/ibc-go/blob/1ff0da6694d08e494a9c87d5b9737ab5312677a0/modules/core/02-client/keeper/client.go#L81) to set the consensus state.

That's the only significant use case of `GetHeight` I found

mkaczanowski:
@AdityaSripal, thanks a lot for the feedback!

> From my understanding, this information would be in data, but you would need to retrieve the smart contract code from WASM store to parse it and return the value.

Once the Wasm light client binary (aka. `data blob` or `data []byte`) is uploaded, the `CosmWasm` library gives us the `codeId` we use to reference the code.
So, whenever we need to execute a call, for instance `CheckHeaderAndUpdateState`, we would have to specify the `codeId`, say:
```
// abstract syntax
WasmVM.Execute(codeID, "CheckHeaderAndUpdateState", args...)
```
And pass all required arguments to the Wasm method itself. Right now, the `codeId` is only present within the `ClientState` structure.

>  If you can't retrieve code, then you will have to store directly in the client state so that go handlers can implement the interfaces. Is that correct?

Yes. If the code can't be retrieved (i.e. client code without VM access, no `codeId` available). Then we would store directly in the relevant structure like `ClientState` to satisfy the interface. To parse the `data []byte`, we need to execute Wasm VM light client method that would "understand the blob of data" and return the Go understandable struct (ie. Height, ConsesnsusState, ClientState, Status etc).

> If not, please clarify why WASM client needs storage to implement these interfaces.

As I am reading my comments, I noticed the "storage" term is a bit unclear. Or at least I made it somewhat confusing, sorry. I referred to "needs storage" as "needs access to Wasm code identified by `codeId`".

So for example, `ClientState.ZeroCustomFields`:
1. must have access to both the Wasm VM and `codeId` (present in ClientState struct) to execute light client native method
2. doesn't require `KV-store` to be present as an argument because the methods consume `ClientState/self` and return updated `ClientState`. The Wasm light client code method is idempotent in this case.

Hopefully, that clarifies it a bit.

spec/client/ics-028-wasm-client/README.md
mkaczanowski: > Just pass the store that the light client receives directly into WASM, it is already sandboxed by core IBC.

In the [initial implementation](https://github.com/cosmos/ibc-go/pull/208) we already do that, this is, use the default flow where `02-client/keeper/client.go` gets client-id prefixed store from [ClientStore()](https://github.com/cosmos/ibc-go/blob/2523fdbfd8fc14554d81dc92ce58b3a590964647/modules/core/02-client/keeper/keeper.go#L386), and passes it on, to the client handle.

@AdityaSripal @cwgoes I am not sure how to phrase it (could use your help here), but here's a try:
```
    store = getStore("clients/{identifier}")
    isValid, consensusState = codeHandle.validateHeaderAndCreateConsensusState(store, clientState, header, epoch)
    store.set("/consensusStates/{header.height}", consensusState)
    // save the client
    store.set("", clientState)
```

> We may rollout more conservatively, but I'm of the belief that the spec itself should enable random users uploading wasm light client code. This is a big selling point of WASM clients imo, since you don't need buy in from governance to support your eclectic choice of consensus algorithm.

Agreed.

spec/client/ics-028-wasm-client/README.md
mkaczanowski: "WASM" has been replaced with "Wasm" across the ICS. I think we can resolve this thread :)

spec/client/ics-028-wasm-client/README.md
mkaczanowski: It doesn't.

I've rephrased this:
```
This specification depends on the correct instantiation of the `Wasm client` and is decoupled from any specific implementation of the target `blockchain` consensus algorithm.
```

@cwgoes ^^ does that look good to you?

spec/client/ics-028-wasm-client/README.md
mkaczanowski: updated the `ClientState` interface [here](https://github.com/ParthDesai/ibc/commit/d7c33fc5c01ff4c07000f89ff9da04c31c27ed23#diff-081ac6bcbed0e39a406e9c83aa79a058605a646a4047d5be67cb84dd9b83877bL48)

spec/client/ics-028-wasm-client/README.md
mkaczanowski: updated `ConsensusState` [here](https://github.com/ParthDesai/ibc/commit/d7c33fc5c01ff4c07000f89ff9da04c31c27ed23#diff-081ac6bcbed0e39a406e9c83aa79a058605a646a4047d5be67cb84dd9b83877bR62)

spec/client/ics-028-wasm-client/README.md
mkaczanowski: updated `Misbehaviour` [here](https://github.com/ParthDesai/ibc/commit/d7c33fc5c01ff4c07000f89ff9da04c31c27ed23#diff-081ac6bcbed0e39a406e9c83aa79a058605a646a4047d5be67cb84dd9b83877bL117)

spec/client/ics-028-wasm-client/README.md
mkaczanowski: updated `Header` [here](https://github.com/ParthDesai/ibc/commit/d7c33fc5c01ff4c07000f89ff9da04c31c27ed23#diff-081ac6bcbed0e39a406e9c83aa79a058605a646a4047d5be67cb84dd9b83877bL106)

mkaczanowski:
@cwgoes @AdityaSripal could you guys take another round on the comments + changes I've made? Thanks!

AdityaSripal:
> Once the Wasm light client binary (aka. data blob or data []byte) is uploaded

Is this the same as the `data []byte` you have in the ClientState. It seems like it isn't? 

From the rest of what you say, the `data` in client state is an implementation-specific marshalling of client data that can be interpreted by WASM

spec/client/ics-028-wasm-client/README.md
```diff
@@ -0,0 +1,448 @@
...
+
+Wasm based light clients are decoupled from SDK source code, which enables one to upgrade existing light client or add support for other blockchain without modifying SDK source code.
+
+### Definitions
+
+Functions & terms are as defined in [ICS 2](../ics-002-client-semantics).
+
+`currentTimestamp` is as defined in [ICS 24](../ics-024-host-requirements).
+
+`Wasm Client Code` refers to Wasm bytecode stored in the client store, which provides a target blockchain specific implementation of [ICS 2](../ics-002-client-semantics).
```
AdityaSripal, line 27: Is the wasm client code stored in the clientID-prefixed store provided by IBC or it it stored in a KV store owned by WASM?

AdityaSripal:
The updates look good to me 👍

spec/client/ics-028-wasm-client/README.md
colin-axner: Sorry to interject without reading through all the context, but I think `GetHeight()` will be removed soonish. In a couple months I'd like to start work on #569 which would move setting of IBC client information to light client implementations. This would mean the consensus state would be set by the light client, which could directly cast to its Header definition as opposed to using the header interface. The header interface would only be really used to pass the header from the user message to 02-client and finally to the IBC client (which would not require `GetHeight()` usage

I'm not really sure what this means for the decision being discussed, but whatever is decided should be fine since it'll likely be short lived. It sounds like `GetHeight` will still be stateless in the short term? The `height` field would just be removed from data once `GetHeight` is removed

cwgoes:
I'm a little confused, do we not want the WASM code to be able to write state itself? I don't see changes to that effect.

spec/client/ics-028-wasm-client/README.md
mkaczanowski: the Wasm client code is stored in the KV store owned by 28-wasm module.

Example entry:
```
ibc/code_id/d6537444f6df8686bdfefc3dc42441cd1d332ac608d27ef5903dd083a537e85f
```

Source: https://github.com/ChorusOne/ibc-go/blob/ics_28_wasm_client/modules/core/28-wasm/keeper/keeper.go#L61

mkaczanowski:
Thanks a lot for the comments! From what I can tell the last major thing to tackle in this ICS, is the question of whether Wasm code should be able to update its own client/consensus state, as [pointed out by](https://github.com/cosmos/ibc/pull/571#issuecomment-876764022) @cwgoes.

As @colin-axner [explained](https://github.com/cosmos/ibc/pull/571#discussion_r666225865) that upcoming changes are to give light clients such flexibility, so I need to update this ICS.

I am not so fluent with pseudo-language, so here is what I came up with:
https://gist.github.com/mkaczanowski/06c43592a66453e13871d419870e66eb

@cwgoes @AdityaSripal can you please tell me if ^^ diff is alright?

spec/client/ics-028-wasm-client/README.md
mkaczanowski: @cwgoes bumping the question :)

spec/client/ics-028-wasm-client/README.md
mkaczanowski: To let the light-clients update their state, IBC interface would have to change slightly (it's going to be covered by https://github.com/cosmos/ibc/issues/569).

For example:
```
CheckHeaderAndUpdateState(sdk.Context, codec.BinaryCodec, sdk.KVStore, Header) (ClientState, ConsensusState, error)
```

to smth like this:
```
CheckHeaderAndUpdateState(sdk.Context, codec.BinaryCodec, sdk.KVStore, Header) error
```

because the {client,consensus} states are going to be updated within `CheckHeaderAndUpdateState` lightclient method

colin-axner:
The diffs align with what I had in mind for the future implementation changes :+1:

cwgoes:
Yes, the store should be passed to the WASM code.

mkaczanowski:
@cwgoes @colin-axner @AdityaSripal I applied the remaining changes, plus renamed the ICS-28 to ICS-08 to match current ordering. Please let me know if you see anything else that requires changes, thanks!

mkaczanowski:
@cwgoes @colin-axner @AdityaSripal friendly ping. It would be great if you could go through this ICS one more time :)

cwgoes (commented):
I'm basically fine with this, just checking - is the choice of having Go Merkle path validation (sub-state proofs) instead of letting WASM code do that intentional? It will constrain what WASM light clients can do, though it will probably be more efficient unless there are precompiles for hash functions or the like.

mkaczanowski:
@cwgoes at first that's what we did (proof verification on the Go side), but as you said it's quite limiting.

>  is the choice of having Go Merkle path validation (sub-state proofs) instead of letting WASM code do that intentional

it's not intentional, and thanks for catching this.

### Proposed changeset (please review)
Ideally, the verify methods are just a pass-through to WASM code ([example VerifyConnectionState method in the initial impl](https://github.com/cosmos/ibc-go/pull/208/files#diff-81beb171d0953178a5eb2401cb21c6702420bbd4d932664c0bb1cb796af2ad23R415))

So, here is the proposed changeset:
https://github.com/mkaczanowski/ibc/commit/5750485b64fb4d7f21a3c30a5e8fd94e9abf71c3

please let me know if ^^ this looks fine to you

spec/client/ics-008-wasm-client/README.md
```diff
@@ -0,0 +1,447 @@
...
+}
+```
+
+### Height
+
+The height of a Wasm light client instance consists of two `uint64`s: the epoch number, and the height in the epoch.
+
+```typescript
+interface Height {
+  epochNumber: uint64
```
colin-axner, line 75: This should be revision number and revision height?

spec/client/ics-008-wasm-client/README.md
```diff
@@ -0,0 +1,447 @@
...
+interface Header {
+  data: []byte
+  height: Height
+}
+```
+
+### Misbehaviour
+
+The `Misbehaviour` type is used for detecting misbehaviour and freezing the client - to prevent further packet flow - if applicable.
+Wasm client `Misbehaviour` consists of two headers at the same height both of which the light client would have considered valid.
```
colin-axner, line 111: Wasm clients may have other instances of misbehaviour other than two headers at the same height (time not monotonically increasing)? I think it is sufficient to say at minimum two headers are needed to prove misbehaviour?

colin-axner:
> So, here is the proposed changeset:
> [mkaczanowski@5750485](https://github.com/mkaczanowski/ibc/commit/5750485b64fb4d7f21a3c30a5e8fd94e9abf71c3)

These changes make a lot more sense to me :+1: This means the [wasm client interface](https://github.com/cosmos/ibc/pull/571/files#diff-bef06cb995c3cec7e766687c5bc2028a17f35d0db7ca78a41d3796245fa4319aR344) needs to be expanded as well?

spec/client/ics-008-wasm-client/README.md
mkaczanowski: yes, will update

spec/client/ics-008-wasm-client/README.md
mkaczanowski: Fair point, this statement `at minimum two headers` makes sense to me. However, if we say "at a minimum" then would the `Misbehaviour` interface have to change? (to allow passing more than two headers?)

colin-axner: Good point, lets just go with `Wasm client `Misbehaviour` consists of two conflicting headers both of which the light client would have considered valid.`? It's a little vague, but I don't think it is the responsibility of this spec to define what misbehaviour is exactly

mkaczanowski: :+1:  this works for me, thanks

mkaczanowski:
> > So, here is the proposed changeset:
> > [mkaczanowski@5750485](https://github.com/mkaczanowski/ibc/commit/5750485b64fb4d7f21a3c30a5e8fd94e9abf71c3)
> 
> These changes make a lot more sense to me +1 This means the [wasm client interface](https://github.com/cosmos/ibc/pull/571/files#diff-bef06cb995c3cec7e766687c5bc2028a17f35d0db7ca78a41d3796245fa4319aR344) needs to be expanded as well?

I am looking at this

spec/client/ics-008-wasm-client/README.md
```diff
@@ -0,0 +1,409 @@
...
+```
+
+### Validity predicate
+
+Wasm client validity checking uses underlying Wasm Client code. If the provided header is valid, the client state is updated & the newly verified commitment written to the store.
+
+```typescript
+function checkValidityAndUpdateState(
+  clientState: ClientState,
+  revision: uint64,
```
AdityaSripal, line 155: revison should be derivable from `header.Height` correct?

AdityaSripal, line 155:
```suggestion
```

spec/client/ics-008-wasm-client/README.md
```diff
@@ -0,0 +1,409 @@
...
+  header: Header) {
+    store = getStore("clients/{identifier}")
+    codeHandle = clientState.codeHandle()
+
+    // verify that provided header is valid and state is saved
+    assert(codeHandle.validateHeaderAndCreateConsensusState(store, clientState, header, revision))
```
AdityaSripal, line 161:
```suggestion
    assert(codeHandle.validateHeaderAndCreateConsensusState(store, clientState, header))
```

spec/client/ics-008-wasm-client/README.md
```diff
@@ -0,0 +1,409 @@
...
+function verifyClientConsensusState(
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  clientIdentifier: Identifier,
+  consensusStateHeight: Height,
+  consensusState: ConsensusState) {
+    codeHandle = getCodeHandleFromClientID(clientIdentifier)
+    assert(codeHandle.isValidClientState(clientState, height, prefix, clientIdentifier, proof, consensusStateHeight, consensusState))
```
AdityaSripal, line 216:
```suggestion
    assert(codeHandle.verifyClientConsensusState(clientState, height, prefix, clientIdentifier, proof, consensusStateHeight, consensusState))
```

spec/client/ics-008-wasm-client/README.md
```diff
@@ -0,0 +1,409 @@
...
+  clientState: ClientState,
+  height: Height,
+  prefix: CommitmentPrefix,
+  proof: CommitmentProof,
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  sequence: uint64,
+  data: bytes) {
+    codeHandle = clientState.codeHandle()
+    assert(codeHandle.verifyPacketData(clientState, height, prefix, proof, portportIdentifier, channelIdentifier, sequence, data))
```
AdityaSripal, line 252:
```suggestion
function verifyPacketCommitment(
  clientState: ClientState,
  height: Height,
  prefix: CommitmentPrefix,
  proof: CommitmentProof,
  portIdentifier: Identifier,
  channelIdentifier: Identifier,
  sequence: uint64,
  commitment: bytes) {
    codeHandle = clientState.codeHandle()
    assert(codeHandle.verifyPacketCommitment(clientState, height, prefix, proof, portIdentifier, channelIdentifier, sequence, commitment))
```

spec/client/ics-008-wasm-client/README.md
```diff
@@ -0,0 +1,409 @@
...
+#### Wasm Client interface
+Every Wasm client code need to support ingestion of below messages in order to be used as light client.
+
+```rust
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct MisbehaviourMessage {
+    pub client_state: Vec<byte>,
+    pub consensus_state: Vec<byte>,
+    pub height: u64,
```
AdityaSripal, line 315: These heights should be of type `Height` correct? ditto for all other heights

AdityaSripal (changes requested):
Great work!! Small nits to address before this is ready to merge

AdityaSripal:
Thanks @mkaczanowski , pending @cwgoes ack and then will merge

AdityaSripal:
Note: We will remove `GetProofSpecs` from the interface as well.

Since `ValidateSelfClient` in the connection handshake must cast to specific client type(s) to perform all its checks, we can remove `GetProofSpecs` from the interface and check that each individual client type is doing the proofs correctly.

ref: https://github.com/confio/ics23/pull/12#issuecomment-958945297
cc: @colin-axner

PR #576, opened 2021-05-24
ICS20-1 cleanup

ethanfrey:
When preparing to add ICS20-2 documentation, I realized the spec deviated from the implementation, and a few items were underspecified.

In particular:
* There is no `source` field in the packet, but important info is encoded in `denomination`, which is not clearly specified in human words (just pseudocode).
* The acknowledgement type was incorrect (it is enum with strings, not always `success: boolean`)
* It is not defined that the packet data and acknowledgement are JSON-encoded (at least one independent implementation used Protobuf encoding until they debugged compatibility issues)
* The prefix in the pseudo-code is missing a trailing `/` that is present in the implementation (and correct in the implementation I believe)

I added these suggests on a "best effort" basis and if the core devs disagree with any of the description, I would ask them to kindly correct this PR. In any case, this PR should be "more correct" than the current master.


spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -45,15 +45,25 @@ interface FungibleTokenPacketData {
 }
 ```
 
+Whether the sending chain is the original source of funds, whether it is redeeming funds received over this channel, or whether it is relaying funds received from a 3rd chain is encoded in the `denomination` field. If the sending chain is the original source of funds, `denomination` must not contain the character `/`. If it contains `/`, then `denomination` must be of the form `{ics20Port}/{ics20Channel}/{remote_denom}`, where `ics20Port` and `ics20Channel` are the ics20 port and channel on the remote chain where the funds originated. If `ics20Port == packet.sourcePort` and `ics20Channel == packet.sourceChannel`, then the remote blockchain is returning/redeeming funds originally sent from this blockchain over this connection.
```
colin-axner, line 48: Well written and thanks for cleaning up the spec. 

The only thing that confuses me is the usage of remote. There seems to be a reference to a send from a remote chain to this blockchain. The `remote_denom` confuses me since it sounds like a denom which originated on the remote chain, but in reality it is likely from this blockchain (or some other blockchain). `remote_denom` would only ever be on the remote chain if sent via localhost or the chain incorrectly mints fake IBC tokens

Does that make sense? Maybe I'm misinterpreting the meaning?

ethanfrey: You are right, this should be clarified. I think `{ics20Port}/{ics20Channel}/{denom}` should be used.

Note that there are two cases when this may be set.

1. I send `foo` from A -> B on channel `channel-7`. When B returns this token to me, it will declare it as `transfer/channel-7/foo` (and I will correctly map it to the one I sent before.
2. I send `bar` from C -> B on `channel-3`. Then I send this token from B -> A on `channel-7`. (I believe) this token will be sent as `transfer/channel-3/bar`

I am not sure how to express the second case and it doesn't seem to be explained clearly in the existing spec. I would be happy if you could extend this (maybe 2 paragraphs?) to ensure both (1) and (2) are covered, as well as the basic "no `/` in denom" case.

colin-axner: Yea, I'd be happy to write up something tomorrow. I find this area of ics20 always confusing, so I'll do my best to provide a concise and clear explanation

ethanfrey: Even more confusing is:

1. Send `bar` from C -> B
2. Send `transfer/channel-3/bar` (?) from B -> A
3. Return this token A -> B as `transfer/channel-7/transfer/channel-3/bar` ??? which then unescrows `transfer/channel-3/bar` on B
4. Return this token B -> C as `transfer/channel-3/bar` which then unescrows `bar` on C.

This is my guess, I wouldn't know if this is correct until I run a relayer in verbose mode with this setup.

In general multi-hop and stripping off prefixes would be nice to clearly define (maybe make a PR on top of this one if it become a larger issue).

AdityaSripal (approved):
ACK, pending @colin-axner 's clarification.

Thanks for the fix @ethanfrey

spec/app/ics-020-fungible-token-transfer/README.md
ethanfrey: Actually, this will not keep extending like I described above, but hash the value in each hop, right?
I am getting confused just thinking about this flow.

colin-axner: checkout the my [suggestion](https://github.com/confio/ibc/pull/1). Happy to swap some of my wording for other terms if it would provide more clarity

colin-axner: The general format I went for:
- define that tokens some sort of record/history
- define how this is represented
- define how this is updated

ethanfrey: Nice addition. I merged your changes.

I have a question about denom hashing: https://github.com/confio/ibc/pull/1#pullrequestreview-667707214

Maybe adding a line or two to clarify this (even saying it is allowed but optional in the implementations and unspecified by the spec)

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -45,15 +45,29 @@ interface FungibleTokenPacketData {
 }
 ```
 
+As tokens are sent across chains using the ICS 20 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `denomination` field. 
+
+The ics20 token denominations are represented the form `{ics20Port}/{ics20Channel}/{denom}`, where `ics20Port` and `ics20Channel` are an ics20 port and channel on the current chain for which the funds exist. The prefixed port and channel pair indicate which channel the funds were previously sent through. If `{denom}` contains `/`, then it must also be in the ics20 form which indicates that this token has a multi-hop record.
```
cwgoes, line 50:
```suggestion
The ics20 token denominations are represented the form `{ics20Port}/{ics20Channel}/{denom}`, where `ics20Port` and `ics20Channel` are an ics20 port and channel on the current chain for which the funds exist. The prefixed port and channel pair indicate which channel the funds were previously sent through. If `{denom}` contains `/`, then it must also be in the ics20 form which indicates that this token has a multi-hop record - note that this requires that the `/` (slash character) is prohibited in denomination names.
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -45,15 +45,29 @@ interface FungibleTokenPacketData {
...
-  success: boolean
-  error: Maybe<string>
+type FungibleTokenPacketAcknowledgement = FungibleTokenPacketSuccess | FungibleTokenPacketError;
+
+interface FungibleTokenPacketSuccess {
+  // This is binary 0x01 base64 encoded
+  success: "AQ=="
+}
+
+interface FungibleTokenPacketSuccess {
```
cwgoes, line 64: Presumably you intended to name this something else? `FungibleTokenPacketError` ?

cwgoes (approved):
ACK but see comments - thanks for the improvements!

spec/app/ics-020-fungible-token-transfer/README.md
colin-axner: `prohibited in non-IBC token denomination names`

or something like that. We should be explicit when referring to non-IBC tokens since `denomination` encapsulates both IBC and non-IBC tokens

spec/app/ics-020-fungible-token-transfer/README.md
ethanfrey: Yup, copy + paste error. Thanks for catching it.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -45,15 +45,29 @@ interface FungibleTokenPacketData {
...
+
+The ics20 token denominations are represented the form `{ics20Port}/{ics20Channel}/{denom}`, where `ics20Port` and `ics20Channel` are an ics20 port and channel on the current chain for which the funds exist. The prefixed port and channel pair indicate which channel the funds were previously sent through. If `{denom}` contains `/`, then it must also be in the ics20 form which indicates that this token has a multi-hop record. 
+
+A sending chain may be acting as a source or sink zone. When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone. When tokens are sent from a source zone, the destination port and channel will be prefixed onto the denomination adding another hop to a tokens record. When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone. When tokens are sent from a sink zone, the last prefixed port and channel pair on the denomination is removed, undoing the last hop in the tokens record.
```
colin-axner, line 52:
```suggestion
A sending chain may be acting as a source or sink zone. When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone. When tokens are sent from a source zone, the destination port and channel will be prefixed onto the denomination (once the tokens are received) adding another hop to a tokens record. When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone. When tokens are sent from a sink zone, the last prefixed port and channel pair on the denomination is removed (once the tokens are received), undoing the last hop in the tokens record. 

Note, hashing the full denomination is characterstic of implementations, such as [ibc-go](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md#decision). Hasing the full denomination is not a requirement of ics20, but may be useful for internal representation of a token denomination.
```

spec/app/ics-020-fungible-token-transfer/README.md
colin-axner: Do you think [this](https://github.com/cosmos/ibc/pull/576#discussion_r638822798) clarifies those questions

spec/app/ics-020-fungible-token-transfer/README.md
ethanfrey: Fixed in 6ea8f4c

spec/app/ics-020-fungible-token-transfer/README.md
ethanfrey: Fixed in 6ea8f4c (both changes including colin's)

spec/app/ics-020-fungible-token-transfer/README.md
ethanfrey: Fixed in fff202f

cwgoes:
@milosevic Want to take a quick look at this or shall we merge (already 2/3 approvals)?

spec/app/ics-020-fungible-token-transfer/README.md
ethanfrey: Yeah, those changes should resolve this

milosevic (approved):
Really nice work.

cwgoes:
Merged. Thank you @ethanfrey!

Issue #578, opened 2021-05-26
ICS29: General Relayer Fee Protocol


AdityaSripal:
Fees aren't baked into IBC protocol because IBC does not have the concept of fungible tokens in-built to the protocol. This is an important and correct decision to make IBC as general as possible. IBC should still work with ledgers that have no tokens, and in contexts where a relayer fee isn't necessary.

However, we should still have a general approach to incentivizing application packets getting relayed. This issue is a place to collect discussion on possible approaches.

AdityaSripal:
I have a rough sketch for a general idea that I want to pitch here before writing it up into a broader issue/spec. It builds on the sender-processing logic I have above.

I propose we still keep the notion of tokens away from core IBC, and implement a separate module called `ibc-fee` that will handle incentivizing relayers.

This `ibc-fee` module can implement a separate message:

```go
type IBCFeeMsg struct {
     string channelID
     string portID
     uint64 sequence
     receiveFee sdk.Coin
     ackFee sdk.Coin
     timeoutFee sdk.Coin
}
```

This message uniquely identifies the packet to be incentivized and specifies the fees to be paid for each potential action (receive/ack/timeout).

Any IBC application can now "opt-in" to support incentivization by simply calling a `PayFee` callback in `OnAckPacket` and `OnTimeoutPacket`.

```go
func OnAcknowledgeent() {
    // ...
    // get receive relayer from acknowledgement. This only works if counterparty app supports incentivization
    // if empty, refund receiveFee
    // get ack relayer from message signer
    ibcfeeKeeper.PayFee(channelID, portID, sequence, receive_relayer, ack_relayer)
}
```

```go
func OnTimeoutPacket() {
    // ... 
    ibcfeekeeper.PayTimeoutFee(channelID, portID, sequence, timeout_relayer)
}
```

The `ibc-fee` Keeper is in charge of paying out the relevant relayers and refunding any unused fees. This will allow any ibc app to be incentivized so long as the sending chain has some notion of fungible tokens, without introducing the concept into core IBC. Blockchains that have no concept of fungible tokens will need some other method to incentivize relayers.

There will need to be some channel registration so fees aren't sent to channels that do not support incentivization.

[...]

PR #579, opened 2021-05-31
Update ICS-4 to pass relayer/signed to modules

ethanfrey:
This changes the ibc handler code, but no changes to the packets send over the wire, so non-breaking on the protocol layer.

It provides some needed information to allow fee incentivization for IBC packets, both in the ics20-2 proposal #577 as well as the more general relayer fee proposal #578

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -591,11 +591,14 @@ The IBC handler performs the following steps in order:
 - Sets a store path to indicate that the packet has been received (unordered channels only)
 - Increments the packet receive sequence associated with the channel end (ordered channels only)
 
+We pass the address of the `relayer` that signed and submitted the packet to enable a module to optionally provide some rewards. This provides a foundation for fee payment, but can be used for other techniques as well (like calculating a leaderboard).
+
 ```typescript
 function recvPacket(
   packet: OpaquePacket,
   proof: CommitmentProof,
-  proofHeight: Height): Packet {
+  proofHeight: Height,
+  relayer: string): Packet {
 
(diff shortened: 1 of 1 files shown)


cwgoes (approved):
Concept ACK, this seems straightforward.

Just a quick ping @colin-axner - any difficulties here from the `ibc-go` angle (presumably not)?

colin-axner:
> Just a quick ping @colin-axner - any difficulties here from the `ibc-go` angle (presumably not)?

Nope, @ethanfrey has already kindly implemented [the changes](https://github.com/cosmos/ibc-go/pull/206)

TimeoutOnClose needs to be updated in this pr though

ethanfrey:
> TimeoutOnClose needs to be updated in this pr though

There was also the idea to expose this info for the various channel lifetimes and such, but I would request that in another PR as I see it as a different but related PR.

`TimeoutOnClose` seems to be more packet-related however and I will add that

AdityaSripal (approved):
concept ACK

Will this PR include wrapping the relayer address in the ack or will this be in a future PR?

ethanfrey:
> Will this PR include wrapping the relayer address in the ack or will this be in a future PR?

That is a future and more contraversial point. I am trying to do step by step where I understand there to be consensus already

ethanfrey:
What do I need to get this merged? It has 2/3 approvals and support on the IBC call.

I am happy for more review, but I would like to see a bit faster feedback cycle if there are no objections.
That would let me to build a series of small, easy to review PRs on top of each other.
Otherwise, you just encourage mega-PRs.

ethanfrey:
@milosevic I guess this just needs your review

milosevic (approved):

PR #580, opened 2021-06-01
Start defining general relayer fee payment protocol

ethanfrey:
Based on the discussion in #578 

this is an attempt to condense these ideas into a workable form. Please comment or contribute


ethanfrey:
Moved to #581 to allow better collaboration

PR #581, opened 2021-06-02
General Fee Payment Spec

AdityaSripal:
Supersedes #580


emadami (approved):

emadami:
25800

در تاریخ چهارشنبه ۲ ژوئن ۲۰۲۱،‏ ۲۱:۰۷ emadami ***@***.***>
نوشت:

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+protocol.
+
+In general, the Interchain dream will never scale unless there is a clear way to incentivize relayers. We seek to
+define a clear interface that can be easily adopted by any application, but not preclude chains that don't use tokens.
+
+### Desired Properties
+
+- Incentivize timely delivery of the packet (`OnReceivePacket` called)
+- Incentivize relaying acks for these packets (`OnAcknowledgement` called)
+- Incentivize relaying timeouts for these packets when the receive fee was too low (`OnTimeout` called)
```
milosevic, line 34:
```suggestion
- Incentivize relaying timeouts for these packets when the timeout has expired before packet is delivered (for example as receive fee was too low) (`OnTimeout` called)
```

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+We require that the [relayer address is exposed to application modules](https://github.com/cosmos/ibc/pull/579) for
+all packet-related messages, so the modules are able to incentivize the packet relayer. `OnAcknowledgement`, `OnTimeout`,
+and `OnTimeoutClose` messages will therefore have the relayer address and be capable of sending escrowed tokens to such address.
+However, we need a way to reliably get the address of the relayer that submitted `OnReceivePacket` on the destination chain to
+the source chain. In fact, we need a *source address* for this relayer to pay out to, not the *destination address* that signed
+the packet.
+
+Given this, the flow would be:
+
+1. User/module submits a send packet on the `source` chain, along with some tokens and fee information on how to distribute them
```
milosevic, line 60: Do we want to be a bit more precise on how to distribute fees?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+2. Relayer submits `OnReceivePacket` on the `destination` chain. Along with this message, the *forward relayer* will submit a `payToOnSource` address where payment should be sent.
+3. Destination application includes this `payToOnSource` in the acknowledgement (there are multiple approaches to discuss below)
+4. Relayer submits `OnAcknowledgement` which provides the *return relayer* address on the source chain, along with the `payToOnSource` address
+5. Source application can distribute the tokens escrowed in (1) to both the *forward* and the *return* relayers.
+
+Alternate flow:
+
+1. User/module submits a send packet on the `source` chain, along with some tokens and fee information on how to distribute them
+2. Relayer submits `OnTimeout` which provides their address on the source chain
```
milosevic, line 69:
```suggestion
2. Relayer submits `OnTimeout` which provides its address on the source chain
```

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+3. Source application can distribute the tokens escrowed in (1) to this relayer, and potentially return extra tokens to the original packet sender.
```
milosevic, line 70:
```suggestion
3. Source application can distribute the tokens escrowed in (1) to this relayer, and potentially return remainder tokens to the original packet sender.
```

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Fee details
+
+For an example implementation in the Cosmos SDK, we consider 3 potential fee payments, which may be defined. Each one may be
+paid out in a different token. Imagine a connection between IrisNet and the Cosmos Hub. They may define:
+
+- ReceiveFee: 0.003 channel-7/ATOM vouchers (ATOMs already on IrisNet via ICS20)
+- AckFee: 0.001 IRIS
```
milosevic, line 78: As ack and timeout are mutually exclusive, and sender is normally interested more in ack, maybe sender can only allocate some tokens and pay complete amount to a relayer that brings ack, and only portion of it if timeout is brought.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Incentivize timely delivery of the packet (`OnReceivePacket` called)
+- Incentivize relaying acks for these packets (`OnAcknowledgement` called)
+- Incentivize relaying timeouts for these packets when the receive fee was too low (`OnTimeout` called)
+- Produces no extra IBC packets
+- One direction works, even when one chain does not support concept of fungible tokens
+- Opt-in for each chain implementing this. eg. ICS27 with fee support on chain A could connect to ICS27 without fee support on chain B.
+- Standardized interface for each chain implementing this extension
+- Allow each chain/application to customize fee-handling logic
+- Permissionless relaying
```
milosevic, line 40: Do we want to somehow formalise a requirement that an address of the relayer that has submitted `OnReceivePacket` cannot be forged as part of `OnAcknowledgement` execution?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+Since different chains may have different representations for fungible tokens and this information is not being sent to other chains; this ICS does not specify a particular representation for the `Fee`. Each chain may choose its own representation, it is incumbent on relayers to interpret the Fee correctly.
+
+### Connection Negotiation
+
+The chains must agree to enable the incentivization feature during the connection handshake. This can be done by adding an incentivization feature to the connection version.
+
+```{"1", ["ORDER_ORDERED", "ORDER_UNORDERED", "INCENTIVE_V1"]}```
+
+If the negotiated connection includes the incentivization feature, then the ICS-4 `WriteAcknowledgement` function must write the forward relayer address into a structured acknowledgement and the ICS-4 handlers for `AcknowledgePacket` and `TimeoutPacket` must pay fees through the ibc fee module callbacks. If the negotiated connection does not include the incentivization feature, then the ICS-4 handlers must not modify the acknowledgement provided by the application and should not call any fee callbacks even if relayer incentivization is enabled on the chain.
+
```
milosevic, line 125: I guess we don't talk about enabling relayer incentivization on chain, but on a connection level, right?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+  acknowledgement: bytes,
+  proof: CommitmentProof,
+  proofHeight: Height,
+  relayer: string): Packet {
+    // get the forward relayer from the acknowledgement
+    // and pay fees to forward and reverse relayers.
+    // reverse_relayer is submitter of acknowledgement message
+    // provided in function arguments
+    // NOTE: Fee may be zero
+    forward_relayer = getForwardRelayer(acknowledgement)
```
milosevic, line 142: Where do we need to define logic for deciding in incentivization is enabled or not?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+The sender chain will escrow 0.003 channel-7/ATOM and 0.002 IRIS. In the case that a forward relayer submits the `MsgRecvPacket` and a reverse relayer submits the `MsgAckPacket`, the forward relayer is reqarded 0.003 channel-7/ATOM and the reverse relayer is rewarded 0.001 IRIS. In the case where the packet times out, the timeout relayer receives 0.002 IRIS and 0.003 channel-7/ATOM is refunded to the original fee payer.
+
+The logic involved in collecting fees from users and then paying it out to the relevant relayers is encapsulated by a separate fee module and may vary between implementations. However, all fee modules must implement a uniform interface such that the ICS-4 handlers can correctly pay out fees to the right relayers, and so that relayers themselves can easily determine the fees they can expect for relaying a packet.
+
+### Fee Module Contract
+
+```typescript
+function PayFee(packet: Packet, forward_relayer: string, reverse_relayer: string) {
```
milosevic, line 90: I don't understand if we want to enforce token fee escrowing as part of packet sending. Do we need an additional function that can be called as part of sending packet to pass tokens to fee module, or that logic is expected to be implemented as part of IBC application?

ethanfrey: I think there are two design decisions.

1. How the fee module works?
2. Where we call in the fee module? (From ICS-4 directly? From the apps (which can be implemented as an IbcHandler decorator, or just a function call)

I am more interested in spec-ing (1) first and when that is clear, decide where it fits (2) - which requires much more investigation at the code level. I think we can define all functionality of (1) without even looking at Go. And then it limits the design space we discuss later

spec/app/ics-029-fee-payment/README.md
```diff
@@ -96,7 +99,33 @@ function PayTimeoutFee(packet: Packet, timeout_relayer: string) {
 }
 ```
 
-These functions will then be utilized in the ICS-4 handlers like so:
+
+The fee module should also expose the following queries so that relayers may query their expected fee:
+
+```typescript
+// Gets the fee expected for submitting ReceivePacket msg for this packet
+function GetReceiveFee(packet) Fee
```
ethanfrey, line 107: I would simplify this with a `PacketID`, which would uniquely identify the packet without requiring the entire data.

This could be `(port_id, channel_id, sequence)` which should be unique on this chain.

spec/app/ics-029-fee-payment/README.md
ethanfrey: I think we need to discuss this a bit more.

There are 2 requirements.

1. The chain must support an extra `payOnSender` field in the `IbcReceivePacket` message, so the relayer can define that.
2. We must pass that data back in the acknowledgement. This could be done in the application or in a generic Acknowledgement wrapper understood by the connection/channel. It requires both sides understand the new acknowledgement

spec/app/ics-029-fee-payment/README.md
```diff
@@ -96,7 +99,33 @@ function PayTimeoutFee(packet: Packet, timeout_relayer: string) {
...
+function GetTimeoutFee(packet) Fee
+```
+
+Since different chains may have different representations for fungible tokens and this information is not being sent to other chains; this ICS does not specify a particular representation for the `Fee`. Each chain may choose its own representation, it is incumbent on relayers to interpret the Fee correctly.
+
+### Connection Negotiation
+
+The chains must agree to enable the incentivization feature during the connection handshake. This can be done by adding an incentivization feature to the connection version.
+
+```{"1", ["ORDER_ORDERED", "ORDER_UNORDERED", "INCENTIVE_V1"]}```
```
ethanfrey, line 122: I like the `"INCENTIVE_V1"` idea. Adding features to the connection.

However, is this opt-in per channel? How does it work?

`ORDER_ORDERED` means "we can create an ordered channel", but doesn't change the functionality for those not using an ordered channel.

Do we define "ordered, incentivized channel"? Or is this orthogonal?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -122,32 +149,24 @@ function timeoutPacket(
...
-// Gets the fee expected for submitting TimeoutPacket msg for this packet
-function GetTimeoutFee(packet) Fee
-```
+##### Correctness
 
-Since different chains may have different representations for fungible tokens and this information is not being sent to other chains; this ICS does not specify a particular representation for the `Fee`. Each chain may choose its own representation, it is incumbent on relayers to interpret the Fee correctly.
+The fee module is responsible for correctly escrowing and distributing funds to the provided relayers. It is IBC's responsibility to provide the fee module with the correct relayers. The ack and timeout relayers are trivially retrievable since they are the senders of the acknowledgment and timeout message.
 
-#### Reasoning
+The receive relayer submits the message to the counterparty chain. Thus the counterparty chain must communicate the knowledge of who relayed the receive packet to the source chain using the acknowledgement. The address that is sent back **must** be the address of the forward relayer on the source chain.
```
ethanfrey, line 167: This seems like the trickiest part to get right and I would expand on that here.

I think this works, but how do we handle all possible errors?
What if eg. the IbcReceivePacket doesn't have a `payOnSender` fields?
What if the *forward relayer* places an invalid address there? <- this case cannot be proven on the destination chain.

There is some level of "best effort" here. We should be clear under what conditions we are unable to provide that info. "Only if the forward relayer provides correct info" may be a valid caveat.

ethanfrey (commented):
Nice addition.

I am still thinking more about the most secure/failsafe way for the `payOnSender` info provided by the *forward relayer* to be validated and returned to the source chain.

More raising questions than anything else. The idea is good, we should just be more specific.

spec/app/ics-029-fee-payment/README.md
milosevic: It would be great if we can enforce that IBC processing ensures that if forward relayer passes invalid address, it will not pass validation check on source chain. In my view, that's the most important invariant we need to ensure.

spec/app/ics-029-fee-payment/README.md
AdityaSripal: Good question. My thinking is that all channels (whether they are ORDERED or UNORDERED) built on top of a connection with `INCENTIVE_V1` are incentivized channels. The ICS-4 handlers check if this feature is enabled in their underlying connection and do the appropriate logic.
All channels built on top of connection without that feature are unincentivized channels.

I think we want to enable two cases:
- Ability for an incentivized IBC to "downgrade" and connect with an unincentivized chain
- Ability for a module to enforce that it is connected by incentivized channel.

I'm not sure if there's a use for opting to not have incentivization when it is possible. Especially if the ICS-4 handlers are dealing with all fee logic, why would it matter to an app that its packets are getting incentivized?
In the case where fees are unnecessary for an app, users can just set 0 fees. There's no reason the app has to enforce this. Is there a use case you had in mind?

I think handling it this way is the easiest way to not require applications to bump their versions as a way of signalling that their underlying chain supports incentivization.

I'd prefer not to create channel types that grow in size on the order of the cross-product of available features.

An ideal solution I think would be to create a feature at the connection level, and then allow channels to opt-in to their desired subset of features.

This doesn't seem to be enabled by the channel handshake. It would be nice if the channel handshake allowed relayer to specify a subset of connection features it would like to use, but this is not possible since ORDER is the only specific choice provided to relayers on channel handshake. We should look into changing this for future versions of IBC.

However, given the lack of a use case for apps voluntarily barring incentivization and the desire to not communicate a core IBC feature in the channel version which is controlled by apps; I think this makes the most sense as an approach. However, I'm open to better ideas here.

spec/app/ics-029-fee-payment/README.md
AdityaSripal: Yes, this is definitely something I should specify in this doc. I think we can only go with "best efforts" approach here, because we do not want to fail packet processing if relayer makes a mistake.

As far as I can see there are two scenarios where this can happen.

It must either be caused by the forward relayer making a mistake => I see this as not as issue. They can fix their code to do it correctly.
Or it must be the case that the counterparty chain is implemented incorrectly => In this case, we would not want to prevent acknowledgements from getting processed. We would just 

If we go with a "best effort" approach which is where I'm leaning, then relayers who incorrectly set their address simply don't get paid until they fix the issue.
An incorrect counterparty implementation will initially burn some forward relayers, but they can learn quickly and the channel becomes a still functioning but unincentivized channel in forward direction until a fix is made.

In the case where we return error
A packet will effectively fail if forward address is incorrect. This is bad because it's already processed on counterparty and acknowledgment has been written. Thus we cannot retry with a better relayer, and acknowledgement can never be processed.
In case of incorrect implementation, acknowledgements can never be properly processed on source chain.

I think we should go with a "best efforts" approach here, and clearly define when we will not pay forward fee so that relayers can adapt if they are not getting their expected fees for some reason.

spec/app/ics-029-fee-payment/README.md
ethanfrey: > I'm not sure if there's a use for opting to not have incentivization when it is possible. Especially if the ICS-4 handlers are dealing with all fee logic, why would it matter to an app that its packets are getting incentivized?
In the case where fees are unnecessary for an app, users can just set 0 fees. There's no reason the app has to enforce this. Is there a use case you had in mind?
>
> I think handling it this way is the easiest way to not require applications to bump their versions as a way of signalling that their underlying chain supports incentivization.
>
> I'd prefer not to create channel types that grow in size on the order of the cross-product of available features.

I agree with you here.

ethanfrey: My question was since a channel on a connection with ```{"1", ["ORDER_ORDERED", "ORDER_UNORDERED", "INCENTIVE_V1"]}``` is incompatible with a channel on connection with ```{"1", ["ORDER_ORDERED", "ORDER_UNORDERED"]}```... why not just bump the version and call this

```{"2", ["ORDER_ORDERED", "ORDER_UNORDERED"]}```

File: spec/app/ics-029-fee-payment/README.md (hunk 12 above)
Comment (ethanfrey) [reply spec/app/ics-029-fee-payment/README.md:167]: Good reasoning.
This should definitely be made explicit in this doc. (Ideally with the reasoning, like you just did above)

File: spec/app/ics-029-fee-payment/README.md (hunk 5 above)
Comment (AdityaSripal) [reply spec/app/ics-029-fee-payment/README.md:78]: Sure, but the example ethan has works better to show how escrowing will work. As the minimum necessary escrow is 0.003 channel-7/ATOM and 0.002 IRIS

Comment (ethanfrey) [reply spec/app/ics-029-fee-payment/README.md:78]: > As ack and timeout are mutually exclusive, and sender is normally interested more in ack, maybe sender can only allocate some tokens and pay complete amount to a relayer that brings ack, and only portion of it if timeout is brought.

For ICS20, I am most interested in Receive, and not really interested in Ack... unless if is a failure (to release my tokens). So far, most "failure cases" are really timeouts. So for ICS20, I would incentivize receive and timeout. The *forward relayer* would only relay the ack to claim the receive fee. (not my example here, but if we try to model some "rational actor")

Comment (AdityaSripal) [reply spec/app/ics-029-fee-payment/README.md:79]: As a rational actor, I would set a high receive fee and just see if that's enough to get my receive and ack relayed by the same relayer. Maybe I would do a small ack fee.

I'd only set a timeout fee at all once the packet has already timed out

File: spec/app/ics-029-fee-payment/README.md (hunk 11 above)
Comment (AdityaSripal) [reply spec/app/ics-029-fee-payment/README.md:123]: Ahh yes, this would work.

In that case, a chain that supports incentivization would send over possible connection versions: `{{"1", ["ORDER_ORDERED", "ORDER_UNORDERED"]}, {"2", ["ORDER_ORDERED", "ORDER_UNORDERED"]}}`

And the counterparty chain will pick the highest that it can support.

File: spec/app/ics-029-fee-payment/README.md [hunk 13]
```diff
@@ -123,6 +124,8 @@ The chains must agree to enable the incentivization feature during the connectio
 
 If the negotiated connection includes the incentivization feature, then the ICS-4 `WriteAcknowledgement` function must write the forward relayer address into a structured acknowledgement and the ICS-4 handlers for `AcknowledgePacket` and `TimeoutPacket` must pay fees through the ibc fee module callbacks. If the negotiated connection does not include the incentivization feature, then the ICS-4 handlers must not modify the acknowledgement provided by the application and should not call any fee callbacks even if relayer incentivization is enabled on the chain.
 
+Thus, a chain can support incentivization while still maintaining connections that do not have the incentivization feature. This is crucial to enable a chain with incentivization to connect with a chain that does not have incentivization feature, as the acknowledgements need to be sent over the wire without the forward relayer.
```
ethanfrey, line 137: :100:

spec/app/ics-029-fee-payment/README.md
```diff
@@ -134,13 +137,18 @@ function acknowledgePacket(
...
+    connection = getConnection()
+    // only call the fee module callbacks if the connection supports incentivization.
+    if IsSupportedFeature(connection, "INCENTIVE_V1") {
+        // get the forward relayer from the acknowledgement
+        // and pay fees to forward and reverse relayers.
+        // reverse_relayer is submitter of acknowledgement message
+        // provided in function arguments
+        // NOTE: Fee may be zero
+        forward_relayer = getForwardRelayer(acknowledgement)
+        PayFee(packet, forward_relayer, relayer)
```
ethanfrey, line 160: I brought up something similar in #582 

It is good you define the interface between ICS-4 and the `PayFee` fee handler.

However, it doesn't seem like the call from ICS-4 to the application is specified anywhere. This seems a bit asymetric. How should we add that?

spec/app/ics-029-fee-payment/README.md
ethanfrey: > I'd only set a timeout fee at all once the packet has already timed out

You mean you can change a fee after sending? That would be a huge complexity increase (also on the relayer, which must re-check every packet even after seeing "too low fee"). There is no changes to increasing tx fees, and I would like to remove such an approach.

spec/app/ics-029-fee-payment/README.md
ethanfrey: Thinking more (and looking at pseudocode below), it seems we *should* specify the *interface* for storing the fee for a packet. Whether this is called from ICS-4 in send-packet or in the application can be discussed separately. But the interface on the fee handler can be specified.

`function SetPacketFee(packet: Packet, receiveFee: Fee, ackFee: Fee, timeoutFee: Fee)`

And then define it may be called once, atomically with committing the packet. If it is not called (or called with 0) then the packet is un-incentivized, which is a valid state.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+Since different chains may have different representations for fungible tokens and this information is not being sent to other chains; this ICS does not specify a particular representation for the `Fee`. Each chain may choose its own representation, it is incumbent on relayers to interpret the Fee correctly.
+
+### Connection Negotiation
+
+The chains must agree to enable the incentivization feature during the connection handshake. This can be done by bumping the connection version.
+
+```{"2", ["ORDER_ORDERED", "ORDER_UNORDERED"]}```
+
+Since most chains that support incentivization will wish to be compatible with chains that do not, a chain with `V2` enabled will send its possible connection versions: `{{"1", ["ORDER_ORDERED", "ORDER_UNORDERED"]}, {"2", ["ORDER_ORDERED", "ORDER_UNORDERED"]}}` in `ConnOpenInit`. The counterparty chain will select the highest version that it can support.
```
ethanfrey, line 133: :+1:

ethanfrey (commented):
Nice improvements.

Added a few more comments on clarifying fee flow

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+- Incentivize timely delivery of the packet (`OnReceivePacket` called)
+- Incentivize relaying acks for these packets (`OnAcknowledgement` called)
+- Incentivize relaying timeouts for these packets when the timeout has expired before packet is delivered (for example as receive fee was too low) (`OnTimeout` called)
+- Produces no extra IBC packets
+- One direction works, even when one chain does not support concept of fungible tokens
+- Opt-in for each chain implementing this. eg. ICS27 with fee support on chain A could connect to ICS27 without fee support on chain B.
+- Standardized interface for each chain implementing this extension
+- Support custom fee-handling logic within the same framework
+- Relayer addresses should not be forgable
```
cwgoes, line 40:
```suggestion
- Relayer addresses should not be forgeable
```

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+- Permissionless relaying
+
```
cwgoes, line 42: Are we considering potential race conditions between relayers (i.e. where multiple relayers try to submit a packet in order to claim the fee and most of their transactions fail)? This is already a bit of a concern with IBC, but certain ways of paying fees would probably make it much worse, I think we should keep this in mind.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+In general, the Interchain dream will never scale unless there is a clear way to incentivize relayers. We seek to
+define a clear interface that can be easily adopted by any application, but not preclude chains that don't use tokens.
+
+### Desired Properties
+
+- Incentivize timely delivery of the packet (`OnReceivePacket` called)
+- Incentivize relaying acks for these packets (`OnAcknowledgement` called)
+- Incentivize relaying timeouts for these packets when the timeout has expired before packet is delivered (for example as receive fee was too low) (`OnTimeout` called)
+- Produces no extra IBC packets
+- One direction works, even when one chain does not support concept of fungible tokens
```
cwgoes, line 36: When either chain, or when the destination chain?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+- Relayer addresses should not be forgable
+- Permissionless relaying
+
+## Technical Specification
+
+### General Design
+
+In order to avoid extra packets on the order of the number of fee packets, as well as provide an opt-in approach, we
+store all fee payment info only on the source chain. The source chain is the one location where the sender can provide tokens
+to incentivize the packet. The fee distribution may be implementation specific and thus does not need to be in the ibc spec
```
cwgoes, line 49:
```suggestion
to incentivize the packet. The fee distribution may be implementation specific and thus does not need to be in the IBC spec
```

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+all packet-related messages, so the modules are able to incentivize the packet relayer. `OnAcknowledgement`, `OnTimeout`,
+and `OnTimeoutClose` messages will therefore have the relayer address and be capable of sending escrowed tokens to such address.
+However, we need a way to reliably get the address of the relayer that submitted `OnReceivePacket` on the destination chain to
+the source chain. In fact, we need a *source address* for this relayer to pay out to, not the *destination address* that signed
+the packet.
+
+Given this, the flow would be:
+
+1. User/module submits a send packet on the `source` chain, along with some tokens and fee information on how to distribute them. The fee tokens are all escrowed by the fee module.
+2. Relayer submits `OnReceivePacket` on the `destination` chain. Along with this message, the *forward relayer* will submit a `payToOnSource` address where payment should be sent.
```
cwgoes, line 62:
```suggestion
2. The first relayer, known in this spec as the _forward relayer_, submits `OnReceivePacket` on the `destination` chain. Along with this message, the *forward relayer* will submit a `payToOnSource` address where payment should be sent.
```

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+3. Destination application includes this `payToOnSource` in the acknowledgement (there are multiple approaches to discuss below)
+4. Relayer submits `OnAcknowledgement` which provides the *return relayer* address on the source chain, along with the `payToOnSource` address
```
cwgoes, line 64: This is _not necessarily the same relayer_, right? We should clarify (e.g. name them relayers 1, 2 or A, B)

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+### Fee details
+
+For an example implementation in the Cosmos SDK, we consider 3 potential fee payments, which may be defined. Each one may be
+paid out in a different token. Imagine a connection between IrisNet and the Cosmos Hub. They may define:
+
+- ReceiveFee: 0.003 channel-7/ATOM vouchers (ATOMs already on IrisNet via ICS20)
+- AckFee: 0.001 IRIS
+- TimeoutFee: 0.002 IRIS
+
+Ideally the fees can easily be redeemed in native tokens on both sides, but relayers may select others. In this example, the relayer collects a fair bit of IRIS, covering it's costs there and more. It also collects channel-7/ATOM vouchers from many packets. After relaying a few thousand packets, the account on the Cosmos Hub is running low, so the relayer will send those channel-7/ATOM vouchers back over channel-7 to it's account on the Hub to replenish the supply there.
```
cwgoes, line 82:
```suggestion
Ideally the fees can easily be redeemed in native tokens on both sides, but relayers may select others. In this example, the relayer collects a fair bit of IRIS, covering its costs there and more. It also collects channel-7/ATOM vouchers from many packets. After relaying a few thousand packets, the account on the Cosmos Hub is running low, so the relayer will send those channel-7/ATOM vouchers back over channel-7 to it's account on the Hub to replenish the supply there. 
```

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+// Gets the fee expected for submitting TimeoutPacket msg for the given packet
+function GetTimeoutFee(portID, channelID, sequence) Fee
+```
+
+Since different chains may have different representations for fungible tokens and this information is not being sent to other chains; this ICS does not specify a particular representation for the `Fee`. Each chain may choose its own representation, it is incumbent on relayers to interpret the Fee correctly.
+
+### Connection Negotiation
+
+The chains must agree to enable the incentivization feature during the connection handshake. This can be done by bumping the connection version.
```
cwgoes, line 129: Wait, why at the connection level instead of at the channel level? Different channels for different applications - but on the same connection - might have totally different fee models.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+The sender chain will escrow 0.003 channel-7/ATOM and 0.002 IRIS. In the case that a forward relayer submits the `MsgRecvPacket` and a reverse relayer submits the `MsgAckPacket`, the forward relayer is reqarded 0.003 channel-7/ATOM and the reverse relayer is rewarded 0.001 IRIS. In the case where the packet times out, the timeout relayer receives 0.002 IRIS and 0.003 channel-7/ATOM is refunded to the original fee payer.
+
+The logic involved in collecting fees from users and then paying it out to the relevant relayers is encapsulated by a separate fee module and may vary between implementations. However, all fee modules must implement a uniform interface such that the ICS-4 handlers can correctly pay out fees to the right relayers, and so that relayers themselves can easily determine the fees they can expect for relaying a packet.
```
cwgoes, line 86: Why do we want to do this from the ICS-4 handlers? Is it possible to split this out into middleware? I don't really like including fee logic in core IBC since it has nothing to do with correctness of transport, authentication, or ordering semantics.

cwgoes (changes requested):
Thank you for this very clear proposal. Most of what the specification discusses makes sense; however, I have two primary concerns which I haven't seen discussed (though feel free to route me to discussions if I have missed them):

- This is more of an ergonomic preference, but would it be possible to split this logic out of ICS 4 instead of stuffing it in the core protocol? I think _standardising_ fees definitely makes sense, but I hesitate to introduce more complexity into the core handlers for anything non-essential to correctness or liveness. Could this be a middleware (behaving appropriately on the callbacks) instead?
- It seems to me like including fees like this - without any restriction on the relayer - greatly intensifies the "race condition" - any relayer which sees a potential high fee for relaying a packet will want to try, many relayers will try at once, and a lot of gas / fees will be wasted as a result. IBC already has this problem in a sense, but without fees paid directly for relaying there's much less reason for relayers to compete. I realise there may be legal concerns with specifying a particular relayer who is to receive a fee, but I think we should also at the very least be aware of and conduct a simple analysis of this issue. Maybe @ancazamfir or @milosevic could chime in here as I know Informal has thought about this in the context of Hermes.

spec/app/ics-029-fee-payment/README.md
AdityaSripal: Could you explain your reasoning here? I don't understand.

> You mean you can change a fee after sending? That would be a huge complexity increase

I only would allow the ability to increase a fee after sending. I don't think it's a huge complexity increase at all. All it would mean is the ability to provide a message that escrows more fee funds against a particular packetID.
In fact, my original proposal had the fee be a completely separate message that could be sent synchronously or asynchronously with the packet.
I do think most apps (and maybe in ICS-4) would like to set fee on Send, but I don't see a reason why those fees can't be increased if the packet is just not getting relayed by anyone.

From the blockchain standpoint, this would simply be implementing a message on the fee-module that allows you to set fees asynchronously from the packet send. The logic of the handler would simply call `SetFee`, which itself will add fees onto any existing escrow or set the escrow if it doesn't already exist.
Applications (or ICS-4) may set the initial fee synchronously during execution by calling the `SetFee` callback, but it may be asynchronously set through the ibc-fee handler itself.

From the relayer point of view, re-checking fees is not required. If you ignore something due to initially low fee, you lose out to smarter relayers. I don't think it's that difficult to create a smart relayer:
Emit events whenever the escrow amount for a packet is increased, relayers can monitor these to see if a packet is worth relaying (regardless of if they've checked before) and relay if they believe it is.

I don't think it's that complex from the blockchain or relayer view.

Additionally, I think it is a big UX boost for users. We want to make sure that users can get their packets relayed.

In the regular transaction case, if your tx gets rejected from mempool you can easily just resubmit with no issues. In the IBC case, if your packet is successfully committed but no relayers want to relay your packet, there's no recourse except to relay your own packet. Until then any funds (ICS-20 case) are stuck. From a user perspective, it means I have to make sure I have enough fees (probably overestimate) to ensure a good UX experience.

I think in the case where a user puts in zero fees or too low fees during packet send, I don't want their funds indefinitely stuck. Nor do I want to force them to set up a relayer and relay their own fees to recover their funds, just because they were hoping for an altruistic relay. Ideally, they wait for some time; realize no one's picking up their packet with those fees, and decide to bump fees in the hope that it gets relayed then.

This was discussed as a strength of the proposal here: https://github.com/cosmos/ibc/issues/578#issuecomment-851622881

spec/app/ics-029-fee-payment/README.md
AdityaSripal: I agree we should standardize `SetPacketFee` interface.

> Whether this is called from ICS-4 in send-packet or in the application can be discussed separately. But the interface on the fee handler can be specified.

I'm a little resistant to doing this in ICS-4 in send-packet. Ideally, I'd like this spec to be a natural fit for most custom fee modules.

The usecase you've mentioned above works really well for the default case where sender pays for the packet flow.

However, consider the case of the altruistic module funds packets on a particular channel.
Here, the module does not escrow some additional funds on each packet send. Instead, the funds have been escrowed in advance for the purpose of paying for future packets. Having a `SetPacketFee` being called in ICS-4 handler here would be unnatural. Sure, we could just set 0 fee in ICS-4 and then have the fees come from the previously created escrow account, but this would of course require the ability to set more than once. In this case, there could be no explicit `SetPacketFee` at all, as the fund could just release some constant amount from its escrow on `PayFee` and `PayTimeoutFee` callbacks.

> And then define it may be called once, atomically with committing the packet. If it is not called (or called with 0) then the packet is un-incentivized, which is a valid state.

For reasons I've stated above, I support the ability to call `SetPacketFee` multiple times with each subsequent call adding onto the previous escrow.

I think `SetPacketFee` should be an open function that can be called into by IBC applications, third-party users, DAOs, governance, whatever. And they should all be able to parallelly incentivize a particular packet without interfering with each other. The relayer will reap collective rewards from all incentivizers. I don't believe this is hard to do because of the way we've split responsibilities in this spec. It's a natural usecase of the functions in this spec.

spec/app/ics-029-fee-payment/README.md
AdityaSripal: Added, let me know if you had something else in mind

spec/app/ics-029-fee-payment/README.md
AdityaSripal: Yes we've looked at permissionless relaying since that seems easiest from legal compliance standpoint, but I agree with you.

I also think there was a comment by @zmanian about allowing any altruistic relayer to relay packet, but only rewarding fee if it is a specific relayer potentially.

I've changed the interfaces to support "permissioned" relaying, so that if it is legal we can trivially support it from a technical standpoint

cc: @charleenfei

spec/app/ics-029-fee-payment/README.md
AdityaSripal: The connection is just negotiating the general fee feature. It is in effect asking if the counterparty chain will send the forward_relayer back in the acknowledgement. That's it. This is not a feature of the application, this is a feature of the chain; which is why we're negotiating on connection so we have a uniform way to negotiate and send/receive relayer information in ICS-4 handlers rather than having all IBC apps implement this feature themselves and encode core IBC feature support into their app-specific channel versions.
 
Since this proposal handles fee payment logic completely on a single chain, we are not negotiating anything specific about fee-handling logic. Both apps can implement completely different fee models using the interfaces in this spec. So long as each chain is propogating relayer information to each other correctly, they can operate their different fee models as much as they like. It does not need to be agreed upon even between apps on opposite ends of a channel. One side could do sender-incentivized packets while the other side uses an altruistic fund.

spec/app/ics-029-fee-payment/README.md
AdityaSripal: I think the main question is do we enable this feature on a per-app basis, or on a per-chain basis.

I think there's consensus to enable it per-chain: https://github.com/cosmos/ibc/pull/581#discussion_r644758984

I'm agnostic on whether we implement this uniform behaviour directly into ICS-4 handlers, or create some middleware that wraps all application callbacks.

cc: @ethanfrey

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+- It must pay the timeout fee for a packet to the timeout relayer specified in `PayTimeoutFee` callback
+- It must refund any remainder fees in escrow to the original fee payer(s) if applicable
+
+```typescript
+function PayFee(packet: Packet, forward_relayer: string, reverse_relayer: string) {
+    // pay the forward fee to the forward relayer address
+    // pay the reverse fee to the reverse relayer address
+    // refund extra tokens to original fee payer(s)
+}
+
```
ancazamfir, line 117: Given that the fees are per packet and could be different accross different messages for the same packet, relayers could play some strategy games resulting in some low fee packets neither being ack-ed nor timed-out. How do we unstuck these packets? there should exist a mechanism ala `Replace by Fee`

ancazamfir:
> * It seems to me like including fees like this - without any restriction on the relayer - greatly intensifies the "race condition" - any relayer which sees a potential high fee for relaying a packet will want to try, many relayers will try at once, and a lot of gas / fees will be wasted as a result. IBC already has this problem in a sense, but without fees paid directly for relaying there's much less reason for relayers to compete. I realise there may be legal concerns with specifying a particular relayer who is to receive a fee, but I think we should also at the very least be aware of and conduct a simple analysis of this issue. Maybe @ancazamfir or @milosevic could chime in here as I know Informal has thought about this in the context of Hermes.

Yes, with the per packet and per packet type fees, "optimal" relayer strategies will result in packets messages being relayed multiple time or not relayed at all. We have not done a proper analysis yet as the proposal was not clear until recently with some details still needing clarification.

spec/app/ics-029-fee-payment/README.md
ethanfrey: > I only would allow the ability to increase a fee after sending. I don't think it's a huge complexity increase at all. All it would mean is the ability to provide a message that escrows more fee funds against a particular packetID.
In fact, my original proposal had the fee be a completely separate message that could be sent synchronously or asynchronously with the packet.

This is not much complexity increase on the blockchain side. And only allowing increases makes it safe for relayers.

The complexity increase is that relayers must be conscious of all future changes to this packet state. No longer checking the packet the first time and putting in "relay" or "ignore" buckets, but they must re-evaluate all "ignores" in future iterations.

This could get complex.

> From the relayer point of view, re-checking fees is not required. If you ignore something due to initially low fee, you lose out to smarter relayers. I don't think it's that difficult to create a smart relayer:

Fair enough. Supporting that on the server does seem the "right" thing to do, and relayers can start with ignoring fee increases at start and only later add that functionality. I am not sure how `hermes` works, but `ts-relayer` checks new packets and handles them right away, it doesn't have some sort of holding pattern for them.

spec/app/ics-029-fee-payment/README.md
ethanfrey: Nice argument.

If we allow `SetPacketFee` to be called multiple times, we need a new name.

`IncrementPacketFee`? `EscrowPacketFee`? 

Something that makes clear it is additive.

spec/app/ics-029-fee-payment/README.md
ethanfrey: I agree with pulling the actual fee handling logic out of ICS-4, but rather define an interface for it.

We could just have a middleware that fulfills the application interface and then calls into another application. Or we could have something like an `AnteHandler` interface, which is called before dispatching to the application.

There have been requests to keep the application interface stable (and not expose the `payOnSource` info there, just the raw acknowledgement bytes). Thus, we couldn't just wrap one application with a fee handler decorator in the same interface. If we do pass all info down, this would be a valid approach IMO. 

If we let app "decorate" it's handlers to allow fee-processing, incentivization may be missing on many cases, which could be bad or good. (I have heard mention of it being opt-in)

I am more in favor of defining some "AnteHandler" interface in ICS4/ICS26 that gets the full metadata on packet and acknoweldgement and relayer/signer. The fee handler must fit into such an interface (but functionality be defined in it's own application-level spec). We could later add other "Ante Handlers" without requiring changes in the core handlers or logic.

ethanfrey: "AnteHandler" allows us to enable this on a per-chain basis, without requiring more than an optional callback in ICS-4.

spec/app/ics-029-fee-payment/README.md
ethanfrey: > I don't think running relayers will be a profitable enough business (at this point and volume of ibc tx) for a hugely competitive relayer market, so really the incentivisation is to allow for relayer operators to at least cover their costs and will be more of a way for projects to offset costs than for a independent business model at the moment.

I agree here.

If it does become hugely profitable at some point, then I would assume that it would become like DeFi front-running. And in the end the validators will run the relayers and just relay on the blocks they propose, putting their messages first. Which is actually quite efficient and still relatively decentralized.

I assume we would only have 2 or 3 relayers trying not to go broke. But I also think the end-game of fee competition will not be so horrible. 

I also agree with technically allowing optional permissioned relaying. But think users should not be able to set the "selected relayer" on their packets unless it is enabled by chain governance (the legal burden is not on the users)

spec/app/ics-029-fee-payment/README.md
charleenfei: i agree with @AdityaSripal here, the user should have the ability to replay the tx with a higher fee.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+function acknowledgePacket(
+  packet: OpaquePacket,
+  acknowledgement: bytes,
+  proof: CommitmentProof,
+  proofHeight: Height,
+  relayer: string): Packet {
+    // get the underlying connection for this channel
+    connection = getConnection()
+    // only call the fee module callbacks if the connection supports incentivization.
+    if IsSupportedFeature(connection, "INCENTIVE_V1") {
```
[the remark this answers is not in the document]
charleenfei: i guess it should be V2 if it's the incentivisation enabled connection version (following the logic in line 152).

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,252 @@
...
+function timeoutPacket(
+  packet: OpaquePacket,
+  proof: CommitmentProof,
+  proofHeight: Height,
+  nextSequenceRecv: Maybe<uint64>,
+  relayer: string): Packet {
+    // get the underlying connection for this channel
+    connection = getConnection()
+    // only call the fee module callbacks if the connection supports incentivization.
+    if IsSupportedFeature(connection, "INCENTIVE_V1") {
```
charleenfei, line 202: same v1/v2 comment here

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,223 @@
...
+### General Design
+
+In order to avoid extra packets on the order of the number of fee packets, as well as provide an opt-in approach, we
+store all fee payment info only on the source chain. The source chain is the one location where the sender can provide tokens
+to incentivize the packet. The fee distribution may be implementation specific and thus does not need to be in the ibc spec
+(just high-level requirements are needed in this doc).
+
+We require that the [relayer address is exposed to application modules](https://github.com/cosmos/ibc/pull/579) for
+all packet-related messages, so the modules are able to incentivize the packet relayer. `OnAcknowledgement`, `OnTimeout`,
+and `OnTimeoutClose` messages will therefore have the relayer address and be capable of sending escrowed tokens to such address.
```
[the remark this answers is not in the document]
AdityaSripal: This original comment does seem to be referring to the ability for app-specific callbacks to use the relayer address. Is that still the intention even after pulling fee-handling logic outside of applications @ethanfrey ?

zmanian:
The middleware version looks good to me and I support it

spec/app/ics-029-fee-payment/README.md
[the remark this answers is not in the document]
ethanfrey: > This original comment does seem to be referring to the ability for app-specific callbacks to use the relayer address. Is that still the intention even after pulling fee-handling logic outside of applications @ethanfrey ?

There was also the request from Dean and Bucky that receiving apps be able to incentive packets (like an IBC-enabled DEX paying to receive orders, as it makes more from the fees on executing the order). For that we would need to pass the relayer to the app (even if we don't need it for fee handling any more).

And if we make fee handler a middleware, we definitely need it.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+Since different chains may have different representations for fungible tokens and this information is not being sent to other chains; this ICS does not specify a particular representation for the `Fee`. Each chain may choose its own representation, it is incumbent on relayers to interpret the Fee correctly.
+
+### IBC Module Wrapper
+
+The fee module will implement its own ICS-26 callbacks that wrap the application-specific module callbacks. This fee module middleware will ensure that the counterparty module supports incentivization and will implement all fee-specific logic. It will then pass on the request to the embedded application module for further callback processing.
+
+In this way, custom fee-handling logic can be hooked up to the IBC packet flow logic without placing the code in the ICS-4 handlers or the application code. This is valuable since the ICS-4 handlers should only be concerned with IBC correctness, and the application handlers should not be handling fee logic that is universal amongst all other incentivized applications. In fact, a given application module should be able to be hooked up to any fee module with no further changes to the application itself.
```
cwgoes, line 150:
```suggestion
In this way, custom fee-handling logic can be hooked up to the IBC packet flow logic without placing the code in the ICS-4 handlers or the application code. This is valuable since the ICS-4 handlers should only be concerned with correctness of core IBC (transport, authentication, and ordering), and the application handlers should not be handling fee logic that is universal amongst all other incentivized applications. In fact, a given application module should be able to be hooked up to any fee module with no further changes to the application itself.
```

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+As mentioned above, the fee module will implement the ICS-26 callbacks, and can embed an application IBC Module. All ICS-26 callbacks in the fee module will call into the embedded application's callback. Thus, only the fee callbacks that do something additional are explicitly specified here.
+
+#### Fee Protocol Negotiation
+
+The fee middleware will negotiate its fee protocol version with the counterparty module by prepending its own version to the application version.
```
cwgoes, line 156: This seems a little awkward to me, although workable - should we encode a list of strings instead in some standard way?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+It is crucial that this proposal maintains backwards compatibility with a chain that does not support relayer incentivization. Thus, we must support the case where one side of the channel does not support the incentivization feature, the version will only contain the app-specific version. In the case, where `ChanOpenInit` is initiated on the chain that supports incentivization; the relayer must explicitly create the channel with just the app version (e.g. `ics20-1`), otherwise the handshake will fail on counterparty during channel version negotiation since counterparty module does not understand the fee version prefix. In the case, where `ChanOpenInit` is initiated on the chain that does not support incentivization, relayer for `ChanOpenTry` may choose to submit just application version or the fee-prefixed version. The fee middleware will inspect counterparty version and "downgrade" channel to unincentivized logic if the counterparty version does not have fee prefix.
```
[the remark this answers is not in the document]
cwgoes: I think the version passthrough is OK for now, I don't have better ideas which aren't way more complicated.

I'm not sure about downgrading, though - shouldn't we require the relayer to figure out which version should be used, and just fail the handshake if they try to use the fee middleware but it isn't supported? Can you explain the last sentence in more detail? Maybe I misunderstand.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,391 @@
...
+function getForwardRelayer(ack: Acknowledgement): string {
+    // TODO: see https://github.com/cosmos/ibc/pull/582
+}
+
+function getAppAcknowledgement(ack: Acknowledgement): bytes {
+    // TODO: see https://github.com/cosmos/ibc/pull/582
+}
+```
+
+Note that if the embedded application uses asynchronous acks then, the `WriteAcknowledgement` call in the application must call the fee middleware's `WriteAcknowledgement` rather than calling the ICS-4 handler's `WriteAcknowledgement` function directly.
```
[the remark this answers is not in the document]
cwgoes: As long as we make this clear by exposed interfaces in the implementations (e.g. SDK), I think this is fine.

cwgoes (commented):
Thank you for writing up the middleware version! It looks reasonably clear to me; I've commented on the parts you requested feedback on, I think.

spec/app/ics-029-fee-payment/README.md
AdityaSripal: Do you mean changing channel version to be a channel list?

I'm in favor, though this would be a breaking change, so something to do in future.

If you mean creating a list of strings inside the string that gets deserialized in a particular way, that is what my proposal is doing. Though there may be a much better way to do it.

cwgoes: I think this is fine for now.

cwgoes (approved):

spec/app/ics-030-middleware/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+When nesting an application, the module must make sure that it is in the middle of communication between core IBC and the application in both directions. Developers should do this by registering the top-level module directly with the IBC router (not any nested applications). The nested applications in turn, must be given access only to the middleware's `WriteAcknowledgement` and `SendPacket` rather than to the core IBC handlers directly.
+
+Additionally, the middleware must take care to ensure that the application logic can execute its own port and version negotiation without interference from the nesting middleware. In order to do this, the middleware will prepend the portID and version with its own portID and version. In the application callbacks, the middleware must do its own version and port negotation and then strip out the prefixes before handing over the data to the nested application's callback. Middleware SHOULD always prepend the portID with its own port. This will allow the original application to also exist as a top-level module connected to the IBC Router.
+
+PortID: `{middleware_port}:{app_port}`
+
+Version: `{middleware_version}:{app_version}`
```
ethanfrey, line 44: I fully support the wrapped app-versions. This basically means each stack is a different app, but still allows version negotiation. Eg. we have `fee-1`, and `whitelist-1` middlewares.

You could build `fee-1:ics20-1` or `fee-1:whitelist-1:ics27-1` applications, just by connecting a few different stacks in code. Each of these apps is just composition, but has a unique version (it must talk to the same remote stack).

I like this idea.

spec/app/ics-030-middleware/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+When nesting an application, the module must make sure that it is in the middle of communication between core IBC and the application in both directions. Developers should do this by registering the top-level module directly with the IBC router (not any nested applications). The nested applications in turn, must be given access only to the middleware's `WriteAcknowledgement` and `SendPacket` rather than to the core IBC handlers directly.
+
+Additionally, the middleware must take care to ensure that the application logic can execute its own port and version negotiation without interference from the nesting middleware. In order to do this, the middleware will prepend the portID and version with its own portID and version. In the application callbacks, the middleware must do its own version and port negotation and then strip out the prefixes before handing over the data to the nested application's callback. Middleware SHOULD always prepend the portID with its own port. This will allow the original application to also exist as a top-level module connected to the IBC Router.
+
+PortID: `{middleware_port}:{app_port}`
```
ethanfrey, line 42: On the other hand, this seems very odd to me. We can have any port for an app.
`ics20` could listen on `transfer` or `custom` or `wasm.cosmos1dj8fifj3ljfliw7nfh8w87cb` (I have tests with all these in `ts-relayer`).

The application logic has no hardcoded port and should not concern itself with that. That should be a param that can be set in genesis (or elsewhere). 

Basically, I would remain that port can be configured to anything and remove this line, just saying that different middleware stacks would be different ports.

You could even do `transfer` => `fee-1:ics20-1`
 and `old-transfer` => `ics20-1`

Forcing people to connect `old-transfer` with older chains.

spec/app/ics-030-middleware/README.md
```diff
@@ -0,0 +1,190 @@
...
+```typescript
+function onChanOpenInit(
+  order: ChannelOrder,
+  connectionHops: [Identifier],
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  counterpartyPortIdentifier: Identifier,
+  counterpartyChannelIdentifier: Identifier,
+  version: string) {
+    middlewarePort, appPort = splitMiddlewarePort(portID)
```
ethanfrey, line 57: This (and all calls of `splitMiddlewarePort`) is what I would remove

ethanfrey (commented):
I do like the middleware idea and do think it is a very elegant alternative to #582 (which I have basically dropped now). Thank you very much for writing this down explicitly. My only remaining questions on the fee middleware was exactly how middleware was supposed to work, so if we all agree on this piece, then the rest of the PR should be quite solid.

From the ics-perspective, we can now easily make 2 apps: `ics20-1` and `fee-1:ics20-1` which are:

* incompatible with each other
* use the same ics20 logic
* allow compatible chains to connect to either (or both variants)
* allow reuse of `fee-1` with other modules (eg `ics27-1`).

This is a great acheivement. I have a few points on port selection, but from the chain-chain perspective, this is quite clean. What I do wonder about is `user-chain` communication. I will break this into multiple points (below):

ethanfrey:
On application-specific "send packet" type messages:

`MsgTransfer` will currently work with `ics20-1`. What message do I need to use `fee-1:ics20-1`. I assume it will need a new message type. And where would that message live? 

It is aware of both `fee-1` and `ics20-1`? This needs to be well defined protobuf with good cli, cosmjs, and rust support ideally so the relayers can work with it well. This applies to any application level "send" message that creates a packet. And I assume we will need one such message for any combination (one for `ics20-1`, another for `fee-1:ics20-1`, another for `ics27-1`, another for `fee-1:ics27-1`, and maybe yet another for `fee-1:gov-whitelist-1:ics27-1`).

How do we handle composition here?

ethanfrey:
Following up on that, how would we extend Smart Contracts...

In CosmWasm, `x/wasm` registers a number of ports on behalf of the contracts and routes messages to them (including channel handshake). I could make it such that it wraps them all with `fee-1` transparently, and they just act the same but allow for incentivized packets - that would be awesome.

However, the same issue as above came up. What happens when a contract wishes to create a packet. It must know whether or not it is fee-enabled to make the proper fee payment calls. This no longer becomes transparent to the module, yet it is ignorant of whether or not it is wrapped. This will become quite difficult (and I welcome any suggestions).

This is how a CosmWasm contract [currently makes an IBC packet](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/ibc.rs#L37-L45). There is still a little time to expand this before we freeze for 1.0, but how?

The same applies to CosmWasm contracts [using `ics20-1` to send tokens](https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/ibc.rs#L24-L28). We actually assumed there is only one possible port and do not expose fee payment options. 

We need good interfaces not just for blockchain clients, but for other modules.

ethanfrey:
Finally, this whole fee middleware will need to add a user-settable field `payToSource` field on the generic `MsgReceivePacket`, which will also be exposed to all middlewares and applications. Again, this breaks these nice "transparent middleware" abstractions for the blockchain users. Can other middlewares just as easily add more fields to these protobuf messages?

(Aside)

Although there is ideally "backwards compatibility" the way it has been implemented in the Cosmos SDK is so strict, that CosmJS struggles with it. First, they version the type names, and a chain only seems to support one of `v1-beta`, `v1`, `v2` of a given message structures. So, the client must know the exact type and adding another field and making a "backwards compatible" version bump will break all clients due to this string.

Also, to ensure the integrity of messages, the SDK will reject any incoming proto message with unknown fields. So even if we use multiple backwards compatible message types as `v1` (eg `MsgReceivePacket` with and without the optional `payToSender` field), if you do fill out the field and send it to a chain that doesn't know about the field, it will reject the message for containing unknown fields.

Maybe ibc spec doesn't care about this, but this is a HUGE real world issue that needs to be clarified.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,405 @@
...
+- It must have in escrow the maximum fees that all outstanding packets may pay out (or it must have ability to mint required amount of tokens)
+- It must pay the receive fee for a packet to the forward relayer specified in `PayFee` callback
+- It must pay the ack fee for a packet to the reverse relayer specified in `PayFee` callback
+- It must pay the timeout fee for a packet to the timeout relayer specified in `PayTimeoutFee` callback
+- It must refund any remainder fees in escrow to the original fee payer(s) if applicable
+
+```typescript
+// EscrowPacketFee is an open callback that may be called by any module/user that wishes to escrow funds in order to
+// incentivize the relaying of the given packet.
+// NOTE: These fees are escrowed in addition to any previously escrowed amount for the packet.
```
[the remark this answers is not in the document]
AdityaSripal: Correct

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,405 @@
...
+// Gets the fee expected for submitting AcknowledgePacket msg for the given packet
+// Caller should provide the intended relayer address in case the fee is dependent on specific relayer(s).
+function GetAckFee(portID, channelID, sequence, relayer) Fee
+
+// Gets the fee expected for submitting TimeoutPacket msg for the given packet
+// Caller should provide the intended relayer address in case the fee is dependent on specific relayer(s).
+function GetTimeoutFee(portID, channelID, sequence, relayer) Fee
+```
+
+Since different chains may have different representations for fungible tokens and this information is not being sent to other chains; this ICS does not specify a particular representation for the `Fee`. Each chain may choose its own representation, it is incumbent on relayers to interpret the Fee correctly.
```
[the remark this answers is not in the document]
AdityaSripal: Good point, though I used a format more in-line with ICS-20

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,405 @@
...
+
+Thus, a relayer looking to create an incentivized channel between two incentivized chains can do so by creating channel between their `fee:transfer` modules. A relayer looking to create an unincentivized channel on a backwards-compatible incentivized chain may do so by creating the channel on the `transfer` port.
+
+#### Reasoning
+
+This proposal satisfies the desired properties. All parts of the packet flow (receive/acknowledge/timeout) can be properly incentivized and rewarded. The protocol does not specify the relayer beforehand, thus the incentivization is permissionless. The escrowing and distribution of funds is completely handled on source chain, thus there is no need for additional IBC packets or the use of ICS-20 in the fee protocol. The fee protocol only assumes existence of fungible tokens on the source chain. Using the connection version, the protocol enables opt-in incentivization and backwards compatibility. The fee module can implement arbitrary custom logic so long as it respects the callback interfaces that IBC expects.
```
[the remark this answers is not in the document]
AdityaSripal: Application version only, changed

AdityaSripal:
^Note the above comments have been addressed in the `User Interaction` sections of the docs

AdityaSripal:
Merging this PR since I believe all concerns have been addressed. Please open an issue if there is anything missing

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,433 @@
...
+(just high-level requirements are needed in this doc).
+
+We require that the [relayer address is exposed to application modules](https://github.com/cosmos/ibc/pull/579) for
+all packet-related messages, so the modules are able to incentivize the packet relayer. `acknowledgePacket`, `timeoutPacket`,
+and `timeoutOnClose` messages will therefore have the relayer address and be capable of sending escrowed tokens to such address.
+However, we need a way to reliably get the address of the relayer that submitted `recvPacket` on the destination chain to
+the source chain. In fact, we need a *source address* for this relayer to pay out to, not the *destination address* that signed
+the packet.
+
+The fee payment mechanism will be implemented as IBC Middleware (see ICS-30) in order to provide maximum flexibility for application developers and blockchains.
```
milosevic, line 67: Can we add a link to ICS-30 here? I was not able to find it.

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,433 @@
...
+}
+```
+
+
+The fee module should also expose the following queries so that relayers may query their expected fee:
+
+```typescript
+// Gets the fee expected for submitting ReceivePacket msg for the given packet
+// Caller should provide the intended relayer address in case the fee is dependent on specific relayer(s).
+function GetReceiveFee(portID, channelID, sequence, relayer) Fee
```
milosevic, line 145: Why we need `sequence` here?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,433 @@
...
+  counterpartyChannelIdentifier: Identifier,
+  version: string,
+  counterpartyVersion: string) {
+      // select mutually compatible fee version
+      cpFeeVersion, cpAppVersion = splitFeeVersion(counterpartyVersion)
+      feeVersion, appVersion = splitFeeVersion(version)
+      if !isCompatible(cpFeeVersion, feeVersion) {
+          return error
+      }
+      selectFeeVersion(cpFeeVersion, feeVersion)
```
milosevic, line 227: What is supposed to happen here?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,433 @@
...
+
+```typescript
+function onChanOpenInit(
+  order: ChannelOrder,
+  connectionHops: [Identifier],
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  counterpartyPortIdentifier: Identifier,
+  counterpartyChannelIdentifier: Identifier,
+  version: string) {
```
milosevic, line 193: I don't understand exactly the implications on app developer/end user here re this change; do they need to know the version of fee middleware?Can someone clarify this a bit?More general, why do we need handshake between handlers if fee protocol does not require the other party to support it?

spec/app/ics-029-fee-payment/README.md
```diff
@@ -0,0 +1,433 @@
...
+    // otherwise return an empty fee version and full version as app version
+    return []string{"", version}
+}
+```
+
+#### Packet Callbacks
+
+```typescript
+function onRecvPacket(packet: Packet, relayer: string): bytes {
+    app_acknowledgement = app.onRecvPacket(packet, relayer)
```
milosevic, line 281: I assume `relayer` here is source address? Why we are passing it to the application?

spec/core/ics-026-routing-module/README.md
```diff
@@ -91,19 +91,19 @@ function onChanCloseConfirm(
     // defined by the module
 }
 
-function onRecvPacket(packet: Packet): bytes {
+function onRecvPacket(packet: Packet, relayer: string): bytes {
```
milosevic, line 94: Where the `relayer` comes from? Do we need to add this field to IBC messages?If yes, are there some properties we expect to hold, i.e., is it a problem if a relayer puts a string that is different from  it's address?

spec/app/ics-030-middleware/README.md
```diff
@@ -0,0 +1,237 @@
...
+    return ics4.writeAcknowledgement(packet, ack_bytes)
+}
+```
+
+```typescript
+function sendPacket(app_packet: Packet) {
+    // middleware may modify packet
+    packet = doCustomLogic(app_packet)
+
+    return ics4.sendPacket(packet)
```
milosevic, line 203: If I am not wrong, we assume that chain of middleware wrappers function as a processing pipeline only upon message reception, but not in other direction (on packet and ack send). Wouldn't it be cleaner if we go through pipeline in both directions?

spec/app/ics-030-middleware/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+NOTE: Middleware that does not need to negotiate with a counterparty middleware on the remote stack will not implement the version splitting and negotiation, and will simply perform its own custom logic on the callbacks without relying on the counterparty behaving similarly.
+
+#### Packet Callbacks
+
+```typescript
+function onRecvPacket(packet: Packet, relayer: string): bytes {
+    doCustomLogic()
+
+    app_acknowledgement = app.onRecvPacket(packet, relayer)
```
milosevic, line 147: Where `app` comes from?I guess that what we mean here by app is actually anything implementing ICS26 interface (`ModuleCallbacks`) as we have potentially several wrappers chained together. If I am not wrong, we might need to make some changes to a router module to be able to support chain of wrappers.

spec/app/ics-029-fee-payment/README.md
AdityaSripal: This is a middleware-specific function to negotiate the fee versions on each end of the channel. Given the counterparty fee version and the fee version, a final version for both modules is selected.

spec/app/ics-029-fee-payment/README.md
AdityaSripal: No they will not, since the middleware will strip out its fee-specific version string before passing onto application. Thus, underlying application does not need to know it is being wrapped by middleware

spec/app/ics-029-fee-payment/README.md
AdityaSripal: The interface is standardized. The Fee middleware is an ICS-26 module, and so is the underlying application. We extended the `OnRecvPacket` to take a relayer to enable applications to implement custom fee logic and to enable fee middleware that implements ICS26

spec/core/ics-026-routing-module/README.md
AdityaSripal: Yes we would have to modify `RecvPacketMsg` to allow relayers to include their source address. If a relayer inputs an invalid address, the forward fee will simply be refunded.

It is incumbent on forward relayer to correctly provide address. If this does not happen, then packet processing will still continue as we do not allow relayers to prevent packet processing

spec/app/ics-030-middleware/README.md
AdityaSripal: This is explaining exactly that concept. On packet/ack receiving; processing goes from ICS-4 to top-layer middleware, through the chain, to base app. On packet/ack sending; processing goes from base application, from lowest to top middleware, and then to ics4.

So the pipeline exists in both directions. The above is an example of middleware that gets called by an underlying base application's `SendPacket` function. Note it is calling ics4, not the base app

spec/app/ics-030-middleware/README.md
AdityaSripal: The app here is the underlying application that implements ICS26. It is embedded by the given middleware (how it is embedded is up to specific languages/implementation. In ibc-go, it would be a field in the middleware struct)

The router module should have no changes. The goal is for ICS-4 and the underlying app to be completely oblivious to the existence of the middleware. This is accomplished on the ICS-4 side, since it is still only talking to some ICS26 app which is bound to a unique port. In the middleware case, it is talking to some top-level middleware but it doesn't matter. It's just an ICS-26 app. The middleware itself will route the call further down the chain.

spec/core/ics-026-routing-module/README.md
AdityaSripal: This relayer is just the message signer (destination address of relayer). The fee middleware holds a mapping from the destination address to relayer source address which it can put in acknowledgement. If they put wrong source address, they don't get paid. Packet processing is not affected

AdityaSripal:
One question I have is if wallets like Keplr and hardware wallets (Ledger) can support signing multiMsg transactions. This is necessary for good UX on user side for incentivizing their packets. cc: @charleenfei

charleenfei:
> One question I have is if wallets like Keplr and hardware wallets (Ledger) can support signing multiMsg transactions. This is necessary for good UX on user side for incentivizing their packets. cc: @charleenfei

both Keplr and Ledger do support multiMsg tx, I am checking in on Rainbow Wallet and Cosmostation.

milosevic (approved):
Thanks a lot for amazing work Aditya!Super clear and clean spec!

PR #582, opened 2021-06-03
Extend acknowledgement type

ethanfrey:
Note: on hold pending #581 discussion. If that is merged, this will likely be closed. If that is rejected, I would integrate all feedback and revive this proposal.

This is discussing changing the `acknowledgement` type from application-dependent bytes to actually care more protocol-level metadata, like the `Packet` type.

So far, I work on defining the type and update some functions to get general feedback.

How a chain / channel / app should handle different types in a backwards compatible way is the huge open question:

* Can we require all applications / modules on a v2 chain to use the new acknowledgement format? Or do we do this on a channel/connection-level basis?
* We must handle the case where a v2 implementation talks with a v1 implementation. How do we handle backwards-compatibility?

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -635,12 +635,16 @@ The IBC handler performs the following steps in order:
 
 We pass the address of the `relayer` that signed and submitted the packet to enable a module to optionally provide some rewards. This provides a foundation for fee payment, but can be used for other techniques as well (like calculating a leaderboard).
 
+Note: `relayer` and `payOnSource` were added in v2 (v1?) of the IBC spec. They will not be present on older implementations.
+`payOnSource` requires another field on the `IBCReceivePacketMsg` which can be set by the relayer to any value it chooses.
+
 ```typescript
 function recvPacket(
   packet: OpaquePacket,
   proof: CommitmentProof,
   proofHeight: Height,
-  relayer: string): Packet {
+  relayer: string,
+  payOnSource: string /* Maybe<string>? */): Packet {
 
     channel = provableStore.get(channelPath(packet.destPort, packet.destChannel))
     abortTransactionUnless(channel !== null)
@@ -712,20 +716,30 @@ The IBC handler performs the following steps in order:
 ```typescript
 function writeAcknowledgement(
   packet: Packet,
-  acknowledgement: bytes): Packet {
+  acknowledgement: Acknowledgement | bytes): Packet {
 
     // cannot already have written the acknowledgement
     abortTransactionUnless(provableStore.get(packetAcknowledgementPath(packet.destPort, packet.destChannel, packet.sequence) === null))
 
+    // we need some way to handle older formats?
+    if (typeof acknowledgement === "bytes") {
+      acknowledgement = Acknowledgement{
+        data: acknowledgement,
+      }
+    }
+
     // write the acknowledgement
+    commitment = hash(makeOpaque(acknowledgement))
     provableStore.set(
       packetAcknowledgementPath(packet.destPort, packet.destChannel, packet.sequence),
-      hash(acknowledgement)
+      commitment
     )
 
     // log that a packet has been acknowledged
-    emitLogEntry("writeAcknowledgement", {sequence: packet.sequence, timeoutHeight: packet.timeoutHeight, port: packet.destPort, channel: packet.destChannel,
-                                timeoutTimestamp: packet.timeoutTimestamp, data: packet.data, acknowledgement})
+    emitLogEntry("writeAcknowledgement", {sequence: packet.sequence, timeoutHeight: packet.timeoutHeight, port: packet.destPort, 
+                  channel: packet.destChannel, timeoutTimestamp: packet.timeoutTimestamp, data: packet.data,
+                  acknowledgement: acknowledgement.data, payOnSource: acknowledgement.payOnSource, forwardRelayer: acknowledgement.forwardRelayer
+                  })
 }
 ```
 
@@ -742,7 +756,7 @@ We pass the `relayer` address just as in [Receiving packets](#receiving-packets)
 ```typescript
 function acknowledgePacket(
   packet: OpaquePacket,
-  acknowledgement: bytes,
+  acknowledgement: bytes | Acknowledgement,
   proof: CommitmentProof,
   proofHeight: Height,
   relayer: string): Packet {
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -124,10 +124,52 @@ interface Packet {
 
 Note that a `Packet` is never directly serialised. Rather it is an intermediary structure used in certain function calls that may need to be created or processed by modules calling the IBC handler.
 
+Question: why is this here? Isn't there a protobuf definition for Packet serialization which must be understood by all
+chains that participate in IBC?
+
 An `OpaquePacket` is a packet, but cloaked in an obscuring data type by the host state machine, such that a module cannot act upon it other than to pass it to the IBC handler. The IBC handler can cast a `Packet` to an `OpaquePacket` and vice versa.
 
 ```typescript
 type OpaquePacket = object
+```
+
+**Update:**
+
+The first version of IBC considered `Acknowledgement` simple application dependent bytes. This stood in contrast to the
+`Packet` types, which includes a number of well-defined metadata fields in addition to an application-dependent
+`data` field, which contains the opaque binary data. While discussing relayer fee incentivization the lack of including
+any protocol-level information in the acknowledgement came up as a limiting factor. Thus, we propose the following format
+for a "V2" acknowledgement. Information on how to transition between the two will be discussed below during the packet
+samples.
+
+```typescript
+interface Acknowledgement {
+  data: bytes
+  payOnSource: string  // Maybe<string> ???
+  forwardRelayer: string
+}
+```
+
+- The `data` is an opaque value which can be defined by the application logic of the associated modules. The contents are
+  identical to a v1 acknowledgement.
+- The `payOnSource` indicates an address on the **source chain** that the *forward relayer* can specify, which should receive
+  any existing reward for relaying the IbcReceivePacket.
+- The `forwardRelayer` field represents the signer of the IbcReceivePacket message. This is always set and equal to the
+  address on the **destination chain** for the *forward relayer*.
+
+Open discussion:
+
+- Is `forwardRelayer` useful info? This information is available without any work by the *forward relayer* and seems
+  more generic than just fee incentivization.
+- Is `payOnSource` required? What happens if a relayer doesn't set this field? Is the packet rejected? What if it is an
+  invalid address on source (this is hard to validate on destination chain)
+
+Note that a `Acknowledgement` is never directly serialized. Rather it is an intermediary structure used in certain function calls that may need to be created or processed by modules calling the IBC handler.
+
+An `OpaqueAcknowledgement` is an acknowledgement, but cloaked in an obscuring data type by the host state machine, such that a module cannot act upon it other than to pass it to the IBC handler. The IBC handler can cast an `Acknowledgement` to an `OpaqueAcknowledgement` and vice versa.
+
+```typescript
+type OpaqueAcknowledgement = object
 ```
 
 ### Desired Properties
@@ -163,7 +205,7 @@ The architecture of clients, connections, channels and packets:
 
 ### Preliminaries
 
-#### Store paths 
+#### Store paths
 
 Channel structures are stored under a store path prefix unique to a combination of a port identifier and channel identifier:
 
@@ -215,7 +257,7 @@ function packetReceiptPath(portIdentifier: Identifier, channelIdentifier: Identi
 }
 ```
 
-Packet acknowledgement data are stored under the `packetAcknowledgementPath`:
+Constant-size commitments to packet acknowledgement data are stored under the `packetAcknowledgementPath`:
 
 ```typescript
 function packetAcknowledgementPath(portIdentifier: Identifier, channelIdentifier: Identifier, sequence: uint64): Path {


ethanfrey:
One idea for backwards compatibility is to do this on a chain-by-chain basis.

* v1 - v1 uses the current acknowledgement bytes both in the application interface and on the wire
* v2 - v2 uses the new acknowledgement object both in the application interface and on the wire
* v1 - v2 must be supported and a bit tricky. This must be detected somewhere (eg. establishing a connection), and we need to use v1 acknowledgements on the wire.

In this last case, the v2 application can return the entire `Acknowledgement` object on `writeAcknowledgement` and the IBC handler (which detects this is sent on a v1 connection) can drop the extra fields and just return the `acknowledgement.data` on the wire.

When they receive an acknowledgement from v1, they can wrap it in an Acknowledgement object before passing to the applications. In this case, all fields but `data` will be unset, but the applications can be shielded from this mainly.

Note: the implications is that limited fees would theoretically be possible on a v2-v1 connection. That is packets sent from the v2 chain to the v1 chain could incentivize ack and timeout, but not receive. Maybe they pay extra on ack and hope that gets picked up by the forward relayer (at least to some decent approximation over many runs)

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -593,12 +635,16 @@ The IBC handler performs the following steps in order:
...
+`payOnSource` requires another field on the `IBCReceivePacketMsg` which can be set by the relayer to any value it chooses.
+
 ```typescript
 function recvPacket(
   packet: OpaquePacket,
   proof: CommitmentProof,
   proofHeight: Height,
-  relayer: string): Packet {
+  relayer: string,
+  payOnSource: string /* Maybe<string>? */): Packet {
```
ethanfrey, line 647: Actual use of it is for another spec like #581

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -670,20 +716,30 @@ The IBC handler performs the following steps in order:
 ```typescript
 function writeAcknowledgement(
   packet: Packet,
-  acknowledgement: bytes): Packet {
+  acknowledgement: Acknowledgement | bytes): Packet {
```
ethanfrey, line 719: Not sure how to write both versions here....

This is a "v2" handler that may be talking on a  "v1" or "v2" connection

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -124,12 +124,54 @@ interface Packet {
...
+`Packet` types, which includes a number of well-defined metadata fields in addition to an application-dependent
+`data` field, which contains the opaque binary data. While discussing relayer fee incentivization the lack of including
+any protocol-level information in the acknowledgement came up as a limiting factor. Thus, we propose the following format
+for a "V2" acknowledgement. Information on how to transition between the two will be discussed below during the packet
+samples.
+
+```typescript
+interface Acknowledgement {
+  data: bytes
+  payOnSource: string  // Maybe<string> ???
```
ethanfrey, line 148: I think this should be optional.
Failure to include it (relayer unaware of fees or doesn't care) shouldn't prevent packet/ack relaying

ethanfrey (commented):
Just a first draft, these were some areas where I am most unclear about

AdityaSripal:
Nice work, I think from a backwards compatibility standpoint. It is easier for the applications to still only create and process opaque bytes as the acknowledgement.

Regardless of what version the chain is on, the modules just return and receive the application-specific acknowledgement bytes.

If the connection version is on V2, then the ICS-4 handlers just wrap/unwrap the acknowledgement bytes in the structure you have and send it over the wire but the application is not aware of this at all.

On write acknowledgement, app send ack bytes and ICS-4 places it in ack struct along with forward relayer address.
During acknowledgePacket, ICS-4 unwraps ack struct and pays forward relayer, it then passes the raw ack bytes to the application.

If the connection version is on V1, then the ICS-4 handlers just commit the acknowledgement bytes directly as it does already.

The only reason I see to bring the structured Acknowledgement all the way to the application is to enable future field extensions that might be used directly by apps but it seems unnecessary for this case.

AdityaSripal:
In order to keep things even easier for relayers to integrate this, we could keep the proto definiton of `Acknowledgement` still have `bytes` in all the relayer messages. It is just required that the bytes unmarshal to the structured Acknowledgement if the connection is `v2`.

In this case, the relayer doesn't even need to send a different type of acknowledgement message. Though this may be too loose of an implementation, and relayers will have to adapt anyway to include forward address.

AdityaSripal:
> On write acknowledgement, app returns ack bytes and ICS-4 places it in ack struct along with forward relayer address.

Note that this requires some storage in core IBC in order to support asynchronous acks

ethanfrey:
> Nice work, I think from a backwards compatibility standpoint. It is easier for the applications to still only create and process opaque bytes as the acknowledgement.
> 
> Regardless of what version the chain is on, the modules just return and receive the application-specific acknowledgement bytes.
> 
> If the connection version is on V2, then the ICS-4 handlers just wrap/unwrap the acknowledgement bytes in the structure you have and send it over the wire but the application is not aware of this at all.
> 
> On write acknowledgement, app returns ack bytes and ICS-4 places it in ack struct along with forward relayer address.
> During acknowledgePacket, ICS-4 unwraps ack struct and pays forward relayer, it then passes the raw ack bytes to the application callback.
> 
> If the connection version is on V1, then the ICS-4 handlers just commit the acknowledgement bytes directly as it does already.
> 
> The only reason I see to bring the structured Acknowledgement all the way to the application is to enable future field extensions that might be used directly by apps but it seems unnecessary for this case.

This seems like a good design.

My biggest issue in this pseudocode is that it seems to be defining the ICS-4 handlers, but doesn't explicit show what it passes into the application-level.

You propose:

* OnReceivePacket - ICS-4 receives all data, stores `payOnSource` (and `forwardRelayer`?)` based on "packet id". It then calls the application without the `payOnSource`
* WriteAcknowledgement - ICS-4 takes `bytes` from the application. It checks if it is v2 and if so, loads the data stored above and constructs a proper `Acknowledgement` struct, which is will proto-encode and then commit to that. If the connection is v1, then it just commits the raw acknowledgement.
  * For better relayer compatibility, we can just return the pre-hashed commitment as an event, regardless if it is raw app data or wrapper. Does ibc-go emit `acknowledgement_hex` event as well (like with packet) for better binary compatibility?
* OnAcknowledge - ICS-4 will check the connection version. If it is v1, then it will pass those raw ack bytes to the application. If it is v2, it will parse the acknowledgement bytes into Acknowledgement message, call the fee handler with ack relayer and `payOnSource` from ack packet, and then pass the `acknowledgement.data` field to the applicaton.

To make this all more clear, it would be nice to define somewhere in this psuedo-code the flow and call from ICS-4 handlers into the application modules. Or is that only done in ibc-go?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -124,12 +124,54 @@ interface Packet {
 
 Note that a `Packet` is never directly serialised. Rather it is an intermediary structure used in certain function calls that may need to be created or processed by modules calling the IBC handler.
 
+Question: why is this here? Isn't there a protobuf definition for Packet serialization which must be understood by all
```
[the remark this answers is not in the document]
cwgoes: No, there is not a protobuf for a `Packet`, because it is never sent over the wire, the fields are implicitly committed to by Merkle path (store) keys.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -124,12 +124,54 @@ interface Packet {
...
 An `OpaquePacket` is a packet, but cloaked in an obscuring data type by the host state machine, such that a module cannot act upon it other than to pass it to the IBC handler. The IBC handler can cast a `Packet` to an `OpaquePacket` and vice versa.
 
 ```typescript
 type OpaquePacket = object
 ```
 
+**Update:**
+
+The first version of IBC considered `Acknowledgement` simple application dependent bytes. This stood in contrast to the
+`Packet` types, which includes a number of well-defined metadata fields in addition to an application-dependent
```
cwgoes, line 139: No, these work the same way. The _contents_ of a packet may be well-defined, e.g. by ICS 20, but not by core IBC.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -124,12 +124,54 @@ interface Packet {
...
+
+```typescript
+interface Acknowledgement {
+  data: bytes
+  payOnSource: string  // Maybe<string> ???
+  forwardRelayer: string
+}
+```
+
+- The `data` is an opaque value which can be defined by the application logic of the associated modules. The contents are
```
cwgoes, line 153: Is it still committed to (by a hash function)? Or is the whole `Acknowledgement` structure now hash-committed-to?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -124,12 +124,54 @@ interface Packet {
...
+  identical to a v1 acknowledgement.
+- The `payOnSource` indicates an address on the **source chain** that the *forward relayer* can specify, which should receive
+  any existing reward for relaying the IbcReceivePacket.
+- The `forwardRelayer` field represents the signer of the IbcReceivePacket message. This is always set and equal to the
+  address on the **destination chain** for the *forward relayer*.
+
+Open discussion:
+
+- Is `forwardRelayer` useful info? This information is available without any work by the *forward relayer* and seems
```
cwgoes, line 162: Seems useful to me, as long as it's not too expensive to store.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -124,12 +124,54 @@ interface Packet {
...
+  data: bytes
+  payOnSource: string  // Maybe<string> ???
+  forwardRelayer: string
+}
+```
+
+- The `data` is an opaque value which can be defined by the application logic of the associated modules. The contents are
+  identical to a v1 acknowledgement.
+- The `payOnSource` indicates an address on the **source chain** that the *forward relayer* can specify, which should receive
+  any existing reward for relaying the IbcReceivePacket.
```
cwgoes, line 156: What's an existing reward - a fee that has already been pre-paid on the source chain?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -124,12 +124,54 @@ interface Packet {
...
+  address on the **destination chain** for the *forward relayer*.
+
+Open discussion:
+
+- Is `forwardRelayer` useful info? This information is available without any work by the *forward relayer* and seems
+  more generic than just fee incentivization.
+- Is `payOnSource` required? What happens if a relayer doesn't set this field? Is the packet rejected? What if it is an
+  invalid address on source (this is hard to validate on destination chain)
+
+Note that a `Acknowledgement` is never directly serialized. Rather it is an intermediary structure used in certain function calls that may need to be created or processed by modules calling the IBC handler.
```
cwgoes, line 167: Ah, similarly to `Packet`, then?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -670,20 +716,30 @@ The IBC handler performs the following steps in order:
...
 
+    // we need some way to handle older formats?
+    if (typeof acknowledgement === "bytes") {
+      acknowledgement = Acknowledgement{
+        data: acknowledgement,
+      }
+    }
+
     // write the acknowledgement
+    commitment = hash(makeOpaque(acknowledgement))
```
cwgoes, line 732: What is `makeOpaque` doing?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -670,20 +716,30 @@ The IBC handler performs the following steps in order:
...
 function writeAcknowledgement(
   packet: Packet,
-  acknowledgement: bytes): Packet {
+  acknowledgement: Acknowledgement | bytes): Packet {
 
     // cannot already have written the acknowledgement
     abortTransactionUnless(provableStore.get(packetAcknowledgementPath(packet.destPort, packet.destChannel, packet.sequence) === null))
 
+    // we need some way to handle older formats?
+    if (typeof acknowledgement === "bytes") {
```
cwgoes, line 725: This is pseudocode, we can't actually do this in Go.

cwgoes (changes requested):
Thanks! See comments.

spec/core/ics-004-channel-and-packet-semantics/README.md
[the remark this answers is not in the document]
ethanfrey: Thank you for the review. I will respond completely on Monday, but one question Chris. 

Please correct me if I am wrong here. In my understanding, when a Packet is passed to the ics-4 handler to send. That will:

- emit events that correspond to each field of the packet
  - one of those fields is app- specific data. The rest what I call "meta data" (defined by core protocol not app packet)
- The packet is serialized as protobuf and then hashed
- This hash is committed to in the Merkle store

In order to relay the packet, the relayer must parse the fields from the events and reform them into a protobuf message, which serialized to a preimage of the commitment on the source chain.

Is this correct? I am pretty sure there is a protobuf encoding used for the proofs/hashes and that is what I was referring to.

I want to do a similar thing with Acknowledgement as you do with Packet now. As close as I can do without breaking something. But mainly I need to understand better what the process is for Packet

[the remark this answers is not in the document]
cwgoes: > The packet is serialized as protobuf and then hashed

No, this does not happen. The only thing that is stored is the commitment to the packet data, timeout height, and timeout timestamp. See [here](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#sending-packets) in the specification and [here](https://github.com/cosmos/ibc-go/blob/1ff0da6694d08e494a9c87d5b9737ab5312677a0/modules/core/04-channel/types/packet.go#L18) in the code. Perhaps we should consider changing this, but in any case we _do not_ need to serialise other fields (e.g. concerning channel source and destination) because they are implicitly committed to by the Merkle store keys.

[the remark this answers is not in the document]
ethanfrey: Thank you for that clarification. Very useful for me.

I feel that [hash function you linked](https://github.com/cosmos/ibc-go/blob/1ff0da6694d08e494a9c87d5b9737ab5312677a0/modules/core/04-channel/types/packet.go#L18):

`sha256(bigEndian(timeoutTimestamp) || bigEndian(timeoutRevNumber)  || bigEndian(timeoutRevHeight) || sha256(data))`

needs to be in the IBC spec, as it is required for a compatible implementation. At least it would clear a few points up for me.

It would also help me understand this better if the actual flow of commitments and reconstructing the proofs (eg. IBC from PoV of the relayer) was documented in another section. I feel this is critical use case and important for correctness/liveness arguments as well.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -124,12 +124,54 @@ interface Packet {
...
+  identical to a v1 acknowledgement.
+- The `payOnSource` indicates an address on the **source chain** that the *forward relayer* can specify, which should receive
+  any existing reward for relaying the IbcReceivePacket.
+- The `forwardRelayer` field represents the signer of the IbcReceivePacket message. This is always set and equal to the
+  address on the **destination chain** for the *forward relayer*.
+
+Open discussion:
+
+- Is `forwardRelayer` useful info? This information is available without any work by the *forward relayer* and seems
+  more generic than just fee incentivization.
```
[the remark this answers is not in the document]
ethanfrey: > What is forwardRelayer meant for?

I threw it in as the info was there, and I thought it might be useful (future proofing).
Then again, it may be totally useless and premature optimization.

Happy not to include it as well.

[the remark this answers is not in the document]
ethanfrey: >  Can the relayer query/ decode the ack bytes and verify that the payOnSource is in fact its address (hasn't been altered)?

This info must be exposed in the events, so the relayer can provide this info on the remote side for proving.
I was a bit confused by commitments, so I haven't finalized the hashes and commitments below

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -593,12 +635,16 @@ The IBC handler performs the following steps in order:
 
 We pass the address of the `relayer` that signed and submitted the packet to enable a module to optionally provide some rewards. This provides a foundation for fee payment, but can be used for other techniques as well (like calculating a leaderboard).
 
+Note: `relayer` and `payOnSource` were added in v2 (v1?) of the IBC spec. They will not be present on older implementations.
```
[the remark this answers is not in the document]
ethanfrey: > The relayer was already in v1

I am a bit confused by version numbers.
Cosmos SDK 0.42 has no `relayer`
Cosmos SDK 0.43 / IBC-Go 1.0 has `relayer`
Proposed changes have `payOnSource` also.

I take your point that v1 channels can support `payOnSource` as well, not sure how to explain these were added later on.

ethanfrey:
I would deprecate this for the new ics-30 middleware solution being developed in #581 

If that middleware solution is not followed, I would be happy to update this spec with all comments. But the 2 are mutually exclusive in the current design.

mpoke:
@ethanfrey Is this PR still relevant?

ethanfrey:
Maybe.

I think a discussion of actually providing structure to the acknowledgement would be a good thing.

There is no concrete action item here

PR #587, opened 2021-07-12
use defined terms in relayer incentive spec

colin-axner:
It seemed to me "return relayer" is synonymous with "reverse relayer" and likely came from an old commit. I think it is most clear to use the defined terminology


cwgoes (approved):
ACK; makes sense to me

AdityaSripal (approved):

PR #588, opened 2021-07-12
Improve ics-023 spec

robert-zaremba:
Adding link to data definitions (defined as code - the only thing we have). 
The ICS-23 spec presents only high level requirements, but not a technical requirements (structure definitions, serialization, etc...).
ref: https://github.com/cosmos/cosmos-sdk/discussions/9331#discussioncomment-993092

### DIFF
--- a/spec/core/ics-023-vector-commitments/README.md
+++ b/spec/core/ics-023-vector-commitments/README.md
@@ -38,7 +38,7 @@ This document only defines desired properties, not a concrete implementation —
 
 ## Technical Specification
 
-Below we define a behaviour and an overview of datatypes. For data type definition look at [confio/ics23](https://github.com/confio/ics23/blob/master/go/ics23.go) repository.
+Below we define a behaviour and an overview of datatypes. For data type definition look at [confio/ics23](https://github.com/confio/ics23/blob/master/proofs.proto) repository.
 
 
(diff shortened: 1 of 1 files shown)


spec/core/ics-023-vector-commitments/README.md
```diff
@@ -38,6 +38,9 @@ This document only defines desired properties, not a concrete implementation —
 
 ## Technical Specification
 
+Below we define a behaviour and an overview of datatypes. For data type definition look at [confio/ics23](https://github.com/confio/ics23/blob/master/go/ics23.go) repository.
```
ethanfrey, line 41: Please link to the actual protobuf definitions, which are the data spec, not the Go code, which is derived from them (as is the Rust and TS code)

https://github.com/confio/ics23/blob/master/proofs.proto

Otherwise, glad to see this link

ethanfrey (commented):
Glad to see this link. Also would be happy for doc PRs on ics23 repo

spec/core/ics-023-vector-commitments/README.md
robert-zaremba: done

cwgoes (approved):
ACK

AdityaSripal (approved):

PR #590, opened 2021-07-20
fix: updating contract interface params

seantking:
Updating the `EscrowPacketFee` / `PayFee` / `PayTimeoutFee` to take a `string packetId` instead of a `Packet` type as parameter.


ethanfrey (commented):
The idea looks good to me and glad to require less in the API. However, packetId must be precisely defined (is it `${packet.portId}/${packet.channelId}/${packet.sequence}`?)

seantking:
Thanks for the review :+1: Agree it should be defined. 

@AdityaSripal has the idea to use a struct. See here: https://github.com/cosmos/ibc-go/pull/272/files#diff-fc6caa5ba0a543336f3c02d081f09660a36e55412cca742e619e94ba959bf4c5R136

wdyt?

cwgoes:
Same comment, what precisely is this packet identifier? A packet is identified by the combination of port, channel, and sequence.

PR #598, opened 2021-08-13
Add Markdown link checker to ensure link integrity

thanethomson:
Adds the same link checker that's used in Tendermint. See [here](https://github.com/tendermint/tendermint/blob/master/.github/workflows/linkchecker.yml).


cwgoes (approved):
AKC

AdityaSripal (approved):

PR #599, opened 2021-08-23
increase port identifier limit to 128 characters

### LINKED ISSUE #594: Increase port identifier limit
ICS 24 specifies the port identifier as being between 2-64 characters. I believe these limits are somewhat arbitrary. Interchain Accounts is including an address in the portID along with a prefix and 2 connection identifier sequences. These sequences have a maximum values of 20 characters. While unlikely that either sequence would get above length 5-7 anytime soon, the current length is cutting it close

This doesn't account for middleware. If interchain accounts wanted to be middleware or use relayer incentivization, it'd need to wrap it's portID with another string such as "ics29". If interchain accounts attaches to 2-3 middleware modules, this could be problematic

I propose we increase the upper limit to 128 characters. This is still arbitrary, but seems like a reasonable next step

colin-axner:
closes #594

### DIFF
--- a/spec/core/ics-024-host-requirements/README.md
+++ b/spec/core/ics-024-host-requirements/README.md
@@ -58,7 +58,7 @@ By default, identifiers have the following minimum and maximum lengths in charac
 
 | Port identifier | Client identifier | Connection identifier | Channel identifier |
 | --------------- | ----------------- | --------------------- | ------------------ |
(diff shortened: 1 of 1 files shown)


ethanfrey (approved):
yay!

cwgoes (approved):
ACK

AdityaSripal (approved):

PR #601, opened 2021-08-27
Fix Interchain Account bug allowing controller account to use host port

colin-axner:
While implementing the interchain accounts spec, we realized there was no check preventing a controller account from [accidentally](https://github.com/cosmos/ibc-go/pull/355#discussion_r697471820) using the port representing the chain as a host chain

I believe this check is necessary in preventing a chain from allowing the port reserved for host chain actions, to be used


mpoke:
@colin-axner I guess this PR is deprecated, right? It seems that the issue is fixed by the `validateControllerPortParams()` method, i.e.,  https://github.com/cosmos/ibc/blob/master/spec/app/ics-027-interchain-accounts/README.md#channel-lifecycle-management

colin-axner:
yup!

PR #615, opened 2021-11-10
ICS721: Spec for interchain NFT transfer

haifengxi:
Adapted from ICS-020, replacing `bank` with `nft` as the "asset tracking module"

### DIFF
--- a/spec/app/ics-021-nft-transfer/README.md
+++ b/spec/app/ics-021-nft-transfer/README.md
@@ -6,9 +6,11 @@ category: IBC/APP
 requires: 25, 26
 kind: instantiation
 author: Christopher Goes <cwgoes@interchain.berlin>, Haifeng Xi <haifeng@bianjie.ai>
-created: 2021-11-10 
-modified: 2021-11-10
+created: 2021-11-10
+modified: 2021-11-12
 ---
+
+> This spec follows the same design principles of [ICS 20](../ics-020-fungible-token-transfer) and copies most of its content therefrom, while replacing `bank` module based asset tracking logic with that of the `nft` module.
 
 ## Synopsis
 
@@ -47,7 +49,7 @@ interface NonFungibleTokenPacketData {
 }
 ```
 
-As tokens are sent across chains using the ICS 21 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `classId` field. 
+As tokens are sent across chains using the ICS 21 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `classId` field.
 
 The ics21 token classes are represented in the form `{ics21Port}/{ics21Channel}/{classId}`, where `ics21Port` and `ics21Channel` are an ics21 port and channel on the current chain for which the token exists. The prefixed port and channel pair indicate which channel the token was previously sent through. If `{classId}` contains `/`, then it must also be in the ics21 form which indicates that this token has a multi-hop record. Note that this requires that the `/` (slash character) is prohibited in non-IBC token denomination names.
 
@@ -330,7 +332,7 @@ In order to track all of the tokens moving around the network of chains in vario
 
 #### Optional addenda
 
-- Each chain, locally, could elect to keep a lookup table to use short, user-friendly local `classId`s in state which are translated to and from the longer `classId`s when sending and receiving packets. 
+- Each chain, locally, could elect to keep a lookup table to use short, user-friendly local `classId`s in state which are translated to and from the longer `classId`s when sending and receiving packets.
 - Additional restrictions may be imposed on which other machines may be connected to & which channels may be established.
 
 ## Backwards Compatibility
--- a/spec/app/ics-021-nft-transfer/README.md
+++ b/spec/app/ics-021-nft-transfer/README.md
@@ -1,0 +1,360 @@
+---
+ics: 21
+title: Non-Fungible Token Transfer
+stage: draft
+category: IBC/APP
+requires: 25, 26
+kind: instantiation
+author: Christopher Goes <cwgoes@interchain.berlin>, Haifeng Xi <haifeng@bianjie.ai>
+created: 2021-11-10 
+modified: 2021-11-10
+---
+
+## Synopsis
+
+This standard document specifies packet data structure, state machine handling logic, and encoding details for the transfer of non-fungible tokens over an IBC channel between two modules on separate chains. The state machine logic presented allows for safe multi-chain denomination handling with permissionless channel opening. This logic constitutes a "non-fungible token transfer bridge module", interfacing between the IBC routing module and an existing asset tracking module on the host state machine.
+
+### Motivation
+
+Users of a set of chains connected over the IBC protocol might wish to utilise an asset issued on one chain on another chain, perhaps to make use of additional features such as exchange or privacy protection, while retaining non-fungibility with the original asset on the issuing chain. This application-layer standard describes a protocol for transferring non-fungible tokens between chains connected with IBC which preserves asset non-fungibility, preserves asset ownership, limits the impact of Byzantine faults, and requires no additional permissioning.
+
+### Definitions
+
+The IBC handler interface & IBC routing module interface are as defined in [ICS 25](../../core/ics-025-handler-interface) and [ICS 26](../../core/ics-026-routing-module), respectively.
+
+### Desired Properties
+
+- Preservation of uniqueness (two-way peg).
+- Preservation of total supply (maintained on a single source chain & module).
+- Permissionless token transfers, no need to whitelist connections, modules, or denominations.
+- Symmetric (all chains implement the same logic, no in-protocol differentiation of hubs & zones).
+- Fault containment: prevents Byzantine-inflation of tokens originating on chain `A`, as a result of chain `B`'s Byzantine behaviour (though any users who sent tokens to chain `B` may be at risk).
+
+## Technical Specification
+
+### Data Structures
+
+Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the denomination, amount, sending account, and receiving account.
+
+```typescript
+interface NonFungibleTokenPacketData {
+  classId: string
+  classUri: string
+  tokenId: string
+  tokenUri: string
+  sender: string
+  receiver: string
+}
+```
+
+As tokens are sent across chains using the ICS 21 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `classId` field. 
+
+The ics21 token classes are represented in the form `{ics21Port}/{ics21Channel}/{classId}`, where `ics21Port` and `ics21Channel` are an ics21 port and channel on the current chain for which the token exists. The prefixed port and channel pair indicate which channel the token was previously sent through. If `{classId}` contains `/`, then it must also be in the ics21 form which indicates that this token has a multi-hop record. Note that this requires that the `/` (slash character) is prohibited in non-IBC token denomination names.
+
+A sending chain may be acting as a source or sink zone. When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone. When tokens are sent from a source zone, the destination port and channel will be prefixed onto the `classId` (once the tokens are received) adding another hop to a tokens record. When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone. When tokens are sent from a sink zone, the last prefixed port and channel pair on the `classId` is removed (once the tokens are received), undoing the last hop in the tokens record. A more complete explanation is present in the ibc-go implementation (TBD).
+
+The acknowledgement data type describes whether the transfer succeeded or failed, and the reason for failure (if any).
+
+```typescript
+type NonFungibleTokenPacketAcknowledgement = NonFungibleTokenPacketSuccess | NonFungibleTokenPacketError;
+
+interface NonFungibleTokenPacketSuccess {
+  // This is binary 0x01 base64 encoded
+  success: "AQ=="
+}
+
+interface NonFungibleTokenPacketError {
+  error: string
+}
+```
+
+Note that both the `NonFungibleTokenPacketData` as well as `NonFungibleTokenPacketAcknowledgement` must be JSON-encoded (not Protobuf encoded) when they serialized into packet data.
+
+The non-fungible token transfer bridge module tracks escrow addresses associated with particular channels in state. Fields of the `ModuleState` are assumed to be in scope.
+
+```typescript
+interface ModuleState {
+  channelEscrowAddresses: Map<Identifier, string>
+}
+```
+
+### Sub-protocols
+
+The sub-protocols described herein should be implemented in a "non-fungible token transfer bridge" module with access to the IBC routing module.
+
+#### Port & channel setup
+
+The `setup` function must be called exactly once when the module is created (perhaps when the blockchain itself is initialised) to bind to the appropriate port and create an escrow address (owned by the module).
+
+```typescript
+function setup() {
+  capability = routingModule.bindPort("nft", ModuleCallbacks{
+    onChanOpenInit,
+    onChanOpenTry,
+    onChanOpenAck,
+    onChanOpenConfirm,
+    onChanCloseInit,
+    onChanCloseConfirm,
+    onRecvPacket,
+    onTimeoutPacket,
+    onAcknowledgePacket,
+    onTimeoutPacketClose
+  })
+  claimCapability("port", capability)
+}
+```
+
+Once the `setup` function has been called, channels can be created through the IBC routing module between instances of the non-fungible token transfer module on separate chains.
+
+An administrator (with the permissions to create connections & channels on the host state machine) is responsible for setting up connections to other state machines & creating channels
+to other instances of this module (or another module supporting this interface) on other chains. This specification defines packet handling semantics only, and defines them in such a fashion
+that the module itself doesn't need to worry about what connections or channels might or might not exist at any point in time.
+
+#### Routing module callbacks
+
+##### Channel lifecycle management
+
+Both machines `A` and `B` accept new channels from any module on another machine, if and only if:
+
+- The channel being created is unordered.
+- The version string is `ics21-1`.
+
+```typescript
+function onChanOpenInit(
+  order: ChannelOrder,
+  connectionHops: [Identifier],
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  counterpartyPortIdentifier: Identifier,
+  counterpartyChannelIdentifier: Identifier,
+  version: string) {
+  // only unordered channels allowed
+  abortTransactionUnless(order === UNORDERED)
+  // assert that version is "ics21-1"
+  abortTransactionUnless(version === "ics21-1")
+  // allocate an escrow address
+  channelEscrowAddresses[channelIdentifier] = newAddress()
+}
+```
+
+```typescript
+function onChanOpenTry(
+  order: ChannelOrder,
+  connectionHops: [Identifier],
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  counterpartyPortIdentifier: Identifier,
+  counterpartyChannelIdentifier: Identifier,
+  version: string,
+  counterpartyVersion: string) {
+  // only unordered channels allowed
+  abortTransactionUnless(order === UNORDERED)
+  // assert that version is "ics21-1"
+  abortTransactionUnless(version === "ics21-1")
+  abortTransactionUnless(counterpartyVersion === "ics21-1")
+  // allocate an escrow address
+  channelEscrowAddresses[channelIdentifier] = newAddress()
+}
+```
+
+```typescript
+function onChanOpenAck(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  version: string) {
+  // port has already been validated
+  // assert that version is "ics21-1"
+  abortTransactionUnless(version === "ics21-1")
+}
+```
+
+```typescript
+function onChanOpenConfirm(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier) {
+  // accept channel confirmations, port has already been validated, version has already been validated
+}
+```
+
+```typescript
+function onChanCloseInit(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier) {
+  // no action necessary
+}
+```
+
+```typescript
+function onChanCloseConfirm(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier) {
+  // no action necessary
+}
+```
+
+##### Packet relay
+
+In plain English, between chains `A` and `B`:
+
+- When acting as the source zone, the bridge module escrows an existing local non-fungible token on the sending chain and mints a corresponding voucher on the receiving chain.
+- When acting as the sink zone, the bridge module burns the local voucher on the sending chain and unescrows the local non-fungible token on the receiving chain.
+- When a packet times-out, local non-fungible tokens are unescrowed back to the sender or vouchers minted back to the sender appropriately.
+- Acknowledgement data is used to handle failures, such as invalid destination accounts. Returning
+  an acknowledgement of failure is preferable to aborting the transaction since it more easily enables the sending chain to take appropriate action based on the nature of the failure.
+
+`createOutgoingPacket` must be called by a transaction handler in the module which performs appropriate signature checks, specific to the account owner on the host state machine.
+
+```typescript
+function createOutgoingPacket(
+  classId: string,
+  tokenId: string,
+  sender: string,
+  receiver: string,
+  source: boolean,
+  destPort: string,
+  destChannel: string,
+  sourcePort: string,
+  sourceChannel: string,
+  timeoutHeight: Height,
+  timeoutTimestamp: uint64) {
+  prefix = "{sourcePort}/{sourceChannel}/"
+  // we are the source if the denomination is not prefixed
+  source = classId.slice(0, len(prefix)) !== prefix
+  if source {
+    // determine escrow account
+    escrowAccount = channelEscrowAddresses[sourceChannel]
+    // escrow source token (assumed to fail if sender is not owner)
+    nft.Transfer(sender, escrowAccount, classId, tokenId)
+  } else {
+    // receiver is source chain, burn voucher (assumed to fail if sender is not owner)
+    bank.Burn(sender, classId, tokenId)
+  }
+  Class class = nft.getClass(classId)
+  NFT token = nft.getNFT(classId, tokenId)
+  NonFungibleTokenPacketData data = NonFungibleTokenPacketData{classId, class.getUri(), tokenId, token.getUri(), sender, receiver}
+  handler.sendPacket(Packet{timeoutHeight, timeoutTimestamp, destPort, destChannel, sourcePort, sourceChannel, data}, getCapability("port"))
+}
+```
+
+`onRecvPacket` is called by the routing module when a packet addressed to this module has been received.
+
+```typescript
+function onRecvPacket(packet: Packet) {
+  NonFungibleTokenPacketData data = packet.data
+  // construct default acknowledgement of success
+  NonFungibleTokenPacketAcknowledgement ack = NonFungibleTokenPacketAcknowledgement{true, null}
+  prefix = "{packet.sourcePort}/{packet.sourceChannel}/"
+  // we are the source if the packets were prefixed by the sending chain
+  source = data.classId.slice(0, len(prefix)) === prefix
+  if source {
+    // receiver is source chain: unescrow token
+    // determine escrow account
+    escrowAccount = channelEscrowAddresses[packet.destChannel]
+    // unescrow token to receiver
+    err = nft.Transfer(escrowAccount, data.receiver, data.classId.slice(len(prefix)), data.tokenId)
+    if (err !== nil)
+      ack = NonFungibleTokenPacketAcknowledgement{false, "transfer nft failed"}
+  } else {
+    prefix = "{packet.destPort}/{packet.destChannel}/"
+    prefixedClassId = prefix + data.classId
+    // sender was source, mint voucher to receiver
+    err = nft.Mint(data.receiver, prefixedClassId, data.tokenId)
+    if (err !== nil)
+      ack = NonFungibleTokenPacketAcknowledgement{false, "mint nft failed"}
+  }
+  return ack
+}
+```
+
+`onAcknowledgePacket` is called by the routing module when a packet sent by this module has been acknowledged.
+
+```typescript
+function onAcknowledgePacket(
+  packet: Packet,
+  acknowledgement: bytes) {
+  // if the transfer failed, refund the tokens
+  if (!ack.success)
+    refundToken(packet)
+}
+```
+
+`onTimeoutPacket` is called by the routing module when a packet sent by this module has timed-out (such that it will not be received on the destination chain).
+
+```typescript
+function onTimeoutPacket(packet: Packet) {
+  // the packet timed-out, so refund the tokens
+  refundToken(packet)
+}
+```
+
+`refundToken` is called by both `onAcknowledgePacket`, on failure, and `onTimeoutPacket`, to refund escrowed token to the original sender.
+
+```typescript
+function refundToken(packet: Packet) {
+  NonFungibleTokenPacketData data = packet.data
+  prefix = "{packet.sourcePort}/{packet.sourceChannel}/"
+  // we are the source if the classId is not prefixed
+  source = data.classId.slice(0, len(prefix)) !== prefix
+  if source {
+    // sender was source chain, unescrow tokens back to sender
+    escrowAccount = channelEscrowAddresses[packet.srcChannel]
+    nft.Transfer(escrowAccount, data.sender, data.classId, data.tokenId)
+  } else {
+    // receiver was source chain, mint voucher back to sender
+    bank.Mint(data.sender, data.classId, data.tokenId)
+  }
+}
+```
+
+```typescript
+function onTimeoutPacketClose(packet: Packet) {
+  // can't happen, only unordered channels allowed
+}
+```
+
+#### Reasoning
+
+##### Correctness
+
+This implementation preserves both uniqueness & supply.
+
+Uniqueness: If tokens have been sent to the counterparty chain, they can be redeemed back in the same `classId` & `tokenId` on the source chain.
+
+Supply: Redefine supply as unlocked tokens. All send-recv pairs for any given token class sum to net zero. Source chain can change supply.
+
+##### Multi-chain notes
+
+This specification does not directly handle the "diamond problem", where a user sends a token originating on chain A to chain B, then to chain D, and wants to return it through D -> C -> A — since the supply is tracked as owned by chain B (and the `classId` will be "{portOnD}/{channelOnD}/{portOnB}/{channelOnB}/classId"), chain C cannot serve as the intermediary. It is not yet clear whether that case should be dealt with in-protocol or not — it may be fine to just require the original path of redemption (and if there is frequent liquidity and some surplus on both paths the diamond path will work most of the time). Complexities arising from long redemption paths may lead to the emergence of central chains in the network topology.
+
+In order to track all of the tokens moving around the network of chains in various paths, it may be helpful for a particular chain to implement a registry which will track the "global" source chain for each `classId`. End-user service providers (such as wallet authors) may want to integrate such a registry or keep their own mapping of canonical source chains and human-readable names in order to improve UX.
+
+#### Optional addenda
+
+- Each chain, locally, could elect to keep a lookup table to use short, user-friendly local `classId`s in state which are translated to and from the longer `classId`s when sending and receiving packets. 
+- Additional restrictions may be imposed on which other machines may be connected to & which channels may be established.
+
+## Backwards Compatibility
+
+Not applicable.
+
+## Forwards Compatibility
+
+This initial standard uses version "ics21-1" in the channel handshake.
+
+A future version of this standard could use a different version in the channel handshake, and safely alter the packet data format & packet handler semantics.
+
+## Example Implementation
+
+Coming soon.
+
+## Other Implementations
+
+Coming soon.
+
+## History
+
+Nov 10, 2021 - Initial draft adapted from ICS20 spec
+
+## Copyright
+
+All content herein is licensed under [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0).
--- a/spec/app/ics-021-nft-transfer/README.md
+++ b/spec/app/ics-021-nft-transfer/README.md
@@ -14,7 +14,7 @@ modified: 2021-11-12
 
 ## Synopsis
 
-This standard document specifies packet data structure, state machine handling logic, and encoding details for the transfer of non-fungible tokens over an IBC channel between two modules on separate chains. The state machine logic presented allows for safe multi-chain `classId` handling with permissionless channel opening. This logic constitutes a "non-fungible token transfer bridge module", interfacing between the IBC routing module and an existing asset tracking module on the host state machine.
+This standard document specifies packet data structure, state machine handling logic, and encoding details for the transfer of non-fungible tokens over an IBC channel between two modules on separate chains. The state machine logic presented allows for safe multi-chain `classId` handling with permissionless channel opening. This logic constitutes a "non-fungible token transfer bridge module", interfacing between the IBC routing module and an existing asset tracking module on the host state machine, which could be either a Cosmos-style "native" module or a smart contract running in a virtual machine.
 
 ### Motivation
 
@@ -48,6 +48,13 @@ interface NonFungibleTokenPacketData {
   receiver: string
 }
 ```
+`classId` uniquely identifies the class/collection to which this NFT belongs in the originating host environment.  In the case of an ERC-1155 compliant smart contract, for example, this could be a string representation of the top 128 bits of the token ID.
+
+`classUri` is optional, but will be extremely beneficial for cross-chain interoperability with NFT marketplaces like OpenSea, where [class/collection metadata](https://docs.opensea.io/docs/contract-level-metadata) can be added for better user experience.
+
+`tokenId` uniquely identifies the NFT within the given class.  In the case of an ERC-1155 compliant smart contract, for example, this could be a string representation of the bottom 128 bits of the token ID.
+
+`tokenUri` refers to an off-chain resource, typically an immutable JSON file containing the NFT's metadata.
 
 As tokens are sent across chains using the ICS 21 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `classId` field.
 
--- a/spec/app/ics-021-nft-transfer/README.md
+++ b/spec/app/ics-021-nft-transfer/README.md
@@ -7,7 +7,7 @@ requires: 25, 26
 kind: instantiation
 author: Christopher Goes <cwgoes@interchain.berlin>, Haifeng Xi <haifeng@bianjie.ai>
 created: 2021-11-10
-modified: 2021-11-12
+modified: 2021-11-17
 ---
 
 > This spec follows the same design principles of [ICS 20](../ics-020-fungible-token-transfer) and copies most of its content therefrom, while replacing `bank` module based asset tracking logic with that of the `nft` module.
@@ -367,6 +367,7 @@ Coming soon.
 ## History
 
 Nov 10, 2021 - Initial draft adapted from ICS20 spec
+Nov 17, 2021 - Revisions to better accommodate smart contracts
 
 ## Copyright
 
--- a/spec/app/ics-021-nft-transfer/README.md
+++ b/spec/app/ics-021-nft-transfer/README.md
@@ -14,11 +14,11 @@ modified: 2021-11-12
 
 ## Synopsis
 
-This standard document specifies packet data structure, state machine handling logic, and encoding details for the transfer of non-fungible tokens over an IBC channel between two modules on separate chains. The state machine logic presented allows for safe multi-chain denomination handling with permissionless channel opening. This logic constitutes a "non-fungible token transfer bridge module", interfacing between the IBC routing module and an existing asset tracking module on the host state machine.
+This standard document specifies packet data structure, state machine handling logic, and encoding details for the transfer of non-fungible tokens over an IBC channel between two modules on separate chains. The state machine logic presented allows for safe multi-chain `classId` handling with permissionless channel opening. This logic constitutes a "non-fungible token transfer bridge module", interfacing between the IBC routing module and an existing asset tracking module on the host state machine.
 
 ### Motivation
 
-Users of a set of chains connected over the IBC protocol might wish to utilise an asset issued on one chain on another chain, perhaps to make use of additional features such as exchange or privacy protection, while retaining non-fungibility with the original asset on the issuing chain. This application-layer standard describes a protocol for transferring non-fungible tokens between chains connected with IBC which preserves asset non-fungibility, preserves asset ownership, limits the impact of Byzantine faults, and requires no additional permissioning.
+Users of a set of chains connected over the IBC protocol might wish to utilise an asset issued on one chain on another chain, perhaps to make use of additional features such as exchange or privacy protection, while retaining uniqueness with the original asset on the issuing chain. This application-layer standard describes a protocol for transferring non-fungible tokens between chains connected with IBC which preserves asset uniqueness, preserves asset ownership, limits the impact of Byzantine faults, and requires no additional permissioning.
 
 ### Definitions
 
@@ -28,7 +28,7 @@ The IBC handler interface & IBC routing module interface are as defined in [ICS 
 
 - Preservation of uniqueness (two-way peg).
 - Preservation of total supply (maintained on a single source chain & module).
-- Permissionless token transfers, no need to whitelist connections, modules, or denominations.
+- Permissionless token transfers, no need to whitelist connections, modules, or `classId`s.
 - Symmetric (all chains implement the same logic, no in-protocol differentiation of hubs & zones).
 - Fault containment: prevents Byzantine-inflation of tokens originating on chain `A`, as a result of chain `B`'s Byzantine behaviour (though any users who sent tokens to chain `B` may be at risk).
 
@@ -36,7 +36,7 @@ The IBC handler interface & IBC routing module interface are as defined in [ICS 
 
 ### Data Structures
 
-Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the denomination, amount, sending account, and receiving account.
+Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the class id, class uri, token id, token uri, sending account, and receiving account.
 
 ```typescript
 interface NonFungibleTokenPacketData {
@@ -51,9 +51,9 @@ interface NonFungibleTokenPacketData {
 
 As tokens are sent across chains using the ICS 21 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `classId` field.
 
-The ics21 token classes are represented in the form `{ics21Port}/{ics21Channel}/{classId}`, where `ics21Port` and `ics21Channel` are an ics21 port and channel on the current chain for which the token exists. The prefixed port and channel pair indicate which channel the token was previously sent through. If `{classId}` contains `/`, then it must also be in the ics21 form which indicates that this token has a multi-hop record. Note that this requires that the `/` (slash character) is prohibited in non-IBC token denomination names.
+The ics21 token classes are represented in the form `{ics21Port}/{ics21Channel}/{classId}`, where `ics21Port` and `ics21Channel` are an ics21 port and channel on the current chain for which the token exists. The prefixed port and channel pair indicate which channel the token was previously sent through. If `{classId}` contains `/`, then it must also be in the ics21 form which indicates that this token has a multi-hop record. Note that this requires that the `/` (slash character) is prohibited in non-IBC token `classId`s.
 
-A sending chain may be acting as a source or sink zone. When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone. When tokens are sent from a source zone, the destination port and channel will be prefixed onto the `classId` (once the tokens are received) adding another hop to a tokens record. When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone. When tokens are sent from a sink zone, the last prefixed port and channel pair on the `classId` is removed (once the tokens are received), undoing the last hop in the tokens record. A more complete explanation is present in the ibc-go implementation (TBD).
+A sending chain may be acting as a source or sink zone. When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone. When tokens are sent from a source zone, the destination port and channel will be prefixed onto the `classId` (once the tokens are received) adding another hop to a tokens record. When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone. When tokens are sent from a sink zone, the last prefixed port and channel pair on the `classId` is removed (once the tokens are received), undoing the last hop in the tokens record. A more complete explanation is present in the [ibc-go implementation]() (TBD).
 
 The acknowledgement data type describes whether the transfer succeeded or failed, and the reason for failure (if any).
 
@@ -108,9 +108,7 @@ function setup() {
 
 Once the `setup` function has been called, channels can be created through the IBC routing module between instances of the non-fungible token transfer module on separate chains.
 
-An administrator (with the permissions to create connections & channels on the host state machine) is responsible for setting up connections to other state machines & creating channels
-to other instances of this module (or another module supporting this interface) on other chains. This specification defines packet handling semantics only, and defines them in such a fashion
-that the module itself doesn't need to worry about what connections or channels might or might not exist at any point in time.
+An administrator (with the permissions to create connections & channels on the host state machine) is responsible for setting up connections to other state machines & creating channels to other instances of this module (or another module supporting this interface) on other chains. This specification defines packet handling semantics only, and defines them in such a fashion that the module itself doesn't need to worry about what connections or channels might or might not exist at any point in time.
 
 #### Routing module callbacks
 
@@ -219,17 +217,19 @@ function createOutgoingPacket(
   sourceChannel: string,
   timeoutHeight: Height,
   timeoutTimestamp: uint64) {
+  // assert that sender is token owner
+  abortTransactionUnless(sender === nft.getOwner(classId, tokenId))
   prefix = "{sourcePort}/{sourceChannel}/"
-  // we are the source if the denomination is not prefixed
+  // we are the source if the classId is not prefixed
   source = classId.slice(0, len(prefix)) !== prefix
   if source {
     // determine escrow account
     escrowAccount = channelEscrowAddresses[sourceChannel]
-    // escrow source token (assumed to fail if sender is not owner)
-    nft.Transfer(sender, escrowAccount, classId, tokenId)
+    // escrow source token
+    nft.Transfer(classId, tokenId, escrowAccount)
   } else {
-    // receiver is source chain, burn voucher (assumed to fail if sender is not owner)
-    bank.Burn(sender, classId, tokenId)
+    // receiver is source chain, burn voucher
+    bank.Burn(classId, tokenId)
   }
   Class class = nft.getClass(classId)
   NFT token = nft.getNFT(classId, tokenId)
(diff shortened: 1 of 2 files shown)


spec/app/ics-021-nft-transfer/README.md
```diff
@@ -0,0 +1,360 @@
...
+  tokenId: string
+  tokenUri: string
+  sender: string
+  receiver: string
+}
+```
+
+As tokens are sent across chains using the ICS 21 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `classId` field. 
+
+The ics21 token classes are represented in the form `{ics21Port}/{ics21Channel}/{classId}`, where `ics21Port` and `ics21Channel` are an ics21 port and channel on the current chain for which the token exists. The prefixed port and channel pair indicate which channel the token was previously sent through. If `{classId}` contains `/`, then it must also be in the ics21 form which indicates that this token has a multi-hop record. Note that this requires that the `/` (slash character) is prohibited in non-IBC token denomination names.
```
robert-zaremba, line 52: Let's take an example:
+ On chain A we have 2 NFTs (n1, n2) with same class `k`.
+ Chain A has two channels to chain B: r1 and r2.
+ If we sent `n1` from A to B using r1 and `n2` using r2, then in chain B they will have different class id

Correct? This smells bad, as we increase fragmentation.

spec/app/ics-021-nft-transfer/README.md
```diff
@@ -0,0 +1,360 @@
...
+
+A sending chain may be acting as a source or sink zone. When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone. When tokens are sent from a source zone, the destination port and channel will be prefixed onto the `classId` (once the tokens are received) adding another hop to a tokens record. When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone. When tokens are sent from a sink zone, the last prefixed port and channel pair on the `classId` is removed (once the tokens are received), undoing the last hop in the tokens record. A more complete explanation is present in the ibc-go implementation (TBD).
```
robert-zaremba, line 54: Is this part of the `ibc-go` implementation or IBC standard?

spec/app/ics-021-nft-transfer/README.md
```diff
@@ -0,0 +1,360 @@
...
+## Technical Specification
+
+### Data Structures
+
+Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the denomination, amount, sending account, and receiving account.
+
+```typescript
+interface NonFungibleTokenPacketData {
+  classId: string
+  classUri: string
```
robert-zaremba, line 42: do we need to send URI?

spec/app/ics-021-nft-transfer/README.md
```diff
@@ -0,0 +1,360 @@
...
+
+This implementation preserves both uniqueness & supply.
+
+Uniqueness: If tokens have been sent to the counterparty chain, they can be redeemed back in the same `classId` & `tokenId` on the source chain.
+
+Supply: Redefine supply as unlocked tokens. All send-recv pairs for any given token class sum to net zero. Source chain can change supply.
+
+##### Multi-chain notes
+
+This specification does not directly handle the "diamond problem", where a user sends a token originating on chain A to chain B, then to chain D, and wants to return it through D -> C -> A — since the supply is tracked as owned by chain B (and the `classId` will be "{portOnD}/{channelOnD}/{portOnB}/{channelOnB}/classId"), chain C cannot serve as the intermediary. It is not yet clear whether that case should be dealt with in-protocol or not — it may be fine to just require the original path of redemption (and if there is frequent liquidity and some surplus on both paths the diamond path will work most of the time). Complexities arising from long redemption paths may lead to the emergence of central chains in the network topology.
```
robert-zaremba, line 327: this is a copy from https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer

robert-zaremba (commented):
Some sections are copied from: https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer

Instead of copying, maybe we can reference those sections? What do you think? CC: @crodriguezvega

haifengxi:
> Some sections are copied from: https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer
> 
> Instead of copying, maybe we can reference those sections? What do you think? CC: @crodriguezvega

Will add a reference to ICS20 and some comments at the beginning of ICS21; as discussed earlier, the quickest and most reasonable approach to ICS21 will be an adaptation of ICS20 with NFT specifics.

spec/app/ics-021-nft-transfer/README.md
haifengxi: Will add a reference to ICS20 and some comments at the beginning of ICS21

spec/app/ics-021-nft-transfer/README.md
haifengxi: The same thing happens with fungible tokens of the same denom; that is the way it is with IBC, and the fragmentation will be handled by UI providers.

spec/app/ics-021-nft-transfer/README.md
haifengxi: It is part of IBC standard, and a hyperlink will be added to reference the Go implementation when the latter is ready.

spec/app/ics-021-nft-transfer/README.md
haifengxi: Class URI is optional but will be very useful for cross-chain interoperability with NFT marketplace like OpenSea where class/collection metadata can be configured for better user experience.

JakeHartnell: > Class URI is optional but will be very useful for cross-chain interoperability with NFT marketplace like OpenSea where class/collection metadata can be configured for better user experience.

https://docs.opensea.io/docs/contract-level-metadata

spec/app/ics-021-nft-transfer/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+## Technical Specification
+
+### Data Structures
+
+Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the class id, class uri, token id, token uri, sending account, and receiving account.
+
+```typescript
+interface NonFungibleTokenPacketData {
+  classId: string
```
JakeHartnell, line 43: `classId` seems to be very centered on the Cosmos SDK NFT module implementation.

From the proto files:
```
class_id defines the unique identifier of the NFT classification, similar to the contract address of ERC721
```

So for smart contract based chains, this would be a contract address?

Would be great to add some wording to this standard for how Smart Contract based NFTs can conform to this standard.

the-frey: +1 - would like to understand this as well

MikeSofaer:
Pylons is primarily concerned with royalties when it comes to NFT IBC transfer. I'd like to make sure that this has been thought about. We have our own modules for item creation, but we would like to have standards for things like image_url in addition to royalties.

the-frey:
> Pylons is primarily concerned with royalties when it comes to NFT IBC transfer. I'd like to make sure that this has been thought about. We have our own modules for item creation, but we would like to have standards for things like image_url in addition to royalties.

Out of interest, is there anything I can read on your approach to royalties? I think getting royalties interop right is _super_ important, and something we'd want smart contracts and native to line up on as well...

spec/app/ics-021-nft-transfer/README.md
robert-zaremba: I know. But maybe there is a way to fix it / make it better.

spec/app/ics-021-nft-transfer/README.md
robert-zaremba: Makes sense, could you add a note / comment about it in the file, please?

MikeSofaer:
> > Pylons is primarily concerned with royalties when it comes to NFT IBC transfer. I'd like to make sure that this has been thought about. We have our own modules for item creation, but we would like to have standards for things like image_url in addition to royalties.
> 
> Out of interest, is there anything I can read on your approach to royalties? I think getting royalties interop right is _super_ important, and something we'd want smart contracts and native to line up on as well...

I don't have anything written, but I can maybe draft something we can work on. I think encumbrances across IBC are actually a larger question that royalties are only a part of. Superfluid staking faces IMO a similar problem with its cross-chain slashing, as does the question of enabling people to get usage value out of locked-up lending collateral, which I think is critical for on-chain economics to compete with off-chain.

spec/app/ics-021-nft-transfer/README.md
robert-zaremba: IMO, it should be a general term. Chains can define a class id as a composed things:
+ `smart-contract/class-in-smart-contract`
+ `app-name/nft-type` (eg `axies/monsters`)
Or simple, for a specialized app chains, eg: `nft-type`.

Does it make sense? If yes, then let's expand about that in the document (readme).

robert-zaremba:
@MikeSofaer  - could you share you proto files?

MikeSofaer:
> @MikeSofaer - could you share you proto files?

https://github.com/Pylons-tech/pylons_dart_sdk/tree/c29ea8cacb3a93dd92daf1776b1a4f0b73e84429/proto

robert-zaremba:
Re Royalties - I think sooner or later it will be a part of almost every NFT project. Especially when going crosschain. So let's reconsider how this will be integrated, what kind of information we need to carry over IBC and if we need to add something to existing `NFT` or `Class ` proto structures.
Here is a related Ethereum ERC: https://eips.ethereum.org/EIPS/eip-2981

the-frey:
> Re Royalties - I think sooner or later it will be a part of almost every NFT project. Especially when going crosschain. So let's reconsider how this will be integrated, what kind of information we need to carry over IBC and if we need to add something to existing `NFT` or `Class ` proto structures. Here is a related Ethereum ERC: https://eips.ethereum.org/EIPS/eip-2981

We've got a small port of EIP-2981 in CW actually, it seems like a good minimal starting point, but imagine there might be other considerations or requirements that we can bring in that the eth folks couldn't - given that they already had a bunch of live marketplaces out in the wild 🤔

robert-zaremba:
@michaelfig  Is royalty a fixed or linear (eg %) amount for each token, or it is a non linear function?
If an NFT is transferred over IBC, how do you expect to carry the royalties? What information do we need to share over IBC?

MikeSofaer:
> @michaelfig Is royalty a fixed or linear (eg %) amount for each token, or it is a non linear function? If an NFT is transferred over IBC, how do you expect to carry the royalties? What information do we need to share over IBC?

If that's intended for me, we currently support linear royalties, and minimum prices. For payment, I'd expect the royalty to be sent to the creator's account on Pylons over IBC, in whatever token was used to pay, but I haven't really thought about any other options.

spec/app/ics-021-nft-transfer/README.md
haifengxi: comment added to the spec

spec/app/ics-021-nft-transfer/README.md
haifengxi: wording added to clarify/generalize the semantics of packet data fields

crodriguezvega:
Seems like the general preference would be to rename this specification to ICS-721 (to name it after Ethereum's ERC-721). Would you mind doing that please?

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,375 @@
...
+### Data Structures
+
+Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the class id, class uri, token id, token uri, sending account, and receiving account.
+
+```typescript
+interface NonFungibleTokenPacketData {
+  classId: string
+  classUri: string
+  tokenId: string
+  tokenUri: string
```
michaelfig, line 46: Would it be possible to transfer multiple tokens of the same class in one packet?  That would help Agoric's use case.
```suggestion
  tokenIds: []string
  tokenUris: []string
```
Otherwise, unlike ICS-20, we can only send a single value of the type at a time.

spec/app/ics-021-nft-transfer/README.md
okwme: I think this design is pretty fundamental to how IBC works. Probably not the time to focus on revising such a core component when the fastest path would be to re-use the norm. That problem should likely be focused on holistically and then applied to both ICS-20 and ICS-721 at a future date. Would be curious to hear whether it's been discussed by the IBC team before (cc @crodriguezvega )

spec/app/ics-021-nft-transfer/README.md
okwme: In the event that a token URI is updated on the origin chain after the NFT has been transferred to another chain, will it cause a problem when the NFT is transferred back since the IBC packet's TokenURI will be different from the origin? Or is it not possible to update the token URI on the origin chain when it is locked in escrow?

okwme:
I'm hesitant to complicate the ICS standard with encumbrances at this point. One goal of this latest batch of work is getting the simplest version working and live as soon as possible and part of the reason prior development has taken a year longer than expected is because so much time was spent trying to get encumbrances right on the first try when I think we can all agree they're a complicated thing to get wrong.

If we look at OpenSea as the example for the most popular marketplace they allow royalties to be added to sales in the following manner:
* prove you control the account that deployed the original smart contract
* once done you may:
  * set a percentage fee to be applied to all sales
  * set an address to receive the royalties
  * set the design of the page on their site that lists your token class
If that were the goal then it should only be the account that is sent with the token; basically who is allowed to configure royalties wherever this token goes. The ability to configure things on behalf of the token will only be as functional as the destination allows them.

Not sure if this actually the best path but consideration should be made for the account address that has this kind of "admin" rights on the asset. It should not be a groups module account because then it would not be able to submit transactions to the various destinations, only in a local setting. It could be a single private key account assuming the key is derived with the same path and the proper bech32 prefix on the destination. Same for cryptographic multisig account although certainly more work to set up in the new context.

@MikeSofaer can you confirm what your IBC user story is? From my understanding the bulk of your product story is for NFTs that start and stay on Pylons, correct? Do you have destination marketplace partners in place who are already aware and willing and able to support your royalty structure?

@haifengxi IRIS is the only team who has already moved live NFTs across networks AFAIK. Can you speak to the features that you found necessary with regard to encumbrances like royalties?

I'd like to understand if a simplified encumbrance field that is essentially allows for destination chains to enforce with the following logic: "who can configure encumbrances wherever this NFT is located?", is sufficient to cover the expected use cases at this point. Or if we should entirely defer to future versions of the ICS that allow for a more complex encumbrance metadata transfer.

In any case, it should be possible to begin with a very simple ICS that uses the NFT module under the hood. A future version of the ICS could include encumbrances and still use the NFT module under the hood. This would result in some NFTs that have the same class but came over different channels. Depending on the channel they could have encumbrance information or not. NFTs could migrate to the updated version by going back to the destination and then through the new encumbrance channel. I think that's a permissible path to allow for iteration over the ICS to support the simplest first and the more complex as user demands shape the direction.

MikeSofaer:
> @MikeSofaer can you confirm what your IBC user story is? From my understanding the bulk of your product story is for NFTs that start and stay on Pylons, correct? Do you have destination marketplace partners in place who are already aware and willing and able to support your royalty structure?

We're interested in moving NFTs to marketplace zones, stargaze is the first zone we've talked to about this. Royalties are core to this story, and Shane and I have talked a bit about how to approach building an IBC connector. I also think the OpenSea flow described above is a crazy hack and we should do better.

I'm also not totally sure how compatible we are with a standard that uses the NFT module under the hood anyway, as we don't use it.

spec/app/ics-021-nft-transfer/README.md
haifengxi: Once a token is transferred out, the original token has been locked in the escrow account at source, and the correct business logic at source, I believe, should be that no updates can be made to locked tokens.

spec/app/ics-721-nft-transfer/README.md
haifengxi: Actually I was thinking about the same thing!  It's definitely a very useful feature.

Muggle-Du:
the locked token in the escrow account owned by IBC indicates that NFT has a third status as "locked", so should we describe this status in NFT data structure?

spec/app/ics-721-nft-transfer/README.md
haifengxi: Done.

spec/app/ics-021-nft-transfer/README.md
haifengxi: > the locked token in the escrow account owned by IBC indicates that NFT has a third status as "locked", so should we describe this status in NFT data structure?

NFT status is a local attribute that is tracked by the application layer at source and participating chains. I think we don't need to include it in the transport protocol.

haifengxi:
> 
> @haifengxi IRIS is the only team who has already moved live NFTs across networks AFAIK. Can you speak to the features that you found necessary with regard to encumbrances like royalties?
> 
The use case we are dealing with does not involve royalties setting -- at least not for now.  Using our customer as an example, however, they can easily set up a royalty scheme that suits them by using OpenSea's infrastructure.  The accrued payments will be in local "currencies" such as ETH, but they can always choose to swap them into other tokens and then transfer them to the Cosmos world.

charleenfei:
> > @MikeSofaer can you confirm what your IBC user story is? From my understanding the bulk of your product story is for NFTs that start and stay on Pylons, correct? Do you have destination marketplace partners in place who are already aware and willing and able to support your royalty structure?
> 
> We're interested in moving NFTs to marketplace zones, stargaze is the first zone we've talked to about this. Royalties are core to this story, and Shane and I have talked a bit about how to approach building an IBC connector. I also think the OpenSea flow described above is a crazy hack and we should do better.
> 
> I'm also not totally sure how compatible we are with a standard that uses the NFT module under the hood anyway, as we don't use it.

a few comments on this:
1) a royalties module could be covered by creating a custom ibc middleware which wraps the ibc nft transfer module and implements whatever types of royalties logic you need for Pylons. I think this is a cleaner and more generic way of implementing an interoperable IBC NFT standard which guarantees we can interoperate with ERC721s, Cosmos NFTs, and cw721s. 

2) the ibc standard would probably able to cover both cw721s and sdk.NFT the same way that this standard for cosmwasm https://docs.cosmwasm.com/docs/0.14/ibc/cw20-ics20/ allows cw20s to be transferred using the ics20 spec. but on the app level, you would probably need something to make sure that you can turn cw721s into sdk.NFTs if you are working with both, or viceversa. @the-frey can correct me if I am wrong on this?

spec/app/ics-021-nft-transfer/README.md
robert-zaremba: OK, wanted to check if there are already solutions for that.

spec/app/ics-021-nft-transfer/README.md
robert-zaremba: > I think the correct logic should be: only tokens that have not been transferred out via IBC can have their tokenURI updated by owner. 

Makes sense - let's add it to the spec. How we can ensure it in x/nft? Today, there is no functionality to update NFT. Maybe we should add it on the keeper level (not msg server level)? The keeper will check if nft is owned by x/ibc to allow edits. What do you think @haifengxi ? If yes then we need to create a new task.

robert-zaremba:
> I'm hesitant to complicate the ICS standard with encumbrances at this point.

Standards are super hard to change. The goal of defining an architecture or standard is to be future proof. Bad standard creates tonnes of a technical debt, often very hard to change. See ERC-20 discussions. 

Here we need to try to find potential technical debts we are aware about and asses how expensive they are. Updating IBC or a core module is often a multi-year process, especially if a network level effect is expected.

For example, not having yet DID / IDD to uniquely identify Cosmos assets is one of the technical debt which is not that hard to solve, but creates more holes in the Interchain Designs.

robert-zaremba:
Re _royalties_, during the last call we were thinking if another module (`x/royalties`) can be used to handle that? 
I was thinking about it, and have the following idea with DIDs. x/royalties module would work as follows:
* associate royalty record with interchain account and asset did
* Add royalty flag into the NFT object or an information in DID Document. The former is easier to consider now, and it will always be shipped with NFT.
* Whenever we know that a royalty is required, then we use DID to  source the royalty information using IBC.

I didn't read the [middleware](https://github.com/cosmos/ibc/tree/master/spec/app/ics-030-middleware) spec, but maybe it's a good solution. @charleenfei  could you confirm if the IBC middleware will work for the following use case: Chan A received NFT `X` from chain B,  The NFT `X` was issued by chain C. How Chain A will know that `X` is bound to some royalty contract?

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+  channelIdentifier: Identifier) {
+  // no action necessary
+}
+```
+
+##### Packet relay
+
+In plain English, between chains `A` and `B`:
+
+- When acting as the source zone, the bridge module escrows an existing local non-fungible token on the sending chain and mints a corresponding voucher on the receiving chain.
```
AdityaSripal, line 206:
```suggestion
- When the sender is source zone, the bridge module escrows an existing local non-fungible token on the sending chain and mints a corresponding voucher on the receiving chain.
```

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+- When acting as the sink zone, the bridge module burns the local voucher on the sending chain and unescrows the local non-fungible token on the receiving chain.
```
AdityaSripal, line 207:
```suggestion
- When sender is the sink zone, the bridge module burns the local voucher on the sending chain and unescrows the local non-fungible token on the receiving chain.
```

I actually find it clearer to say instead:

`When receiver is the source zone`

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+- When a packet times-out, local non-fungible tokens are unescrowed back to the sender or vouchers minted back to the sender appropriately.
```
AdityaSripal, line 208: Is there a cleaner way to word this?

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+  receiver: string,
+  source: boolean,
+  destPort: string,
+  destChannel: string,
+  sourcePort: string,
+  sourceChannel: string,
+  timeoutHeight: Height,
+  timeoutTimestamp: uint64) {
+  prefix = "{sourcePort}/{sourceChannel}/"
+  // we are the source if the classId is not prefixed
```
AdityaSripal, line 228:
```suggestion
  // we are the source if the classId is not prefixed by the sourcePort and sourceChannelID
```

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+    // assert that sender is token owner
+    abortTransactionUnless(sender === nft.getOwner(classId, tokenId))
+    if source {
+      // determine escrow account
+      escrowAccount = channelEscrowAddresses[sourceChannel]
+      // escrow source token
+      nft.Transfer(classId, tokenId, escrowAccount)
+    } else {
+      // receiver is source chain, burn voucher
+      bank.Burn(classId, tokenId)
```
AdityaSripal, line 241:
```suggestion
      nft.Burn(classId, tokenId)
```

The voucher is also an NFT

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+
+`onRecvPacket` is called by the routing module when a packet addressed to this module has been received.
+
+```typescript
+function onRecvPacket(packet: Packet) {
+  NonFungibleTokenPacketData data = packet.data
+  // construct default acknowledgement of success
+  NonFungibleTokenPacketAcknowledgement ack = NonFungibleTokenPacketAcknowledgement{true, null}
+  prefix = "{packet.sourcePort}/{packet.sourceChannel}/"
+  // we are the source if the packets were prefixed by the sending chain
```
AdityaSripal, line 258:
```suggestion
  // we are the source if the classId was prefixed by the sending chain
```

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+  source = data.classId.slice(0, len(prefix)) === prefix
+  for (var i in data.tokenIds) {
+    if source {
+      // receiver is source chain: unescrow token
+      // determine escrow account
+      escrowAccount = channelEscrowAddresses[packet.destChannel]
+      // assert that escrow account is token owner
+      abortTransactionUnless(escrowAccount === nft.getOwner(data.classId.slice(len(prefix)), data.tokenIds[i]))
```
AdityaSripal, line 266: We should remove this line so that if transfer fails because module account is not owner, we will write failed acknowledgement rather than just abort tx

AdityaSripal (changes requested):
ACK on general architecture. Nice work translating from the ICS-20 design to this usecase.

I have some comments to make the spec clearer

spec/app/ics-021-nft-transfer/README.md
AdityaSripal: This is core to how IBC works. Agreed to keep as is

AdityaSripal:
I'm also in favor of not keeping royalties in scope for this ICS. It makes sense to create a standard that just focuses on the bare simple function of transferring NFT from an address on chainA to an address on chainB.

When we created ICS-20 we didn't optimize for the DEX usecase. The interchain DEX's currently deployed are complicated application that are built on top of a very simple ICS.

I think in order to serve the broadest use cases, we should build a simple standard that everyone can build complexity on top of. As @robert-zaremba mentioned, standards are very hard to change so I'm hesitant to bake royalties fundamentally into the concept of an NFT since that serves a particular subset of usecases and may not be the preferred approach forever.

I would agree with @charleenfei that royalties should probably be built as middleware on top of the simpler ICS. I could imagine that there is a separate royalties module that can store the original royalty addresses as the NFT moves across the ecosystem. Then each chain would be responsible for paying royalties to the original creator if the NFT gets sold on their zone. I would also reject as the sender chain any NFT transfer of royalty NFTs if they did not go through the royalty middleware. This doesn't quite solve the issue if the NFT IBC transfer itself is considered a sale.

robert-zaremba:
+ let's add a section _Further Discussion_ with a short information that this design supports extended use cases such as royalties or permissioned transfers. A solution could use  hooks, [030-middleware](https://github.com/cosmos/ibc/tree/master/spec/app/ics-030-middleware) etc.. Designing a guideline is out of the scope of this ICS.

spec/app/ics-021-nft-transfer/README.md
haifengxi: Will update the spec.  However, this is something that can only be enforced by the application layer of participating chains instead of the IBC protocol itself.  `x/nft` is at the "storage layer", we probably cannot enforce it there without incurring a circular dependency between `x/nft` and `x/ibc`.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+  channelIdentifier: Identifier,
+  counterpartyPortIdentifier: Identifier,
+  counterpartyChannelIdentifier: Identifier,
+  version: string) {
+  // only unordered channels allowed
+  abortTransactionUnless(order === UNORDERED)
+  // assert that version is "ics721-1"
+  abortTransactionUnless(version === "ics721-1")
+  // allocate an escrow address
+  channelEscrowAddresses[channelIdentifier] = newAddress()
```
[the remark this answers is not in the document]
cwgoes: I agree

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+
+#### Reasoning
+
+##### Correctness
+
+This implementation preserves both uniqueness & supply.
+
+Uniqueness: If tokens have been sent to the counterparty chain, they can be redeemed back in the same `classId` & `tokenId` on the source chain.
+
+Supply: Redefine supply as unlocked tokens. All send-recv pairs for any given token class sum to net zero. Source chain can change supply.
```
cwgoes, line 346: this is probably not quite the right property for NFTs

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+  classUri: string
+  tokenIds: []string
+  tokenUris: []string
+  sender: string
+  receiver: string
+}
+```
+`classId` uniquely identifies the class/collection to which this NFT belongs in the originating host environment.  In the case of an ERC-1155 compliant smart contract, for example, this could be a string representation of the top 128 bits of the token ID.
+
+`classUri` is optional, but will be extremely beneficial for cross-chain interoperability with NFT marketplaces like OpenSea, where [class/collection metadata](https://docs.opensea.io/docs/contract-level-metadata) can be added for better user experience.
```
cwgoes, line 53: why exactly do we want to pass this along as on-chain data, if it's just a pointer to off-chain data anyways? can't anyone interested in the off-chain data look up the source chain, look up the URI there, and then look up the data? this introduces the possibility of divergence (e.g. a URI on some receiving chain not matching the URI associated with the asset on the original source chain), which seems like a recipe for problems

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+
+`tokenIds` uniquely identifies some NFTs within the given class that are being transferred.  In the case of an ERC-1155 compliant smart contract, for example, a `tokenId` could be a string representation of the bottom 128 bits of the token ID.
+
+Each `tokenId` has a corresponding entry in `tokenUris`, which refers to an off-chain resource that is typically an immutable JSON file containing the NFT's metadata.
```
cwgoes, line 57: ditto (see above comment)

cwgoes (changes requested):
Thanks for the submission; see comments (and I second Aditya's).

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+Users of a set of chains connected over the IBC protocol might wish to utilise an asset issued on one chain on another chain, perhaps to make use of additional features such as exchange or privacy protection, while retaining uniqueness with the original asset on the issuing chain. This application-layer standard describes a protocol for transferring non-fungible tokens between chains connected with IBC which preserves asset uniqueness, preserves asset ownership, limits the impact of Byzantine faults, and requires no additional permissioning.
+
+### Definitions
+
+The IBC handler interface & IBC routing module interface are as defined in [ICS 25](../../core/ics-025-handler-interface) and [ICS 26](../../core/ics-026-routing-module), respectively.
+
+### Desired Properties
+
+- Preservation of uniqueness (two-way peg).
+- Preservation of total supply (maintained on a single source chain & module).
```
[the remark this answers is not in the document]
haifengxi: "Preservation of total supply..." removed.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
+---
+ics: 721
+title: Non-Fungible Token Transfer
+stage: draft
+category: IBC/APP
+requires: 25, 26
+kind: instantiation
+author: Christopher Goes <cwgoes@interchain.berlin>, Haifeng Xi <haifeng@bianjie.ai>
```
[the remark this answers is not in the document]
haifengxi: Point taken.

spec/app/ics-721-nft-transfer/README.md
haifengxi: "Supply: ..." removed.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+- Preservation of non-fungibility and uniqueness (i.e., only one instance of the token is *live* across all the IBC-connected blockchains).
+- Permissionless token transfers, no need to whitelist connections, modules, or `classId`s.
+- Symmetric (all chains implement the same logic, no in-protocol differentiation of hubs & zones).
+- Fault containment: prevents Byzantine-creation of tokens originating on chain `A`, as a result of chain `B`'s Byzantine behavior.
+
+## Technical Specification
+
+### Data Structures
+
+Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the class id, class uri, token id's, token uri's, sending account, and receiving account.
```
AdityaSripal, line 38:
```suggestion
Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the class id, class uri, token id's, token uri's, sender address, and receiver address.
```

spec/app/ics-721-nft-transfer/README.md
AdityaSripal: Is it possible for the classUri to change after initialization? Similar question regarding tokenUris?

> Avoiding metadata divergence is something that can and should be enforced at the application layer.

This can be done at the point of transfer, but if the metadata changes at a later date, it may be difficult to propogate this information over to wherever the NFT is currently

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+`tokenIds` uniquely identifies some NFTs within the given class that are being transferred.  In the case of an ERC-1155 compliant smart contract, for example, a `tokenId` could be a string representation of the bottom 128 bits of the token ID.
+
+Each `tokenId` has a corresponding entry in `tokenUris`, which refers to an off-chain resource that is typically an immutable JSON file containing the NFT's metadata.
+
+As tokens are sent across chains using the ICS 721 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `classId` field.
+
+The ics721 token classes are represented in the form `{ics721Port}/{ics721Channel}/{classId}`, where `ics721Port` and `ics721Channel` identifiy the channel on the current chain from which the token arrived. The prefixed port and channel pair indicate which channel the token was previously sent through. If `{classId}` contains `/`, then it must also be in the ics721 form which indicates that this token has a multi-hop record. Note that this requires that the `/` (slash character) is prohibited in non-IBC token `classId`s.
```
AdityaSripal, line 60:
```suggestion
The ics721 token classes are represented in the form `{ics721Port}/{ics721Channel}/{classId}`, where `ics721Port` and `ics721Channel` identify the channel on the current chain from which the token arrived. The prefixed port and channel pair indicate which channel the token was previously sent through. If `{classId}` contains `/`, then it must also be in the ics721 form which indicates that this token has a multi-hop record. Note that this requires that the `/` (slash character) is prohibited in non-IBC token `classId`s.
```

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+```typescript
+function onChanOpenTry(
+  order: ChannelOrder,
+  connectionHops: [Identifier],
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier,
+  counterpartyPortIdentifier: Identifier,
+  counterpartyChannelIdentifier: Identifier,
+  version: string,
+  counterpartyVersion: string) {
```
AdityaSripal, line 153: update TRY callback to latest update. see changes on master

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,386 @@
...
+  channelIdentifier: Identifier) {
+  // accept channel confirmations, port has already been validated, version has already been validated
+}
+```
+
+```typescript
+function onChanCloseInit(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier) {
+  // no action necessary
```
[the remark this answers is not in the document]
AdityaSripal: It doesn't look like this is resolved? Seems like this is also a problem in ICS20, but it should be fixed here. I'll make a separate pr for ics20

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+function onChanCloseConfirm(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier) {
+  // no action necessary
+}
+```
+
+##### Packet relay
+
+In plain English:
```
AdityaSripal, line 203:
```suggestion
```

AdityaSripal (changes requested):
A few nits, and a question about how mutable the NFT URIs are. I believe that will have an effect on the final design of the spec. Getting close though!

spec/app/ics-721-nft-transfer/README.md
hxrts: In ERC721 URIs may be changed after minting if an account is authorised to make that change (can also be locked on or after mint).

If we were to use a middleware approach and someone opted to clone metadata, would want to pass the origin clone time.

spec/app/ics-721-nft-transfer/README.md
haifengxi: Do you mean the logic here -> https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/ibc_module.go

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+#### Reasoning
+
+##### Correctness
+
+This implementation preserves token uniqueness.
+
+Uniqueness: If tokens have been sent to the counterparty chain, they can be redeemed back in the same `classId` & `tokenId` on the source chain.
+
+##### Multi-chain notes
```
mpoke, line 341: Maybe refer here to ICS20 instead. It's exactly the same content (except `classId` instead of denomination).

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+This specification does not directly handle the "diamond problem", where a user sends a token originating on chain A to chain B, then to chain D, and wants to return it through D -> C -> A — since the supply is tracked as owned by chain B (and the `classId` will be "{portOnD}/{channelOnD}/{portOnB}/{channelOnB}/classId"), chain C cannot serve as the intermediary. It is not yet clear whether that case should be dealt with in-protocol or not — it may be fine to just require the original path of redemption (and if there is frequent liquidity and some surplus on both paths the diamond path will work most of the time). Complexities arising from long redemption paths may lead to the emergence of central chains in the network topology.
+
+In order to track all of the tokens moving around the network of chains in various paths, it may be helpful for a particular chain to implement a registry which will track the "global" source chain for each `classId`. End-user service providers (such as wallet authors) may want to integrate such a registry or keep their own mapping of canonical source chains and human-readable names in order to improve UX.
+
+#### Optional addenda
```
mpoke, line 347: ditto (see above comment)

mpoke (changes requested):
Thanks for the submission. Also ACK on general architecture. See comments bellow.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+  // construct default acknowledgement of success
+  NonFungibleTokenPacketAcknowledgement ack = NonFungibleTokenPacketAcknowledgement{true, null}
+  prefix = "{packet.sourcePort}/{packet.sourceChannel}/"
+  // we are the source if the classId is prefixed with the packet's sourcePort and sourceChannel
+  source = data.classId.slice(0, len(prefix)) === prefix
+  for (var i in data.tokenIds) {
+    if source {
+      // receiver is source chain: unescrow token
+      // determine escrow account
+      escrowAccount = channelEscrowAddresses[packet.destChannel]
```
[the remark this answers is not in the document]
haifengxi: Will fix this.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+      // unescrow token to receiver
+      err = nft.Transfer(data.classId.slice(len(prefix)), data.tokenIds[i], data.receiver)
+      if (err !== nil) {
+        ack = NonFungibleTokenPacketAcknowledgement{false, "transfer nft(" + data.classId + ", " + data.tokenIds[i] + ") failed"}
```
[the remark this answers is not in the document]
haifengxi: Will fix this.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+  channelEscrowAddresses[channelIdentifier] = newAddress()
+}
+```
+
+```typescript
+function onChanCloseInit(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier) {
+  // abort and return error to prevent channel closing by user
+  abortTransaction
```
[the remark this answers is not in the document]
haifengxi: Will fix.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+  tokenUris: []string
+  sender: string
+  receiver: string
+}
+```
+`classId` uniquely identifies the class/collection to which this NFT belongs in the originating host environment.  In the case of an ERC-1155 compliant smart contract, for example, this could be a string representation of the top 128 bits of the token ID.
+
+`classUri` is optional, but will be extremely beneficial for cross-chain interoperability with NFT marketplaces like OpenSea, where [class/collection metadata](https://docs.opensea.io/docs/contract-level-metadata) can be added for better user experience.
+
+`tokenIds` uniquely identifies some NFTs within the given class that are being transferred.  In the case of an ERC-1155 compliant smart contract, for example, a `tokenId` could be a string representation of the bottom 128 bits of the token ID.
```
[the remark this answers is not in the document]
haifengxi: Nice catch, will fix.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+A sending chain may be acting as a source or sink zone. When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone. When tokens are sent from a source zone, the destination port and channel will be prefixed onto the `classId` (once the tokens are received) adding another hop to a tokens record. When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone. When tokens are sent from a sink zone, the last prefixed port and channel pair on the `classId` is removed (once the tokens are received), undoing the last hop in the tokens record. A more complete explanation is present in the [ibc-go implementation]() (TBD).
```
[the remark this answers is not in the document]
haifengxi: This prefix concept is by now very well understood, I think the explanation here and the well commented pseudo code should suffice.

spec/app/ics-721-nft-transfer/README.md
haifengxi: Done fixing.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+  source = classId.slice(0, len(prefix)) !== prefix
+  tokenUris = []
+  for (let tokenId in tokenIds) {
+    // assert that sender is token owner
+    abortTransactionUnless(sender === nft.getOwner(classId, tokenId))
+    if source {
+      // determine escrow account
+      escrowAccount = channelEscrowAddresses[sourceChannel]
+      // escrow source token
+      nft.Transfer(classId, tokenId, escrowAccount)
```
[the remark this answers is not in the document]
haifengxi: Will add reference to [adr-043](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-043-nft-module.md) where the NFT module is spec'ed out.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+  for (var i in data.tokenIds) {
+    if source {
+      // receiver is source chain: unescrow token
+      // determine escrow account
+      escrowAccount = channelEscrowAddresses[packet.destChannel]
+      // unescrow token to receiver
+      err = nft.Transfer(data.classId.slice(len(prefix)), data.tokenIds[i], data.receiver)
+      if (err !== nil) {
+        ack = NonFungibleTokenPacketAcknowledgement{false, "transfer nft(" + data.classId + ", " + data.tokenIds[i] + ") failed"}
+        break
```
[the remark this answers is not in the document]
haifengxi: NFTs that were previously transferred out had been locked in the escrow account at the sending chain.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+      }
+    } else {
+      prefix = "{packet.destPort}/{packet.destChannel}/"
+      prefixedClassId = prefix + data.classId
+      // sender was source, mint voucher to receiver
+      err = nft.Mint(prefixedClassId, data.classUri, data.tokenIds[i], data.tokenUris[i], data.receiver)
+      if (err !== nil) {
+        ack = NonFungibleTokenPacketAcknowledgement{false, "mint nft(" + data.classId + ", " + data.tokenIds[i] + ") failed"}
+        break
```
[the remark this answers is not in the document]
haifengxi: There would be no previously minted NFTs on the receiving chain; all the other copies of the token being minted (including the original one at the source chain) are locked in corresponding escrow accounts on the _upstream_ blockchains along the IBC path.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+  sourceChannel: string,
+  timeoutHeight: Height,
+  timeoutTimestamp: uint64) {
+  prefix = "{sourcePort}/{sourceChannel}/"
+  // we are the source if the classId is not prefixed with the sourcePort and sourceChannel
+  source = classId.slice(0, len(prefix)) !== prefix
+  tokenUris = []
+  for (let tokenId in tokenIds) {
+    // assert that sender is token owner
+    abortTransactionUnless(sender === nft.getOwner(classId, tokenId))
```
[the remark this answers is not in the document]
haifengxi: NFT ownership is maintained in the `x/nft` module (see [adr-043](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-043-nft-module.md)).  An NFT successfully transferred out is owned by the escrow account and cannot be transferred by the previous owner or anyone else for that matter.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+  NonFungibleTokenPacketData data = packet.data
+  prefix = "{packet.sourcePort}/{packet.sourceChannel}/"
+  for (let tokenId in data.tokenIds) {
+    // we are the source if the classId is not prefixed with the packet's sourcePort and sourceChannel
+    source = data.classId.slice(0, len(prefix)) !== prefix
+    if source {
+      // sender was source chain, unescrow tokens back to sender
+      escrowAccount = channelEscrowAddresses[packet.srcChannel]
+      // assert that escrow account is token owner
+      abortTransactionUnless(escrowAccount === nft.getOwner(data.classId, tokenId))
```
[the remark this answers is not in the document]
haifengxi: Ditto.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,385 @@
...
+}
+```
+
+#### Reasoning
+
+##### Correctness
+
+This implementation preserves token uniqueness.
+
+Uniqueness: If tokens have been sent to the counterparty chain, they can be redeemed back in the same `classId` & `tokenId` on the source chain.
```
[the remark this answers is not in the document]
haifengxi: The current spec indeed guarantees that.  Please see above answer about ownership.

spec/app/ics-721-nft-transfer/README.md
haifengxi: ICS20 talks about fungibility and supply here, which are two concepts that really don't apply to NFTs, as Chris commented earlier.

spec/app/ics-721-nft-transfer/README.md
haifengxi: This paragraph is inherited from ICS20 (implemented as _denom traces_). In the case of ICS-721, this will be _class traces_, in the exact same way as denom traces have been implemented.

spec/app/ics-721-nft-transfer/README.md
haifengxi: One way to enforce the immutability of NFT metadata would be to update the implementation in [x/nft](https://github.com/cosmos/cosmos-sdk/blob/master/x/nft/keeper/nft.go) to prevent anyone from updating the metadata (i.e., `uri` and `uri_hash`) of **IBC tokens** they own.

We shouldn't worry about IBC tokens locked in (owned by) escrow accounts along the IBC path because no one can update them anyway.

haifengxi: > Is it possible for the classUri to change after initialization? Similar question regarding tokenUris?
> 
> > Avoiding metadata divergence is something that can and should be enforced at the application layer.
> 
> This can be done at the point of transfer, but if the metadata changes at a later date, it may be difficult to propogate this information over to wherever the NFT is currently

It's possible that the `classUri` and other metadata (e.g., name, symbol) of an NFT class could change after associated NFTs have been transferred out via IBC.  However, this is not a new thing, we have the same "problem" with fungible tokens, as the metadata (name, symbol etc.) of a fungible token can change at any time.  Such issues are being resolved out-of-band by UI teams and the same should be done with NFT classes.

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
mpoke: > NFTs that were previously transferred out had been locked in the escrow account at the sending chain.

Here you're breaking the loop and returning a negative ACK. This means that the IBC packet was unsuccessful. However, for all the NFTs before the current one (that stopped the loop) `nft.Transfer()` was called. Are these calls changing state?Are these changes persisted in the next state or are they rolled back?

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
mpoke: Same as the above comment. Is `nft.Mint()` changing state?

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
mpoke: > Will add reference to [adr-043](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/adr-043-nft-module.md) where the NFT module is spec'ed out.

Adding a reference is good. However, the specification should contain all the information necessary to reason about correctness. This information can be in the form of assumptions made about different modules of the ABCI application (e.g., the NFT module). The reason for this is twofold. First, the reader of this spec shouldn't have to read the spec of the NFT module. Second, more importantly, if the spec of the NFT module changes, those changes need to be reflected in this spec. By explicitly stating the assumptions about the NFT module, this spec will be linked to a certain version of the NFT module.  Once the NFT module changes to the point that the assumptions need to be modified, this spec will be updated to a new version.

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
mpoke: Ownership is an important concept for reasoning about the correctness of this spec. Thus, it needs to be described in this specification. At the moment, it is not clear in the spec how the ownership changes.

spec/app/ics-721-nft-transfer/README.md
mpoke: My comment was about the "Optional addenda" section.

This seems to be an answer for the "Multi-chain notes" section. My suggestion was to try to avoid the redundancy; it's minor though, so you can ignore.

spec/app/ics-721-nft-transfer/README.md
robert-zaremba: I think it's too limited for a general usage, if all dapps will need to freeze a token metadata when IBCing it. I think adding a timestamp makes most of sense, and clarifying on the protocol side - that it should be taken with consideration that it's not a guaranteed by the standard that the token metadata (in particular the URI) won't change. 
So that would be a not required , maybe redundant information, but maybe useful for some use cases.

robert-zaremba: TBH, ideally we would like to use DID here, an reference the NFT with DID. I was talking already about it during the initial NFT proposal last summer. For functioning we would need a resolver, but this is not needed to ship the NFT standard, and that functionality could be added later without changing the standard or the base protocol.

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
robert-zaremba: BTW, what do you mean with "last prefixed port and channel pair"? 
"Last" in which sense? Recently used? Could you add a context?

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
robert-zaremba: Probably, the word we want to use here is _Redeemability_

Uniqueness: It's guaranteed that for any chain C1, and C2, C3 and any NFT  N1 created \in C1, N2 \in C2 and N3 \in C3: when IBC transferring N1 and N2 to C3, then all N1, N2 and N3 will have different ID and won't collide.

robert-zaremba:
How is the PO for this work? We need to put more pressure to finalize it because x/nft is being release with 0.46

haifengxi:
> How is the PO for this work? We need to put more pressure to finalize it because x/nft is being release with 0.46

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
haifengxi: Will add my comment above and reference to adr-043 to the spec.

spec/app/ics-721-nft-transfer/README.md
haifengxi: Will update the Multi-chain notes section to make it more clear

spec/app/ics-721-nft-transfer/README.md
haifengxi: > I think it's too limited for a general usage, if all dapps will need to freeze a token metadata when IBCing it. I think adding a timestamp makes most of sense, and clarifying on the protocol side - that it should be taken with consideration that it's not a guaranteed by the standard that the token metadata (in particular the URI) won't change. So that would be a not required , maybe redundant information, but maybe useful for some use cases.

Essentially, the NFT metadata -- a piece of descriptive information stored off-chain -- tells dApps what the NFT looks like (image/logo) and what traits (properties) are visually presentable to human users.

For an NFT that represents digital artwork, its metadata is all that matters. For other more "business-heavy" scenarios where, for example, an NFT represents digital security, the token's on-chain **data** holds the security's business state and matters more than its metadata.

Now, let's look at why people want to transfer their NFTs to other chains. Nowadays in the domain of public blockchains, most if not all of the motivation is to transfer tokens to a popular NFT marketplace for trading, in which case NFT metadata in a format that is compatible with the target marketplace (OpenSea being the dominant standard today) is of utmost importance.

What we had in mind, maybe unconsciously, for this ICS721 spec is just for that purpose -- sending tokens to popular NFT marketplace for trading, that's why our IBC packet only carries along `tokenUri` (pointing to metadata) but not NFT data itself.

If we were to send NFTs to another chain to take advantage of its enhanced on-chain capabilities that are not available on the source chain, then we must pass along NFT data in our IBC packets as well (of course, the NFT data must be in a format the target chain can understand). For example, we may want to send our digital security tokens to a chain that support security mortgaging.

IMHO, in the context of NFT trading, metadata represents the unique value people trade for and therefore should be change along the IBC path.

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
haifengxi: Will clarify the prefix concept with an example.

haifengxi:
> How is the PO for this work? We need to put more pressure to finalize it because x/nft is being release with 0.46

IMO, this spec is almost ready for implementation.

spec/app/ics-721-nft-transfer/README.md
robert-zaremba: I agree. That's why I propose to include the timestamp ( the time of tracked the last modification) as well. Or hash of the whole onchain object.

haifengxi: Thanks for the comment. As you said, this timestamp/hash is not required and may be redundant; the cost and trouble involved -- adding a field in NFT object (updating `x/nft`) and adding a field in the IBC packet -- most likely outweighs the benefits.

Updating IBC token metadata (`tokenUri`) by some sink chain is done through a local tx, so there is really no need for a separate timestamp as the information is already available in the on-chain tx. (redundant)

On the other hand, even if we pass the timestamp of latest metadata update back to the source chain in the IBC packet, we can't do much about it because there is no way for us to verify, in the protocol, whether the off-chain resource referenced by the updated uri is compatible/usable by the source-chain dApps; therefore, we can't and shouldn't override `tokenUri` in the un-escrowed token.

So, maybe it's safe to skip this timestamp/hash field for now, you think?

robert-zaremba: x/nft is not yet released, so there is no much cost to add additional field. I would like to hear more opinions if there is a benefit from the business side for including timestamp or hash. PS: for CID links, hash is not needed, so that could be a workaround.

spec/app/ics-721-nft-transfer/README.md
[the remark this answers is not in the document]
AdityaSripal: ditto

AdityaSripal (approved):
Great work on adopting the ICS-20 logic.  I think the only open question for me is the mutability of the original NFT at the source. It needs to be clarified how this should be handled if the "live" version of the NFT is living on a different chain.

This can be addressed in a separate PR since it is a longer discussion. The spec outside of the mutability question looks good to me so unblocking.

mpoke (approved):
Great work. Some minor comments below.

spec/app/ics-721-nft-transfer/README.md
haifengxi: I believe this is the last issue to be resolved for this spec: do we enforce the freezing of NFT token `uri` in ICS-721 spec? @AdityaSripal

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,395 @@
...
+
+Each send to any chain other than the one from which the token was previously received is a movement forward in the token's timeline. This causes trace to be added to the token's history and the destination port and destination channel to be prefixed to the `classId`. In these instances the sender chain is acting as the *source zone*. When the token is sent back to the chain it was previously received from, the prefix is removed. This is a backward movement in the token's timeline and the sender chain is acting as the *sink zone*.
```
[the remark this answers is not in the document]
mpoke: The previous paragraph starts with "A sending chain may be acting as a source or sink zone." Then it goes on explaining what a source zone is, i.e., 
> When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone

and what a sink zone is, i.e., 
> When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone.

spec/app/ics-721-nft-transfer/README.md
cwgoes: Maybe we don't need to try to answer this question in the specification, but I still think it's a recipe for ecosystem-level disaster if NFT infrastructure ends up not checking the source chain. Even if there is some option to freeze the metadata here, the infrastructure then needs to check that every intermediate chain implemented this ICS correctly and didn't malleate the metadata... of course, there is a similar concern for ICS20 multi-hop transfer - but the infrastructure should be checking there too!

I don't want to block too long on this concern, however. Could we add a paragraph writeup explaining the risks involved with using this field?

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,460 @@
...
+##### Correctness
+
+This implementation preserves token non-fungibility and redeemability.
+
+* Non-fungibility: Only one instance of any token is *live* across all the IBC-connected blockchains.
+* Redeemability: If tokens have been sent to the counterparty chain, they can be redeemed back in the same `classId` & `tokenId` on the source chain.
+
+##### Multi-chain notes
+
+This specification does not directly handle the "diamond problem", where a user sends a token originating on chain A to chain B, then to chain D, and wants to return it through D -> C -> A — since the token is tracked as owned by chain B (and the `classId` will be "{portOnD}/{channelOnD}/{portOnB}/{channelOnB}/classId"), chain C cannot serve as the intermediary. The original path, in reverse order, is required to redeem the token at its source chain. Complexities arising from long redemption paths may lead to the emergence of central chains in the network topology.
```
cwgoes, line 416: This text looks mostly lifted from the ICS20 spec. Better to remove it here and reference the ICS20 specs, as the problem is the same in both cases.

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,460 @@
...
+
+In order to track all of the tokens moving around the network of chains in various paths, it may be helpful for a particular chain to implement a registry which will track the "global" source chain for each `classId`. End-user service providers (such as wallet authors) may want to integrate such a registry or keep their own mapping of canonical source chains and human-readable names in order to improve UX.
```
cwgoes, line 418: Ditto (we should even consider removing this from the ICS20 specs)

cwgoes (approved):
ACK, but see comments

aofengli:
How is the current progress? When will it be possible to merge?

AdityaSripal:
@haifengxi , this passes the requisite checks. Please check @cwgoes comments to see if there are any last changes you would like to incorporate. If not, then let me know that this is mergable and I will merge, thanks.

spec/app/ics-721-nft-transfer/README.md
AdityaSripal: Agreed with Chris, that for now we can just write a paragraph that users are responsible for verifying metadata at the source chain

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -0,0 +1,460 @@
...
+- Fault containment: prevents Byzantine-creation of tokens originating on chain `A`, as a result of chain `B`'s Byzantine behavior.
+
+## Technical Specification
+
+### Data Structures
+
+Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the class id, class uri, token id's, token uri's, sender address, and receiver address.
+
+```typescript
+interface NonFungibleTokenPacketData {
```
humanalgorithm, line 41: Not sure if this has been asked before but
Is there any reason for not* having a flexible field here?
that way if someone wants to put in royalty or <whichever> data they can add it at their discretion.

haifengxi: Decision has been made earlier to implement business logic such as royalties as IBC middleware or application module.

Raulnori77:
Cosmos de caballeros del zodiaco?
Importantes necesito trabajo, work, empleo, chamba, me entienden?  O
quieren seguir jugando? Cuando quieran vemos todo el cosmos que quieran.
🗺️

PR #616, opened 2021-11-11
spec: updating ics27 spec to ensure parity with implementation

seantking:
This PR is updating the spec to have parity with the latest implementation. See [here](https://github.com/cosmos/ibc-go/pull/380).

Some improvements to the UX have been made by making use of ICS-04 channel negotiation (shoutout to @colin-axner  & @AdityaSripal for these ideas) as well as taking advantage of ICS-30 middleware architecture which replaces the previously existing `hooks` functionality for custom logic on `Success/Fail` of ICS-27 packets. 

I have also updated the introduction to be more clear. As well as various small updates throughout in order to align with the latest changes to the implementation. 

To be clear: **These changes are already implemented in code** & primarily effect the UX of the interchain accounts API & developer experience.

### DIFF
--- a/spec/app/ics-027-interchain-accounts/README.md
+++ b/spec/app/ics-027-interchain-accounts/README.md
@@ -16,7 +16,7 @@ This standard document specifies packet data structure, state machine handling l
 
 ### Motivation
 
-On Ethereum, there are two types of accounts: externally owned accounts, controlled by private keys, and contract accounts, controlled by their contract code ([ref](https://github.com/ethereum/wiki/wiki/White-Paper)). Similar to Ethereum's contract accounts, interchain accounts are controlled by another chain (not a private key) while retaining all the capabilities of a normal account (i.e. stake, send, vote, etc). While an Ethereum CA's contract logic is performed within Ethereum's EVM, interchain accounts are managed by a separate chain via IBC in a way such that the owner of the account retains full control over how it behaves. ICS27-1 primarily targets the use cases of DAO investing and staking derivatives over IBC.
+ICS-27 Interchain Accounts outlines a cross-chain account management protocol built upon IBC. ICS-27 enabled chains can programmatically create accounts on other ICS-27 enabled chains & control these accounts via IBC transactions (instead of signing with a private key). Interchain accounts retain all of the capabilities of a normal account (i.e. stake, send, vote) but instead are managed by a separate chain via IBC in a way such that the owner of the accounts retain full control over how the accounts behave. 
 
 ### Definitions 
 
@@ -30,16 +30,19 @@ The IBC handler interface & IBC relayer module interface are as defined in [ICS 
 ### Desired Properties
 
 - Permissionless 
+- Fault tolerance: A controller chain must not be able to control accounts registered by other controller chains. For example, in the case of a fork attack on a controller chain, only the interchain accounts registered by the forked chain will be vulnerable
 - The ordering of transactions sent to an interchain account on a host chain must be maintained. Transactions should be executed by an interchain account in the order in which they are sent by the controller chain
 - If a channel closes, the controller chain must be able to regain access to registered interchain accounts by simply opening a new channel
-- Each interchain account is owned by a single account on the controller chain. Only the owner account on the controller chain is authorized to control the interchain account
-- The controller chain must store the account address of the interchain account
+- Each interchain account is owned by a single account on the controller chain. Only the owner account on the controller chain is authorized to control the interchain account. The controller chain is responsible for enforcing this logic
+- The controller chain must store the account address of any owned interchain account's registered on host chains
+- A host chain must have the ability to limit interchain account functionality on its chain as necessary (e.g. a host chain can decide that interchain accounts registered on the host chain cannot take part in staking)
+
 
 ## Technical Specification
 
 ### General Design 
 
-A chain can implement one or both parts of the interchain accounts protocol (*controlling* and *hosting*). A controller chain that registers accounts on other host chains (that support interchain accounts) does not necessarily have to allow other controller chains to register accounts on its chain, and vice versa. 
+A chain can utilize one or both parts of the interchain accounts protocol (*controlling* and *hosting*). A controller chain that registers accounts on other host chains (that support interchain accounts) does not necessarily have to allow other controller chains to register accounts on its chain, and vice versa. 
 
 This specification defines the general way to register an interchain account and transfer tx bytes to control the account. The host chain is responsible for deserializing and executing the tx bytes, and the controller chain should know how the host chain will handle the tx bytes in advance (Cosmos SDK chains will deserialize using Protobuf). 
 
@@ -49,52 +52,38 @@ This specification defines the general way to register an interchain account and
 
 // * CONTROLLER CHAIN *
 
-// InitInterchainAccount is the entry point to registering an interchain account.
-// It generates a new port identifier using the owner address, connection identifiers
-// The port id will look like: ics27-1-{connection-number}-{counterparty-connection-number}-{owner-address}
-// and counterparty connection identifier. It will bind to the port identifier and
+// InitInterchainAccount is the entry point to registering an interchain account 
+// It generates a new port identifier using the owner address & connection identifiers
+// The port id will look like: ics27-1.{connection-id}.{counterparty-connection-id}.{owner-address}
+// It will bind to the port identifier and
 // call 04-channel 'ChanOpenInit'. An error is returned if the port identifier is already in use
 // An `OnChannelOpenInit` event is emitted which can be picked up by an offchain process such as a relayer
-function InitInterchainAccount(connectionId: string, counterPartyConnectionId: string, ownerAddress: string) returns (nil){
+// The account will be registered during the OnChanOpenTry step on the host chain
+function InitInterchainAccount(connectionId: string, counterPartyConnectionId: string, owner: string) returns (error){
 }
 
-// TrySendTx is used to send an IBC packet containing instructions (messages) to an interchain account on a host chain for a given interchain account owner. 
-function TrySendTx(connectionId: string, counterPartyConnectionId: string, ownerAddress: string, messages: []bytes) returns ([]bytes, error){
-    // A port id string will be generated from the connectionIds + ownerAddress. This is the port-id that the IBC packet will be sent on
-    // A call to GetActiveChannel() checks if there is a currently active channel for this port-id which also implies an interchain account has previously be registered
-    // A call to check if the channel status is OPEN (an additional, albeit slightly redundant check)
+// TrySendTx is used to send an IBC packet containing instructions (messages) to an interchain account on a host chain for a given interchain account owner 
+function TrySendTx(channelCapability: ChannelCapability, portID: string, connectionId: string, counterPartyConnectionId: string, icaPacketData: InterchainAccountPacketData) returns (uint64, error){
+    // A call to GetActiveChannel() checks if there is a currently active channel for this port-id which also implies an interchain account has been registered using this port identifier
     // if there are no errors CreateOutgoingPacket() is called and the IBC packet will be sent to the host chain on the active channel
 }
 
-// This helper function is required for the controller chain to get the address of a newly registered interchain account on a host chain.
-// Because the registration of an interchain account happens during the channel creation handshake, there is no way for the controller chain to know what the address of the interchain account is on the host chain in advance. 
-// This function sends an IBC packet to the host chain, on the owner port + active channel with the sole intention of eventually parsing the interchain account address from the Acknowledgement packet on the controller chain side.
-// The OnAcknowledgePacket function on the controller chain will handle the parsing + setting the interchain account address in state.
-// The controller chain builds the messages (before sending via IBC in the TrySendTx fn) that the host side will eventually execute. Therefore, the interchain account address must be known by the controller chain.
-function GetInterchainAccountAddressFromAck(connectionId: string, counterPartyConnectionId: string, ownerAddress: string) returns (nil){
-    // Sends a generic IBC packet to the host chain with the intention of parsing the interchain account address associated with this port/connection/channel from the Acknowledgement packet.
-}
 
 // Opens a new channel on a particular port given a connection
 // This is a helper function to open a new channel 
 // This is a safety function in case of a channel closing and the controller chain needs to regain access to an interchain account on the host chain 
 function InitChannel(portId: string, connectionId: string) returns (nil){
-  // An `OnChannelOpenInit` event is emitted which can be picked up by an offchain process such as a relayer which will finish the channel opening handshake
+  // An `OnChannelOpenInit` event is emitted which can be picked up by an off-chain process such as a relayer which will finish the channel opening handshake
+  // The active channel will be set to the newly opened channel on the OnChanOpenAck & OnChanOpenConfirm steps
 }
 
 // * HOST CHAIN *
 
-// RegisterInterchainAccount is called on the OnOpenTry step during the channel creation handshake 
-function RegisterInterchainAccount(counterPartyPortId: string) returns (error){
-   // generates an address for the interchain account 
+// RegisterInterchainAccount is called on the OnChanOpenTry step during the channel creation handshake 
+function RegisterInterchainAccount(accAddr: string, counterPartyPortId: string) returns (nil){
    // checks to make sure the account has not already been registered
    // creates a new address on chain 
    // calls SetInterchainAccountAddress()
-   // returns the address of the newly created account
-}
-
-// DeserializeTx is used to deserialize message bytes parsed from an IBC packet into a format that the host chain can recognize & execute i.e. cosmos SDK messages
-function DeserializeTx(txBytes []byte) returns ([]Any, error){
 }
 
 // AuthenticateTx is called before ExecuteTx.
@@ -113,11 +102,11 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
 // * UTILITY FUNCTIONS *
 
 // Sets the active channel
-function SetActiveChannel(portId: string, channelId: string) returns (error){
+function SetActiveChannelID(portId: string, channelId: string) returns (error){
 }
 
 // Returns the id of the active channel if present
-function GetActiveChannel(portId: string) returns (string, boolean){
+function GetActiveChannelID(portId: string) returns (string, boolean){
 }
 
 // Stores the address of the interchain account in state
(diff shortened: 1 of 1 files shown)


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -30,16 +30,19 @@ The IBC handler interface & IBC relayer module interface are as defined in [ICS
 ### Desired Properties
 
 - Permissionless 
+- Fault tolerance: A controller chain must not be able to control accounts registered by other controller chains. For example, in the case of a fork attack on a controller chain, only the interchain accounts registered by the forked chain will be vulnerable
 - The ordering of transactions sent to an interchain account on a host chain must be maintained. Transactions should be executed by an interchain account in the order in which they are sent by the controller chain
 - If a channel closes, the controller chain must be able to regain access to registered interchain accounts by simply opening a new channel
-- Each interchain account is owned by a single account on the controller chain. Only the owner account on the controller chain is authorized to control the interchain account
-- The controller chain must store the account address of the interchain account
+- Each interchain account is owned by a single account on the controller chain. Only the owner account on the controller chain is authorized to control the interchain account. The controller chain is responsible for enforcing this logic
```
AdityaSripal, line 36:
```suggestion
- Each interchain account is owned by a single account on the controller chain. Only the owner account on the controller chain is authorized to control the interchain account. The controller chain is responsible for enforcing this logic.
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -49,52 +52,38 @@ This specification defines the general way to register an interchain account and
 
 // * CONTROLLER CHAIN *
 
-// InitInterchainAccount is the entry point to registering an interchain account.
-// It generates a new port identifier using the owner address, connection identifiers
-// The port id will look like: ics27-1-{connection-number}-{counterparty-connection-number}-{owner-address}
-// and counterparty connection identifier. It will bind to the port identifier and
+// InitInterchainAccount is the entry point to registering an interchain account 
+// It generates a new port identifier using the owner address & connection identifiers
```
AdityaSripal, line 56:
```suggestion
// It generates a new port identifier using the owner address and connection identifiers
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
...
+5. During the `OnChanOpenAck` callback on the controller chain a record of the interchain account address registered on the host chain during `OnChanOpenTry` is set in state with a mapping from portID -> interchain account address. See [version negotiation](#Version-Negotiation) section below for how to implement this
+6. During the `OnChanOpenAck` & `OnChanOpenConfirm` callbacks on the controller & host chains respectively, the [active-channel](#Active-Channels) for this interchain account/owner pair, is set in state
+
+#### Active Channels
 
 The controller and host chain should keep track of an `active-channel` for each registered interchain account. The `active-channel` is set during the channel creation handshake process. This is a safety mechanism that allows a controller chain to regain access to an interchain account on a host chain in case of a channel closing. 
 
-An active channel can look like this:
+An active channel on the controller chain will look like this:
```
AdityaSripal, line 146:
```suggestion
An example of an active channel on the controller chain can look like this:
```

AdityaSripal: This is not the struct for all active channels but just a single example

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
...
+
+2. The controller chain emits an event signaling to open a new channel on this port given a connection 
+3. A relayer listening for `OnChannelOpenInit` events will begin the channel creation handshake
+4. During the `OnChanOpenTry` callback on the host chain an interchain account will be registered and a mapping of the interchain account address to the owner account address will be stored in state (this is used for authenticating transactions on the host chain at execution time). 
+5. During the `OnChanOpenAck` callback on the controller chain a record of the interchain account address registered on the host chain during `OnChanOpenTry` is set in state with a mapping from portID -> interchain account address. See [version negotiation](#Version-Negotiation) section below for how to implement this
+6. During the `OnChanOpenAck` & `OnChanOpenConfirm` callbacks on the controller & host chains respectively, the [active-channel](#Active-Channels) for this interchain account/owner pair, is set in state
```
AdityaSripal, line 140: Conclude each statement with `.`

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
...
 
-An active channel can look like this:
+An active channel on the controller chain will look like this:
 
 
 ```typescript
 {
  // Controller Chain
- SourcePortId: `ics-27-0-0-cosmos1mjk79fjjgpplak5wq838w0yd982gzkyfrk07am`,
+ SourcePortId: `ics27-1.0.0.<owner-id>`,
```
AdityaSripal, line 152: Define the structure of `owner-id`

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
...
+To register an interchain account we require an off-chain process (relayer) to listen for `OnChannelOpenInit` events with the capability to finish a channel creation handshake on a given connection. 
+
+1. The controller chain binds a new IBC port with an id composed of the *source/counterparty connection-ids* & the *interchain account owner address*
+
+The IBC port identifier will look like this:
+```
+ics27-1.{connection-id}.{counterparty-connection-id}.{owner-address}
+```
+
+This port will be used to create channels between the controller & host chain for a specific owner/interchain account pair. Only the account with `{owner-address}` matching the bound port will be authorized to send IBC packets over channels created with `ics27-1.{connection-id}.{counterparty-connection-id}.{owner-address}`. It is up to each controller chain to enforce this port registration and access on the controller side.
```
AdityaSripal, line 134: `owner-address` is the account address on the controller chain, correct?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
...
 
-Once an interchain account is registered on the host chain a controller chain can begin sending instructions (messages) to control this account. 
+#### Version negotiation summary
 
-1. The controller chain calls `GetInterchainAccountAddressFromAck()` to get the address of the interchain account on the host chain and sets the address in state mapped to the respective portID. 
-2. The controller chain calls `TrySendTx` and passes message(s) that will be executed on the host side by the associated interchain account (determined by the source port identifer)
+| Initiator | Datagram         | Chain acted upon |  Version (Controller, Host) | Port ID (Controller, Host) |
+| --------- | ---------------- | ---------------- |  ---------------------- | ---------------------- |
+| Controller| ChanOpenInit     | Controller       |  (ics27-1, none)        | (ics27-1.0.0.<owner-id>, interchain-account)           |
+| Relayer   | ChanOpenTry      | Host             |  (ics27-1, ics27-1.<owner-id>)        | (ics27-1.0.0.<owner-id>, interchain-account)       |
```
AdityaSripal, line 178: This table seems incorrect. `owner-id` sounds like the account address on the controller chain since that is the "owner" of the interchain account.

But from my understanding, the version set by host chain contains the address of the host chain that is being controlled by the owner address on the controller chain.

I think the address in version string should have a different name: `host-address`?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
...
+| Relayer   | ChanOpenAck      | Controller       |  (ics27-1, ics27-1.<owner-id>)        | (ics27-1.0.0.<owner-id>, interchain-account)        | (ics27-1.0.0.<owner-id>, interchain-account)        |
+| Relayer   | ChanOpenConfirm  | Host             |  (ics27-1, ics27-1.<owner-id>)        | (ics27-1.0.0.<owner-id>, interchain-account)           | (ics27-1.0.0.<owner-id>, interchain-account)           |
+
+
```
AdityaSripal, line 182: Briefly describe how controller chain can retrieve host address from counterparty version

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
...
 msg := &banktypes.MsgSend{FromAddress: interchainAccountAddress, ToAddress: ToAddress, Amount: amount}
+icaPacketData = InterchainAccountPacketData{
+   Type: types.EXECUTE_TX,
+   Data: serialize(msg),
+   Memo: "memo",
+}
+
 // Sends the message to the host chain, where it will eventually be executed 
-TrySendTx(ownerAddress, connectionId, counterPartyConnectionId, msg)
+TrySendTx(chanCapability, ownerAddress, connectionId, counterPartyConnectionId, data)
```
AdityaSripal, line 202:
```suggestion
TrySendTx(ownerAddress, connectionId, counterPartyConnectionId, data)
```

The capabilities aren't mentioned in the spec since it is SDK specific authentication schema. Similar to how we don't include `ctx sdk.Context` in our spec documents.

I would just say that it is responsibility of ICA module to authenticate that the caller of `TrySendTx` is the owner of the interchain account

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
 // * UTILITY FUNCTIONS *
 
 // Sets the active channel
```
AdityaSripal, line 104: for a given portID

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -113,86 +102,119 @@ function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, ms
...
-function SetActiveChannel(portId: string, channelId: string) returns (error){
+function SetActiveChannelID(portId: string, channelId: string) returns (error){
 }
 
 // Returns the id of the active channel if present
```
AdityaSripal, line 108: for a given portID

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal, line 152: You have `owner-id` in some places, and `owner-address` in some places. Standardize the terminology and define it to be the address on the controller chain that owns the interchain account.

AdityaSripal, line 152: Also it doesn't make sense here to give an example `connection-id` and `counterparty-connection-id` but leave the owner address as a variable.

Either give them all an example value, or leave them all as variables

AdityaSripal (changes requested):
The version negotiation table seems incorrect, and it would be good to clarify some terminology.

You may take or ignore the nits

spec/app/ics-027-interchain-accounts/README.md
seantking: great catch! this is indeed incorrect. <owner-address> here should be the interchain account address registered on the host chain.

spec/app/ics-027-interchain-accounts/README.md
seantking: Correct. I've been thinking lately that we should maybe use the terminology `owner-id` as it's up to the controller chain to manage this, and there's not any enforcement on the host side that this needs to be an account address. WDYT?

spec/app/ics-027-interchain-accounts/README.md
seantking: Cool. Good to know :handshake:

spec/app/ics-027-interchain-accounts/README.md
seantking: https://github.com/cosmos/ibc/pull/616#discussion_r747497839

seantking: Yeah, I agree, a good point. I removed the cosmos address as it was specific to the hub but I can put it back in.

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: Sounds fair, so long as it is consistent

spec/app/ics-027-interchain-accounts/README.md
seantking: I updated the variable name to be `interchain-account-address` as I'm hesitant to add further terminology here. Open to discussion. 

I also removed the port-id column as I noticed that I wasn't using variable names for the connection identifiers and adding them in makes the column difficult to parse. I think it's fine without.

spec/app/ics-027-interchain-accounts/README.md
seantking: I decided to keep this as the address of the account owner on the controller chain. I've replaced any reference of `owner-address` or `owner-id` with `owner-account-address`. 

I think the opening definition for `Interchain Account Owner` should make it obvious what the `owner-account-address` is representing. WDYT?

spec/app/ics-027-interchain-accounts/README.md
seantking: The reason I replaced the `<>` with `{}` is simply because using `<>` inside a table makes the content invisible on github README for some reason.

spec/app/ics-027-interchain-accounts/README.md
seantking: Done :+1:

spec/app/ics-027-interchain-accounts/README.md
seantking: update: https://github.com/cosmos/ibc/pull/616#discussion_r750099314

seantking: I actually put them all as variables in the end. :handshake:

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,127 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
+Combined with the one channel per interchain account approach, this method of version negotiation allows us to pass the address of the interchain account back to the controller chain and create a mapping from controller port ID -> interchain account address during the `OnChanOpenAck` callback. As outlined in the [controlling flow](#Controlling-Flow), a controller chain will need to know the address of a registered interchain account in order to send transactions to the account on the host chain.
 
-**Controlling Flow**
+#### Version Negotiation Summary
 
-Once an interchain account is registered on the host chain a controller chain can begin sending instructions (messages) to control this account. 
+| Initiator | Datagram         | Chain acted upon |  Version (Controller, Host) |
+| --------- | ---------------- | ---------------- |  ---------------------- | 
+| Controller| ChanOpenInit     | Controller       |  (ics27-1, none)        | 
+| Relayer   | ChanOpenTry      | Host             |  (ics27-1, ics27-1.{interchain-account-address})        |
```
AdityaSripal, line 177: Define somewhere (either above or below this table), that `interchain-account-address` is the address on the host chain that is controlled by the owner through the given ICA channel

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -308,58 +332,61 @@ function onChanOpenAck(
...
 
 ```typescript
 function onChanCloseInit(
   portIdentifier: Identifier,
-  channelIdentifier: Identifier) {
-    // users should not be able to close channels
-    return nil
+  channelIdentifier: Identifier) { \
+    // unset the active channel for this owner/interchain account pair
+    DeleteActiveChannel(portIdentifier, channelIdentifier)
```
AdityaSripal, line 345: Shouldn't this return an error always? Ahh, is this because the ICA-auth may decide to close the channel? If so, then the line above about ICA channels only closing on timeouts should be changed

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -308,58 +332,61 @@ function onChanOpenAck(
...
+	// was successfully decoded
+	if ack.Success() {
+		switch data.Type {
+	          case types.EXECUTE_TX:
+		        msgs, err := types.DeserializeTx(data.Data)
+		        if err != nil {
+			      return err
+		        }
+
+		        if err = executeTx(ctx, packet.SourcePort, packet.DestinationPort, packet.DestinationChannel, msgs); err != nil {
```
AdityaSripal, line 375: Does `executeTx` include authentication? If so that should be commented here.

AdityaSripal (changes requested):
Only required changes are clarification on voluntary channel closing, and documentation of `executeTx`

spec/app/ics-027-interchain-accounts/README.md
seantking: This should return an error you are correct. I've updated this.

AdityaSripal (approved):
Nice work!!

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,135 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
+To register an interchain account we require an off-chain process (relayer) to listen for `OnChannelOpenInit` events with the capability to finish a channel creation handshake on a given connection. 
+
+1. The controller chain binds a new IBC port with an id composed of the *source/counterparty connection-ids* & the *interchain account owner address*
+
+The IBC port identifier will look like this:
+```
+ics27-1.{connection-id}.{counterparty-connection-id}.{owner-account-address}
+```
+
+This port will be used to create channels between the controller & host chain for a specific owner/interchain account pair. Only the account with `{owner-account-address}` matching the bound port will be authorized to send IBC packets over channels created with `ics27-1.{connection-id}.{counterparty-connection-id}.{owner-account-address}`. It is up to each controller chain to enforce this port registration and access on the controller side.
```
adizere, line 140:
```suggestion
This port will be used to identify the channel(s) created between the controller & host chain for a specific owner/interchain account. Only the account with `{owner-account-address}` matching the bound port will be authorized to send IBC packets over channels created with port id `ics27-1.{connection-id}.{counterparty-connection-id}.{owner-account-address}`. It is up to each controller chain to enforce this port registration and access on the controller side. 
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,135 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
+
+The IBC port identifier will look like this:
+```
+ics27-1.{connection-id}.{counterparty-connection-id}.{owner-account-address}
+```
+
+This port will be used to create channels between the controller & host chain for a specific owner/interchain account pair. Only the account with `{owner-account-address}` matching the bound port will be authorized to send IBC packets over channels created with `ics27-1.{connection-id}.{counterparty-connection-id}.{owner-account-address}`. It is up to each controller chain to enforce this port registration and access on the controller side. 
+
+2. The controller chain emits an event signaling to open a new channel on this port given a connection 
+3. A relayer listening for `OnChannelOpenInit` events will begin the channel creation handshake
```
adizere, line 143:
```suggestion
3. A relayer listening for `OnChannelOpenInit` events will continue the channel creation handshake
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,135 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
-An active channel can look like this:
+An example of an active channel on the controller chain can look like this:
 
 
 ```typescript
 {
  // Controller Chain
- SourcePortId: `ics-27-0-0-cosmos1mjk79fjjgpplak5wq838w0yd982gzkyfrk07am`,
- SourceChannelId: `channel-1`,
+ SourcePortId: `ics27-<version>.<source-connection-id>.<counterparty-connection-id>.<owner-account-address>`,
```
adizere, line 158: It may be confusing to mix the "source" terminology with "counterparty". Consider using _**one**_ of these pairs of terms:

- `sourcePortId` and `destinationPortId`
- `portId` and `counterpartyPortId`

This is also in line with other ICS-es ([example](https://github.com/cosmos/ibc/tree/b25d9b8b2bae739821703404266256bb9a741027/spec/core/ics-004-channel-and-packet-semantics)).

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,135 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
 
-To register an interchain account we require an off-chain process (relayer) to listen for `OnChannelOpenInit` events with the capability to finish a channel creation handshake on a given connection. 
+#### Version Negotiation
 
-1. The controller chain binds a new IBC port with an id composed of the *source/counterparty connection-ids* & the *interchain account owner address*
+ICS-27 takes advantage of [ISC-04 channel version negotiation](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#versioning) to store a mapping of the controller chain port ID to the newly registered interchain account address, on both the host chain & controller chains, during the channel creation handshake ([account registration flow](#Register-Account-Flow)).
```
adizere, line 170:
```suggestion
ICS-27 takes advantage of [ICS-004 channel version negotiation](https://github.com/cosmos/ibc/tree/master/spec/core/ics-004-channel-and-packet-semantics#versioning) to store a mapping of the controller chain port ID to the newly registered interchain account address, on both the host chain & controller chains, during the channel creation handshake ([account registration flow](#Register-Account-Flow)). 
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,135 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
-5. During the `OnChanOpenAck` & `OnChanOpenConfirm` callbacks on the controller & host chains respectively, the `active-channel` for this interchain account/owner pair, is set in state
+Due to how the mechanics of ICS-004 channel version negotiation operate the version passed into the host chain side (OnChanOpenTry) must match the counterparty version passed into the controller side (OnChanOpenAck) otherwise, there will be a resulting error at the IBC protocol level. 
+
+Combined with the one channel per interchain account approach, this method of version negotiation allows us to pass the address of the interchain account back to the controller chain and create a mapping from controller port ID -> interchain account address during the `OnChanOpenAck` callback. As outlined in the [controlling flow](#Controlling-Flow), a controller chain will need to know the address of a registered interchain account in order to send transactions to the account on the host chain.
+
+#### Version Negotiation Summary
+
+`interchain-account-address` is the address of the interchain account registered on the host chain by the controller chain.
```
adizere, line 180: Very clear!

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,135 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
+| Relayer   | ChanOpenAck      | Controller       |  (ics27-1, ics27-1.{interchain-account-address})        | 
+| Relayer   | ChanOpenConfirm  | Host             |  (ics27-1, ics27-1.{interchain-account-address})        | 
 
+The channel `version` string passed into the `OnChannelOpenTry` callback by a relayer will contain the account address of the interchain account that will be registered on the host chain. A relayer will then pass this version into the `OnChannelOpenAcknowledgement` step on the controller chain as the `counterpartyVersion`. The controller chain will then proceed to parse the interchain account address from the `counterpartyVersion` string and store this address in state with a map to the associated port id.  
 
-**Controlling Flow**
+#### Controlling Flow
 
-Once an interchain account is registered on the host chain a controller chain can begin sending instructions (messages) to control this account. 
+Once an interchain account is registered on the host chain a controller chain can begin sending instructions (messages) to the host chain to control the account.
```
adizere, line 193: What does _registered_ mean? Does it mean that `ChanOpenConfirm` executed at the host chain?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,135 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
+#### Controlling Flow
 
-Once an interchain account is registered on the host chain a controller chain can begin sending instructions (messages) to control this account. 
+Once an interchain account is registered on the host chain a controller chain can begin sending instructions (messages) to the host chain to control the account. 
 
-1. The controller chain calls `GetInterchainAccountAddressFromAck()` to get the address of the interchain account on the host chain and sets the address in state mapped to the respective portID. 
-2. The controller chain calls `TrySendTx` and passes message(s) that will be executed on the host side by the associated interchain account (determined by the source port identifer)
+1. The controller chain calls `TrySendTx` and passes message(s) that will be executed on the host side by the associated interchain account (determined by the controller side port identifier)
 
 Cosmos SDK psuedo code example:
```
adizere, line 197:
```suggestion
Cosmos SDK pseudo-code example:
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -104,95 +93,135 @@ function AuthenticateTx(msgs []Any, portId string) error {
...
-TrySendTx(ownerAddress, connectionId, counterPartyConnectionId, msg)
+TrySendTx(ownerAddress, connectionId, counterPartyConnectionId, data)
 ```
 
-4. The host chain upon receiving the IBC packet will call `DeserializeTx` and then call `AuthenticateTx` for each message. If either of these steps fails an error will be returned.
-5. The host chain will then call `ExecuteTx` for each message and return an acknowledgment
-
+4. The host chain upon receiving the IBC packet will call `DeserializeTx`. 
+    
+5. The host chain will then call `ExecuteTx` & `AuthenticateTx` for each message and return an acknowledgment containing a success or error.
```
adizere, line 214:
```suggestion
5. The host chain will then call `AuthenticateTx` and `ExecuteTx` for each message and return an acknowledgment containing a success or error.  
```

adizere (approved):
Cool work Sean! Left some comments, mostly nits and questions.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,30 +16,33 @@ This standard document specifies packet data structure, state machine handling l
...
+ICS-27 Interchain Accounts outlines a cross-chain account management protocol built upon IBC. ICS-27 enabled chains can programmatically create accounts on other ICS-27 enabled chains & control these accounts via IBC transactions (instead of signing with a private key). Interchain accounts retain all of the capabilities of a normal account (i.e. stake, send, vote) but instead are managed by a separate chain via IBC in a way such that the owner of the accounts retain full control over how the accounts behave.
```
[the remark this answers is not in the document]
cwgoes: Can you clarify the last part of this sentence to indicate that there are two separate accounts involved (one on the controller chain, one on the controlled chain, or whatever the terminology is)?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,30 +16,33 @@ This standard document specifies packet data structure, state machine handling l
...
 
 ### Desired Properties
 
 - Permissionless 
-- The ordering of transactions sent to an interchain account on a host chain must be maintained. Transactions should be executed by an interchain account in the order in which they are sent by the controller chain
-- If a channel closes, the controller chain must be able to regain access to registered interchain accounts by simply opening a new channel
-- Each interchain account is owned by a single account on the controller chain. Only the owner account on the controller chain is authorized to control the interchain account
-- The controller chain must store the account address of the interchain account
+- Fault tolerance: A controller chain must not be able to control accounts registered by other controller chains. For example, in the case of a fork attack on a controller chain, only the interchain accounts registered by the forked chain will be vulnerable.
+- The ordering of transactions sent to an interchain account on a host chain must be maintained. Transactions should be executed by an interchain account in the order in which they are sent by the controller chain.
```
cwgoes, line 34: What if transactions fail? Do we want a sort of causal execution dependency here, or is that out of scope?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -16,30 +16,33 @@ This standard document specifies packet data structure, state machine handling l
...
+- If a channel closes, the controller chain must be able to regain access to registered interchain accounts by simply opening a new channel.
```
cwgoes, line 35: Really? Why, and how is this made safe?

cwgoes: I see that you tie accounts to the connection identifiers. I think this is fine, but I don't think it really gives any additional "fault liveness recovery", which I presume is what you want. If clients expire a new connection will need to be created in addition to new channels.

cwgoes (approved):
ACK to these changes

However, I would like to see a clearer design rationale for "alternate-channel recovery" (see comment), since it is not clear to me. If there is prior discussion elsewhere explaining this, I'm happy to take a look (but perhaps it should be linked in the spec as well).

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: So we are considering two possibilities for the channel to close:

1. Timeout on ORDERED channel

In this case, the user can simply open a new channel on the same connection with the same port and that new channel will be able to control the original interchain account.

2. Underlying client expires or gets frozen for misbehaviour.

Here ICA does not have any in-protocol method for recovery, you are right. However, the underyling layers may have their own recovery mechanisms such that we don't have to create new connections/channels.
In fact, this is exactly the case in the ibc-go implementation. If a client expires or is frozen, then governance can come in and revive the client such that the connections/channels can resume working without having to create a whole new one.
So here, we can rely on governance to hopefully resolve the issue.
It's possible of course that governance does not choose to do this, in which case, there is nothing we can do.

Perhaps this reasoning should be written down succinctly @seantking

seantking:
@cwgoes @adizere Thanks for the review! I will get back to your questions tomorrow as well as create a follow-up PR with further changes based on your feedback.

We have an audit beginning today from an external company so I would like to merge this in for now.

spec/app/ics-027-interchain-accounts/README.md
cwgoes: Right, OK, so the case that we're really concerned about here is (1), since that is the case where using connection identifiers allows for recovery in this case of timeouts of packets _but not_ the client. That makes sense to me. However, the semantics seem identical to (and more complicated than) an ordered channel which just doesn't close on timeouts, so maybe we should aim to add support for that instead and avoid the unnecessary extra handshake?

AFAIK (2) doesn't benefit from the possibility of using multiple channels so isn't relevant to this design question.

cwgoes:
Did you have a chance to create a follow-up PR? I'd definitely like to see clearer reasoning on the connection ID usage (see above comment) since it blurs the boundary of intended abstractions a bit imo (though I think it's fine from a safety perspective).

seantking:
> Did you have a chance to create a follow-up PR? I'd definitely like to see clearer reasoning on the connection ID usage (see above comment) since it blurs the boundary of intended abstractions a bit imo (though I think it's fine from a safety perspective).

Thank you for this feedback!! We discussed some of it on a call yesterday and agree with your rationale behind pushing for ordered channels that do not close on timeout (although, ideally this will be in a future upgrade as we are close to releasing, assuming everything is secure).

We will follow up with a PR to address some more of your concerns next week at the latest. :handshake:

spec/app/ics-027-interchain-accounts/README.md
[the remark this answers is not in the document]
seantking: > Can you clarify the last part of this sentence to indicate that there are two separate accounts involved (one on the controller chain, one on the controlled chain, or whatever the terminology is)?

I think it's important to be more clear on this, I agree. Perhaps, it is better to simply say that the controller chain is the owner of all accounts it registers on a host chain. In terms of the current implementation, the `owner-id` (which is in the portID on the controller chain) is arbitrary and does not need to be mapped to an account address per se but it is what we have been recommending. We have been dancing around the semantics of this for a while. In the end, it's always down to the controller chain to determine what the `owner-id` is exactly. 

The controller chain will always be responsible for building some level of authentication around the interchain accounts API (i.e. not exposing the API publically) and the 'owner' will often times not be an end-user but instead controlled via some on-chain protocol / governance. 

I think in a follow up PR we should change the semantics around this @AdityaSripal

PR #621, opened 2021-12-01
Add Connection Upgrade Spec

AdityaSripal:
This is a rough draft proposal for the ability to upgrade a connection once it has already been established

### DIFF
--- a/spec/core/ics-003-connection-semantics/UPGRADES.md
+++ b/spec/core/ics-003-connection-semantics/UPGRADES.md
@@ -48,14 +48,6 @@ interface ConnectionEnd {
 }
 ```
 
-```typescript
-interface UpgradeStatus {
-    state: ConnectionState
-    timeoutHeight: Height
-    timeoutTimestamp: uint64
-}
-```
-
 The desired property that the connection upgrade protocol may not modify the underlying clients or connection identifiers, means that only some fields are upgradable by the upgrade protocol.
 
 - `state`: The state is specified by the handshake steps of the upgrade protocol.
@@ -66,6 +58,18 @@ The desired property that the connection upgrade protocol may not modify the und
 - `version`: The version MAY be modified by the upgrade protocol. The same version negotiation that happens in the initial connection handshake can be employed for the upgrade handshake.
 - `delayPeriodTime`: The delay period MAY be modified by the upgrade protocol. The counterparty MUST accept the new proposed value or return an error during the upgrade handshake.
 - `delayPeriodBlocks`: The delay period MAY be modified by the upgrade protocol. The counterparty MUST accept the new proposed value or return an error during the upgrade handshake.
+
+```typescript
+interface UpgradeStatus {
+    state: ConnectionState
+    timeoutHeight: Height
+    timeoutTimestamp: uint64
+}
+```
+
+- UpgradeStatus contains the state of the upgrade. If upgrade is successful it will contain the `ConnectionState` of the upgrading connection on the executing chain. If the executing chain wants to abort the upgrade, it will restore its previous connection under its connection path with state `OPEN` and write an `UpgradeStatus` with state `UPGRADE_ERR`.
+- `timeoutHeight`: Timeout height indicates the height at which the counterparty must no longer proceed with the upgrade handshake. The chains will then preserve their original connection and the upgrade handshake is aborted.
+- `timeoutTimestamp`: Timeout timestamp indicates the time on the counterparty at which the counterparty must no longer proceed with the upgrade handshake. The chains will then preserve their original connection and the upgrade handshake is aborted.
 
 ```typescript
 interface UpgradeConnectionState {
@@ -98,7 +102,7 @@ function restorePath(id: Identifier): Path {
 The upgrade status path is a public path that can signal the status of the upgrade to the counterparty. It does not store anything in the successful case, but it will store a sentinel abort value in the case that a chain does not accept the proposed upgrade.
 
 ```typescript
-function upgradePath(id: Identifier): Path {
+function statusPath(id: Identifier): Path {
     return "connections/{id}/upgradeStatus"
 
 }
@@ -113,7 +117,7 @@ A successful protocol execution flows as follows (note that all calls are made t
 | Initiator | Datagram             | Chain acted upon | Prior state (A, B)          | Posterior state (A, B)      |
 | --------- | -------------------- | ---------------- | --------------------------- | --------------------------- |
 | Actor     | `ConnUpgradeInit`    | A                | (OPEN, OPEN)                | (UPGRADE_INIT, OPEN)        |
-| Relayer   | `ConnUpgradeTry`     | B                | (UPGRADE_INIT, OPEN)        | (UPGRADE_INIT, UPGRADE_TRY) |
+| Actor     | `ConnUpgradeTry`     | B                | (UPGRADE_INIT, OPEN)        | (UPGRADE_INIT, UPGRADE_TRY) |
 | Relayer   | `ConnUpgradeAck`     | A                | (UPGRADE_INIT, UPGRADE_TRY) | (OPEN, UPGRADE_TRY)         |
 | Relayer   | `ConnUpgradeConfirm` | B                | (OPEN, UPGRADE_TRY)         | (OPEN, OPEN)                |
 
@@ -151,7 +155,14 @@ function connUpgradeInit(
     // either timeout height or timestamp must be non-zero
     abortTransactionUnless(proposedUpgrade.TimeoutHeight != 0 || proposedUpgrade.TimeoutTimestamp != 0)
 
-    provableStore.set(connectionPath(identifier), proposedUpgrade)
+    upgradeStatus = UpgradeStatus{
+        state: UPGRADE_INIT,
+        timeoutHeight: proposedUpgrade.timeoutHeight,
+        timeoutTimestamp: proposedUpgrade.timeoutTimestamp,
+    }
+
+    provableStore.set(statusPath(identifier), upgradeStatus)
+    provableStore.set(connectionPath(identifier), proposedUpgrade.connection)
     privateStore.set(restorePath(identifier), currentConnection)
 }
 ```
@@ -159,5 +170,185 @@ function connUpgradeInit(
 ```typescript
 function connUpgradeTry(
     identifier: Identifier,
-    proposedUpgrade
-)
+    proposedUpgrade: UpgradeConnectionState,
+    counterpartyConnection: ConnectionEnd,
+    counterpartyUpgradeStatus: UpgradeStatus,
+    proofConnection: CommitmentProof,
+    proofUpgradeStatus: CommitmentProof,
+    proofHeight: Height
+) {
+    // current connection must be OPEN or UPGRADE_INIT (crossing hellos)
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == (OPEN || UPGRADE_INIT))
+
+    // abort transaction if an unmodifiable field is modified
+    // upgraded connection state must be in `UPGRADE_TRY`
+    abortTransactionUnless(
+        proposedUpgrade.connection.state == UPGRADE_TRY &&
+        proposedUpgrade.connection.counterpartyConnectionIdentifier == currentConnection.counterpartyConnectionIdentifier &&
+        proposedUpgrade.connection.clientIdentifier == currentConnection.clientIdentifier &&
+        proposedUpgrade.connection.counterpartyClientIdentifier == currentConnection.counterpartyClientIdentifier &&
+    )
+
+    // either timeout height or timestamp must be non-zero
+    abortTransactionUnless(proposedUpgrade.TimeoutHeight != 0 || proposedUpgrade.TimeoutTimestamp != 0)
+
+
+    // verify proofs of counterparty state
+    abortTransactionUnless(currentConnection.client.verifyConnectionState(proofHeight, proofConnection, counterpartyConnection))
+    abortTransactionUnless(currentConnection.client.verifyUpgradeStatus(proofHeight, proofUpgradeStatus, counterpartyUpgradeStatus))
+
+    // verify that counterparty connection unmodifiable fields have not changed and counterparty state
+    // is UPGRADE_INIT
+    restoreConnectionUnless(
+        counterpartyConnection.state == UPGRADE_INIT &&
+        counterpartyConnection.counterpartyConnectionIdentifier == identifier &&
+        counterpartyConnection.clientIdentifier == currentConnection.counterpartyClientIdentifier &&
+        counterpartyConnection.counterpartyClientIdentifier == currentConnection.clientIdentifier
+    )
+    restoreConnectionUnless(counterpartyUpgradeStatus.state == UPGRADE_INIT)
+
+    // counterparty-specified timeout must not have exceeded
+    restoreConnectionUnless(
+        counterparty.UpgradeStatus.TimeoutHeight < currentHeight() &&
+        counterparty.UpgradeStatus.TimeoutTimestamp < currentTimestamp()
+    )
+
+    // verify chosen versions are compatible
+    versionsIntersection = intersection(counterpartyConnection.version, proposedUpgrade.Connection.version)
+    version = pickVersion(versionsIntersection) // throws if there is no intersection
+
+    // both connection ends must be mutually compatible.
+    restoreConnectionUnless(IsCompatible(counterpartyConnection, proposedUpgrade.Connection))
+
+    upgradeStatus = UpgradeStatus{
+        state: UPGRADE_TRY,
+        timeoutHeight: proposedUpgrade.timeoutHeight,
+        timeoutTimestamp: proposedUpgrade.timeoutTimestamp,
+    }
+
+    provableStore.set(statusPath(identifier), upgradeStatus)
+    provableStore.set(connectionPath(identifier), proposedUpgrade.connection)
+    privateStore.set(restorePath(identifier), currentConnection)
+}
+```
+
+```typescript
+function onChanUpgradeAck(
+    identifier: Identifier,
+    counterpartyConnection: ConnectionEnd,
+    counterpartyStatus: UpgradeStatus,
+    proofConnection: CommitmentProof,
+    proofUpgradeStatus: CommitmentProof,
+    proofHeight: Height
+) {
+    // current connection is in UPGRADE_INIT
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == UPGRADE_INIT)
+
+
+    // verify proofs of counterparty state
+    abortTransactionUnless(currentConnection.client.verifyConnectionState(proofHeight, proofConnection, counterpartyConnection))
+    abortTransactionUnless(currentConnection.client.verifyUpgradeStatus(proofHeight, proofUpgradeStatus, counterpartyUpgradeStatus))
+
+    // counterparty must be in TRY state
+    restoreConnectionUnless(counterpartyStatus.state == UPGRADE_TRY)
+    restoreConnectionUnless(counterpartyConnection.State == UPGRADE_TRY)
+
+    // verify connections are mutually compatible
+    // this will also check counterparty chosen version is valid
+    restoreConnectionUnless(IsCompatible(counterpartyConnection, connection))
+
+    // counterparty-specified timeout must not have exceeded
+    restoreConnectionUnless(
+        counterparty.UpgradeStatus.TimeoutHeight < currentHeight() &&
+        counterparty.UpgradeStatus.TimeoutTimestamp < currentTimestamp()
+    )
+
+    // upgrade is complete
+    // set connection to OPEN and remove unnecessary state
+    currentConnection.state = OPEN
+    provableStore.set(connectionPath(identifier), currentConnection)
+    provableStore.delete(statusPath(identifier))
+    privateStore.delete(restorePath(identifier))
+}
+```
+
+```typescript
+function onChanUpgradeConfirm(
+    identifier: Identifier,
+    counterpartyConnection: ConnectionEnd,
+    proofConnection: CommitmentProof,
+    proofHeight: Height,
+) {
+    // current connection is in UPGRADE_TRY
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == UPGRADE_TRY)
+
+    // counterparty must be in OPEN state
+    abortTransactionUnless(counterpartyConnection.State == OPEN)
+
+    // verify proofs of counterparty state
+    abortTransactionUnless(currentConnection.client.verifyConnectionState(proofHeight, proofConnection, counterpartyConnection))
+    
+    // upgrade is complete
+    // set connection to OPEN and remove unnecessary state
+    currentConnection.state = OPEN
+    provableStore.set(connectionPath(identifier), currentConnection)
+    provableStore.delete(statusPath(identifier))
+    privateStore.delete(restorePath(identifier))
+}
+```
+
+Note: In the `TRY` and `ACK` steps, the chain has the ability to abort the upgrade process if the upgrade parameters are not suitable or the timeout has exceeded. In this case, the chain is expected to write `UPGRADE_ERR` as the state in the `UPGRADE_STATUS`, and restore the original connection with state `OPEN`. The counterparty can then verify that the upgrade status is `UPGRADE_ERR` and restore its original connection to `OPEN` as well, thus cancelling the upgrade.
+
+```typescript
+function restoreConnectionUnless(condition: bool) {
+    if !condition {
+        // cancel upgrade
+        // set UpgradeStatus to `UPGRADE_ERR`
+        // and restore original conneciton
+        errStatus = UpgradeStatus{UPGRADE_ERR}
+        provableStore.set(statusPath(identifier), errStatus)
+        originalConnection = privateStore.get(restorePath(identifier))
+        provableStore.set(connectionPath(identifier), originalConnection)
+        privateStore.delete(restorePath(identifier))
+        // caller should return as well
+    } else {
+        // caller should continue execution
+    }
+}
+```
+
+### Cancel Upgrade Process
+
+As discussed, during the upgrade handshake a chain may cancel the upgrade by writing `UPGRADE_ERR` into the `UpgradeStatus` and restoring the original connection to `OPEN`. The counterparty must then restore its connection to `OPEN` as well. A relayer can facilitate this by calling `CancelConnectionUpgrade`
+
+```typescript
+function cancelConnectionUpgrade(
+    identifier: Identifer,
+    counterpartyUpgradeStatus: UpgradeStatus,
+    proofUpgradeStatus: CommitmentProof,
+    proofHeight: Height,
+) {
+    // current connection is in UPGRADE_INIT or UPGRADE_TRY
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == (UPGRADE_INIT || UPGRADE_TRY))
+
+    abortTransactionUnless(counterpartyUpgradeStatus.state == UPGRADE_ERR)
+
+    abortTransactionUnless(currentConnection.client.verifyUpgradeStatus(proofHeight, proofUpgradeStatus, counterpartyUpgradeStatus))
+
+    // cancel upgrade
+    // and restore original conneciton
+    // delete unnecessary state
+    originalConnection = privateStore.get(restorePath(identifier))
+    provableStore.set(connectionPath(identifier), originalConnection)
+    provableStore.delete(statusPath(identifier))
+    privateStore.delete(restorePath(identifier))
+}
+```
+
+It is possible that there is an extraordinary delay in the upgrade handshake occurs because of liveness issues. In this case, we do not want to indefinitely stay in the upgrade process and we will revert to the original connection by calling `CancelConnectionUpgradeTimeout`.
+
+// TODO
--- a/spec/core/ics-003-connection-semantics/UPGRADES.md
+++ b/spec/core/ics-003-connection-semantics/UPGRADES.md
@@ -1,0 +1,163 @@
+# Upgrading Connections
+
+### Synopsis
+
+This standard document specifies the interfaces and state machine logic that IBC implementations must implement in order to enable existing connections to upgrade after initial connection handshake.
+
+### Motivation
+
+As new features get added to IBC, chains may wish the take advantage of new connection features without abandoning the accumulated state and network effect(s) of an already existing connection. The upgrade protocol proposed would allow chains to renegotiate an existing connection to take advantage of new features without having to create a new connection.
+
+### Desired Properties
+
+- Both chains must agree to the renegotiated connection parameters
+- Connection state and logic on both chains should either be using the old parameters or the new parameters, but should not be in an in-between state. i.e. It should not be possible for a chain to write state to an old proof path, while the counterparty expects a new proof path.
+- If the upgrade handshake is unsuccessful, the connection must fall-back to the original connection parameters
+- If the upgrade handshake is successful, then both connection ends have adopted the new connection parameters and process IBC data appropriately.
+- The connection upgrade protocol should have the ability to change all connection-related parameters; however the connection upgrade protocol MUST NOT be able to change the underlying `ClientState`.
+- The connection upgrade protocol may not modify the connection identifiers.
+
+## Technical Specification
+
+### Data Structures
+
+```typescript
+enum ConnectionState {
+  INIT,
+  TRYOPEN,
+  OPEN,
+  UPGRADE_INIT,
+  UPGRADE_TRY,
+  UPGRADE_ERR,
+}
+```
+
+- The chain that is proposing the upgrade should set the connection state from `OPEN` to `UPGRADE_INIT`
+- The counterparty chain that accepts the upgrade should set the connection state from `OPEN` to `UPGRADE_TRY`
+
+```typescript
+interface ConnectionEnd {
+  state: ConnectionState
+  counterpartyConnectionIdentifier: Identifier
+  counterpartyPrefix: CommitmentPrefix
+  clientIdentifier: Identifier
+  counterpartyClientIdentifier: Identifier
+  version: string | []string
+  delayPeriodTime: uint64
+  delayPeriodBlocks: uint64
+}
+```
+
+```typescript
+interface UpgradeStatus {
+    state: ConnectionState
+    timeoutHeight: Height
+    timeoutTimestamp: uint64
+}
+```
+
+The desired property that the connection upgrade protocol may not modify the underlying clients or connection identifiers, means that only some fields are upgradable by the upgrade protocol.
+
+- `state`: The state is specified by the handshake steps of the upgrade protocol.
+- `counterpartyConnectionIdentifier`: The counterparty connection identifier CAN NOT be modified by the upgrade protocol.
+- `counterpartyPrefix`: The prefix MAY be modified in the upgrade protocol. The counterparty must accept the new proposed prefix value, or it must return an error during the upgrade handshake.
+- `clientIdentifier`: The client identifier CAN NOT be modified by the upgrade protocol
+- `counterpartyClientIdentifier`: The counterparty client identifier CAN NOT be modified by the upgrade protocol
+- `version`: The version MAY be modified by the upgrade protocol. The same version negotiation that happens in the initial connection handshake can be employed for the upgrade handshake.
+- `delayPeriodTime`: The delay period MAY be modified by the upgrade protocol. The counterparty MUST accept the new proposed value or return an error during the upgrade handshake.
+- `delayPeriodBlocks`: The delay period MAY be modified by the upgrade protocol. The counterparty MUST accept the new proposed value or return an error during the upgrade handshake.
+
+```typescript
+interface UpgradeConnectionState {
+    connection: ConnectionEnd
+    timeoutHeight: Height
+    timeoutTimestamp: uint64
+}
+```
+
+- `proposedConnectionState`: Proposed `ConnectionState` to replace the current `ConnectionState`, it MUST ONLY modify the fields that are permissable to 
+- `timeoutHeight`: Timeout height indicates the height at which the counterparty must no longer proceed with the upgrade handshake. The chains will then preserve their original connection and the upgrade handshake is aborted.
+- `timeoutTimestamp`: Timeout timestamp indicates the time on the counterparty at which the counterparty must no longer proceed with the upgrade handshake. The chains will then preserve their original connection and the upgrade handshake is aborted.
+
+NOTE: One of the timeout height or timeout timestamp must be non-zero.
+
+### Store Paths
+
+##### Restore Connection Path
+
+The chain must store the previous connection end so that it may restore it if the upgrade handshake fails. This may be stored in the private store.
+
+```typescript
+function restorePath(id: Identifier): Path {
+    return "connections/{id}/restore"
+}
+```
+
+##### UpgradeStatus Path
+
+The upgrade status path is a public path that can signal the status of the upgrade to the counterparty. It does not store anything in the successful case, but it will store a sentinel abort value in the case that a chain does not accept the proposed upgrade.
+
+```typescript
+function upgradePath(id: Identifier): Path {
+    return "connections/{id}/upgradeStatus"
+
+}
+```
+
+### Upgrade Handshake
+
+The upgrade handshake defines four datagrams: *ConnUpgradeInit*, *ConnUpgradeTry*, *ConnUpgradeAck*, and *ConnUpgradeConfirm*
+
+A successful protocol execution flows as follows (note that all calls are made through modules per ICS 25):
+
+| Initiator | Datagram             | Chain acted upon | Prior state (A, B)          | Posterior state (A, B)      |
+| --------- | -------------------- | ---------------- | --------------------------- | --------------------------- |
+| Actor     | `ConnUpgradeInit`    | A                | (OPEN, OPEN)                | (UPGRADE_INIT, OPEN)        |
+| Relayer   | `ConnUpgradeTry`     | B                | (UPGRADE_INIT, OPEN)        | (UPGRADE_INIT, UPGRADE_TRY) |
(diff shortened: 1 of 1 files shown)


AdityaSripal:
Some notes:

- Current idea is that upgrading a connection will require bilateral agreement. i.e. Both sides must explicitly allow the initial parts of handshake (INIT and TRY). Ack and Confirm can be completed by trustless relayer.
- There is currently a timeout specified on INIT and TRY, so that the connection does not hang indefinitely in a closed state. Not fully specified yet. 
- The details of how to handle crossing hellos is not fully specified yet, so ignore that possibility on initial review.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,360 @@
...
+
+- `state`: The state is specified by the handshake steps of the upgrade protocol.
+- `counterpartyConnectionIdentifier`: The counterparty connection identifier CAN NOT be modified by the upgrade protocol.
+- `counterpartyPrefix`: The prefix MAY be modified in the upgrade protocol. The counterparty must accept the new proposed prefix value, or it must return an error during the upgrade handshake.
+- `clientIdentifier`: The client identifier CAN NOT be modified by the upgrade protocol
+- `counterpartyClientIdentifier`: The counterparty client identifier CAN NOT be modified by the upgrade protocol
+- `version`: The version MAY be modified by the upgrade protocol. The same version negotiation that happens in the initial connection handshake can be employed for the upgrade handshake.
+- `delayPeriodTime`: The delay period MAY be modified by the upgrade protocol. The counterparty MUST accept the new proposed value or return an error during the upgrade handshake.
+- `delayPeriodBlocks`: The delay period MAY be modified by the upgrade protocol. The counterparty MUST accept the new proposed value or return an error during the upgrade handshake.
+
```
colin-axner, line 61: maybe split into two sections:

MAY be modified

CAN NOT be modified

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,360 @@
...
+At the end of an opening handshake between two chains implementing the sub-protocol, the following properties hold:
+
+- Each chain is running their new upgraded connection end and is processing upgraded logic and state according to the upgraded parameters.
+- Each chain has knowledge of and has agreed to the counterparty's upgraded connection parameters.
+
+If a chain does not agree to the proposed counterparty `UpgradeConnectionState`, it may abort the upgrade handshake by writing a sentinel abort value into the upgrade-failed connection path.
+
+`connection/{identifier}/proposedUpgradeFailed` => `[]byte(0x1)`.
+
+The counterparty receiving proof of this must cancel the upgrade and resume the
```
colin-axner, line 133: unfinished sentence

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,360 @@
...
+    restoreConnectionUnless(IsCompatible(counterpartyConnection, proposedUpgrade.Connection))
+
+    upgradeStatus = UpgradeStatus{
+        state: UPGRADE_TRY,
+        timeoutHeight: proposedUpgrade.timeoutHeight,
+        timeoutTimestamp: proposedUpgrade.timeoutTimestamp,
+    }
+
+    provableStore.set(statusPath(identifier), upgradeStatus)
+    provableStore.set(connectionPath(identifier), proposedUpgrade.connection)
```
colin-axner, line 234: Interesting, I'd think you'd want to leave the connection unchanged until the upgrade process successfully completes. Otherwise you could potentially disrupt packet flow if the connection is in UPGRADE_INIT instead of OPEN?

I'd expect you to store the proposed upgrade connection under a proposed upgrade connection path. Use that for proofs and then if successful update the connection on ack/confirm?

colin-axner (commented):
Excellent!! Concept makes sense to me. Architecturally not difficult to introduce into existing ibc implementations

spec/core/ics-003-connection-semantics/UPGRADES.md
AdityaSripal: I don't want the case where chainA has its connectionEnd OPEN with the old parameters, and chainB has its connectionEnd OPEN with the new parameters as that may cause unexpected consequences.

If a connection is `OPEN` on both ends it should be a valid connection always.

So either both ends are `OPEN` running the old parameters, or both ends are `OPEN` running the new parameters.

The only way to accomplish this is to temporarily close the connection on both sides (INIT and TRY), then reopen after the upgrade is complete on either end (ACK and CONFIRM).

This does mean packet flow is temporarily halted but i believe this is fine. Once the connection is back up, the relayers can either receive or timeout the pending packets

AdityaSripal: This is also why I specify a timeout, because I don't want the "temporary pause" to last arbitrarily long

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,365 @@
+# Upgrading Connections
+
+### Synopsis
+
+This standard document specifies the interfaces and state machine logic that IBC implementations must implement in order to enable existing connections to upgrade after initial connection handshake.
+
+### Motivation
+
+As new features get added to IBC, chains may wish the take advantage of new connection features without abandoning the accumulated state and network effect(s) of an already existing connection. The upgrade protocol proposed would allow chains to renegotiate an existing connection to take advantage of new features without having to create a new connection.
```
cwgoes, line 9: and presumably preserve any channels etc. associated with that connection

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,365 @@
...
+
+- `state`: The state is specified by the handshake steps of the upgrade protocol.
+
+CAN BE MODIFIED:
+- `counterpartyPrefix`: The prefix MAY be modified in the upgrade protocol. The counterparty must accept the new proposed prefix value, or it must return an error during the upgrade handshake.
+- `version`: The version MAY be modified by the upgrade protocol. The same version negotiation that happens in the initial connection handshake can be employed for the upgrade handshake.
+- `delayPeriodTime`: The delay period MAY be modified by the upgrade protocol. The counterparty MUST accept the new proposed value or return an error during the upgrade handshake.
+- `delayPeriodBlocks`: The delay period MAY be modified by the upgrade protocol. The counterparty MUST accept the new proposed value or return an error during the upgrade handshake.
+
+CAN NOT BE MODIFIED:
```
cwgoes, line 61:
```suggestion
CANNOT BE MODIFIED:
```

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,365 @@
...
+
+```typescript
+interface UpgradeConnectionState {
+    connection: ConnectionEnd
+    timeoutHeight: Height
+    timeoutTimestamp: uint64
+}
+```
+
+- `proposedConnectionState`: Proposed `ConnectionState` to replace the current `ConnectionState`, it MUST ONLY modify the fields that are permissable to
```
cwgoes, line 86:
```suggestion
- `proposedConnectionState`: Proposed `ConnectionState` to replace the current `ConnectionState`, it MUST ONLY modify the fields that are permissible to 
```

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,365 @@
...
+function connUpgradeInit(
+    identifier: Identifier,
+    proposedUpgrade: UpgradeConnectionState
+) {
+    // current connection must be OPEN
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == OPEN)
+
+    // abort transaction if an unmodifiable field is modified
+    // upgraded connection state must be in `UPGRADE_INIT`
```
cwgoes, line 152: note that if we add fields to the connection struct this will by default allow them to be modified

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,365 @@
...
+    proofConnection: CommitmentProof,
+    proofUpgradeStatus: CommitmentProof,
+    proofHeight: Height
+) {
+    // current connection must be OPEN or UPGRADE_INIT (crossing hellos)
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == (OPEN || UPGRADE_INIT))
+
+    // abort transaction if an unmodifiable field is modified
+    // upgraded connection state must be in `UPGRADE_TRY`
```
cwgoes, line 193: ditto on above note

spec/core/ics-003-connection-semantics/UPGRADES.md
cwgoes: yeah, this makes sense to me

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,365 @@
...
+    originalConnection = privateStore.get(restorePath(identifier))
+    provableStore.set(connectionPath(identifier), originalConnection)
+    provableStore.delete(statusPath(identifier))
+    privateStore.delete(restorePath(identifier))
+}
+```
+
+It is possible that there is an extraordinary delay in the upgrade handshake occurs because of liveness issues. In this case, we do not want to indefinitely stay in the upgrade process and we will revert to the original connection by calling `CancelConnectionUpgradeTimeout`.
+
+// TODO
```
cwgoes, line 365: What is TODO?

cwgoes (changes requested):
Concept ACK; most changes ACK (modulo open question of crossing hellos); see a few minor comments.

I will also note that I would like to see re-verification of IBC safety / liveness with this logic incorporated.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+    proofUpgradeStatus: CommitmentProof,
+    proofHeight: Height
+) {
+    // current connection is in UPGRADE_INIT or UPGRADE_TRY (crossing hellos)
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(currentConnection.state == UPGRADE_INIT || currentConnection.state == UPGRADE_TRY)
+
+    // verify proofs of counterparty state
+    abortTransactionUnless(currentConnection.client.verifyConnectionState(proofHeight, proofConnection, counterpartyConnection))
+    abortTransactionUnless(currentConnection.client.verifyUpgradeStatus(proofHeight, proofUpgradeStatus, counterpartyUpgradeStatus))
```
mpoke, line 274: Same comment as in `connUpgradeTry`.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+) {
+    // current connection is in UPGRADE_TRY
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == UPGRADE_TRY)
+
+    // counterparty must be in OPEN state
+    abortTransactionUnless(counterpartyConnection.State == OPEN)
+
+    // verify proofs of counterparty state
+    abortTransactionUnless(currentConnection.client.verifyConnectionState(proofHeight, proofConnection, counterpartyConnection))
```
mpoke, line 308: Same comment as in `connUpgradeTry`.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+    // upgrade is complete
+    // set connection to OPEN and remove unnecessary state
+    currentConnection.state = OPEN
+    provableStore.set(connectionPath(identifier), currentConnection)
+    provableStore.delete(statusPath(identifier))
+    privateStore.delete(restorePath(identifier))
+}
+```
+
+Note: In the `TRY` and `ACK` steps, the chain has the ability to abort the upgrade process if the upgrade parameters are not suitable or the timeout has exceeded. In this case, the chain is expected to write `UPGRADE_ERR` as the state in the `UPGRADE_STATUS`, and restore the original connection with state `OPEN`. The counterparty can then verify that the upgrade status is `UPGRADE_ERR` and restore its original connection to `OPEN` as well, thus cancelling the upgrade.
```
mpoke, line 319: This contains redundant info from the Cancel Upgrade Process section.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+```typescript
+function cancelConnectionUpgrade(
+    identifier: Identifer,
+    counterpartyUpgradeStatus: UpgradeStatus,
+    proofUpgradeStatus: CommitmentProof,
+    proofHeight: Height,
+) {
+    // current connection is in UPGRADE_INIT or UPGRADE_TRY
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == (UPGRADE_INIT || UPGRADE_TRY))
```
mpoke, line 352: Confusing notation.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+
+    abortTransactionUnless(counterpartyUpgradeStatus.state == UPGRADE_ERR)
+
+    abortTransactionUnless(currentConnection.client.verifyUpgradeStatus(proofHeight, proofUpgradeStatus, counterpartyUpgradeStatus))
```
mpoke, line 356: Same comment as in `connUpgradeTry`.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+) {
+    upgradeStatus = provableStore.get(statusPath(identifier))
+
+    // proof must be from a height after timeout has elapsed. Either timeoutHeight or timeoutTimestamp must be defined.
+    // if timeoutHeight is defined and proof is from before timeout height
+    // then abort transaction
+    abortTransactionUnless(upgradeStatus.TimeoutHeight.IsZero() || proofHeight >= upgradeStatus.TimeoutHeight)
+    // if timeoutTimestamp is defined then the consensus time from proof height must be greater than timeout timestamp
+    connection = getConnection(identifer)
+    consensusState = getConsensusState(connection.clientIdentifer, proofHeight)
```
mpoke, line 391: More of a general comment, `getConsensusState(id, height)` is not defined in the IBC specs.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+    abortTransactionUnless(upgradeStatus.TimeoutTimestamp.IsZero() || consensusState.Timestamp >= upgradeStatus.Timestamp)
```
mpoke, line 392: `consensusState.Timestamp` is an ICS7 concept. From ICS2,
> The ConsensusState MUST define a getTimestamp() method which returns the timestamp associated with that consensus state:

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+    // counterparty connection must be proved to still be in OPEN state
+    abortTransactionUnless(counterpartyConnection.State === OPEN)
+    abortTransactionUnless(connection.client.verifyConnectionState(proofHeight, proofConnection, counterpartyConnection))
+
+    // we must restore the connection since the timeout verification has passed
+    restoreConnectionUnless(false)
+}
+```
+
+Note that the timeout logic only applies to the INIT step. This is to protect an upgrading chain from being stuck in a non-OPEN state if the counterparty cannot execute the TRY successfully. Once the TRY step succeeds, then both sides are guaranteed to have the upgrade feature enabled. Liveness is no longer an issue, because we can wait until liveness is restored to execute the ACK step which will move the connection definitely into an OPEN state (either a successful upgrade or a rollback).
```
mpoke, line 403: > Note that the timeout logic only applies to the INIT step

Then why is it necessary to pass the timeouts params to `connUpgradeTry()`?

mpoke (changes requested):
Concept ACK. I agree with Christopher that we need to make sure that the IBC safety and liveness are preserved with this logic incorporated. See comments below.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+function connUpgradeTry(
+    identifier: Identifier,
+    proposedUpgrade: UpgradeConnectionState,
+    counterpartyConnection: ConnectionEnd,
+    counterpartyUpgradeStatus: UpgradeStatus,
+    proofConnection: CommitmentProof,
+    proofUpgradeStatus: CommitmentProof,
+    proofHeight: Height
+) {
+    // current connection must be OPEN or UPGRADE_INIT (crossing hellos)
```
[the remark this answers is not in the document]
mpoke: Actually, the chain not ignoring the TRY messages still needs to handle the crossing hellos logic, so this would not really solve anything. You can just ignore my comment.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+  UPGRADE_TRY,
+  UPGRADE_ERR,
+}
+```
+
+- The chain that is proposing the upgrade should set the connection state from `OPEN` to `UPGRADE_INIT`
+- The counterparty chain that accepts the upgrade should set the connection state from `OPEN` to `UPGRADE_TRY`
+
+```typescript
+interface ConnectionEnd {
```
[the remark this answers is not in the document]
mpoke: Then I would say that `ConnectionEnd` is defined as in ICS 3, i.e., 
```
interface ConnectionEnd {
...
}
```
And then just summarize what MUST NO be modified and what MAY be modified.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+    if !condition {
+        // cancel upgrade
+        // set UpgradeStatus to `UPGRADE_ERR`
+        // and restore original conneciton
+        errStatus = UpgradeStatus{UPGRADE_ERR}
+        provableStore.set(statusPath(identifier), errStatus)
+        originalConnection = privateStore.get(restorePath(identifier))
+        provableStore.set(connectionPath(identifier), originalConnection)
+        privateStore.delete(restorePath(identifier))
+        // caller should return as well
```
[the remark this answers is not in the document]
AdityaSripal: This is to continue with the pattern of `abortTransactionUnless`, and this is semantically similar since we are aborting the handshake. So i think it makes sense to follow the same pattern

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+    // then we must ensure that the proposedUpgrade by the counterparty is the same as the currentConnection
+    // except for the connection state (upgrade connection will be in UPGRADE_TRY and current connection will be in UPGRADE_INIT)
+    // if the proposed upgrades on either side are incompatible, then we will restore the connection and cancel the upgrade.
+    currentConnection.state = UPGRADE_TRY
+    restoreConnectionUnless(currentConnection.IsEqual(proposedUpgrade.connection))
+
+    // either timeout height or timestamp must be non-zero
+    // if the upgrade feature is implemented on the TRY chain, then a relayer may submit a TRY transaction after the timeout.
+    // this will restore the connection on the executing chain and allow counterparty to use the CancelUpgradeMsg to restore their connection.
+    restoreTransactionUnless(proposedUpgrade.TimeoutHeight != 0 || proposedUpgrade.TimeoutTimestamp != 0)
```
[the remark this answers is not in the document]
AdityaSripal: Correct, thanks for the catch.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+
+```typescript
+function restoreConnectionUnless(condition: bool) {
+    if !condition {
+        // cancel upgrade
+        // set UpgradeStatus to `UPGRADE_ERR`
+        // and restore original conneciton
+        errStatus = UpgradeStatus{UPGRADE_ERR}
+        provableStore.set(statusPath(identifier), errStatus)
+        originalConnection = privateStore.get(restorePath(identifier))
```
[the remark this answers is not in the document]
AdityaSripal: I think these sorts of programming errors should be handled in implementation, but doesn't make as much sense to pollute spec with these because then we would have to add them everywhere and it should never happen with correct spec anyway

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+
+- UpgradeStatus contains the state of the upgrade. If upgrade is successful it will contain the `ConnectionState` of the upgrading connection on the executing chain. If the executing chain wants to abort the upgrade, it will restore its previous connection under its connection path with state `OPEN` and write an `UpgradeStatus` with state `UPGRADE_ERR`.
+- `timeoutHeight`: Timeout height indicates the height at which the counterparty must no longer proceed with the upgrade handshake. The chains will then preserve their original connection and the upgrade handshake is aborted.
+- `timeoutTimestamp`: Timeout timestamp indicates the time on the counterparty at which the counterparty must no longer proceed with the upgrade handshake. The chains will then preserve their original connection and the upgrade handshake is aborted.
+
+```typescript
+interface UpgradeConnectionState {
+    connection: ConnectionEnd
+    timeoutHeight: Height
+    timeoutTimestamp: uint64
```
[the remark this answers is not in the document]
AdityaSripal: You're right. I've removed `UpgradeConnectionState` and `UpgradeStatus`. Replaced with `UpgradeTimeout` and I believe the protocol is much cleaner.

Leaving the struct fields as is, since they are the same semantically as the packet timeouts. But changed the naming in the OpenInit function

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+        counterpartyConnection.counterpartyConnectionIdentifier == identifier &&
+        counterpartyConnection.clientIdentifier == currentConnection.counterpartyClientIdentifier &&
+        counterpartyConnection.counterpartyClientIdentifier == currentConnection.clientIdentifier
+    )
+    restoreConnectionUnless(counterpartyUpgradeStatus.state == UPGRADE_INIT)
+
+    // counterparty-specified timeout must not have exceeded
+    restoreConnectionUnless(
+        counterparty.UpgradeStatus.TimeoutHeight < currentHeight() &&
+        counterparty.UpgradeStatus.TimeoutTimestamp < currentTimestamp()
```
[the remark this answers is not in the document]
AdityaSripal: This suggestion is not the correct logic. since it is `restoreConnectionUnless(condition)

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+    )
+
+    // verify chosen versions are compatible
+    versionsIntersection = intersection(counterpartyConnection.version, proposedUpgrade.Connection.version)
+    version = pickVersion(versionsIntersection) // throws if there is no intersection
+
+    // both connection ends must be mutually compatible.
+    restoreConnectionUnless(IsCompatible(counterpartyConnection, proposedUpgrade.Connection))
```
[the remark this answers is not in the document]
AdityaSripal: Added documentation. This is dependent on the specifics of the upgrade

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+### Upgrade Handshake
+
+The upgrade handshake defines four datagrams: *ConnUpgradeInit*, *ConnUpgradeTry*, *ConnUpgradeAck*, and *ConnUpgradeConfirm*
+
+A successful protocol execution flows as follows (note that all calls are made through modules per ICS 25):
+
+| Initiator | Datagram             | Chain acted upon | Prior state (A, B)          | Posterior state (A, B)      |
+| --------- | -------------------- | ---------------- | --------------------------- | --------------------------- |
+| Actor     | `ConnUpgradeInit`    | A                | (OPEN, OPEN)                | (UPGRADE_INIT, OPEN)        |
+| Actor     | `ConnUpgradeTry`     | B                | (UPGRADE_INIT, OPEN)        | (UPGRADE_INIT, UPGRADE_TRY) |
```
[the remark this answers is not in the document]
AdityaSripal: Documented in the TRY step. Up to implementations if they allow relayer or actor

spec/core/ics-003-connection-semantics/UPGRADES.md
AdityaSripal: queryConsensusState exists, so i replaced it with that

spec/core/ics-003-connection-semantics/UPGRADES.md
AdityaSripal: Documented clarification

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+  upgradeErrorReceipt: []byte, 
+) {
+    client = queryClient(connection.clientIdentifier)
+    client.verifyUpgradeError(height, connection.counterpartyPrefix, proof, connection.counterpartyConnectionIdentifier, upgradeErrorReceipt)
+}
+```
+
+```typescript
+// Client VerifyUpgradeError
+function verifyConnectionUpgradeError(
```
mpoke, line 122:
```suggestion
function VerifyUpgradeError(
```

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+  upgradeTimeout: UpgradeTimeout, 
+) {
+    client = queryClient(connection.clientIdentifier)
+    client.verifyUpgradeTimeout(height, connection.counterpartyPrefix, proof, connection.counterpartyConnectionIdentifier, upgradeTimeout)
+}
+```
+
+```typescript
+// Client VerifyUpgradeTimeout
+function verifyConnectionUpgradeTimeout(
```
mpoke, line 163:
```suggestion
function VerifyUpgradeTimeout(
```

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,407 @@
...
+    provableStore.set(connectionPath(identifier), currentConnection)
+    provableStore.delete(statusPath(identifier))
+    privateStore.delete(restorePath(identifier))
+}
+```
+
+Note: In the `TRY` and `ACK` steps, the chain has the ability to abort the upgrade process if the upgrade parameters are not suitable or the timeout has exceeded. In this case, the chain is expected to write `UPGRADE_ERR` as the state in the `UPGRADE_STATUS`, and restore the original connection with state `OPEN`. The counterparty can then verify that the upgrade status is `UPGRADE_ERR` and restore its original connection to `OPEN` as well, thus cancelling the upgrade.
+
+```typescript
+function restoreConnectionUnless(condition: bool) {
```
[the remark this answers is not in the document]
mpoke: > I will leave this as-is for now since I didn't find a better place to put it, and maybe its ok to just treat as a black-box until you reach the function definition. But open to suggestions

Since the function is used in multiple subprotocols, you could just add it to the overview (i.e., before the "Upgrade Handshake" section) and mention that e.g., it is a utility function used when the upgrade fails and it restores the connection to its pre-upgrade state.

spec/core/ics-003-connection-semantics/UPGRADES.md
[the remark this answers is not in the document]
mpoke: So if the condition is true, you don't restore the connection. This means that the current height exceeds `TimeoutHeight` and the current timestamp exceeds `TimeoutTimestamp`. Am I missing something?

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+    // if timeoutHeight is defined and proof is from before timeout height
+    // then abort transaction
+    abortTransactionUnless(upgradeTimeout.timeoutHeight.IsZero() || proofHeight >= upgradeTimeout.timeoutHeight)
+    // if timeoutTimestamp is defined then the consensus time from proof height must be greater than timeout timestamp
+    consensusState = queryConsensusState(currentConnection.clientIdentifer, proofHeight)
+    abortTransactionUnless(upgradeTimeout.timeoutTimestamp.IsZero() || consensusState.getTimestamp() >= upgradeTimeout.timestamp)
+
+    // counterparty connection must be proved to still be in OPEN state
+    abortTransactionUnless(counterpartyConnection.State === OPEN)
+    abortTransactionUnless(connection.client.verifyConnectionState(proofHeight, proofConnection, counterpartyConnection))
```
mpoke, line 470: Use the connection helper function, i.e., https://github.com/cosmos/ibc/blob/master/spec/core/ics-003-connection-semantics/README.md#helper-functions

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+```typescript
+function cancelConnectionUpgrade(
+    identifier: Identifer,
+    errorReceipt: []byte,
+    proofUpgradeError: CommitmentProof,
+    proofHeight: Height,
+) {
+    // current connection is in UPGRADE_INIT or UPGRADE_TRY
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == UPGRADE_INIT || connection.state == UPGRADE_TRY)
```
mpoke, line 421: Use `currentConnection` instead of `connection`.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+
+    abortTransactionUnless(!isEmpty(errorReceipt))
+
+    abortTransactionUnless(verifyUpgradeError(currentConnection, proofHeight, proofUpgradeError, currentConnection.counterpartyConnectionIdentifier, errorReceipt))
```
mpoke, line 425:
```suggestion
    abortTransactionUnless(verifyConnectionUpgradeError(currentConnection, proofHeight, proofUpgradeError, errorReceipt))
```

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+```typescript
+function connUpgradeConfirm(
+    identifier: Identifier,
+    counterpartyConnection: ConnectionEnd,
+    proofConnection: CommitmentProof,
+    proofHeight: Height,
+) {
+    // current connection is in UPGRADE_TRY
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(connection.state == UPGRADE_TRY)
```
mpoke, line 372:
```suggestion
    abortTransactionUnless(currentConnection.state == UPGRADE_TRY)
```

mpoke (changes requested):
Thanks for the changes. The protocol seems correct. I left some comments, mostly minor.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+) {
+    client = queryClient(connection.clientIdentifier)
+    client.verifyUpgradeError(height, connection.counterpartyPrefix, proof, connection.counterpartyConnectionIdentifier, upgradeErrorReceipt)
+}
+```
+
+```typescript
+// Client VerifyUpgradeError
+function verifyConnectionUpgradeError(
+    clientState: ClientState,
```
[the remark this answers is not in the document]
AdityaSripal: This is the same pattern used in ICS3 and ICS2. I believe it is the standard for Typescript.

https://github.com/cosmos/ibc/blob/master/spec/core/ics-003-connection-semantics/README.md#helper-functions

https://github.com/cosmos/ibc/blob/master/spec/core/ics-002-client-semantics/README.md#required-functions

spec/core/ics-003-connection-semantics/UPGRADES.md
AdityaSripal: I've added Connection here to the name because there are identical functions for channel upgradability.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+    height: Height,
+    prefix: CommitmentPrefix,
+    proof: CommitmentProof,
+    counterpartyConnectionIdentifier: Identifier,
+    upgradeTimeout: UpgradeTimeout,
+) {
+    path = applyPrefix(prefix, timeoutPath(counterpartyConnectionIdentifier))
+    abortTransactionUnless(!clientState.frozen)
+    timeoutBytes = protobuf.marshal(upgradeTimeout)
+    return clientState.verifiedRoots[height].verifyMembership(path, timeoutBytes, proof)
```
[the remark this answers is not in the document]
AdityaSripal: This does not assume anything about Tendermint, but just says we will verify membership against a consensus state at provided height. This is taken from ICS2.

https://github.com/cosmos/ibc/blob/master/spec/core/ics-002-client-semantics/README.md#example-implementation

spec/core/ics-003-connection-semantics/UPGRADES.md
AdityaSripal: ditto

spec/core/ics-003-connection-semantics/UPGRADES.md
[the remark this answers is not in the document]
mpoke: Yeah, I've seen that. I find it very confusing. Until now I thought it's an inconsistency. Still not convinced that it's not :)

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+) {
+    client = queryClient(connection.clientIdentifier)
+    client.verifyUpgradeTimeout(height, connection.counterpartyPrefix, proof, connection.counterpartyConnectionIdentifier, upgradeTimeout)
+}
+```
+
+```typescript
+// Client VerifyUpgradeTimeout
+function verifyConnectionUpgradeTimeout(
+    clientState: ClientState,
```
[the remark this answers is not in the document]
AdityaSripal: ditto

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+  - either it is unsuccessful and then the connection MUST fall-back to the original connection parameters; 
+  - or it is successful and then both connection ends MUST adopt the new connection parameters and process IBC data appropriately.
+- The connection upgrade protocol should have the ability to change all connection-related parameters; however the connection upgrade protocol MUST NOT be able to change the underlying `ClientState`.
+The connection upgrade protocol MUST NOT modify the connection identifiers.
+
+## Technical Specification
+
+### Data Structures
+
+The `ConnectionState` and `ConnectionEnd` are defined in [ICS-3](./README.md), they are reproduced here for the reader's convenience. `UPGRADE_INIT`, `UPGRADE_TRY`, `UPGRADE_ERR` are additional states added to enable the upgrade feature.
```
colin-axner, line 25: Is `UPGRADE_ERR` still needed? It is missing below

spec/core/ics-003-connection-semantics/UPGRADES.md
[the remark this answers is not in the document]
AdityaSripal: Ahh yes I had the wrong logic and you are correct. Sorry for the confusion!

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+  counterpartyPrefix: CommitmentPrefix
+  clientIdentifier: Identifier
+  counterpartyClientIdentifier: Identifier
+  version: string | []string
+  delayPeriodTime: uint64
+  delayPeriodBlocks: uint64
+}
+```
+
+The desired property that the connection upgrade protocol MUST NOT modify the underlying clients or connection identifiers, means that only some fields of `ConnectionEnd` are upgradable by the upgrade protocol.
```
colin-axner, line 53: Can you discuss removal of connection fields (such as counterpartyPrefix)

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,483 @@
...
+- `counterpartyClientIdentifier`: The counterparty client identifier CAN NOT be modified by the upgrade protocol
+
+NOTE: If the upgrade adds any fields to the `ConnectionEnd` these are by default modifiable, and can be arbitrarily chosen by an Actor (e.g. chain governance) which has permission to initiate the upgrade.
+
+```typescript
+interface UpgradeTimeout {
+    timeoutHeight: Height
+    timeoutTimestamp: uint64
+}
+```
```
colin-axner, line 75: I find the jump in concepts to be abrupt. Maybe you should use a subheader for each Data Structure?

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,487 @@
...
+ 
+    provableStore.set(connectionPath(identifier), proposedUpgradeConnection)
+}
+```
+
+NOTE: It is up to individual implementations how they will provide access-control to the `ConnUpgradeTry` function. E.g. chain governance, permissioned actor, DAO, etc. A chain may decide to have permissioned **or** permissionless `UpgradeTry`. In the permissioned case, both chains must explicitly consent to the upgrade, in the permissionless case; one chain initiates the upgrade and the other chain agrees to the upgrade by default. In the permissionless case, a relayer may submit the `ConnUpgradeTry` datagram.
+
+
+```typescript
+function connUpgradeAck(
```
colin-axner, line 354: so the timeout value only applies to the first two steps of the handshake?

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,487 @@
...
+
+    // cancel upgrade
+    // and restore original conneciton
+    // delete unnecessary state
+    originalConnection = privateStore.get(restorePath(identifier))
+    provableStore.set(connectionPath(identifier), originalConnection)
+
+    // delete auxilliary upgrade state
+    provableStore.delete(timeoutPath(identifier))
+    privateStore.delete(restorePath(identifier))
```
colin-axner, line 439: should this just call `restoreConnectionUnless`?

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,487 @@
...
+
+    upgradeTimeout = provableStore.get(timeoutPath(identifier))
+
+    // proof must be from a height after timeout has elapsed. Either timeoutHeight or timeoutTimestamp must be defined.
+    // if timeoutHeight is defined and proof is from before timeout height
+    // then abort transaction
+    abortTransactionUnless(upgradeTimeout.timeoutHeight.IsZero() || proofHeight >= upgradeTimeout.timeoutHeight)
+    // if timeoutTimestamp is defined then the consensus time from proof height must be greater than timeout timestamp
+    consensusState = queryConsensusState(currentConnection.clientIdentifer, proofHeight)
+    abortTransactionUnless(upgradeTimeout.timeoutTimestamp.IsZero() || consensusState.getTimestamp() >= upgradeTimeout.timestamp)
```
colin-axner, line 470: just as a note, I think the timestamp should be obtained via the client state (passing in the desired consensus state height). This doesn't work for solo machines

spec/core/ics-003-connection-semantics/UPGRADES.md
AdityaSripal: Thanks it was removed

spec/core/ics-003-connection-semantics/UPGRADES.md
AdityaSripal: Yes because you need liveness to be restored for timeouts to be processed, and if liveness is restored you may as well just continue with the handshake anyway.

The timeout is really only useful in this case, when it is possible that we stall indefinitely which is only possible on TRY step

AdityaSripal: See Timeout Upgrade Process

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,495 @@
...
+) {
+    // current connection is in UPGRADE_INIT or UPGRADE_TRY (crossing hellos)
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(currentConnection.state == UPGRADE_INIT || currentConnection.state == UPGRADE_TRY)
+
+    // counterparty must be in TRY state
+    if counterpartyConnection.State != UPGRADE_TRY {
+        restoreConnection()
+        return
+    }
```
colin-axner, line 359: we have to be careful to not use relayer provided information unless verified (especially if we are restoring/cancelling an upgrade)

mpoke (changes requested):
Looks great. I have just one concern  regarding the transition from `UPGRADE_TRY` to `OPEN` (i.e., `connUpgradeConfirm`). The rest of the comments are just nits.

spec/core/ics-003-connection-semantics/UPGRADES.md
```diff
@@ -0,0 +1,495 @@
...
+    // current connection is in UPGRADE_TRY
+    currentConnection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless(currentConnection.state == UPGRADE_TRY)
+
+    // counterparty must be in OPEN state
+    abortTransactionUnless(counterpartyConnection.State == OPEN)
+
+    // verify proofs of counterparty state
+    abortTransactionUnless(verifyConnectionState(currentConnection, proofHeight, proofConnection, currentConnection.counterpartyConnectionIdentifier, counterpartyConnection))
+
```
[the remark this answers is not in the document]
AdityaSripal: Ahh thank you, nice catch

mpoke (approved):

spec/core/ics-003-connection-semantics/UPGRADES.md
AdityaSripal: Fixed

Issue #628, opened 2021-12-16
Modify App Callbacks to improve version negotiation flow for Relayers


AdityaSripal:
Thanks to @ethanfrey for his feedback here: https://github.com/CosmWasm/wasmvm/issues/269

To be clear about the original solution: This was a proposal intended to allow a more complex version negotiation for ICA with minimal disruption to core IBC. `NegotiateAppVersion` was a helper function added to the Application interface that relayers could **optionally** use to determine the correct versions to place into the handshake messages. It was not a necessary part of the application semantics, but made the handshake process smoother for automated relayers. Thanks to @damiannolan for his work on this.

But as @ethanfrey points out, if we are willing to make more substantial changes to core IBC; we can still simplify the process for relayers and remove this helper function `NegotiateAppVersion` from the required interface for application developers.

Taking the feedback into account we propose the following changes:

1. Deprecate `Version` from the `ChannelOpenTry` message. If the relayer is going to specify a particular version, it should be the INIT relayer choice that determines the ultimate version; not a combination of INIT relayer choice and TRY relayer choice. Note: we will not remove the field as this is a breaking change, we will simply ignore the inputted value.
2. Change `OnChanOpenTry` to return `version string, err error`. Core IBC will then set the channel struct with the version chosen by `OnChanOpenTry`

This makes `NegotiateAppVersion` unnecessary for the ICA workflow. We plan to make the above changes before the final release of ICA to ensure that we can remove `NegotiateAppVersion` from the interface before it gets real world usage. This will minimize the technical debt, as we can remove it immediately without going through deprecation.

Another set of changes can be added to further simplify the process but can be added at a later time:

1. Change `OnChanOpenInit` to return `version string, err error`. If the relayer passes in a non-empty version string, then INIT callback should return error if that version is not supported. If the relayer passes in an empty string, then INIT callback should return the default version string signifying all of the versions that it can support. If there is no "default version" for a particular application, the application should return an error.

Currently relayers hardcode the versions that they pass in for apps like ICS20. As the number of applications on-chain increase, this is unsustainable. Relayers will want to open channels for a particular port, without even understanding what application that port corresponds to and what version string they should be passing in. Passing in an empty string is effectively an "unopinionated" relayer opening a channel. The channel version may simply be negotiated from the set of versions INIT chain supports and the set of versions the TRY chain supports.

This still leaves the option of "opinionated" relayers choosing a specific version(s) by providing a non-empty version string.

[...]

PR #629, opened 2021-12-16
Improve handshake callbacks

AdityaSripal:
Closes: #628

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -315,7 +315,6 @@ function chanOpenTry(
   order: ChannelOrder,
   connectionHops: [Identifier],
   portIdentifier: Identifier,
-  previousIdentifier: Identifier,
   counterpartyChosenChannelIdentifer: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
@@ -323,22 +322,8 @@ function chanOpenTry(
   counterpartyVersion: string,
   proofInit: CommitmentProof,
   proofHeight: Height): CapabilityKey {
-    if (previousIdentifier !== "") {
-      previous = provableStore.get(channelPath(portIdentifier, channelIdentifier))
-      abortTransactionUnless(
-        (previous !== null) &&
-        (previous.state === INIT &&
-         previous.order === order &&
-         previous.counterpartyPortIdentifier === counterpartyPortIdentifier &&
-         previous.counterpartyChannelIdentifier === "" &&
-         previous.connectionHops === connectionHops &&
-         previous.version === counterpartyVersion
-        )
-      channelIdentifier = previousIdentifier
-    } else {
-      // generate a new identifier if the provided identifier was the sentinel empty-string
-      channelIdentifier = generateIdentifier()
-    }
+    channelIdentifier = generateIdentifier()
+
     abortTransactionUnless(validateChannelIdentifier(portIdentifier, channelIdentifier))
     abortTransactionUnless(connectionHops.length === 1) // for v1 of the IBC protocol
     abortTransactionUnless(authenticateCapability(portPath(portIdentifier), portCapability))
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -343,12 +343,12 @@ function chanOpenTry(
                          counterpartyChannelIdentifier, connectionHops, version}
     provableStore.set(channelPath(portIdentifier, channelIdentifier), channel)
     channelCapability = newCapability(channelCapabilityPath(portIdentifier, channelIdentifier))
-    // only reset sequences if the previous channel didn't exist, else we might overwrite optimistically-sent packets
-    if (previous === null) {
-      provableStore.set(nextSequenceSendPath(portIdentifier, channelIdentifier), 1)
-      provableStore.set(nextSequenceRecvPath(portIdentifier, channelIdentifier), 1)
-      provableStore.set(nextSequenceAckPath(portIdentifier, channelIdentifier), 1)
-    }
+
+    // initialize channel sequences
+    provableStore.set(nextSequenceSendPath(portIdentifier, channelIdentifier), 1)
+    provableStore.set(nextSequenceRecvPath(portIdentifier, channelIdentifier), 1)
+    provableStore.set(nextSequenceAckPath(portIdentifier, channelIdentifier), 1)
+
     return channelCapability
 }
 ```
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -319,7 +319,7 @@ function chanOpenTry(
   counterpartyChosenChannelIdentifer: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  version: string,
+  version: string, // deprecated
   counterpartyVersion: string,
   proofInit: CommitmentProof,
   proofHeight: Height): CapabilityKey {
@@ -332,7 +332,7 @@ function chanOpenTry(
          previous.counterpartyPortIdentifier === counterpartyPortIdentifier &&
          previous.counterpartyChannelIdentifier === "" &&
          previous.connectionHops === connectionHops &&
-         previous.version === version)
+         previous.version === counterpartyVersion
         )
       channelIdentifier = previousIdentifier
     } else {
--- a/spec/core/ics-026-routing-module/README.md
+++ b/spec/core/ics-026-routing-module/README.md
@@ -41,6 +41,20 @@ The functions `newCapability` & `authenticateCapability` are defined as in [ICS 
 
 Modules must expose the following function signatures to the routing module, which are called upon the receipt of various datagrams:
 
+#### **ChanOpenInit**
+
+ChanOpenInit will verify that the relayer-chosen parameters
+are valid and perform any custom INIT logic.
+It may return an error if the chosen parameters are invalid
+in which case the handshake is aborted.
+If the provided version string is non-empty, ChanOpenInit should return
+the version string if valid or an error if the provided version is invalid.
+If the version string is empty, ChanOpenInit is expected to
+return a default version string representing the version(s)
+it supports.
+If there is no default version string for the application,
+it should return an error if provided version is empty string.
+
 ```typescript
 function onChanOpenInit(
   order: ChannelOrder,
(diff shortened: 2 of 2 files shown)


spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -332,7 +332,7 @@ function chanOpenTry(
          previous.counterpartyPortIdentifier === counterpartyPortIdentifier &&
          previous.counterpartyChannelIdentifier === "" &&
          previous.connectionHops === connectionHops &&
-         previous.version === version)
+         previous.version === counterpartyVersion
```
AdityaSripal, line 335: For simplicity, we enforce that in case of crossing hellos, previous version must be exactly same as counterparty version.

We could relax this requirement by passing the previous version into the callback, and letting the callback decide if they are compatible. However this is a more complicated solution to support a rather small edgecase, so it will not be supported for now.

colin-axner (approved):

spec/core/ics-026-routing-module/README.md
```diff
@@ -41,6 +41,20 @@ The functions `newCapability` & `authenticateCapability` are defined as in [ICS
...
+It may return an error if the chosen parameters are invalid
+in which case the handshake is aborted.
+If the provided version string is non-empty, ChanOpenInit should return
+the version string if valid or an error if the provided version is invalid.
+If the version string is empty, ChanOpenInit is expected to
+return a default version string representing the version(s)
+it supports.
+If there is no default version string for the application,
+it should return an error if provided version is empty string.
+
```
colin-axner, line 57:
```suggestion
onChanOpenInit will verify that the relayer-chosen parameters
are valid and perform any custom INIT logic.
It may return an error if the chosen parameters are invalid
in which case the handshake is aborted.
If the provided version string is non-empty, onChanOpenInit should return
the version string if valid or an error if the provided version is invalid.
If the version string is empty, onChanOpenInit is expected to
return a default version string representing the version(s)
it supports.
If there is no default version string for the application,
it should return an error if provided version is empty string.

```

colin-axner: ChanOpenInit -> onChanOpenInit

spec/core/ics-026-routing-module/README.md
```diff
@@ -49,35 +63,63 @@ function onChanOpenInit(
...
+ChanOpenTry will verify the relayer-chosen parameters along with the
+counterparty-chosen version string and perform custom TRY logic.
+If the relayer-chosen parameters
+are invalid, the callback must return an error to abort the handshake.
+If the counterparty-chosen version is not compatible with this modules
+supported versions, the callback must return an error to abort the handshake.
+If the versions are compatible, the try callback must select the final version
+string and return it to core IBC.
+ChanOpenTry may also perform custom initialization logic
```
colin-axner, line 81:
```suggestion
onChanOpenTry will verify the relayer-chosen parameters along with the
counterparty-chosen version string and perform custom TRY logic.
If the relayer-chosen parameters
are invalid, the callback must return an error to abort the handshake.
If the counterparty-chosen version is not compatible with this modules
supported versions, the callback must return an error to abort the handshake.
If the versions are compatible, the try callback must select the final version
string and return it to core IBC.
onChanOpenTry may also perform custom initialization logic
```
ChanOpenTry -> onChanOpenTry

colin-axner:
should this pr update the ICS apps as well (ICS20 and ICS27)? Or will that be done separately?

spec/core/ics-004-channel-and-packet-semantics/README.md
ethanfrey: I have a question about this whole "crossing hellos" logic.

I believe that made sense when the relayer selected the channel, and it was possible to init the exact same channel on both sides, which then try against each other.

With the channel auto-gen, there is no possibility that an init and a try on the same chain create the same (port, channel) identifier.

I think this case can be removed, which will simplify the code.

ethanfrey: Please explain to me where in practice this case can still occur. I may be missing something

colin-axner:
middleware spec probably needs to be updated as well

spec/core/ics-004-channel-and-packet-semantics/README.md
ethanfrey: Bump. 

I think this concept is no longer possible with auto-assigned channel_ids.
Can someone please verify this or correct me if I am wrong (and explain how this case occurs).

If we are cleaning up specs, removing or explaining this case would be a huge help.

spec/core/ics-026-routing-module/README.md
```diff
@@ -41,6 +41,20 @@ The functions `newCapability` & `authenticateCapability` are defined as in [ICS
 
 Modules must expose the following function signatures to the routing module, which are called upon the receipt of various datagrams:
 
+#### **ChanOpenInit**
+
+ChanOpenInit will verify that the relayer-chosen parameters
+are valid and perform any custom INIT logic.
```
crodriguezvega, line 47:
```suggestion
are valid and perform any custom `INIT` logic.
```

spec/core/ics-026-routing-module/README.md
crodriguezvega: onChanOpenInit -> `onChanOpenInit`

spec/core/ics-026-routing-module/README.md
crodriguezvega: onChanOpenTry -> `onChanOpenTry`

spec/core/ics-026-routing-module/README.md
```diff
@@ -49,35 +63,63 @@ function onChanOpenInit(
...
-  version: string) {
+  version: string) => (version: string, err: Error) {
     // defined by the module
 }
+```
+
+#### **ChanOpenTry**
+
+ChanOpenTry will verify the relayer-chosen parameters along with the
+counterparty-chosen version string and perform custom TRY logic.
```
crodriguezvega, line 74:
```suggestion
counterparty-chosen version string and perform custom `TRY` logic.
```

spec/core/ics-026-routing-module/README.md
```diff
@@ -49,35 +63,63 @@ function onChanOpenInit(
...
-  version: string,
-  counterpartyVersion: string) {
+  counterpartyVersion: string) => (version: string, err: Error) {
     // defined by the module
 }
+```
+
+#### **OnChanOpenAck**
 
+OnChanOpenAck will error if the counterparty selected version string
```
crodriguezvega, line 98:
```suggestion
`onChanOpenAck` will error if the counterparty selected version string
```

spec/core/ics-026-routing-module/README.md
```diff
@@ -49,35 +63,63 @@ function onChanOpenInit(
...
   portIdentifier: Identifier,
   channelIdentifier: Identifier,
   version: string) {
     // defined by the module
 }
+```
+
+#### **OnChanOpenConfirm**
+
+OnChanOpenConfirm will perform custom CONFIRM logic and may error to abort the handshake.
```
crodriguezvega, line 112:
```suggestion
`onChanOpenConfirm` will perform custom CONFIRM logic and may error to abort the handshake.
```

spec/core/ics-004-channel-and-packet-semantics/README.md
colin-axner: I think the existence of "crossing hellos" scenarios now acts as an optimization as opposed to a safety mechanism.

We have chainA and chainB. For some reason both chains want to open a channel with X set of parameters. 

chainA calls ChanOpenInit with X parameters (counterparty channel identifier unspecified)
chainB call ChanOpenInit with X parameters (counterparty channel identifier unspecified)

In the existing specification, a relayer could notice both sides wish to create the same channel. It can call ChanOpenTry specifying the existing counterparty channel identifier, thus connecting the two sides without creating additional channel identifiers.

If the case above is removed, a new channel identifier/end would need to be created on both chains to complete the handshakes even if both chains created channels which are compatible. 

To the best of my knowledge, there aren't any attacks possible if we removed crossing hello support (I really haven't thought about it much).

Given that probability of two chains wanting to open the same channel in the same moment is very low, I think it makes sense to explicitly disallow "crossing hellos". A fair amount of our bugs have come from crossing hello edge cases. IBC applications are currently required to understand the "crossing hello" concept (in relation to claiming channel capability). Crossing hellos makes onboarding onto handshake logic more difficult. 

cc @cwgoes

ethanfrey: Okay, this is an optimisation for an edge case that makes the main case more complex to implement and bug-prone?

I agree with explicitly removing support for this. I mean, if no one uses it and no one outside of IG and Chris really understand it...

AdityaSripal: Agreed, simply removing this code block on TRY and generating new channel identifier each time

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -315,30 +315,15 @@ function chanOpenTry(
...
-         previous.counterpartyChannelIdentifier === "" &&
-         previous.connectionHops === connectionHops &&
-         previous.version === counterpartyVersion
-        )
-      channelIdentifier = previousIdentifier
-    } else {
-      // generate a new identifier if the provided identifier was the sentinel empty-string
-      channelIdentifier = generateIdentifier()
-    }
+    channelIdentifier = generateIdentifier()
```
ethanfrey, line 325: So much simpler :smile:

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -315,30 +315,15 @@ function chanOpenTry(
...
-         previous.counterpartyPortIdentifier === counterpartyPortIdentifier &&
-         previous.counterpartyChannelIdentifier === "" &&
-         previous.connectionHops === connectionHops &&
-         previous.version === version)
-        )
-      channelIdentifier = previousIdentifier
-    } else {
-      // generate a new identifier if the provided identifier was the sentinel empty-string
-      channelIdentifier = generateIdentifier()
-    }
```
colin-axner, line 341: You need to remove `previous == null` [checks](https://github.com/cosmos/ibc/pull/629/files#diff-6d4b263fc3c751e8e3fbd765b2817241db22c8e3083c61685900fa4783669220R347)

colin-axner (commented):
Including the removal of crossing hellos in this pr is fine, but I'd prefer it to be done in a separate pr. I think mashing together changes which aren't necessarily associated is bad practice. It is easier to understand what changes are made and the justification for those changes if they are logically isolated in the commits pushed to the master branch. Each IBC implementor would (hopefully) update their implementations with these changes and we should make it as easy as possible for them to do so. If there isn't already, there should be some sort of CHANGELOG which can inform implementors of changes they should make

Ideally we would have 3 pr's:
- removing crossing hellos from connection
- removing crossing hellos from channel
- improvement of version negotiation in channel handshake

As an implementor, this is how I would breakup the issues/pr's from the proposed changes

As a side note, I'd love to see the justification for removing crossing hellos logged somewhere, otherwise there will be discussion of why crossing hellos was removed in the future

AdityaSripal:
I understand what you're saying but we also have to balance the fact that the spec repo moves much much slower than implementation (we are actively working to fix this!). But in the interim, I think it makes sense to respond to things that are semantically linked in the same PR to avoid long response times.

I'm planning to merge this PR as it currently is, and then write the following PRs:
- Connection handshake fixes
- Fixes to applications/middleware

cwgoes (approved):
ACK, all these simplifications make sense to me

PR #636, opened 2022-01-10
ICS4: Add ORDERED_ALLOW_TIMEOUT channel type

AdityaSripal:
Allow for ORDERED_ALLOW_TIMEOUT, by writing a timeout receipt on destination chain if packet is received after timeout. Once the timeout receipt is written, it can be proven on counterparty chain to timeout packet without closing the channel

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -42,7 +42,9 @@ A *bidirectional* channel is a channel where packets can flow in both directions
 
 A *unidirectional* channel is a channel where packets can only flow in one direction: from `A` to `B` (or from `B` to `A`, the order of naming is arbitrary).
 
-An *ordered* channel is a channel where packets are delivered exactly in the order which they were sent.
+An *ordered* channel is a channel where packets are delivered exactly in the order which they were sent. This channel type offers a very strict guarantee of ordering. Either, the packets are received in the order they were sent, or if a packet in the sequence times out; then all future packets are also not receivable and the channel closes.
+
+An *ordered_allow_timeout* channel is a less strict version of the *ordered* channel. Here, the channel logic will take a *best effort* approach to delivering the packets in order. In a stream of packets, the channel will relay all packets in order and if a packet in the stream times out, the timeout logic for that packet will execute and the rest of the later packets will continue processing in order. Thus, we **do not close** the channel on a timeout with this channel type.
 
 An *unordered* channel is a channel where packets can be delivered in any order, which may differ from the order in which they were sent.
 
@@ -50,6 +52,7 @@ An *unordered* channel is a channel where packets can be delivered in any order,
 enum ChannelOrder {
   ORDERED,
   UNORDERED,
+  ORDERED_ALLOW_TIMEOUT,
 }
 ```
 
@@ -73,7 +76,7 @@ interface ChannelEnd {
 ```
 
 - The `state` is the current state of the channel end.
-- The `ordering` field indicates whether the channel is ordered or unordered.
+- The `ordering` field indicates whether the channel is `unordered`, `ordered`, or `ordered_allow_timeout`.
 - The `counterpartyPortIdentifier` identifies the port on the counterparty chain which owns the other end of the channel.
 - The `counterpartyChannelIdentifier` identifies the channel end on the counterparty chain.
 - The `nextSequenceSend`, stored separately, tracks the sequence number for the next packet to be sent.
@@ -145,8 +148,9 @@ type OpaquePacket = object
 
 #### Ordering
 
-- On ordered channels, packets should be sent and received in the same order: if packet *x* is sent before packet *y* by a channel end on chain `A`, packet *x* must be received before packet *y* by the corresponding channel end on chain `B`.
-- On unordered channels, packets may be sent and received in any order. Unordered packets, like ordered packets, have individual timeouts specified in terms of the destination chain's height.
+- On *ordered* channels, packets should be sent and received in the same order: if packet *x* is sent before packet *y* by a channel end on chain `A`, packet *x* must be received before packet *y* by the corresponding channel end on chain `B`. If packet *x* is sent before packet *y* by a channel and packet *x* is timed out; then packet *y* and any packet sent after *x* cannot be received.
+- On *ordered_allow_timeout* channels, packets should be sent and received in the same order: if packet *x* is sent before packet *y* by a channel end on chain `A`, packet *x* must be received **or** timed out before packet *y* by the corresponding channel end on chain `B`.
+- On *unordered* channels, packets may be sent and received in any order. Unordered packets, like ordered packets, have individual timeouts specified in terms of the destination chain's height.
 
 #### Permissioning
 
@@ -585,11 +589,11 @@ The IBC handler performs the following steps in order:
 - Checks that the channel & connection are open to receive packets
 - Checks that the calling module owns the receiving port
 - Checks that the packet metadata matches the channel & connection information
-- Checks that the packet sequence is the next sequence the channel end expects to receive (for ordered channels)
+- Checks that the packet sequence is the next sequence the channel end expects to receive (for ordered and ordered_allow_timeout channels)
 - Checks that the timeout height has not yet passed
 - Checks the inclusion proof of packet data commitment in the outgoing chain's state
 - Sets a store path to indicate that the packet has been received (unordered channels only)
-- Increments the packet receive sequence associated with the channel end (ordered channels only)
+- Increments the packet receive sequence associated with the channel end (ordered and ordered_allow_timeout channels only)
 
 We pass the address of the `relayer` that signed and submitted the packet to enable a module to optionally provide some rewards. This provides a foundation for fee payment, but can be used for other techniques as well (like calculating a leaderboard).
 
@@ -624,7 +628,7 @@ function recvPacket(
 
     // all assertions passed (except sequence check), we can alter state
 
-    if (channel.order === ORDERED) {
+    if (channel.order === ORDERED || channel.order == ORDERED_ALLOW_TIMEOUT) {
       nextSequenceRecv = provableStore.get(nextSequenceRecvPath(packet.destPort, packet.destChannel))
       abortTransactionUnless(packet.sequence === nextSequenceRecv)
       nextSequenceRecv = nextSequenceRecv + 1
@@ -732,7 +736,7 @@ function acknowledgePacket(
     ))
 
     // abort transaction unless acknowledgement is processed in order
-    if (channel.order === ORDERED) {
+    if (channel.order === ORDERED || channel.order == ORDERED_ALLOW_TIMEOUT) {
       nextSequenceAck = provableStore.get(nextSequenceAckPath(packet.sourcePort, packet.sourceChannel))
       abortTransactionUnless(packet.sequence === nextSequenceAck)
       nextSequenceAck = nextSequenceAck + 1
@@ -823,7 +827,7 @@ function timeoutPacket(
     abortTransactionUnless(provableStore.get(packetCommitmentPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
            === hash(packet.data, packet.timeoutHeight, packet.timeoutTimestamp))
 
-    if channel.order === ORDERED {
+    if channel.order === ORDERED || channel.order == ORDERED_ALLOW_TIMEOUT {
       // ordered channel: check that packet has not been received
       abortTransactionUnless(nextSequenceRecv <= packet.sequence)
       // ordered channel: check that the recv sequence is as claimed
@@ -849,6 +853,7 @@ function timeoutPacket(
     // delete our commitment
     provableStore.delete(packetCommitmentPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
 
+    // only close on strictly ORDERED channels
     if channel.order === ORDERED {
       // ordered channel: close the channel
       channel.state = CLOSED
@@ -903,7 +908,7 @@ function timeoutOnClose(
       expected
     ))
 
-    if channel.order === ORDERED {
+    if channel.order === ORDERED || channel.order == ORDERED_ALLOW_TIMEOUT {
       // ordered channel: check that the recv sequence is as claimed
       abortTransactionUnless(connection.verifyNextSequenceRecv(
         proofHeight,
@@ -928,12 +933,6 @@ function timeoutOnClose(
 
     // delete our commitment
     provableStore.delete(packetCommitmentPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
-
-    if channel.order === ORDERED {
-      // ordered channel: close the channel
-      channel.state = CLOSED
-      provableStore.set(channelPath(packet.sourcePort, packet.sourceChannel), channel)
-    }
 
     // return transparent packet
     return packet
@@ -1011,6 +1010,8 @@ Jul 29, 2019 - Revisions to handle timeouts after connection closure
 Aug 13, 2019 - Various edits
 
 Aug 25, 2019 - Cleanup
+
+Jan 10, 2022 - Add ORDERED_ALLOW_TIMEOUT channel type and appropriate logic
 
 ## Copyright
 
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -643,7 +643,7 @@ function recvPacket(
             TIMEOUT_RECEIPT
           )
         }
-        break;
+        return;
 
       default:
         // unsupported channel type
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -650,7 +650,7 @@ function recvPacket(
         // for ORDERED_ALLOW_TIMEOUT, we do not abort on timeout
         // instead increment next sequence recv and write the sentinel timeout value in packet receipt
         // then return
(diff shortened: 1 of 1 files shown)


spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -42,14 +42,17 @@ A *bidirectional* channel is a channel where packets can flow in both directions
 
 A *unidirectional* channel is a channel where packets can only flow in one direction: from `A` to `B` (or from `B` to `A`, the order of naming is arbitrary).
 
-An *ordered* channel is a channel where packets are delivered exactly in the order which they were sent.
+An *ordered* channel is a channel where packets are delivered exactly in the order which they were sent. This channel type offers a very strict guarantee of ordering. Either, the packets are received in the order they were sent, or if a packet in the sequence times out; then all future packets are also not receivable and the channel closes.
```
ethanfrey, line 45: This is the current semantics of Ordered Channels? I did not know that.

That definitely makes sense to add a version with looser requirements.

(This will require a hard fork on both sides to add it?)

ethanfrey (commented):
Interesting

spec/core/ics-004-channel-and-packet-semantics/README.md
AdityaSripal: Correct it will require both sides to have the new type enabled

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -929,12 +934,6 @@ function timeoutOnClose(
     // delete our commitment
     provableStore.delete(packetCommitmentPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
 
-    if channel.order === ORDERED {
```
cwgoes, line 932: what does this change have to do with the new channel type?

cwgoes (changes requested):
ACK, but see comment

spec/core/ics-004-channel-and-packet-semantics/README.md
AdityaSripal: This is just in the wrong place originally.

This is pseudocode for when an ORDERED channel times out. But that is already specified on line 856.

But this function is for when a channel has already closed, and we want to time out all pending packets. So there is no  special case for ORDERED channels here.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -624,7 +628,7 @@ function recvPacket(
 
     // all assertions passed (except sequence check), we can alter state
 
-    if (channel.order === ORDERED) {
+    if (channel.order === ORDERED || channel.order == ORDERED_ALLOW_TIMEOUT) {
       nextSequenceRecv = provableStore.get(nextSequenceRecvPath(packet.destPort, packet.destChannel))
```
colin-axner, line 632: I don't understand. If packet sequence 5 times out, even if the channel doesn't close, I don't see any changes here that allow packet sequence 6 to be received

AdityaSripal: Yes great catch, this makes the changes less trivial as we need to somehow communicate to destination chain the sequences that have been timed out

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -859,6 +859,11 @@ function timeoutPacket(
       channel.state = CLOSED
       provableStore.set(channelPath(packet.sourcePort, packet.sourceChannel), channel)
     }
+    // on ORDERED_ALLOW_TIMEOUT, increment NextSequenceRecv so that next packet sequence can be received once this packet times out.
+    if channel.order === ORDERED_ALLOW_TIMEOUT {
+      nextSequenceRecv = nextSequenceRecv + 1
+      provableStore.set(nextSequenceRecvPath(packet.destPort, packet.destChannel), nextSequenceRecv)
```
colin-axner, line 865: this should be the nextSequenceAck (as it is on the sender side)

AdityaSripal:
As colin mentioned this feature will be more complicated than first envisioned.

The ack sequence can be correctly incremented when a timeout occurs. However, the receive sequence needs to be incremented on the counterparty chain.

Thus, the receiving chain needs to have this information propogated back to it somehow. This will probably require some slightly different relayer behaviour to enable.

e.g. We might have to relay the ack sequence to the receiving chain before being able to receive the next packet after timeout. Though, async acks complicates this

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -610,8 +624,31 @@ function recvPacket(
     abortTransactionUnless(connection !== null)
     abortTransactionUnless(connection.state === OPEN)
 
-    abortTransactionUnless(packet.timeoutHeight === 0 || getConsensusHeight() < packet.timeoutHeight)
-    abortTransactionUnless(packet.timeoutTimestamp === 0 || currentTimestamp() < packet.timeoutTimestamp)
+    switch channel.order {
+      case ORDERED:
```
colin-axner, line 628: Is it possible to do something like? I get that this is a typescript fall through, but coming from a go lens it performs completely different behaviour 
```go
case ORDERED || UNORDERED:
```
or
```go
case ORDERED, UNORDERED:
```

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -610,8 +624,31 @@ function recvPacket(
     abortTransactionUnless(connection !== null)
     abortTransactionUnless(connection.state === OPEN)
 
-    abortTransactionUnless(packet.timeoutHeight === 0 || getConsensusHeight() < packet.timeoutHeight)
-    abortTransactionUnless(packet.timeoutTimestamp === 0 || currentTimestamp() < packet.timeoutTimestamp)
+    switch channel.order {
```
colin-axner, line 627: I imagine the logic like so:
```
if packet.TimedOut() {
    switch channel.Order {
    case ORDERED_ALLOW_TIMEOUT:
        // write timeout reciept
    default:
        abort()
    }   
}
```
Main request is that logically, we are only performing certain logic if the packet has timed out

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -823,37 +869,64 @@ function timeoutPacket(
...
+        abortTransactionUnless(nextSequenceRecv <= packet.sequence)
+        // ordered channel: check that the recv sequence is as claimed
+        abortTransactionUnless(connection.verifyNextSequenceRecv(
+          proofHeight,
+          proof,
+          packet.destPort,
+          packet.destChannel,
+          nextSequenceRecv
+        ))
+        break;
```
colin-axner, line 884: what's the break for?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -823,37 +869,64 @@ function timeoutPacket(
...
+    // only close on strictly ORDERED channels
     if channel.order === ORDERED {
       // ordered channel: close the channel
       channel.state = CLOSED
       provableStore.set(channelPath(packet.sourcePort, packet.sourceChannel), channel)
     }
+    // on ORDERED_ALLOW_TIMEOUT, increment NextSequenceAck so that next packet can be acknowledged after this packet timed out.
+    if channel.order === ORDERED_ALLOW_TIMEOUT {
+      nextSequenceAck = nextSequenceAck + 1
+      provableStore.set(nextSequenceAckPath(packet.destPort, packet.destChannel), nextSequenceAck)
```
colin-axner, line 928: This should be sourcePort/sourceChannel

colin-axner (commented):
Concept ACK. Wonderful work. Clever solution

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -624,22 +661,31 @@ function recvPacket(
...
-      )
-    }
+    switch channel.order {
+      case ORDERED:
+      case ORDERED_ALLOW_TIMEOUT:
+        nextSequenceRecv = provableStore.get(nextSequenceRecvPath(packet.destPort, packet.destChannel))
+        abortTransactionUnless(packet.sequence === nextSequenceRecv)
+        nextSequenceRecv = nextSequenceRecv + 1
+        provableStore.set(nextSequenceRecvPath(packet.destPort, packet.destChannel), nextSequenceRecv)
+        break;
```
mpoke, line 671:
```suggestion
        // for ORDERED_ALLOW_TIMEOUT, write the sentinel timeout value in packet receipt
        if (channel.order == ORDERED_ALLOW_TIMEOUT  && getConsensusHeight() > packet.timeoutHeight && packet.timeoutHeight != 0) || (currentTimestamp() > packet.timeoutTimestamp && packet.timeoutTimestamp != 0) {
          packetReceipt = TIMEOUT_RECEIPT
        } 
        break
```

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -624,22 +661,31 @@ function recvPacket(
 
     // all assertions passed (except sequence check), we can alter state
```
mpoke, line 663:
```suggestion
    // assume the packet is successfully received 
    packetReceipt = SUCCESSFUL_RECEIPT
```

spec/core/ics-004-channel-and-packet-semantics/README.md
mpoke: I'd use the first variant, i.e., `case ORDERED || UNORDERED`. Seems less ambiguous.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -624,22 +661,31 @@ function recvPacket(
...
+      case UNORDERED:
+        // for unordered channels we must set the receipt so it can be verified on the other side
+        // this receipt does not contain any data, since the packet has not yet been processed
+        // it's the sentinel success receipt: []byte{0x01}
+        abortTransactionUnless(provableStore.get(packetReceiptPath(packet.destPort, packet.destChannel, packet.sequence) === null))
+        provableStore.set(
+          packetReceiptPath(packet.destPort, packet.destChannel, packet.sequence),
+          SUCCESFUL_RECEIPT
+        )
+      break;
```
mpoke, line 682:
```suggestion
      break
```
Set the PacketReceipt only once after the switch.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -624,22 +661,31 @@ function recvPacket(
...
 
+      default:
+        // unsupported channel type
+        abortTransactionUnless(true)
+    }
+
```
mpoke, line 688:
```suggestion
    provableStore.set(
      packetReceiptPath(packet.destPort, packet.destChannel, packet.sequence),
      packetReceipt
    )
    if (packetReceipt == TIMEOUT_RECEIPT) {
      return
    }
```

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -624,22 +661,31 @@ function recvPacket(
...
+          SUCCESFUL_RECEIPT
+        )
+      break;
 
+      default:
+        // unsupported channel type
+        abortTransactionUnless(true)
+    }
+    
     // log that a packet has been received
```
mpoke, line 689: As a side note about ICS4, shouldn't we call here the `OnRecvPacket()` callback function and in case it returns a non-nil ACK, invoke `writeAcknowledgement`?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -823,37 +869,64 @@ function timeoutPacket(
...
-        proofHeight,
-        proof,
-        packet.destPort,
-        packet.destChannel,
-        packet.sequence
-      ))
+    switch channel.order {
+      case ORDERED:
+        // ordered channel: check that packet has not been received
+        abortTransactionUnless(nextSequenceRecv <= packet.sequence)
```
mpoke, line 875: Is not related to the `ORDERED_ALLOW_TIMEOUT` channels, but shouldn't this condition be `abortTransactionUnless(nextSequenceRecv == packet.sequence)`? Why do we allow timeout messages regarding future packets. Maybe the packet with  `nextSequenceRecv` will still be received on time, even if another packet with e.g.,  `nextSequenceRecv+1` has timed out.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -823,37 +869,64 @@ function timeoutPacket(
...
+          packet.destPort,
+          packet.destChannel,
+          packet.sequence
+        ))
+        break;
+
+      // NOTE: For ORDERED_ALLOW_TIMEOUT, the relayer must first attempt the receive on the destination chain
+      // before the timeout receipt can be written and subsequently proven on the sender chain in timeoutPacket
+      case ORDERED_ALLOW_TIMEOUT:
+        abortTransactionUnless(connection.verifyPacketReceipt(
```
mpoke, line 900: In order for the following property to hold, we need to check for the sequence here, similar to the ORDERED case.
> On *ordered_allow_timeout* channels, packets should be sent and received in the same order: if packet *x* is sent before packet *y* by a channel end on chain `A`, packet *x* must be received **or** timed out before packet *y* by the corresponding channel end on chain `B`.

This means that a packet with sequence e.g., 5 cannot be timed out while we are still waiting for an ACK from sequence  e.g., 3

mpoke (changes requested):
Cool idea. Great work. See my comments below.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -610,8 +624,31 @@ function recvPacket(
...
+        // then return
+        if (getConsensusHeight() > packet.timeoutHeight && packet.timeoutHeight != 0) || (currentTimestamp() > packet.timeoutTimestamp && packet.timeoutTimestamp != 0) {
+          nextSequenceRecv = nextSequenceRecv + 1
+          provableStore.set(nextSequenceRecvPath(packet.destPort, packet.destChannel), nextSequenceRecv)
+          provableStore.set(
+          packetReceiptPath(packet.destPort, packet.destChannel, packet.sequence),
+            TIMEOUT_RECEIPT
+          )
+        }
+        return;
```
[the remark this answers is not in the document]
AdityaSripal: reordered

spec/core/ics-004-channel-and-packet-semantics/README.md
AdityaSripal: The packet is successfully received at this point. There's no assumption being made

spec/core/ics-004-channel-and-packet-semantics/README.md
AdityaSripal: Hmm yea that makes sense to me.

spec/core/ics-004-channel-and-packet-semantics/README.md
mpoke: What I meant by the comment is that we set a variable to `SUCCESSFUL_RECEIPT` to avoid multiple if conditions below, i.e.,`packetReceipt =  SUCCESSFUL_RECEIPT` unless stated otherwise.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -823,37 +869,68 @@ function timeoutPacket(
...
-        proofHeight,
-        proof,
-        packet.destPort,
-        packet.destChannel,
-        packet.sequence
-      ))
+    switch channel.order {
+      case ORDERED:
+        // ordered channel: check that packet has not been received
+        // only allow timeout on next sequence so all sequences before the timed out packet can be received
```
mpoke, line 875: This comment is a bit unclear. Did you mean to say "all sequences **AFTER** the timed out packet can be received"?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -823,37 +869,68 @@ function timeoutPacket(
...
+          packet.destChannel,
+          packet.sequence
+        ))
+        break;
+
+      // NOTE: For ORDERED_ALLOW_TIMEOUT, the relayer must first attempt the receive on the destination chain
+      // before the timeout receipt can be written and subsequently proven on the sender chain in timeoutPacket
+      case ORDERED_ALLOW_TIMEOUT:
+        // ordered channel: check that packet has not been received
+        // only allow timeout on next sequence so all sequences before the timed out packet can be received
```
mpoke, line 902: ditto

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -823,37 +869,68 @@ function timeoutPacket(
...
+        abortTransactionUnless(nextSequenceRecv == packet.sequence)
+        abortTransactionUnless(connection.verifyPacketReceipt(
```
mpoke, line 904: Shouldn't `verifyPacketReceipt` also get the sequence as param?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -823,37 +869,68 @@ function timeoutPacket(
...
+        abortTransactionUnless(connection.verifyPacketReceiptAbsence(
+          proofHeight,
+          proof,
+          packet.destPort,
+          packet.destChannel,
+          packet.sequence
+        ))
+        break;
+
+      // NOTE: For ORDERED_ALLOW_TIMEOUT, the relayer must first attempt the receive on the destination chain
```
mpoke, line 898: Not sure I understand why is this necessary. Why can't the relayer just send a timeout packet to the sender chain and handle the check as in the ORDERED case, i.e., using the `verifyNextSequenceRecv` method?

mpoke (approved):
Nice work. See my comments below; mostly nits. One thing worth looking into is that I don't find the reasoning for the `TIMEOUT_RECEIPT` packet receipt sentinel value to be clear enough.

PR #638, opened 2022-01-17
ICS27 Updates

AdityaSripal:
- remove metadata from portID and move it to version
- create structured JSON map for version and recommend as best practice for negotiating metadata
- remove deleting active channel on channel closing


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -175,40 +173,99 @@ An example of an active channel on the controller chain can look like this:
...
+Datagram: ChanOpenInit
+
+Chain Acted Upon: Controller
+
+Version: 
+```json
+{
+  "Version": "ics27-1",
+  "Encoding": ["supported encoding types", "..."],
+  "Tx_Types": ["supported transaction types", "..."],
```
AdityaSripal, line 213: I've included these two to be very explicit about any assumptions shared between the two chains. However, open to feedback on if we should standardize the encoding to all implementations.

Or if we should allow the tx_type to be implicit.

Note in the implementation there will only be one supported type for each

```
{
...
Encoding: protobuf,
Tx_Types: sdk_multi_msg,
}
```

colin-axner:
This is a lot to review at once. Is it possible to split up?

seantking:
> This is a lot to review at once. Is it possible to split up?

I don't mean to be awkward but I agree with @colin-axner  we should try to split up these IBC spec PRs a bit in the future.

AdityaSripal:
I could move the active channel unsetting out since that is logically isolated, though that is only a small part of the PR.

The rest of the changes are tightly coupled. So I could make the port changes a separate PR, however that on its own would be incorrect. Without the changes to the version. So then separating these, in my opinion, makes correctness hard to track because the changes that would make interchain accounts secure are now in separate PRs and it is harder to verify that the changes together still create a correct specification. I'm open to a different opinion if reviewers still think it is worth separating these two. But I would want to make sure then that people are still considering both PRs in conjunction with each other and making sure that the changes make sense together.

spec/app/ics-027-interchain-accounts/README.md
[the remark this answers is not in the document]
crodriguezvega: These 2 fields [are not part of the implementation in ibc-go](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/types/metadata.pb.go#L28), is that ok?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -175,40 +173,99 @@ An example of an active channel on the controller chain can look like this:
...
+  "Version": "ics27-1",
+  "Encoding": "negotiated_encoding_type",
+  "Tx_Types": "negotiated_tx_type",
+  "ControllerConnectionId": "self_connection_id",
+  "HostConnectionId": "counterparty_connection_id",
+  "Address": "interchain_account_address",
+}
+```
+
+Comments: On the ChanOpenAck step, the ICS27 application on the controller chain must verify the version string chosen by the host chain on ChanOpenTry. The controller chain must verify that it can support the negotiated encoding and tx type selected by the host chain. If either is unsupported, then it must return an error and abort the handshake.
```
crodriguezvega, line 265:
```suggestion
Comments: On the `OnChanOpenAck` step, the ICS-27 application on the controller chain must verify the version string chosen by the host chain on `OnChanOpenTry`. The controller chain must verify that it can support the negotiated encoding and tx type selected by the host chain. If either is unsupported, then it must return an error and abort the handshake.
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -329,21 +396,40 @@ function onChanOpenTry(
...
+  txType = SelectTxType(cpMetadata.TxType)
+  abortTransactionUnless(txType != "")
+
+  abortTransactionUnless(cpMetadata.ControllerConnectionId === counterpartyConnectionId)
+  abortTransactionUnless(cpMetadata.HostConnectionId === connectionId)
+  
+  metadata = {
+    "Version": "ics27-1",
+    "Encoding": encoding,
+    "TxType": txType,
```
crodriguezvega, line 426: `TxTypes`?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -329,21 +396,40 @@ function onChanOpenTry(
...
+    "ControllerConnectionId": cpMetadata.ControllerConnectionId,
+    "HostConnectionId": cpMetadata.HostConnectionId,
+    "Account": account,
```
crodriguezvega, line 429: This should be `Address`, right?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -352,11 +438,20 @@ function onChanOpenTry(
...
+  abortTransactionUnless(metadata.Version === "ics27-1")
+  abortTransactionUnless(IsSupportedEncoding(metadata.Encoding))
+  abortTransactionUnless(IsSupportedTxType(metadata.TxType))
+
+  abortTransactionUnless(metadata.ControllerConnectionId === connectionId)
+  abortTransactionUnless(metadata.HostConnectionId === counterpartyConnectionId)
+
   // state change to keep track of successfully registered interchain account
-  SetInterchainAccountAddress(portID, accAddr)
+  SetInterchainAccountAddress(portID, metadata.Account)
```
crodriguezvega, line 453: `.Address` instead, right?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -81,7 +81,7 @@ This is a helper function to open a new channel.
 This is a safety function in case of a channel closing and the controller chain needs to regain access to an interchain account on the host chain.
 
 ```typescript
-function InitChannel(portId: string, connectionId: string) returns (nil){
+function InitChannel(portId: string, connectionId: string) returns (nil) {
```
seantking, line 84: Can we remove this? We decided not to implement it for the time being in ibc-go and will instead leave it up to ica-auth application developers.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -94,9 +94,9 @@ function InitChannel(portId: string, connectionId: string) returns (nil){
 `RegisterInterchainAccount` is called on the `OnChanOpenTry` step during the channel creation handshake.
 
 ```typescript
-function RegisterInterchainAccount(accAddr: string, counterpartyPortId: string) returns (nil){
+function RegisterInterchainAccount(counterpartyPortId: string) returns (nil) {
    // checks to make sure the account has not already been registered
-   // creates a new address on chain 
+   // creates a new address on chain deterministically given counterpartyPortId
```
seantking, line 99: I guess this should probably also mention the connection-id based on [colins issue](https://github.com/cosmos/ibc-go/issues/767) yesterday? I think the grammar here needs an update either way.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -175,40 +173,99 @@ An example of an active channel on the controller chain can look like this:
...
+ICS-27 takes advantage of [ICS-04 channel version negotiation](../../core/ics-004-channel-and-packet-semantics/README.md#versioning) to negotiate metadata and channel parameters during the channel handshake. The metadata will contain the encoding format along with the transaction type so that the counterparties can agree on the structure and encoding of the interchain transactions. The metadata sent from the host chain on the TRY step will also contain the interchain account address, so that it can be relayed to the controller chain. At the end of the channel handshake, both the controller and host chains will store a mapping of the controller chain portID to the newly registered interchain account address ([account registration flow](#Register-account-flow)).
```
seantking, line 190: This is a big improvement. Nice

AdityaSripal:
Superceded by #641 and #642

PR #639, opened 2022-01-17
ICS27 cleanup

AdityaSripal:
- Minor cleanups and fixes to keep consistent with implementation

### DIFF
--- a/spec/app/ics-027-interchain-accounts/README.md
+++ b/spec/app/ics-027-interchain-accounts/README.md
@@ -107,7 +107,7 @@ function RegisterInterchainAccount(counterpartyPortId: string) returns (nil) {
 `AuthenticateTx` checks that the signer of a particular message is the interchain account associated with the counterparty portID of the channel that the IBC packet was sent on.
 
 ```typescript
-function AuthenticateTx(msgs []Any, portId string) error {
+function AuthenticateTx(msgs []Any, portId string) returns (error) {
     // GetInterchainAccountAddress(portId)
     // interchainAccountAddress != signer.String() return error
 }
@@ -118,10 +118,11 @@ function AuthenticateTx(msgs []Any, portId string) error {
 Executes each message sent by the owner account on the controller chain.
 
 ```typescript
-function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, msgs []Any) error {
+function ExecuteTx(sourcePort: string, msgs []Any) returns (resultString, error) {
   // validate each message
   // verify that interchain account owner is authorized to send each message
   // execute each message
+  // return result of transaction
 }
 ```
 
@@ -142,10 +143,6 @@ function SetInterchainAccountAddress(portId string, address string) returns (str
 
 // Retrieves the interchain account from state.
 function GetInterchainAccountAddress(portId string) returns (string, bool){
-}
-
-// DeleteActiveChannelID removes the active channel keyed by the provided portID stored in state
-function (k Keeper) DeleteActiveChannelID(portId string) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -118,10 +118,11 @@ function AuthenticateTx(msgs []Any, portId string) error {
 Executes each message sent by the owner account on the controller chain.
 
 ```typescript
-function ExecuteTx(sourcePort: string, destPort: string, destChannel: string, msgs []Any) error {
```
AdityaSripal, line 121: These fields are not used

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -501,34 +498,29 @@ function onChanCloseConfirm(
...
-	        if err != nil {
-                    ack = NewErrorAcknowledgement(err.Error())
-                }
-	}
-
-	// NOTE: acknowledgment will be written synchronously during IBC handler execution.
-	return ack
+  switch data.Type {
+  case types.EXECUTE_TX:
+  msgs, err := types.DeserializeTx(data.Data)
```
ethanfrey, line 507: Isn't the norm "throwing exceptions" in this ics spec "pseudo code".

It seems modelled heavily on TypeScript

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -501,34 +498,29 @@ function onChanCloseConfirm(
...
+  if err != nil {
+    return NewErrorAcknowledgement(err)
```
colin-axner, line 509: should the spec discuss the implication of returning the error in the acknowledgement? That is, including the error string in the acknowledgement makes error strings state machine breaking

colin-axner: Maybe this should be optional, as we will not include the error in the ibc-go implementation

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -501,34 +498,29 @@ function onChanCloseConfirm(
...
+  }
+
+  // ExecuteTx calls the AuthenticateTx function defined above 
+  result, err = ExecuteTx(ctx, packet.SourcePort, packet.DestinationPort, packet.DestinationChannel, msgs)
+  if err != nil {
+    return NewErrorAcknowledgement(err)
+  }
+
+  // return acknowledgement containing the transaction result after executing on host chain
+  return NewAcknowledgement(result)
```
colin-axner, line 519: ditto for implications of including the result

seantking (approved):
Looks good. 

Agree with @ethanfrey about the state of the pseudo-code in general. We should have a standard for this.

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: We don't want to throw an exception in this case and abort the transaction, we want to write an error acknowledgement and commit the state changes

ethanfrey: Understood. But the idiom `msg, err := fn()` is very go. The rest of the ibc specs are modelled on TypeScript.

I would say:

```ts
try {
  const msg = deserializeTx(data.data)
  // ... other stuff
} catch (err) {
  // return error packet
}
```

would look more in line with the other specs (but still needs some adjustment).

I believe the specs were intentionally not written in Go to keep a clear distinction between them and the implementation. Whether that was a good decision I will leave to you and @cwgoes to discuss. I just like consistency in the style.

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: > should the spec discuss the implication of returning the error in the acknowledgement? That is, including the error string in the acknowledgement makes error strings state machine breaking

Very good point.
That has proved to be the case for CosmWasm with submessages and we did hit a case where one node included the stack trace and fell out of consensus.

I like to include error message, but this is a very good thing to think through (and check code paths)

spec/app/ics-027-interchain-accounts/README.md
ethanfrey: The result of a transaction is currently hashed and included in the the Tendermint block header (DataHash I believe). The data field and the code (u32) are the only fields currently consensus breaking (not events). I think including Data is fully acceptable, where errors and events may need more consideration.

colin-axner: Great insight! I did not know the data field is included in consensus. I agree, if the result is already included in consensus then it is fully acceptable to include in the acknowledgement

AdityaSripal:
Superceded by #643

PR #641, opened 2022-01-20
ICS27 Port Changes

AdityaSripal:
Changes to PortID from #638 

Only includes removing connectionIDs from the portID and updating relevant APIs.
To ensure correctness, this also includes the recommended change in https://github.com/cosmos/ibc-go/issues/767

### DIFF
--- a/spec/app/ics-027-interchain-accounts/README.md
+++ b/spec/app/ics-027-interchain-accounts/README.md
@@ -51,15 +51,15 @@ This specification defines the general way to register an interchain account and
 #### **InitInterchainAccount**
 
 `InitInterchainAccount` is the entry point to registering an interchain account.
-It generates a new controller portID using the owner account address and connection identifiers.
+It generates a new controller portID using the owner account address.
 It will bind to the controller portID and
 call 04-channel `ChanOpenInit`. An error is returned if the controller portID is already in use.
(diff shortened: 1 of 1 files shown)


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -94,9 +94,9 @@ function InitChannel(portId: string, connectionId: string) returns (nil){
 `RegisterInterchainAccount` is called on the `OnChanOpenTry` step during the channel creation handshake.
 
 ```typescript
-function RegisterInterchainAccount(accAddr: string, counterpartyPortId: string) returns (nil){
+function RegisterInterchainAccount(counterpartyPortId: string, connectionID: string) returns (nil) {
    // checks to make sure the account has not already been registered
-   // creates a new address on chain 
+   // creates a new address on chain deterministically given counterpartyPortId and underlying connectionID
```
AdityaSripal, line 99: Ref: https://github.com/cosmos/ibc-go/issues/767

seantking: Is it worth mentioning that it's the host side connection-id?

seantking (approved):

colin-axner (approved):

PR #642, opened 2022-01-20
Metadata Negotiation in ICS27

AdityaSripal:
Pull Metadata negotiation logic out of #638


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -185,28 +185,85 @@ An example of an active channel on the controller chain can look like this:
...
+Comments: The ICS-27 application on the host chain is responsible for returning this version given the counterparty version set by the controller chain in INIT. The host chain must select a single encoding type and a single tx type that is supported by both the host chain and the controller chain (ie. included in counterparty version). If no such encoding or tx type exists, then the host chain must return an error and abort the handshake.
+The host chain must also generate the interchain account address and populate the address field in the version with the interchain account address string.
+
+**ACK**
+
+Initiator: Relayer
+
+Datagram: ChanOpenTry
+
+Chain Acted Upon: Host
```
colin-axner, line 251:
```suggestion
Datagram: ChanOpenAck

Chain Acted Upon: Controller
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -329,21 +393,38 @@ function onChanOpenTry(
...
   // only allow channels to be created on host chain if the counteparty port ID
   // is in the expected controller portID format.
   abortTransactionUnless(validateControllerPortParams(counterpartyPortIdentifier))
-  // assert that version is expected format `ics27-1.interchain-account-address`
-  abortTransactionUnless(validateVersion(version))
-  // assert that the counterparty version is `ics27-1`  
-  abortTransactionUnless(counterpartyVersion === "ics27-1")
   // create the interchain account 
-  RegisterInterchainAccount(accAddr, counterpartyPortIdentifier)
+  address = RegisterInterchainAccount(accAddr, counterpartyPortIdentifier)
```
colin-axner, line 405: this must use the host connection ID as well. Is this in a followup pr?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -185,28 +185,85 @@ An example of an active channel on the controller chain can look like this:
...
+Datagram: ChanOpenInit
+
+Chain Acted Upon: Controller
+
+Version: 
+```json
+{
+  "Version": "ics27-1",
+  "Encoding": ["supported encoding types", "..."],
+  "TxTypes": ["supported transaction types", "..."],
```
colin-axner, line 213: personally I'd prefer to keep this simple. INIT should propose an encoding/tx type, if the counterparty can't support them, it should abort. It won't always be clear what encoding type is best for this channel for the counterparty to select. The initiating actor should have the best insight, but in this situation also has the least control (outside of only setting one supported type - which is what I'm proposing)

Then we don't need `len(encoding) == 1` checks in the rest of the handshake

Is there a concrete example of when multiple supported encoding types may be useful to propose on INIT?

colin-axner (commented):
Looks great! General ACK, but see comments

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: The spec is as general as possible but I agree that the SDK implementation should be extremely simple. We should just include a single string not even in an array on the SDK.

colin-axner: Both chains need to encode/decode the same way. Wouldn't one chain not using an array cause an issue?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -162,7 +162,7 @@ This port will be used to create channels between the controller & host chain fo
 2. The controller chain emits an event signaling to open a new channel on this port given a connection. 
 3. A relayer listening for `ChannelOpenInit` events will continue the channel creation handshake.
 4. During the `OnChanOpenTry` callback on the host chain an interchain account will be registered and a mapping of the interchain account address to the owner account address will be stored in state (this is used for authenticating transactions on the host chain at execution time). 
-5. During the `OnChanOpenAck` callback on the controller chain a record of the interchain account address registered on the host chain during `OnChanOpenTry` is set in state with a mapping from portID -> interchain account address. See [version negotiation](#Version-negotiation) section below for how to implement this.
+5. During the `OnChanOpenAck` callback on the controller chain a record of the interchain account address registered on the host chain during `OnChanOpenTry` is set in state with a mapping from portID -> interchain account address. See [metadata negotiation](#Metadata-negotiation) section below for how to implement this.
```
seantking, line 165: @colin-axner Did we say that we needed to update this mapping to include connection seq?

seantking (approved):
Great improvements :handshake: 

Would it be worth adding [this check](https://github.com/cosmos/ibc-go/pull/785) in this PR as well? I can also do a small follow-up.

spec/app/ics-027-interchain-accounts/README.md
colin-axner: The portID is unique from the perspective of the controller chain, so no the connection seq is not needed

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: > The initiating actor should have the best insight, but in this situation also has the least control

Yea I suppose we can just limit to 1 type to ensure initiator either gets their preferred type or noting at all.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -185,28 +185,85 @@ An example of an active channel on the controller chain can look like this:
...
+Chain Acted Upon: Host
+
+Version: 
+```json
+{
+  "Version": "ics27-1",
+  "Encoding": "negotiated_encoding_type",
+  "TxTypes": "negotiated_TxType",
+  "ControllerConnectionId": "self_connection_id",
+  "HostConnectionId": "counterparty_connection_id",
```
damiannolan, line 237: For `ChanOpenTry` host should be the "self" connection ID.

```suggestion
  "ControllerConnectionId": "counterparty_connection_id",
  "HostConnectionId": "self_connection_id",
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -173,28 +173,85 @@ An example of an active channel on the controller chain can look like this:
...
+
+Version: 
+```json
+{
+  "Version": "ics27-1",
+  "Encoding": "requested_encoding_types",
+  "TxTypes": "requested_transaction_types",
+  "ControllerConnectionId": "self_connection_id",
+  "HostConnectionId": "counterparty_connection_id",
+  "Address": ""
```
colin-axner, line 204:
```suggestion
  "Version": "ics27-1",
  "Encoding": "requested_encoding_type",
  "TxType": "requested_transaction_type",
  "ControllerConnectionId": "self_connection_id",
  "HostConnectionId": "counterparty_connection_id",
  "Address": ""
```

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -173,28 +173,85 @@ An example of an active channel on the controller chain can look like this:
...
+}
+```
+
+Comments: The address is left empty since this will be generated and relayed back by the host chain.
```
colin-axner, line 208: I think it should be okay to leave encoding type and tx type empty

spec/app/ics-027-interchain-accounts/README.md
colin-axner: would need to be applied below as well

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -317,21 +381,39 @@ function onChanOpenTry(
...
+  cpMetadata = UnmarshalJSON(counterpartyVersion)
+  abortTransactionUnless(cpMetadata.Version === "ics27-1")
+  // select an encoding and tx type. If none are supported by executing chain then
+  // abortTransaction
+  encoding = SelectEncoding(cpMetadata.Encoding)
+  abortTransactionUnless(encoding != "")
+  txType = SelectTxType(cpMetadata.TxType)
+  abortTransactionUnless(txType != "")
+  abortTransactionUnless(cpMetadata.ControllerConnectionId === counterpartyConnectionId)
+  abortTransactionUnless(cpMetadata.HostConnectionId === connectionId)
```
colin-axner, line 405: now that the encoding/tx type are a single value, should this be changed to `IsSupported`? It'd be odd for a host chain to select a encoding not requested by a controller

colin-axner (commented):
Where do we plan to specify what the `TxType` values mean. For example, in ibc-go if we specify the tx type as `sdk.Msg`, where can we put this in the spec?

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: Why? INIT should always choose ask for something in particular. Are you talking about relayers leaving version string empty? Or the chain itself committing a version without encoding and tx type?

colin-axner: Yea it is unnecessary

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -173,28 +173,85 @@ An example of an active channel on the controller chain can look like this:
...
+Datagram: ChanOpenAck
+
+Chain Acted Upon: Controller
+
+CounterpartyVersion: 
+```json
+{
+  "Version": "ics27-1",
+  "Encoding": "negotiated_encoding_type",
+  "TxType": "negotiated_TxType",
```
colin-axner, line 246: nit: there's an inconsistency between:
`negotiated_encoding_type`
`negotiated_TxType`

maybe use `negotiated_tx_type` since the ConnectionId's use `connection_id`

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -301,10 +358,17 @@ function onChanOpenInit(
...
   abortTransactionUnless(activeChannel === nil)
+
+  // validate metadata
+  metadata = UnmarshalJSON(version)
+  abortTransactionUnless(metadata.Version === "ics27-1")
+  // all elements in encoding list and tx type list must be supported
+  abortTransactionUnless(IsSupportedEncoding(metadata.Encoding))
+  abortTransactionUnless(IsSupportedTxType(metadata.TxType))
+  abortTransactionUnless(metadata.ControllerConnectionId === connectionId)
+  abortTransactionUnless(metadata.HostConnectionId === counterpartyConnectionId)
```
colin-axner, line 371: for future reference, it might make sense to define helper functions like `validateMetada`. That's how we have it implemented in the code and it helps ensure if you change validation metadata, you don't have to do it in 3 places

colin-axner (approved):

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: I agree this should be done in implementation. I'm going to pass on doing it here, mainly because there are subtle but important changes depending on whether you are on controller or host chain and i think those get obfuscated if it is put in a helper function that will be visually separate from the rest of the code in the handshake callbacks

colin-axner: We have `validateControllerMetadata` and `validateHostMetadata` fwiw

mpoke (approved):
Looks great. I left some comments bellow.

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -317,21 +381,37 @@ function onChanOpenTry(
   channelIdentifier: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  version: string,
-  counterpartyVersion: string) {
+  counterpartyVersion: string) (version: string) {
```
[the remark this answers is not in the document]
AdityaSripal: Yes we should create a new issue to update ICS26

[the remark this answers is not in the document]
AdityaSripal: It does seem to match? https://github.com/cosmos/ibc/blob/master/spec/core/ics-026-routing-module/README.md#chanopentry

PR #643, opened 2022-01-20
Fix api documentation in ICS27

AdityaSripal:


colin-axner (approved):
ACK, but as noted on #639 the error should probably not be included in the acknowledgement

cwgoes (approved):
ACK

PR #644, opened 2022-01-21
ICS27: rename portID and portprefix

crodriguezvega:
Spec update based on [cosmos/ibc-go#778](https://github.com/cosmos/ibc-go/issues/778).

Since the portID for ICS20 is `transfer`, for naming consistency I propose to rename the `PortID` for the interchain accounts host submodule to `icahost` and the `PortPrefix` for the interchain accounts controller submodule to `icacontroller`.


AdityaSripal (approved):
Thanks for the update @crodriguezvega !

cwgoes (approved):
ACK

what a nice world it would be if the spec were _typechecked_

PR #647, opened 2022-01-25
Interchain query spec draft

schnetzlerjoe:
Draft of interchain query spec to discuss and add to

### DIFF
--- a/spec/app/ics-interchain-queries/README.md
+++ b/spec/app/ics-interchain-queries/README.md
@@ -1,0 +1,146 @@
+---
+ics: TBA
+title: Interchain Query
+stage: draft
+category: IBC/APP
+requires: 25, 26, 27
+kind: instantiation
+author: Joe Schnetzler <schnetzlerjoe@gmail.com>
+created: 2022-01-06
+modified: 2020-01-09
+---
+
+## Synopsis
+
+This documents aims to document the structure, plan and implementation of the Interchain Queries Module allowing for cross-chain querying of state from IBC enabled chains.
+
+### Motivation
+
+Interchain Accounts (ICS-27) brings one of the most important features IBC offers, cross chain transactions (on-chain). Limited in this functionality is the querying of state from one chain, on another chain. Adding interchain querying via the Interchain Query module, gives unlimited flexibility to chains to build IBC enabled protocols around Interchain Accounts and beyond.
+
+### Definitions 
+
+- `Querying Chain`: The chain that is interested in getting data from another chain (Queried Chain). The Querying Chain is the chain that implements the Interchain Query Module into their chain.
+- `Queried Chain`: The chain that's state is being queried via the Querying Chain. The Queried Chain gets queried via a relayer utilizing its RPC client which is then submitted back to the Querying Chain.
+
+### Desired Properties
+
+- Permissionless: A Querying Chain can query a chain and implement cross-chain querying without any approval from a third party or chain governance.
+
+- Minimal Querying Chain Work: A Queried Chain has to do no implementation work or add any module to enable cross chain querying. By utilizing an RPC client on a relayer, this is possible.
+
+- Modular: Adding cross-chain querying should be as easy as implementing a module in your chain.
+
+- Control Queried Data: The Querying Chain should have ultimate control on how to handle queried data.
+
+- Incentivization: In order to incentivize relayers for participating in interchain queries, a bounty is paid.
+
+## Technical Specification
+
+### General Design 
+
+The Querying Chain starts with the implementation of the Interchain Query Module by adding the module into their chain.
+
+The general flow for interchain queries starts with a Cross Chain Query Request from the Querying Chain which is listened to by relayers. Upon recognition of a cross chain query, relayers utilize a ABCI Query Request to query data from the Queried Chain. Upon success, the relayer submits a `MsgSubmitQueryResult` to the Querying chain.
+
+On failure of a query, relayers submit `MsgSubmitQueryErrorResult` to the Querying chain. Alternatively on timeout relayers submit `MsgSubmitQueryTimeoutResult`.
+
+### Data Structures
+
+A CrossChainABCIQueryRequest data type is used to specify the query. Included in this is the.
+
+```go
+type CrossChainABCIQueryRequest struct {
+	Data          []byte
+	Path          string
+	timeoutHeight uint64
+	Bounty        sdk.Coin
+	ChainId       string
+}
+```
+
+```go
+type QueryResult struct {
+	Data    []byte
+	Height  uint64
+	ChainId string
+}
+```
+
+```go
+type QueryErrorResult struct {
+	Error   string
+	ChainId string
+}
+```
+
+```go
+type QueryTimeoutResult struct {
+	timeoutHeight  string
+	ChainId        string
+}
+```
+
+### Messages
+
+
+```go
+func MsgCrossChainABCIQueryRequest(
+	
+) *cobra.Command {
+  //Msg for query request submit message
+}
+```
+
+```go
+func MsgSubmitQueryResult(
+
+) *cobra.Command {
+  //Msg to submit a query result
+}
+```
+
+```go
+func MsgSubmitQueryErrorResult(
+
+) *cobra.Command {
+  //Msg to submit a query error result
+}
+```
+
+```go
+func MsgSubmitQueryTimeoutResult(
+
+) *cobra.Command {
+  //Msg to submit a query timeout result
+}
+```
+
+```go
+func MsgGetQueries(
+
+) *cobra.Command {
+  //Msg to get a list of queries
+}
+```
+
+```go
+// Add -- max-height flag to allow for querying the highest height for a query (most recent)?
+func MsgGetQuery(
+
+) *cobra.Command {
+  //Msg to get a query via id
+}
+```
+
+### Events
+
+```go
+func EmitQueryEvent(ctx sdk.Context, timeoutHeight exported.Height) {
+	//Event to trigger query event on relayer
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+		)
+	})
+}
+```
--- a/spec/app/ics-interchain-queries/README.md
+++ b/spec/app/ics-interchain-queries/README.md
@@ -7,7 +7,7 @@ requires: 25, 26, 27
 kind: instantiation
 author: Joe Schnetzler <schnetzlerjoe@gmail.com>
 created: 2022-01-06
-modified: 2020-01-09
+modified: 2022-02-12
 ---
 
 ## Synopsis
@@ -22,6 +22,9 @@ Interchain Accounts (ICS-27) brings one of the most important features IBC offer
 
 - `Querying Chain`: The chain that is interested in getting data from another chain (Queried Chain). The Querying Chain is the chain that implements the Interchain Query Module into their chain.
 - `Queried Chain`: The chain that's state is being queried via the Querying Chain. The Queried Chain gets queried via a relayer utilizing its RPC client which is then submitted back to the Querying Chain.
+- `Key`: A Key in the querying chain module is a user specified query identifier for a query so a user can identify various query types. For example, if you were to query stakers/delegators on Gaia (Cosmos Hub) and store it in state. You can set the `key` as `stakers` and thus query via the key later on as needed.
+
+``Note:`` that a query in state can only have one unique key-id representation. For example, if a query with key = stakers & id = 0 is already in state, the query will fail if you try to add another query with key = stakers & id = 0. A query with key = stakers & id = 1 will succeed however (as long as that is not in state already).
 
 ### Desired Properties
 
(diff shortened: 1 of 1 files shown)


cwgoes:
IBC packets should not need to be _sent_ in order to conduct a cross-chain state query, that is inefficient. I suggest an approach based on instructing relayers to create a specific cross-chain state proof through some sort of standardised interface instead.

schnetzlerjoe:
> IBC packets should not need to be _sent_ in order to conduct a cross-chain state query, that is inefficient. I suggest an approach based on instructing relayers to create a specific cross-chain state proof through some sort of standardised interface instead.

Hey! Yes, that is the idea. The idea is to utilize ABCI Query already implemented on relayers. So chain submits the transaction to Query module. Relayer detects this query in state via a query event that is sent, the relayer then goes and performs an ABCI Query (includes proof) and takes the results and submits it into state on the querying chain. No packets or IBC work really needed. This also allows querying without other chains having to integrate the module (saves many developer hours).

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,146 @@
+---
+ics: TBA
+title: Interchain Query
+stage: draft
+category: IBC/APP
```
AdityaSripal, line 5: should this be under APP? Considering we're not implementing ICS26 here

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,146 @@
...
+	timeoutHeight uint64
+	Bounty        sdk.Coin
+	ChainId       string
+}
+```
+
+```go
+type QueryResult struct {
+	Data    []byte
+	Height  uint64
```
AdityaSripal, line 65: This should have a proof as well, so it can be verified

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+A CrossChainABCIQueryRequest data type is used to specify the query. Included in this is the.
+
+```go
+type CrossChainABCIQueryRequest struct {
+	Data          []byte
+	Path          string
+	timeoutHeight uint64
+	Bounty        sdk.Coin
+	ChainId       string
```
AdityaSripal, line 58:
```suggestion
	ClientId       string
```

ChainID is not unique. ClientId is how one chain uniquely identifies another chain.

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+On failure of a query, relayers submit `MsgSubmitQueryErrorResult` to the Querying chain. Alternatively on timeout relayers submit `MsgSubmitQueryTimeoutResult`.
+
+### Data Structures
+
+A CrossChainABCIQueryRequest data type is used to specify the query. Included in this is the.
+
+```go
+type CrossChainABCIQueryRequest struct {
+	Data          []byte
```
AdityaSripal, line 54: Why does the request have a data field?

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,146 @@
...
+type QueryResult struct {
+	Data    []byte
+	Height  uint64
+	ChainId string
+}
+```
+
+```go
+type QueryErrorResult struct {
+	Error   string
```
AdityaSripal, line 72: Proof?

spec/app/ics-interchain-queries/README.md
AdityaSripal: Ditto for all other structs

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,146 @@
...
+	ChainId string
+}
+```
+
+```go
+type QueryTimeoutResult struct {
+	timeoutHeight  string
```
AdityaSripal, line 79: How do you prove a timeout of a query on the queried chain? It definitely is not similar to the timeout semantics of a packet, so that will need to be made clear.

Here, I think the only thing that would make sense is to allow this message to be sent to the querying chain if an error or successful query result has not been relayed to the querying chain by the time it reaches the timeout height.

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,146 @@
...
+}
+```
+
+### Messages
+
+
+```go
+func MsgCrossChainABCIQueryRequest(
+	
+) *cobra.Command {
```
AdityaSripal, line 90: Spec should not have a dependency on cobra. Is this meant to show the messages users might submit on the querying chain to send out a query request?

Are we expecting users themselves to make these queries? Why would they, since they can just directly query the queried chain?

I think the query module would only be useful as a way for modules within the querying chain to query state of other chains as part of its state machine logic. Thus, the query module is an Oracle service for the querying chain that can provide information about the state of other chains.

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+) *cobra.Command {
+  //Msg to get a query via id
+}
+```
+
+### Events
+
+```go
+func EmitQueryEvent(ctx sdk.Context, timeoutHeight exported.Height) {
```
AdityaSripal, line 139: Please note that this must be emitted by the querying chain

AdityaSripal (changes requested):

spec/app/ics-interchain-queries/README.md
hxrts: would it make sense to let a user post cross-chain data themselves by allowing both the query and response to be posted together?

cwgoes:
> > IBC packets should not need to be _sent_ in order to conduct a cross-chain state query, that is inefficient. I suggest an approach based on instructing relayers to create a specific cross-chain state proof through some sort of standardised interface instead.
> 
> Hey! Yes, that is the idea. The idea is to utilize ABCI Query already implemented on relayers. So chain submits the transaction to Query module. Relayer detects this query in state via a query event that is sent, the relayer then goes and performs an ABCI Query (includes proof) and takes the results and submits it into state on the querying chain. No packets or IBC work really needed. This also allows querying without other chains having to integrate the module (saves many developer hours).

Splendid - then I misunderstood, my bad. The approach you propose makes total sense to me :) .

spec/app/ics-interchain-queries/README.md
schnetzlerjoe: To me I feel like it falls best under app (it's essentially an extension of interchain accounts) but if you think otherwise then?

spec/app/ics-interchain-queries/README.md
schnetzlerjoe: Good point. Will change

spec/app/ics-interchain-queries/README.md
schnetzlerjoe: I was following the same method via the ABCI query on relayer. Also allows more advanced querying. But you prob are right. Keep it via path.

spec/app/ics-interchain-queries/README.md
schnetzlerjoe: The idea here is that upon a certain height on the querying chain, a timeout occurs and is submitted.

spec/app/ics-interchain-queries/README.md
schnetzlerjoe: Oops. Yeah, no should not rely on cobra. The best way would be to just build keepers so then other modules can import the keepers and query based on their needs?

spec/app/ics-interchain-queries/README.md
joe-bowman: Data here is the json blob of parameters, rights?

spec/app/ics-interchain-queries/README.md
schnetzlerjoe: @joe-bowman agreed. On my local implementation I actually have it as a keeper that runs and checks for timeouts every endblocker.

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,151 @@
...
+- Permissionless: A Querying Chain can query a chain and implement cross-chain querying without any approval from a third party or chain governance.
+
+- Minimal Querying Chain Work: A Queried Chain has to do no implementation work or add any module to enable cross chain querying. By utilizing an RPC client on a relayer, this is possible.
+
+- Modular: Adding cross-chain querying should be as easy as implementing a module in your chain.
+
+- Control Queried Data: The Querying Chain should have ultimate control on how to handle queried data. Like querying for a certain query form/type.
+
+- Incentivization: In order to incentivize relayers for participating in interchain queries, a bounty is paid.
+
```
AdityaSripal, line 40: In addition to desired properties, we should also in this case have a section to underscore what the querying module does not guarantee.

Should be clear to any users of querying module what **is** promised by interchain querying and what **is not** promised.

For example:

Since there's no prior negotiation, we cannot assume that data coming back from queried chain will be in an expected format.

We also cannot assume that the value stored under the key will be stable, since it may change after the query result has been submitted

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,151 @@
...
+On failure of a query, relayers submit `MsgSubmitQueryResult` with the `success` flag as 0 to the Querying chain. Alternatively on timeout based on the height of the querying chain, the querying chain will submit `SubmitQueryTimeoutResult` with the timeout height specified.
+
+### Data Structures
+
+A CrossChainABCIQueryRequest data type is used to specify the query. Included in this is the `Path` which is the path field of the query i.e: /custom/auth/account. `Key` is the data key to name the query i.e: `pools` or `stakers`. `Id` is the id of the query with each key to id being unique i.e: stakers-0 or stakers-1275 (this example follows key-id format). `TimeoutHeight` specifies the timeout height on the querying chain to timeout the query. `Bounty` is a bounty that is given to the relayer for participating in the query. `ClientId` is used to identify the chain of interest. 
+
+```go
+type CrossChainABCIQueryRequest struct {
```
AdityaSripal, line 56: Currently the relayer has complete control over which height on the queried chain to use.

This is not a problem with regular IBC apps, because the IBC protocol ensures that the data queried (commitments, acks) will never change and thus the relayer can query at any height and is guaranteed by the protocol to get the same value.

However, with interchain querying, the keys that are being queried can have different values at different heights.

Thus, a malicious relayer can choose to query a height that has a value that benefits them the most.

In fact, with the current spec a relayer can choose to query at a height before the query was even posted on the querying chain.

Thus, it changes the security model from relayers being purely liveness facilitators to them being able to actively manipulate what data goes on chain.

We should not have a different security model for this feature, since that is bound to cause mistakes. we should remove the ability for relayers to influence the outcome a query by specifying the desired height to be queried in the request itself.

The relayer then **must** return the result from a query at that height, and once again we get back to the original security model where the relayer is just able to affect liveness

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,151 @@
...
+	Path           string
+	Key            string
+	Id             string
+	TimeoutHeight  uint64
```
AdityaSripal, line 60:
```suggestion
        QueryHeight Height // height on queried chain at which result must be queried from
	TimeoutHeight  Height // height on querying chain by which point result must be posted
```

AdityaSripal: Also see: https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/types/height.go#L29

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,151 @@
...
+	ClientId       string
+}
+```
+
+```go
+type QueryResult struct {
+	Data     []byte
+	Key      string
+	Id       string
+	Height   uint64
```
AdityaSripal, line 71:
```suggestion
```

Not needed since Height should be specified by querying agent

AdityaSripal (changes requested):
Nice work. There are two main changes requested.

1. Remove dependence on ABCI
2. Remove ability of relayer to affect result data

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+- Incentivization: In order to incentivize relayers for participating in interchain queries, a bounty is paid.
+
+## Technical Specification
+
+### General Design 
+
+The Querying Chain starts with the implementation of the Interchain Query Module by adding the module into their chain.
+
+The general flow for interchain queries starts with a Cross Chain Query Request from the Querying Chain which is listened to by relayers. Upon recognition of a cross chain query, relayers utilize a ABCI Query Request to query data from the Queried Chain. Upon success, the relayer submits a `MsgSubmitQueryResult` to the Querying chain with the success flag as 1.
```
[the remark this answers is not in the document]
AdityaSripal: This is made even easier with the changes scheduled on the client refactor. See: https://github.com/cosmos/ibc/pull/687

mpoke:
Closing. Replaced by https://github.com/cosmos/ibc/pull/735

PR #659, opened 2022-02-15
ICS28: Extend consumer InitGenesis

mpoke:
Extends the CCV specification to cover the changes made by https://github.com/cosmos/interchain-security/pull/30, i.e., `InitGenesis` on consumer should return a list of validator updates, 
and https://github.com/cosmos/interchain-security/pull/31, i.e., query the provider chain for a consumer chain genesis state. 

Currently, the mechanism of disseminating the consumer chain genesis state is outside the scope of this specification. CCV assumes that the **_same_** genesis state is disseminated to all the correct validators in the initial validator set of the consumer chain. To aid developers, the mechanism implemented (see the above pull requests) is described as a possible approach.


spec/app/ics-028-cross-chain-validation/technical_specification.md
```diff
@@ -28,8 +28,9 @@ Before describing the data structures and sub-protocols of the CCV protocol, we
   (for more details, take a look at the [ABCI documentation](https://docs.tendermint.com/v0.34/spec/abci/abci.html)). 
   In this specification we define the following methods that handle messages that are of particular interest to the CCV protocol:
   - `InitGenesis()` -- Called when the chain is first started, on receiving an `InitChain` message from the consensus engine. 
+    This is also where the application can inform the underlying consensus engine of the initial validator set.
```
AdityaSripal, line 31:
```suggestion
    This is also where the application must inform the underlying consensus engine of the initial validator set.
```

spec/app/ics-028-cross-chain-validation/technical_specification.md
```diff
@@ -139,8 +140,13 @@ The CCV module is initialized through the `InitGenesis` method when the chain is
 - On the consumer chain, the genesis state is described by the following interface:
   ```typescript
   interface ConsumerGenesisState {
-    providerClientState: ClientState
-    providerConsensusState: ConsensusState
+    newChain: Bool // true for new consumer chain, false for chain restart
+    providerClientId: Identifier // empty for a new chain
+    providerChannelId: // empty for a new chain
+    providerClientState: ClientState // nil on restart
```
AdityaSripal, line 146: Can you explain the two cases of an `NewChain` genesis and a restart genesis before this?

AdityaSripal: Personally I don't think the restart genesis is necessary to put in the specification. It is more a quirk of the SDK than anything else

spec/app/ics-028-cross-chain-validation/technical_specification.md
```diff
@@ -510,18 +521,72 @@ function onChanOpenConfirm(
...
+    abortSystemUnless(gs.providerConsensusState.Valid() == true)
+    // - has the providerClientId unspecified
+    abortSystemUnless(gs.providerClientId == "")
+    // - has the providerChannelId unspecified
+    abortSystemUnless(gs.providerChannelId == "")
+    // - has the unbondingPackets empty
+    abortSystemUnless(gs.unbondingPackets IS empty)
+    // - contains the initial validator set that matches 
+    //   the validator set in the providerConsensusState (see ICS 7)
+    abortSystemUnless(gs.initialValSet == gs.providerConsensusState.validatorSet)
```
AdityaSripal, line 543: We check hashes technically

AdityaSripal (changes requested):

spec/app/ics-028-cross-chain-validation/technical_specification.md
AdityaSripal, line 31:
```suggestion
    This is also where the application must inform the underlying consensus engine of the initial validator set.
```

The ABCI application overall MUST return a valset on init genesis. An individual module within the application may or may not return a valset, but something in application must return valset on initgenesis

spec/app/ics-028-cross-chain-validation/technical_specification.md
mpoke: Yes, but this is a detail of implementation. ICS7 defines the ConsensusState as having a validator set, not a hash, i.e., https://github.com/cosmos/ibc/blob/marius%2Fccv-evidence/spec/client/ics-007-tendermint-client/README.md#consensus-state

spec/app/ics-028-cross-chain-validation/technical_specification.md
mpoke: According to the [ABCI documentation on InitChain](https://docs.tendermint.com/v0.34/spec/abci/abci.html#initchain), 
> If ResponseInitChain.Validators is empty, the initial validator set will be the RequestInitChain.Validators

However, in the SDK this [line](https://github.com/cosmos/cosmos-sdk/blob/7869d38e94d5e691f453880d1be40af60121b85a/types/module/module.go#L338) entails that "Exactly one module must return a non-empty validator set update to correctly initialize the chain"

spec/app/ics-028-cross-chain-validation/technical_specification.md
mpoke: I agree. I'll remove anything related to ExportGenesis and restarting a chain.

spec/app/ics-028-cross-chain-validation/technical_specification.md
```diff
@@ -531,10 +550,23 @@ function InitGenesis(state: ConsumerGenesisState) {
...
+  - The initial validator set is returned to the consensus engine.
 - Error condition:
+  - The genesis state contains no valid provider client state, where the validity is defined as in [ICS 7](../../client/ics-007-tendermint-client).
+  - The genesis state contains no valid provider consensus state, where the validity is defined as in [ICS 7](../../client/ics-007-tendermint-client).
+  - The genesis state contains an empty initial validator set.
+  - The genesis state contains an initial validator set that does not match the validator set in the provider consensus state.
   - The capability for the port `ConsumerPortId` cannot be claimed.
 
-> **Note**: CCV assumes that `ConsumerGenesisState` is disseminated to all validators. To ensure that validators receive the correct genesis state, the governance proposal also contains a hash of the `ConsumerGenesisState` (see `CreateConsumerChainProposal`). 
+> **Note**: CCV assumes that the _same_ consumer chain genesis state is disseminated to all the correct validators in the initial validator set of the consumer chain.
```
josef-widder, line 561: "all the correct validators": Does that mean to "all the operators managing keys of validators?". Perhaps this should be more explained in what the operators have to do to launch the chain. Where to get information, and where to put it, etc?

mpoke: I agree that should be clear for operators, but I'm not sure it should be part of the spec. Probably is better to have a tutorial or FAQ for operators on how to deal w/ interchain security. What do you think?

spec/app/ics-028-cross-chain-validation/technical_specification.md
AdityaSripal: Ahh that is quite interesting. I don't know in practice if `RequestInitChain` does pass in a Validators field. This is a good question to clarify with the tendermint team

mpoke: From my discussion with @sergio-mena:

> `RequestInitChain` contains the set of validators coming from the genesis file. When handling  `RequestInitChain` , the app (e.g., SDK) can override that set with a different one. If the app does not return anything in `ResponseInitChain` , then Tendermint will simply reuse the value it used to populate `RequestInitChain` , which is the one found in the genesis info

> If the genesis info contains an empty validator set, and the app doesn't provide anything in `ResponseInitChain` either, then Tendermint [returns an error](https://github.com/tendermint/tendermint/blob/2183d90d05e1886132f71a4cc68c3048c45261a6/internal/consensus/replay.go#L343). 

> The restriction introduced in the SDK is **_further constraining_** this behavior: the SDK must always override the initial validator set .  But an app using directly ABCI, doesn't necessarily need to override the validator set from genesis

@AdityaSripal  I would suggest to keep the IBC spec consistent with the ABCI spec and not Cosmos SDK, i.e., leave "the application MAY inform". What do you say? 
I also think this should be address in Cosmos SDK to avoid confusion and possible bugs.

PR #660, opened 2022-02-17
Marius/ccv evidence

Raulnori77:
Asi


AdityaSripal:
@mpoke will open the PR for CCV slashing when it is ready

PR #662, opened 2022-02-22
ICS8: Add clearer WASM client motivation

AdityaSripal:
From discussion with @ethanfrey , the motiviations for ICS8 are not clearly articulated in the spec and to the wider community.

### DIFF
--- a/spec/client/ics-008-wasm-client/README.md
+++ b/spec/client/ics-008-wasm-client/README.md
@@ -20,9 +20,11 @@ Currently adding a new client implementation or upgrading an existing one requir
 
 This may be acceptable when adding new client types, since the number of unique consensus algorithms that need to be supported is currently small. However, this process will become very tedious when it comes to upgrading light clients.
 
-Without dynamically upgradable light clients, a chain that wishes to upgrade its consensus algorithm (and thus break existing light clients) must wait for all counterparty chains to perform a hard upgrade that adds support for the upgraded light client before it itself can perform an upgrade on its chain. This will inevitably slow the pace of upgrades in the IBC network, since the deployment of an upgrade on even a very experimental, fast-moving chain will be blocked by an upgrade to a high-value chain that will be inherently more conservative.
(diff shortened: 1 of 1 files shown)


spec/client/ics-008-wasm-client/README.md
```diff
@@ -16,7 +16,13 @@ This specification document describes an interface to a client (verification alg
...
+Currently adding a new client implementation or upgrading an existing one requires a hard upgrade because the client implementations are part of the static chain binary. Any change to the on-chain light client code is dependent on chain governance to approve an upgrade before it can be deployed.
+
+This may be acceptable when adding new client types, since the number of unique consensus algorithms that need to be supported is currently small. However, this process will become very tedious when it comes to upgrading light clients.
+
+Without dynamically upgradable light clients, a chain that wishes to upgrade its consensus algorithm (and thus break existing light clients) must wait for all counterparty chains to perform a hard upgrade that adds support for the upgraded light client before it itself can perform an upgrade on its chain. This will inevitably slow the pace of upgrades in the IBC network, since the deployment of an upgrade on even a very experimental, fast-moving chain will be blocked by an upgrade to a high-value chain that will be inherently more conservative.
```
ethanfrey, line 23: Let's make this more clear...

"a chain that wishes to upgrade its consensus algorithm (and thus break existing light clients)" would be if a chain upgraded to a hypothetical tendermint v2, or if they switch from tendermint to honeybadger. 

They could even switch from PoS to liquid staking to PoE without needing such an upgrade (people confuse the x/staking module and consensus algorithm often)

spec/client/ics-008-wasm-client/README.md
```diff
@@ -16,7 +16,13 @@ This specification document describes an interface to a client (verification alg
...
+
+With dynamically upgradable light clients enabled by WASM, a chain may upgrade its consensus algorithm whenever it wishes and relayers may upgrade the client code of all counterparty chains without requiring the counterparty chains to perform an upgrade themselves. This prevents a dependency on counterparty chains when considering upgrading one's own consensus algorithm.
```
ethanfrey, line 25: "With dynamically upgradable light clients enabled by WASM, a chain may upgrade its consensus algorithm whenever it wishes"

Assuming *all counterparty chains* have *previously upgraded* to support Wasm light clients, a chain no longer has to wait for them to add support for it's future light client before upgrading, as it can deploy it's light client to the chains.

ethanfrey (commented):
Thank you for adding this.

Added some comments for clarity. I think adding one concrete example after the motivation would help a lot. This whole repo is full of abstract ideas that are very hard to approach if you don't understand the concrete use cases.

Also, it would be good to add a link/reference to the (future / to be written?) spec that includes how a light client could be upgraded. Because if the counterparties have the go code, we can do tendermint v1 to tendermint v2 without Wasm clients. And understanding how that works in general is an important basis, then seeing how Wasm clients enable that process to be more dynamic and faster.

mpoke (approved):
LGTM.

PR #677, opened 2022-03-07
ICS4: Channel Upgradability Spec

AdityaSripal:


Raulnori77 (approved):

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -0,0 +1,584 @@
...
+}
+```
+
+
+### Cancel Upgrade Process
+
+During the upgrade handshake a chain may cancel the upgrade by writing an error receipt into the error path and restoring the original channel to `OPEN`. The counterparty must then restore its channel to `OPEN` as well. A relayer can facilitate this by calling `CancelChannelUpgrade`:
+
+```typescript
+function cancelChannelUpgrade(
```
colin-axner, line 507: missing application callback

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -0,0 +1,582 @@
...
+}
+```
+
+### Timeout Upgrade Process
+
+It is possible for the channel upgrade process to stall indefinitely on UPGRADE_TRY if the UPGRADE_TRY transaction simply cannot pass on the counterparty; for example, the upgrade feature may not be enabled on the counterparty chain.
+
+In this case, we do not want the initializing chain to be stuck indefinitely in the `UPGRADE_INIT` step. Thus, the `UpgradeInit` message will contain a `TimeoutHeight` and `TimeoutTimestamp`. The counterparty chain is expected to reject `UpgradeTry` message if the specified timeout has already elapsed.
+
+A relayer must then submit an `UpgradeTimeout` message to the initializing chain which proves that the counterparty is still in its original state. If the proof succeeds, then the initializing chain shall also restore its original channel and cancel the upgrade.
```
mpoke, line 530: We should be consistent about the existing messages throughout the spec:
- `ChannelUpgradeTimeoutMsg`
- `ChannelUpgradeCancelMsg`

This should also be the case for the connection update, i.e., 
- `ConnUpgradeTimeoutMsg`
- `ConnUpgradeCancelMsg`

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -0,0 +1,582 @@
...
+    // if timeoutHeight is defined and proof is from before timeout height
+    // then abort transaction
+    abortTransactionUnless(upgradeTimeout.timeoutHeight.IsZero() || proofHeight >= upgradeTimeout.timeoutHeight)
+    // if timeoutTimestamp is defined then the consensus time from proof height must be greater than timeout timestamp
+    connection = queryConnection(currentChannel.connectionIdentifier)
+    abortTransactionUnless(upgradeTimeout.timeoutTimestamp.IsZero() || getTimestampAtHeight(connection, proofHeight) >= upgradeTimeout.timestamp)
+
+    // counterparty channel must be proved to still be in OPEN state
+    abortTransactionUnless(counterpartyChannel.State === OPEN)
+    abortTransactionUnless(channel.client.verifyChannelState(proofHeight, proofChannel, counterpartyChannel))
```
mpoke, line 556: ditto (`verifyChannelState` not matching signature)

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -0,0 +1,582 @@
...
+    // proof must be from a height after timeout has elapsed. Either timeoutHeight or timeoutTimestamp must be defined.
+    // if timeoutHeight is defined and proof is from before timeout height
+    // then abort transaction
+    abortTransactionUnless(upgradeTimeout.timeoutHeight.IsZero() || proofHeight >= upgradeTimeout.timeoutHeight)
+    // if timeoutTimestamp is defined then the consensus time from proof height must be greater than timeout timestamp
+    connection = queryConnection(currentChannel.connectionIdentifier)
+    abortTransactionUnless(upgradeTimeout.timeoutTimestamp.IsZero() || getTimestampAtHeight(connection, proofHeight) >= upgradeTimeout.timestamp)
+
+    // counterparty channel must be proved to still be in OPEN state
+    abortTransactionUnless(counterpartyChannel.State === OPEN)
```
mpoke, line 555:
```suggestion
    // counterparty channel must be proved to still be in OPEN state or UPGRADE_INIT state (crossing hellos)
    abortTransactionUnless(counterpartyChannel.State === OPEN || counterpartyChannel.State == UPGRADE_INIT)
```
Similarly to the connection upgrade.

mpoke (changes requested):
Looks good. ACK the protocol. See my comments below.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -0,0 +1,584 @@
...
+
+    // verify proofs of counterparty state
+    abortTransactionUnless(verifyChannelState(currentChannel, proofHeight, proofChannel, currentChannel.counterpartyChannelIdentifier, counterpartyChannel))
+    abortTransactionUnless(verifyUpgradeTimeout(currentChannel, proofHeight, proofUpgradeTimeout, currentChannel.counterpartyChannelIdentifier, upgradeTimeout))
+
+    if currentChannel.state == UPGRADE_INIT {
+        // if there is a crossing hello, ie an UpgradeInit has been called on both channelEnds,
+        // then we must ensure that the proposedUpgrade by the counterparty is the same as the currentChannel
+        // except for the channel state (upgrade channel will be in UPGRADE_TRY and current channel will be in UPGRADE_INIT)
+        // if the proposed upgrades on either side are incompatible, then we will restore the channel and cancel the upgrade.
```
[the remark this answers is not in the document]
AdityaSripal: Agreed this is a complicated case, and we could design it to handle this such that chainB goes to OPEN. But I think that is quite difficult to engineer and evaluate for correctness. It is better for the first implementation to just revert the channel in this case. This is an inconvenience when we reach this edge case, since we have to restart the handshake but I believe that is fine.

So we will start with:

chainA, UPGRADE_INIT, timeout 5
chainB, UPGRADE_INIT, timeout 10

at time 7, both chains attempt to go to TRY. chainA succeeds and chainB fails. So we have:

chainA, UPGRADE_TRY
chainB, OPEN (reverted and error receipt written. See pseudocode below where we restore channel if timeout has passed)

Then at a time < 10 relayer can send a `CancelUpgradeMsg` to chainA:

chainA, OPEN (reverted)
chainB, OPEN (reverted and error receipt written)

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -0,0 +1,584 @@
...
+
+    // both channel ends must be mutually compatible.
+    // this function has been left unspecified since it will depend on the specific structure of the new channel.
+    // It is the responsibility of implementations to make sure that verification that the proposed new channels
+    // on either side are correctly constructed according to the new version selected.
+    if !IsCompatible(counterpartyChannel, proposedUpgradeChannel) {
+        restoreChannel()
+        return
+    }
+
```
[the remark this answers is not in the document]
AdityaSripal: Sorry my mistake. Needed to define and prove counterpartyChannel earlier. counterpartyChannel is the channel that initiator upgraded to.

proposedUpgradeChannel is the channel that TRY is going to upgrade to.

Similar to the connection upgrade channel there needs to be some check on compatibility that will not be specified here since it is upgrade-specific.

mpoke (approved):
Great work. Just one minor comment.

PR #686, opened 2022-03-14
Application Upgrade Callbacks

AdityaSripal:


colin-axner (approved):
Excellent!

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+#### **OnChanUpgradeInit**
+
+`onChanUpgradeInit` will verify that the upgrade parameters 
+are valid and perform any custom `UpgradeInit` logic.
+It may return an error if the chosen parameters are invalid 
+in which case the upgrade handshake is aborted.
+If the provided version string is empty, `onChanUpgradeInit` should return 
+a default version string or an error if the provided version is invalid.
```
mpoke, line 44: The sentence is unclear.
> or an error if the provided version is invalid.

How can an empty version be invalid?

Also, it seems that the following two sentences are describing the same thing.

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -0,0 +1,141 @@
+# Application Upgrade Callbacks
+
+## Synopsis
+
+This standard document specifies the interfaces and state machine logic that IBC applications must implement in order to enable existing channels to upgrade their applications after the initial channel handshake.
```
mpoke, line 5: Same confusion as with the title. 
> enable existing channels to upgrade their applications

Is this true? Or are the callbacks necessary to enable the endpoint modules to adapt to a channel upgrade.?

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -0,0 +1,141 @@
...
+`onChanUpgradeAck` will error if the counterparty selected version string
+is invalid to abort the handshake. It may also perform custom ACK logic.
+
+After `onChanUpgradeAck` returns, the application upgrade is complete on this end so any 
+auxilliary data stored for the purposes of recovery is no longer needed and may be deleted.
+
+The application MUST have its state fully migrated to start processing packet data according to the new application parameters by the time the callback returns.
+
+```typescript
+function onChanUpgradeAck(
```
mpoke, line 101: `onChanUpgradeAck()` does not return an error as stated above.

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -0,0 +1,141 @@
...
+function onChanUpgradeConfirm(
+  portIdentifier: Identifier,
+  channelIdentifier: Identifier) {
+    // defined by the module
+}
+```
+
+#### **OnChanUpgradeRestore**
+
+`onChanUpgradeRestore` will be called on `cancelChannelUpgrade` and `timeoutChannelUpgrade` to restore the application to its pre-upgrade state.
```
mpoke, line 129: `onChanUpgradeRestore` is also called by `restoreChannel()`, which may be called from `chanUpgradeTry` and `chanUpgradeAck`

mpoke (changes requested):
In general it looks good. I'm not sure I understood the utility of these functions. See my comments below.

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -0,0 +1,141 @@
+# Application Upgrade Callbacks
```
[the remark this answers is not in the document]
AdityaSripal: These are callbacks to upgrade the application during a channel upgrade. If the application does not need to do any upgrade and it is only IBC channel data that is changing, then these can all no-op. I think personally it is clearer as-is, but open to other opinions.

spec/core/ics-026-routing-module/UPGRADES.md
AdityaSripal: > enable existing channels to upgrade their applications

this is true. It will allow a transfer v1 to upgrade to breaking version transfer v2 using the channel upgrade process.

spec/core/ics-026-routing-module/UPGRADES.md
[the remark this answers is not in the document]
mpoke: Maybe just add this explanation for clarity.

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### **OnChanUpgradeAck**
+
+`onChanUpgradeAck` will error if the counterparty selected version string
+is invalid to abort the handshake. It may also perform custom ACK logic.
+
+After `onChanUpgradeAck` returns, the application upgrade is complete on this end so any 
+auxilliary data stored for the purposes of recovery is no longer needed and may be deleted.
+
+The application MUST have its state fully migrated to start processing packet data according to the new application parameters by the time the callback returns.
```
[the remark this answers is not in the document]
AdityaSripal: So the application parameters are any parameters that application needs to run the application logic that is needed for the channel.

This will be the channel parameters like version, but it may also be completely internal parameters defined and maintained by the end application.

These parameters might be created by the application callbacks during the channel upgrade process

mpoke (approved):

Issue #672, opened 2022-03-03
ICS28: Replace "Initiator" w/ "Caller" and “Trigger Event”


mpoke:
In the technical specification of CCV (see https://github.com/cosmos/ibc/blob/marius/ccv/spec/app/ics-028-cross-chain-validation/technical_specification.md), the subprotocols are specified by describing every method. This entails the pseudocode followed by the following fields:
- Initiator
- Expected precondition
- Expected postcondition
- Error condition

To remove ambiguity, these fields should be restructured, i.e.,
- `Caller` The entity invoking the method (e.g., another ABCI module). This is useful to keep track of the interfaces with other modules. 
- `Trigger Event` The event triggering the call to this method (e.g., a message is received). This is useful to reason about the overall logic of the protocol.   
- `Precondition` The internal state that enables the call of this method, i.e., so called "guards" in TLA+. 
- `Postcondition` The modifications to the internal state as a result of executing this method. This should also include `abortTransactionUnless` calls (i.e., reverting a transaction). 
- `Error Condition` The errors triggered as a result of executing this method. This consists of `abortSystemUnless` calls.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #696, opened 2022-03-23
ICS28: Replace "Initiator" w/ "Caller" and “Trigger Event”

mpoke:
Close https://github.com/cosmos/ibc/issues/672

See issue for more details.


jovankomatovic (commented):
I like the new separation.
It would maybe make sense to write explicitly that a "Trigger Event" is an "invocation" event, whereas a "Precondition" represents a "state".
Otherwise, it is pretty clean (in my opinion).

AdityaSripal (approved):
LGTM

josef-widder (approved):
Looks great. I think the caller, trigger event, etc. would also help if added to the core ICSs eventually to better understand the context of all the functions.

Issue #683, opened 2022-03-10
ICS28: Handle pending clients


mpoke:
Handle pending clients (i.e., `pendingClient`) once the spawn time has passed. For more details, see https://github.com/cosmos/interchain-security/issues/51

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #697, opened 2022-03-23
ICS28: Handle pending proposals to spawn consumer chains

mpoke:
Close https://github.com/cosmos/ibc/issues/683

Pending proposals are governance proposals to spawn new consumer chains with `spawnTime` in the future. These proposals must be handle by the provider chain once `spawnTime` passes.


josef-widder (approved):
Looks good. Is there a corresponding high level discussion in the overview (or somewhere else) about this, and that operators need to get the consumer chain running within spawnTime?

mpoke:
> Looks good. Is there a corresponding high level discussion in the overview (or somewhere else) about this, and that operators need to get the consumer chain running within spawnTime?

In the overview, no. I'd assume that all this will be documented together with the code, i.e., how to start a consumer chain.

josef-widder:
> > Looks good. Is there a corresponding high level discussion in the overview (or somewhere else) about this, and that operators need to get the consumer chain running within spawnTime?
> 
> In the overview, no. I'd assume that all this will be documented together with the code, i.e., how to start a consumer chain.

That's what I thought. We should just not forget that someone should write "how to start a consumer chain".  ;-) Perhaps we can open an issue, and collect important points to cover in this document?

PR #704, opened 2022-04-04
ICS28: CCV Reward Distribution subprotocol

mpoke:
The partial specification of Cross-Chain Validation.

Changes to base (i.e., [ccv](https://github.com/cosmos/ibc/tree/marius/ccv)):
- add dependency to ICS20
- add overview of Reward Distribution
- update the signatures of `onChanOpenInit()` and `onChanOpenTry` (cf. https://github.com/cosmos/ibc/pull/629)
- add  `CCVHandshakeMetadata` struct
- add timeouts to `SendPacket()`
- technical spec of the Reward Distribution subprotocol


spec/app/ics-028-cross-chain-validation/overview_and_basic_concepts.md
```diff
@@ -279,3 +281,26 @@ The following figure shows an overview of the Consumer Initiated Slashing operat
...
+
+At the beginning of every block, the rewards for the previous block are pooled into a distribution module account. 
+The Reward Distribution operation of CCV enables every consumer chain to transfer a fraction of these rewards to the provider chain. 
+The operation consists of two steps that are depicted in the following figure:
+
+![Reward Distribution](./figures/ccv-distribution-overview.png?raw=true)
+
+- At the beginning of every block on the consumer chain, a fraction of the rewards are transferred to an account on the consumer CCV module.
+- At regular intervals (e.g., every `100` blocks), the consumer CCV module sends the accumulated rewards to the distribution module account on the provider chain through an IBC token transfer packet (as defined in [ICS 20](../ics-020-fungible-token-transfer/README.md)). 
+  Note that the IBC transfer packet is sent over a separate unordered channel.
```
josef-widder, line 300: This may add some delay / asynchrony compared to the single chain case. For instance, after I undelegate I may still receive fees that I earned before I undelegated. As this is different from the single case system, it may make sense to make this explicit here.

josef-widder (approved):
Looks good. I think the first version is straight forward. I left a comment regarding the differences to the single chain case that might be highlighted

mpoke:
Commit https://github.com/cosmos/ibc/pull/704/commits/2a15d41de95f6b5005e4ce82ae20be5041f0d1ce adds an invariant for the rewards distribution sub-protocol.

spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
```diff
@@ -128,6 +134,10 @@ CCV provides the following system invariants.
   >
   > **Note:** The *Slash Invariant* also ensures that if a delegator starts unbonding an amount `x` of tokens from `val` before height `hi`, then `x` will not be slashed, since `x` is not part of `Token(Power(c,hi,val))`.
 
+- ***Distribution Invariant***: If a consumer chain sends to the provider chain an amount `T` of tokens as reward for providing security, then
+  - `T` (equivalent) tokens are eventually minted on the provider chain and then distributed among the validators that are part of the validator set;
+  - the total supply of tokens is preserved, i.e., the `T` (original) tokens are escrowed on the consumer chain.
```
josef-widder, line 139: I think this mixes a liveness property into the invariant:
- liveness: If a consumer chain sends to the provider chain an amount `T` of tokens as reward for providing security, then`T` (equivalent) tokens are eventually minted on the provider chain and then distributed among the validators that are part of the validator set;
- invariant: It is always the case that the total sum of the minted fee tokens on the provider chain is less than or equal to the total sum of the escrowed tokens on the consumer chain.

spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
```diff
@@ -363,4 +384,7 @@ i.e., we informally prove the properties described in the [previous section](#de
...
   The tokens bonded by `val` at height `hp` that were not in the process of unbonding (i.e., `Token(Power(cc,hi,val))`) could not have completely unbonded by block `B` (cf. `ts(he) < ts(hi) + UnbondingPeriod`, *VSC Maturity and Slashing Order*, *Register Maturity Timeliness*, *Unbonding Safety*).
-  This means that *exactly* the amount of tokens `Token(Power(cc,hi,val))` is slashed on the provider chain. 
\ No newline at end of file
+  This means that *exactly* the amount of tokens `Token(Power(cc,hi,val))` is slashed on the provider chain. 
+
+- ***Distribution Invariant***: The first part of the *Distribution Invariant* (i.e., the tokens are eventually minted on the provider chain and then distributed among the validators) follows directly from *Distribution Liveness* and *Distribution Warranty*. 
+  The second part of the *Distribution Invariant* (i.e., the total supply of tokens is preserved) follows directly from the *Supply* property of the Fungible Token Transfer protocol (see [ICS 20](../ics-020-fungible-token-transfer/README.md)).
```
josef-widder, line 390: If the "invariant" is split into liveness and invariant, then this write-up also needs to be adapted.

josef-widder (commented):
Thanks for the effort to formalize the properties. I suggested splitting the properties a bit.

spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
mpoke: Indeed. Nice catch. I replaced "system invariants" with "system properties", since I believe it's more accurate. Many of the existing "invariants" had hidden temporal properties in them.

spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
mpoke: See above comment.

AdityaSripal (approved):
Nice and succinct!

PR #717, opened 2022-04-21
Clarify property about consumer validator set changes

danwt:
Hey, I was confused about this line so I changed it to the way I understand. Is my understanding wrong?

Thanks

### DIFF
--- a/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
+++ b/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
@@ -169,7 +169,7 @@ As a result, it is possible for multiple VSCs to be received by a consumer chain
 i.e., the validator updates within the VSCs are being *aggregated* by keeping only the latest update per validator. 
 As a consequence, some validator sets on the provider chain are not existing on all consumer chains. 
 In other words, the validator sets on each consumer chain form a *subsequence* of the validator sets on the provider chain. 
-Nonetheless, as a **requirement of CCV**, *all the validator updates on the provider chain MUST be included in the sequence of validator sets on all consumer chains*.
+Nonetheless, as a **requirement of CCV**, *all the validator updates on all the consumer chains MUST be included in the sequence of validator sets on the provider chain*.
 
 This is possible since every validator update contains *the absolute voting power* of that validator. 
 Given a validator `val`, the sequence of validator updates targeting `val` (i.e., updates of the voting power of `val`) is the prefix sum of the sequence of relative changes of the voting power of `val`. 
--- a/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
+++ b/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
@@ -169,7 +169,7 @@ As a result, it is possible for multiple VSCs to be received by a consumer chain
 i.e., the validator updates within the VSCs are being *aggregated* by keeping only the latest update per validator. 
 As a consequence, some validator sets on the provider chain are not existing on all consumer chains. 
 In other words, the validator sets on each consumer chain form a *subsequence* of the validator sets on the provider chain. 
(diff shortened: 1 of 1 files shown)


spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
```diff
@@ -169,7 +169,7 @@ As a result, it is possible for multiple VSCs to be received by a consumer chain
 i.e., the validator updates within the VSCs are being *aggregated* by keeping only the latest update per validator. 
 As a consequence, some validator sets on the provider chain are not existing on all consumer chains. 
 In other words, the validator sets on each consumer chain form a *subsequence* of the validator sets on the provider chain. 
-Nonetheless, as a **requirement of CCV**, *all the validator updates on the provider chain MUST be included in the sequence of validator sets on all consumer chains*.
+Nonetheless, as a **requirement of CCV**, *all the validator sets on all the consumer chains MUST be included in the sequence of validator sets on the provider chain*.
```
mpoke, line 172: The idea here is to make sure that all validator updates that occur on the **_provider chain_** are provided to the consumer chain. Let two updates `U1` and `U2` to the same validator `V` occur in two subsequent blocks on the provider, i.e., 
- `V` has 100 ATOMs bonded
- `U1`: V's stake changes to 80 (a delegator undelegates 20 atoms)
- `U2`: V's stake changes to 50 (a delegator undelegates 30 atoms)

The provider chain sends `U1` and `U2` to the consumer chain, however, due to delays, it may be that both are received by a consumer chain in the same block. Thus, `V`'s power on that consumer chain will go from 100 directly to 50 (I assume here that's a 1 to 1 conversion between stake and power). Basically, `U1` is included in `U1` (see the Validator Update Inclusion property below). 

The point of this sentence is to express that the effect of `U1` is not lost from the consumer chain's perspective.

danwt: I see, I was just confused by the wording. I will suggest another wording.

spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
```diff
@@ -169,7 +169,7 @@ As a result, it is possible for multiple VSCs to be received by a consumer chain
 i.e., the validator updates within the VSCs are being *aggregated* by keeping only the latest update per validator. 
 As a consequence, some validator sets on the provider chain are not existing on all consumer chains. 
 In other words, the validator sets on each consumer chain form a *subsequence* of the validator sets on the provider chain. 
-Nonetheless, as a **requirement of CCV**, *all the validator updates on the provider chain MUST be included in the sequence of validator sets on all consumer chains*.
+Nonetheless, as a **requirement of CCV**, *all the updates to validator power on the provider chain MUST be included in the sequence of updates to validator power on all consumer chains*.
```
mpoke, line 172: I would keep it as it is. The reason is twofold. 

First, "validator updates" is described above, i.e., 
> validator updates (i.e., changes in the validators voting power)

Thus, I find that the phrase  "all the validator updates on the provider chain" does not need further clarification. 

Second, the sequence of validator sets on a consumer chain is the result of applying a sequence of VSCs, which contains a sequence of validator updates. This is described in the following lines
> the sequence of blocks produces a sequence of validator updates and a sequence of validator sets. Furthermore, the sequence of validator updates on the provider chain results in a sequence of VSCs to all consumer chains. Ideally, this sequence of VSCs is applied by every consumer chain, resulting in a sequence of validator sets identical to the one on the provider chain

By saying that a validator update U1 must be "included" in a validator set, I mean that the results of either applying U1 or a subsequent update U2 that includes U1 must be visible in the validator set. I agree that "included" may not be the best word here, but I couldn't come with a better one.

danwt: Hmm, ok, fair enough. It was good to discuss anyway. Thanks.

danwt:
Resolved

PR #722, opened 2022-04-25
Add ChanUpgradeCancelConfirm datagram and handler to reset upgrade state

### LINKED ISSUE #698: Reset UpgradeError Path
Reset UpgradeError path so a stale upgrade error cannot be used to cancel a new upgrade handshake

AdityaSripal:
Closes: #698 for channels

We clear the error receipt by sending a message back to the original aborting chain that proves that the counterparty has restored their channel

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -505,10 +505,14 @@ function chanUpgradeConfirm(
 
 **Protocol**:
 
-During the upgrade handshake an aborting chain may cancel the upgrade by writing an error receipt into the error path and restoring the original channel to `OPEN`. The counterparty must then restore its channel to `OPEN` as well. A relayer can facilitate this by sending `ChanUpgradeCancelInitMsg` to the counterparty chain's handler:
+During the upgrade handshake an aborting chain may cancel the upgrade by writing an error receipt into the error path and restoring the original channel to `OPEN`. The counterparty must then restore its original channel to `OPEN` as well.
+
+A channelEnd may only cancel the upgrade during the upgrade negotiation process (`TRY`, `ACK`). Thus, the counterparty channel state must be either `UPGRADE_INIT` or `UPGRADE_TRY`. An upgrade cannot be cancelled on one end once the other chain has already completed its upgrade and moved to `OPEN` since that will lead the channel to being in an invalid state.
+
+Once the aborting chain has written to the error receipt, a relayer must submit `ChanUpgradeCancelInit` to the counterparty chain:
 
 ```typescript
-function cancelChannelUpgradeInit(
+function chanUpgradeCancelInit(
     portIdentifier: Identifier,
     channelIdentifier: Identifier,
     errorReceipt: []byte,
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -75,6 +75,16 @@ interface UpgradeTimeout {
 
 At least one of the timeoutHeight or timeoutTimestamp MUST be non-zero.
 
+```typescript
+interface ErrorReceipt {
+    counterpartyHeight: Height
+    errorMsg: string
+}
+```
+
+- `counterpartyHeight`: Counterparty height is the proof height of the counterparty when the executing chain aborts the upgrade process. This height will get used to clear the error receipt before future upgrades.
+- `errorMsg`: Error Msg is an arbitrary string that chains may use to provide more information as to why the upgrade failed.
+
 ### Store Paths
 
 #### Restore Channel Path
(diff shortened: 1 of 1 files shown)


spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -499,10 +499,20 @@ function chanUpgradeConfirm(
 
 ### Cancel Upgrade Process
 
-During the upgrade handshake a chain may cancel the upgrade by writing an error receipt into the error path and restoring the original channel to `OPEN`. The counterparty must then restore its channel to `OPEN` as well. A relayer can facilitate this by sending `ChanUpgradeCancelMsg` to the handler:
+**Sub-Definitions**:
```
mpoke, line 502: I don't think it's necessary to have a subsection here for defining an aborting chain. You could just say directly something like this:
> During the upgrade handshake a chain may encounters an error and cancel the upgrade. The first channel to do so is referred to as the *aborting chain*. The aborting chain  may cancel the upgrade by writing an error receipt into the error path and restoring the original channel to `OPEN`. The counterparty must then restore its original channel to `OPEN` as well.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -499,10 +499,20 @@ function chanUpgradeConfirm(
...
+
+`Aborting Chain`: This is the chain that first encounters an error, writes data to the error receipt and restores the channel.
+
+**Protocol**:
+
+During the upgrade handshake an aborting chain may cancel the upgrade by writing an error receipt into the error path and restoring the original channel to `OPEN`. The counterparty must then restore its original channel to `OPEN` as well.
+
+A channelEnd may only cancel the upgrade during the upgrade negotiation process (`TRY`, `ACK`). Thus, the counterparty channel state must be either `UPGRADE_INIT` or `UPGRADE_TRY`. An upgrade cannot be cancelled on one end once the other chain has already completed its upgrade and moved to `OPEN` since that will lead the channel to being in an invalid state.
```
mpoke, line 510:
```suggestion
The aborting chain may cancel the upgrade only during the upgrade negotiation process (`TRY`, `ACK`). Thus, the channel state on the counterparty chain must be either `UPGRADE_INIT` or `UPGRADE_TRY`. The aborting chain cannot cancel the channel upgrade once the counterparty chain has already completed its upgrade and moved to `OPEN`, since that will lead to the channel being in an invalid state.
```

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -499,10 +499,20 @@ function chanUpgradeConfirm(
...
+
+Once the aborting chain has written to the error receipt, a relayer must submit `ChanUpgradeCancelInit` to the counterparty chain:
```
mpoke, line 512:
```suggestion
Once the aborting chain has written to the error receipt, a relayer must submit `ChanUpgradeCancelInitMsg` to the counterparty chain:
```

mpoke (approved):
Looks good. Some nits below. 

The only concern that I have is that if the message `ChanUpgradeCancelConfirmMsg` is never relayed, the aborting chain never clears the error receipt. Thus, both `INIT` and `TRY` should abort unless the error receipt is empty. Which makes me wonder why do we need the `ChanUpgradeCancelConfirmMsg` in the first place, i.e., we could just clear error receipt in `INIT` and `TRY`. However, this would be problematic since both the error receipt and a new upgrade will happen at the same height. Thus, a relayer could relay first `ChanUpgradeTry` and then `ChanUpgradeCancelInit`, which would cancel the upgrade and leave the two chain in an inconsistent state. The issue is that this scenario may happen also with `ChanUpgradeCancelConfirmMsg`, i.e., if the aborting chain clears the error receipt at height h (by executing `chanUpgradeCancelConfirm`) and at the same height it executes `chanUpgradeInit`.

mpoke:
> Looks good. Some nits below.
> 
> The only concern that I have is that if the message `ChanUpgradeCancelConfirmMsg` is never relayed, the aborting chain never clears the error receipt. Thus, both `INIT` and `TRY` should abort unless the error receipt is empty. Which makes me wonder why do we need the `ChanUpgradeCancelConfirmMsg` in the first place, i.e., we could just clear error receipt in `INIT` and `TRY`. However, this would be problematic since both the error receipt and a new upgrade will happen at the same height. Thus, a relayer could relay first `ChanUpgradeTry` and then `ChanUpgradeCancelInit`, which would cancel the upgrade and leave the two chain in an inconsistent state. The issue is that this scenario may happen also with `ChanUpgradeCancelConfirmMsg`, i.e., if the aborting chain clears the error receipt at height h (by executing `chanUpgradeCancelConfirm`) and at the same height it executes `chanUpgradeInit`.

If this is actually a problem, a solution would be to write the height in the error receipt and make both `INIT` and `TRY` conditional on the current height to be higher than the error height (if set), i.e., an upgrade cannot be canceled and restarted in the same height. This would also remove the need of the `ChanUpgradeCancelConfirmMsg`.

AdityaSripal:
> we could just clear error receipt in INIT and TRY

I thought about this. The problem was that if the error receipt is on the TRY chain, then it could be submitted to the initializing chain to cancel upgrade before the TRY message comes in to clear it.

But you are right, that if we include the height in the error receipt maybe we could avoid this issue. But possibly we will need the counterparty height. Will think about this...

PR #724, opened 2022-04-28
Add upgrade sequences to avoid proof collisions

AdityaSripal:
Much cleaner fix to #698 

Supersedes: #722 

Each upgrade attempt is given a unique sequence, the error path is keyed on this sequence to avoid collisions

It also fixes another previously uncaught issue where relayers could submit proofs of channel state from previous upgrade attempts for the current upgrade handshake

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -358,6 +358,8 @@ function chanUpgradeTry(
     // verify proofs of counterparty state
     abortTransactionUnless(verifyChannelState(connection, proofHeight, proofChannel, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, counterpartyChannel))
     abortTransactionUnless(verifyChannelUpgradeTimeout(connection, proofHeight, proofUpgradeTimeout, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, upgradeTimeout))
+    abortTransactionUnless(verifyUpgradeSequence(connection, proofHeight, proofUpgradeSequence, currentChannel.counterpartyPortIdentifier,
+    currentChannel.counterpartyChannelIdentifier, counterpartySequence))
 
     if currentChannel.state == UPGRADE_INIT {
         // if there is a crossing hello, ie an UpgradeInit has been called on both channelEnds,
@@ -437,6 +439,7 @@ function chanUpgradeAck(
     channelIdentifier: Identifier,
     counterpartyChannel: ChannelEnd,
     proofChannel: CommitmentProof,
+    proofSequence: CommitmentProof,
     proofHeight: Height
 ) {
     // current channel is in UPGRADE_INIT or UPGRADE_TRY (crossing hellos)
@@ -448,6 +451,14 @@ function chanUpgradeAck(
 
     // verify proofs of counterparty state
     abortTransactionUnless(verifyChannelState(connection, proofHeight, proofChannel, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, counterpartyChannel))
+
+    // verify that the counterparty sequence is the same as the current sequence to ensure that the proofs were
+    // retrieved from the current upgrade attempt
+    // since all proofs are retrieved from same proof height, and there can not be multiple upgrade states in the store for a given
+    // channel at the same time
+    sequence = provableStore.get(currentSequence(portIdentifier, channelIdentifier))
+    abortTransactionUnless(verifyUpgradeSequence(connection, proofHeight, proofUpgradeSequence, currentChannel.counterpartyPortIdentifier,
+    currentChannel.counterpartyChannelIdentifier, sequence))
 
     // counterparty must be in TRY state
     if counterpartyChannel.State != UPGRADE_TRY {
@@ -495,6 +506,7 @@ function chanUpgradeConfirm(
     counterpartyChannel: ChannelEnd,
     proofChannel: CommitmentProof,
     proofUpgradeError: CommitmentProof,
+    proofSequence: CommitmentProof,
     proofHeight: Height,
 ) {
     // current channel is in UPGRADE_TRY
@@ -515,6 +527,14 @@ function chanUpgradeConfirm(
     // verify counterparty did not abort upgrade handshake by writing upgrade error
     // must have absent value at upgradeError path at the current sequence
     abortTransactionUnless(verifyUpgradeChannelErrorAbsence(connection, proofHeight, proofUpgradeError, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, sequence))
+
+    // verify that the counterparty sequence is the same as the current sequence to ensure that the proofs were
+    // retrieved from the current upgrade attempt
+    // since all proofs are retrieved from same proof height, and there can not be multiple upgrade states in the store for a given
+    // channel at the same time
+    sequence = provableStore.get(currentSequence(portIdentifier, channelIdentifier))
+    abortTransactionUnless(verifyUpgradeSequence(connection, proofHeight, proofUpgradeSequence, currentChannel.counterpartyPortIdentifier,
+    currentChannel.counterpartyChannelIdentifier, sequence))
 
     // call modules onChanUpgradeConfirm callback
     module = lookupModule(portIdentifier)
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -75,6 +75,16 @@ interface UpgradeTimeout {
 
 At least one of the timeoutHeight or timeoutTimestamp MUST be non-zero.
 
+```typescript
+interface ErrorReceipt {
+    sequence: uint64
+    errorMsg: string
+}
+```
+
+- `sequence`: Sequence contains the sequence at which the error occurred. Both chains are expected to increment to the next sequence after the upgrade is aborted.
+- `errorMsg`: ErrorMsg contains an arbitrary string which chains may use to provide additional information as to why the upgrade was aborted.
+
 ### Store Paths
 
 #### UpgradeSequencePath
@@ -117,11 +127,11 @@ function restorePath(portIdentifier: Identifier, channelIdentifier: Identifier):
 
 #### UpgradeError Path
 
-The upgrade error path is a public path that can signal an error of the upgrade to the counterparty for the given upgrade attempt. It does not store anything in the successful case, but it will store a sentinel abort value in the case that a chain does not accept the proposed upgrade. The error path will contain the portIdentifier, channelIdentifier, **and** the upgrade sequence so that the error receipts for different upgrade attempts do not interfere with each other.
+The upgrade error path is a public path that can signal an error of the upgrade to the counterparty for the given upgrade attempt. It does not store anything in the successful case, but it will store the `ErrorReceipt` in the case that a chain does not accept the proposed upgrade.
 
 ```typescript
-function errorPath(portIdentifier: Identifier, channelIdentifier: Identifier, sequence: uint64): Path {
-    return "channelUpgrade/ports/{portIdentifier}/channels/{channelIdentifier}/sequence/{sequence}/upgradeError"
+function errorPath(portIdentifier: Identifier, channelIdentifier: Identifier): Path {
+    return "channelUpgrade/ports/{portIdentifier}/channels/{channelIdentifier}/upgradeError"
 }
 ```
 
@@ -135,11 +145,10 @@ function verifyChannelUpgradeError(
   proof: CommitmentProof,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  sequence: uint64,
-  upgradeErrorReceipt: []byte
+  upgradeErrorReceipt: ErrorReceipt
 ) {
     client = queryClient(connection.clientIdentifier)
-    path = applyPrefix(connection.counterpartyPrefix, channelErrorPath(counterpartyPortIdentifier, counterpartyChannelIdentifier, sequence))
+    path = applyPrefix(connection.counterpartyPrefix, channelErrorPath(counterpartyPortIdentifier, counterpartyChannelIdentifier))
     client.verifyMembership(height, 0, 0, proof, path, upgradeErrorReceipt)
 }
 ```
@@ -152,10 +161,9 @@ function verifyChannelUpgradeErrorAbsence(
   proof: CommitmentProof,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  sequence: uint64
 ) {
     client = queryClient(connection.clientIdentifier)
-    path = applyPrefix(connection.counterpartyPrefix, channelErrorPath(counterpartyPortIdentifier, counterpartyChannelIdentifier, sequence))
+    path = applyPrefix(connection.counterpartyPrefix, channelErrorPath(counterpartyPortIdentifier, counterpartyChannelIdentifier))
     client.verifyNonMembership(height, 0, 0, proof, path)
 }
 ```
@@ -199,11 +207,14 @@ The Channel Upgrade process consists of three sub-protocols: `UpgradeChannelHand
 ```typescript
 function restoreChannel() {
     // cancel upgrade
-    // write an error receipt into the error path at the current sequence
+    // write an error receipt with the current sequence into the error path
     // and restore original channel
     sequence = provableStore.get(upgradeSequencePath(portIdentifier, channelIdentifier))
-    errorReceipt = []byte{1}
-    provableStore.set(errorPath(portIdentifier, channelIdentifier, sequence), errorReceipt)
+    errorReceipt = ErrorReceipt{
+        Sequence: sequence,
+        ErrorMsg: ""
+    }
+    provableStore.set(errorPath(portIdentifier, channelIdentifier), errorReceipt)
     originalChannel = privateStore.get(restorePath(portIdentifier, channelIdentifier))
     provableStore.set(channelPath(portIdentifier, channelIdentifier), originalChannel)
     provableStore.delete(timeoutPath(portIdentifier, channelIdentifier))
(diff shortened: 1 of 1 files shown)


spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -87,18 +87,27 @@ function restorePath(portIdentifier: Identifier, channelIdentifier: Identifier):
 }
 ```
 
-#### UpgradeError Path
+#### CurrentUpgradeSequencePath
```
mpoke, line 90: I'd remove "current", i.e., `UpgradeSequencePath`. Also make it consistent throughout the file.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -87,18 +87,27 @@ function restorePath(portIdentifier: Identifier, channelIdentifier: Identifier):
...
-function errorPath(portIdentifier: Identifier, channelIdentifier: Identifier): Path {
-    return "channelUpgrade/ports/{portIdentifier}/channels/{channelIdentifier}/upgradeError"
+function currentSequencePath(portIdentifier: Identifier, channelIdentifier: Identifier) Path {
+    return "channelUpgrade/ports/{portIdentifier}/channelIdentifier/{channelIdentifier}/currentUpgradeSequence"
+}
+```
+
+#### UpgradeError Path
+
+The upgrade error path is a public path that can signal an error of the upgrade to the counterparty for the given upgrade attempt. It does not store anything in the successful case, but it will store a sentinel abort value in the case that a chain does not accept the proposed upgrade. The error path will contain the portIdentifier, channelIdentifier, **and** the upgrade sequence so that the error receipts for different upgrade attempts do not interfere with each other.
```
mpoke, line 102: Wouldn't it be sufficient to write the upgrade sequence to the same path that contains only the portIdentifier and channelIdentifier, i.e., `channelUpgrade/ports/{portIdentifier}/channels/{channelIdentifier}/upgradeError`?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -285,17 +303,33 @@ function chanUpgradeTry(
...
     currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
     abortTransactionUnless(currentChannel.state == OPEN || currentChannel.state == UPGRADE_INIT)
 
+    // get current sequence on this channel
+    // if the counterparty sequence is greater than the current sequence, we fast forward to the counterparty sequence
+    // so that both channel ends are using the same sequence for the current upgrade
+    // if the counterparty sequence is less than the current sequence, then either the counterparty chain is out-of-sync or
+    // the message is out-of-sync and we write an error receipt for that sequence to avoid a possible collision with previous upgrade attempt
+    currentSequence = provableStore.get(currentSequencePath(portIdentifier, channelIdentifier))
+    abortTransactionUnless()
```
mpoke, line 325: This line probably needs to be removed.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -324,6 +358,8 @@ function chanUpgradeTry(
     // verify proofs of counterparty state
     abortTransactionUnless(verifyChannelState(connection, proofHeight, proofChannel, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, counterpartyChannel))
     abortTransactionUnless(verifyChannelUpgradeTimeout(connection, proofHeight, proofUpgradeTimeout, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, upgradeTimeout))
+    abortTransactionUnless(verifyUpgradeSequence(connection, proofHeight, proofUpgradeSequence, currentChannel.counterpartyPortIdentifier,
```
mpoke, line 361: `verifyUpgradeSequence` is not defined.

mpoke (commented):
@AdityaSripal Looks good at a first glance. Still need to think a bit about your two comments.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -285,17 +303,33 @@ function chanUpgradeTry(
...
     abortTransactionUnless(currentChannel.state == OPEN || currentChannel.state == UPGRADE_INIT)
 
+    // get current sequence on this channel
+    // if the counterparty sequence is greater than the current sequence, we fast forward to the counterparty sequence
+    // so that both channel ends are using the same sequence for the current upgrade
+    // if the counterparty sequence is less than the current sequence, then either the counterparty chain is out-of-sync or
+    // the message is out-of-sync and we write an error receipt for that sequence to avoid a possible collision with previous upgrade attempt
+    currentSequence = provableStore.get(currentSequencePath(portIdentifier, channelIdentifier))
+    abortTransactionUnless()
+    if counterpartySequence > currentSequence {
```
mpoke, line 326: I think there is an issue here. If we keep it `>`, then the following scenario is possible. Chain `A` starts an upgrade with `seq=10` and moves to `UPGRADE_INIT`. A malicious relayer sends to `A` a `ChanUpgradeTry` with the `counterpartySequence` set to `10`, which will cancel the upgrade. In general, the `counterpartySequence` should be verified before updating the state (i.e., `errorPath`). However, this scenario could happen even with correct relayers, if both chain `A` and the counterparty (chain `B`) start an upgrade with the same sequence `10`. 

Alternatively, if we change `>` to `>=`, then the following scenario is possible. Chain `A` just finished executing `chanUpgradeAck` for sequence `10`, i.e., the upgrade is successful and `A` moves to `OPEN`. A relayer sends to `A`  a `ChanUpgradeTry` with the `counterpartySequence` set to `10`, which will trigger a new update for the same sequence.

mpoke (changes requested):
I think there is an issue with the sequence numbers. See comments bellow. 

In general, due to concurrency (multiple chains, multiple relayers, multiple upgrades), this protocol is quite subtle. I think it may be worthwhile to write a TLA+ model for it.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
AdityaSripal: Would it be safer to bump the sequence on ACK and CONFIRM and then switch to `>=`?

I think it would also be nice to track the sequences in state so we don't allow relayers to pass in a sequence at all. But that may end up being too complicated if the individual sequences on each chain go out of sync. I tried to think a few times of how to do it, but there were scenarios in which one chain gets its sequence incremented a few times then initiating from the other chain would fail until the sequences were back in sync

mpoke: > Would it be safer to bump the sequence on ACK and CONFIRM and then switch to `>=`?

It will solve the scenario described above. I'm not sure though if it will not introduce other issues. For example, in CONFIRM the sequence needs to be bumped  before the sequenceProof is verified.

mpoke: What about the following approach that uses heights as sequence?

- Initially, each chain has its own `errorPath` set to `""`. 
- Chain `A` recv INIT -> `seq = currentHeightA`.
- Chain `B` recv TRY(sequenceProof) -> `seq = currentHeightB`, `counterpartySeq = currentHeightA`
- Chain `A` recv ACK(sequenceProof) -> `counterpartySeq = currentHeightB`

Note that 
- INIT does not require a `sequence`.
- CONFIRM does not require `sequenceProof` since the `proofHeight` can be compared to the `counterpartySeq` (i.e., must be `>=`).

`restoreChannel()` is called either on TRY or on ACK. When called, the  `errorPath` is set to the `counterpartySeq`.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -221,6 +233,7 @@ If an upgrade message arrives after the specified timeout, then the message MUST
 function chanUpgradeInit(
     portIdentifier: Identifier,
     channelIdentifier: Identifier,
+    sequence: uint64,
```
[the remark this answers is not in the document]
AdityaSripal: Solved

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -285,17 +303,33 @@ function chanUpgradeTry(
     portIdentifier: Identifier,
     channelIdentifier: Identifier,
     counterpartyChannel: ChannelEnd,
+    counterpartySequence: uint64,
     proposedUpgradeChannel: ChannelEnd,
     timeoutHeight: Height,
     timeoutTimestamp: uint64,
     proofChannel: CommitmentProof,
     proofUpgradeTimeout: CommitmentProof,
+    proofSequence: CommitmentProof,
```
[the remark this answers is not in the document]
AdityaSripal: Used proofUpgradeSequence in method signatures

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
AdityaSripal: In order to do that, we would need to create a new data structure for error receipt so that we can parse out the sequence from the error receipt. With this solution we can keep the error receipt to be arbitrary bytes

The other downside is that error receipts will get overwritten on subsequent upgrades, rather than having some store of past attempts that can then be pruned if desired.

For these reasons, since there's only one piece of data we want to communicate in protocol, it made more sense to encode it in the key

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
AdityaSripal: I believe I've solved this issue in the following changes:

1. Allow upgrade on TRY if counterparty sequence is `>=` to our sequence. We then will set the counterparty sequence as our sequence on TRY
2. If counterparty sequence is `<` our sequence, write our sequence into the error receipt so that counterparty can be "caught up" efficiently and without relayer input
3. Increment sequences on ACK, CONFIRM, and error cases
4. Do not increment sequence on timeout

With these changes, I don't even think it is possible for sequences to go out of sync. But i have the logic to handle it in case it is.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -324,6 +363,22 @@ function chanUpgradeTry(
...
+    // if the counterparty sequence is greater than the current sequence, we fast forward to the counterparty sequence
+    // so that both channel ends are using the same sequence for the current upgrade
+    // if the counterparty sequence is less than the current sequence, then either the counterparty chain is out-of-sync or
+    // the message is out-of-sync and we write an error receipt with our own sequence so that the counterparty can update
+    // their sequence as well.
+    currentSequence = provableStore.get(upgradeSequencePath(portIdentifier, channelIdentifier))
+    if counterpartySequence >= currentSequence {
+        provableStore.set(upgradeSequencePath(portIdentifier, channelIdentifier), counterpartySequence)
+    } else {
+        provableStore.set(errorPath(portIdentifier, channelIdentifier, currentSequence), []byte{1})
```
AdityaSripal, line 379: I don't believe this error condition will be hit, but please review that this is a correct recovery in case the sequences do go out of sync

mpoke: I believe that the error condition is hit in the following scenario involving two chains A and B. Both have sequence `10.` 
- A executes INIT and sends `ChanUpgradeTry(seq=10)`
- B executes TRY, however it results in restore, which entails `errorB = 10` and `seqB=11`
- A does not receive `ChanUpgradeCancelMsg` (`seqA` remains set to `10`)
- A eventually receives `ChanUpgradeTimeoutMsg`, but `seqA` still remains set to `10`
- A executes INIT and sends `ChanUpgradeTry(seq=10)`
- B executes TRY and writes `errorB = 11`

Moreover, 
- A receives `ChanUpgradeCancelMsg(seq=11)`, sets `seqA = 11` and aborts the transaction
- A executes INIT and sends `ChanUpgradeTry(seq=11)`
- B executes TRY and sends `ChanUpgradeAck()`
- a relayer relays `ChanUpgradeCancelMsg(seq=11)` to A
  - A aborts the transaction and sets `seqA = 12`
  - B gets stuck in `UPGRADE_TRY`

mpoke: A solution would be to have the following changes:
```typescript
function chanUpgradeTry {
  // .... 
  currentSequence = provableStore.get(upgradeSequencePath(portIdentifier, channelIdentifier))
  if counterpartySequence >= currentSequence {
    provableStore.set(upgradeSequencePath(portIdentifier, channelIdentifier), counterpartySequence)
  } else {
    provableStore.set(errorPath(portIdentifier, channelIdentifier, currentSequence), []byte{1})
    // CHANGE: increment sequence similarly to restore()
    return
  }
  // .... 
}
```

```typescript
function cancelChannelUpgrade {
  // .... 
  currentSequence = provableStore.get(currentSequencePath(portIdentifier, channelIdentifier))
  abortTransactionUnless(counterpartySequence >= currentSequence)
  if counterpartySequence > currentSequence {
    // the counterparty sequence is higher and thus the channel ends are out of sync
    // setting our sequence to the counterparty sequence will bring the channel ends back in sync
    // CHANGE: additionally increment the sequence, i.e., counterpartySequence+1
    provableStore.set(upgradeSequencePath(portIdentifier, channelIdentifier), counterpartySequence+1)
  } else {
    // current sequence and counterparty sequence are equal so both chains are in sync.
    // We must increment the sequence to prepare for the next upgrade
    provableStore.set(upgradeSequencePath(portIdentifier, channelIdentifier), currentSequence+1)
  }
  // .... 
}
```
Basically, if chain A attempts an upgrade for a sequence lower than the sequence on chan B, then the sequences on both A and B should not be used again. Not sure though if this has further implications on other parts of the protocol.

AdityaSripal: > Moreover,
> A receives ChanUpgradeCancelMsg(seq=11), sets seqA = 11 and aborts the transaction
> A executes INIT and sends ChanUpgradeTry(seq=11)
> B executes TRY and sends ChanUpgradeAck()
> a relayer relays ChanUpgradeCancelMsg(seq=11) to A
> A aborts the transaction and sets seqA = 12
> B gets stuck in UPGRADE_TRY

This is absolutely correct. We have to bump both sequences on an error. This is simple to do and I'll make the requisite changes.

> A executes INIT and sends ChanUpgradeTry(seq=10)
> B executes TRY, however it results in restore, which entails errorB = 10 and seqB=11
> A does not receive ChanUpgradeCancelMsg (seqA remains set to 10)
> A eventually receives ChanUpgradeTimeoutMsg, but seqA still remains set to 10
> A executes INIT and sends ChanUpgradeTry(seq=10)
> B executes TRY and writes errorB = 11

This can never happen precisely because Timeouts can only be processed by proving that the counterparty had not seen the TRY message (same as packet timeouts, also see timeout pseudocode). So regardless of how much time passes after B restores connection, the next step in the handshake must be a `ChanUpgradeCancelMsg` not a `ChanUpgradeTimeout` which would fail. The channel will simply be unlive during that time.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
AdityaSripal: I ended up having to do as you suggested, since in timeout and confirm we want to prove there is no error receipt stored on the counterparty (or it is only at lower sequence) and this would be difficult to do if sequence is in the key

angbrav:
I wonder if there is a way to simplify the specification by restricting concurrency. Crossing hellos cannot be avoided, but I believe once a crossing hello is detected, one could break it by using some logic based on sequence numbers and tiebreakers. 

Assume that a chain A detects a crossing hello between itself and a second chain B, i.e., when a chain A is on UPGRADE_INIT while executing ChanUpgradeInit. Let seqA and seqB be the sequences numbers used by each chain respectively for their channel upgrade.
- If seqA > seqB, then chain A will ignore the ChanUpgradeInit.
- If seqA < seqB, then chain A will give up its channel upgrade and go with channel B upgrade: restore channel and execute ChanUpgradeInit.
- If seqA == seqB, then we need a tiebreaker to decide who wins. The question is what to use as a tiebreaker. One option would be to use a function isGreater that deterministically orders channel identifiers. Given two channel identifiers id1 and id2, isGreat returns true if id1>id2. Thus,
   - if (seqA == seqB)&& isGreater(channelIdentifier, counterpartyChannelIdentifier), then chain A ignores the ChanUpgradeInit;
   - if (seqA == seqB)&& !isGreater(channelIdentifier, counterpartyChannelIdentifier), then chain A give up its channel upgrade
- This of course only works if channelIdentifier and counterpartyChannelIdentifier are guaranteed to be different. My guess is that if this is not the case, we could find another tiebreaker, e.g., the CommitmentRoot of each chain at the height at which they started the upgrade (very unlikely that two chains ever have the same).

Let me know if this makes sense, I could help develop the idea.

mpoke (approved):
LGTM. Nice work @AdityaSripal.

mpoke:
@angbrav I don't think the complexity of the protocol comes from the crossing hellos, but rather from the fact that we have two different chains that can be in various states and based on these states, various events can occur (i.e., relayers sending different messages). The crossing hello logic is quite straightforward, i.e.,  
```typescript
function chanUpgradeTry() {
  // ... 
  if currentChannel.state == UPGRADE_INIT {
      // if there is a crossing hello, ie an UpgradeInit has been called on both channelEnds,
      // then we must ensure that the proposedUpgrade by the counterparty is the same as the currentChannel
      // except for the channel state (upgrade channel will be in UPGRADE_TRY and current channel will be in UPGRADE_INIT)
      // if the proposed upgrades on either side are incompatible, then we will restore the channel and cancel the upgrade.
      currentChannel.state = UPGRADE_TRY
      if !currentChannel.IsEqual(proposedUpgradeChannel) {
          restoreChannel()
          return
      }
  }
  // ... 
}
```
In other words, we allow for crossing hellos only if there are for exactly the same upgrade. Having one of the chains backing out would only enable crossing hellos for different upgrades, which I don't believe we need (since it's very unlikely).

mpoke:
> Unless I misunderstand something, the crossing hellos is the reason to have chains in different states creating the variety of interleavings. 

It's not only that. The other reasons are
- chains in `UPGRADE_INIT` can timeout
- chains can restore the previous channel state in case of an error (while executing TRY or ACK)
- chains in `UPGRADE_INIT` or `UPGRADE_TRY` can receive a cancel message and as a result restore the previous channel state
- relayers can submit any type of message, including outdated ones (i.e., there can be multiple upgrade attempts)

Issue #720, opened 2022-04-25
ICS28: Remove dependency on Tendermint and ABCI


mpoke:
Interchain Security needs neither Tendermint nor ABCI. The necessary assumption is that each of the two chains (i.e., consumer and provider) consist of a state machine (SM) and an underlying consensus engine. Also, each chain is uniquely defined by a chain ID and a validator set. The SM must have a mechanism to decide on the next validator set (e.g., the staking module in Cosmos SDK) and provide this information to the consensus layer (e.g., via ABCI). In case the validator set is selected based on PoS, the provider chain's SM must provide a mechanism for slashing and must not allow unbonding operations to complete before a certain unbonding period elapses.

PR #730, opened 2022-05-02
ICS28: Remove dependency on Tendermint and ABCI

mpoke:
Closes https://github.com/cosmos/ibc/issues/720


AdityaSripal (approved):
Nice work!

PR #735, opened 2022-05-11
Cross-chain query spec draft

angbrav:
Draft of cross-chain query spec. The goal of the PR is to get some feedback to improve the spec.

This PR supersedes #647.

Taken decisions after discussion:

- The bounty system is vulnerable to front-running - any relayer who actually bothers to query the other chain can be front-run by the block proposer. This is true of any permissionless relayer incentivization, e.g., ICS29 will ship v1 with the same problem. However, the APIs are constructed such that we could easily specify a relayer in the future. We have decided to do something similar here. (https://github.com/cosmos/ibc/pull/735#issuecomment-1134674583)

- Add features to enable https://github.com/cosmos/ibc/discussions/759. Future work. (https://github.com/cosmos/ibc/pull/735#issuecomment-1140161175)

- Shall we prevent queryHeight from being in the future to avoid cross-chain MEV? We have decided to add a note to acknowledge the problem. Solving MEV is out of the scope of this spec. (https://github.com/cosmos/ibc/pull/735#discussion_r913638666)


schnetzlerjoe:
> Draft of cross-chain query spec. The goal of the PR is to get some feedback to improve the spec.

Amazing! Live implementation before your great changes btw -> https://github.com/defund-labs/defund/tree/main/x/query (relayer -> https://github.com/defund-labs/relayer). Feel free to fork as the starting point for the modularized version

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+`CommitmentPath` and `CommitmentProof` are as defined in ICS 23.
+
+`Identifier`, `get`, `set`, `delete`, `getCurrentHeight`, and module-system related primitives are as defined in ICS 24.
+
+## System Model and Properties
+
+### Assumptions
+
+- **Safe chains:** Both the Querying and Queried chains are safe.
```
AdityaSripal, line 41: what do you mean by safe?

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,323 @@
...
+### Data Structures
+
+The Cross-chain Querying module stores query requests when it processes them. 
+
+A CrossChainQuery is a particular interface to represent query requests. A request is retrieved when its result is submitted.
+
+```typescript
+interface CrossChainQuery struct {
+    id: Identifier
+	path: CommitmentPath
```
AdityaSripal, line 90: formatting

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,323 @@
...
+    queryHeight: Height
+    clientId: Identifier
+	bounty: sdk.Coin
+}
+```
+
+- The `id` field uniquely identifies the query at the Querying Chain.
+- The `path` field is the path to be queried at the Queried Chain.
+- The `timeoutHeight` field  specifies a height limit at the Querying Chain after which a query is considered to have failed and a timeout result should be returned to the original caller.
+- The `queryHeigth` field is the height at which the relayer must query the Queried Chain
```
AdityaSripal, line 101: spelling

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,323 @@
...
+5) The query caller can then asynchronously retrieve the query result.
+
+#### Normal path methods
+
+The `CrossChainQueryRequest` function is called when the Cross-chain Querying module at the Querying Chain receives a new query request.
+
+```typescript
+function CrossChainQueryRequest(
+  path: CommitmentPath,
+  queryHeigth: Heigth,
```
AdityaSripal, line 179: replace `Heigth` with `Height` in this file

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,323 @@
...
+  clientId: Identifier,
+  bounty: sdk.Coin,
+  ): Identifier {
+
+    // Check that there exists a client of the Queried Chain. The client will be used to verify the query result.
+    abortTransactionUnless(queryClientState(clientId) !== null)
+
+    // Generate a unique query identifier.
+    queryIdentifier = generateQueryIdentifier()
+
```
AdityaSripal, line 190: check that timeoutHeight > currentHeight

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,323 @@
...
+The `CrossChainQueryResult` function is called when the Cross-chain Querying module at the Querying Chain receives a new query reply.
+
+```typescript
+function CrossChainQueryResult(
+  queryId: Identifier,
+  data: []byte
+  proof: CommitmentProof,
+  proofHeight: Height,
+  success: boolean,
```
AdityaSripal, line 223: can't we simplify this by saying that we will verify nonMembership if data is nil?

Seems like the success field is unnecessary

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,323 @@
...
+        abortTransactionUnless(client.verifyMemership(
+            client,
+            proofHeight,
+            delayPeriodTime,
+            delayPeriodBlocks,
+            proof,
+            query.path,
+            data
+        ))
+        result = FOUND
```
AdityaSripal, line 250: using FOUND and NOTFOUND here, but define `SUCCESS` and `FAILURE` earlier

AdityaSripal (commented):
Excellent work!! This is exactly what i had in mind. There are just a few minor points on the current spec.

One potential improvement I think is to have an additional message that allows the original caller to prune the result from the store once they retrieve it. This would require storing the caller address in the query result, but I think that's worth doing so that we don't have too many queried information staying in the store forever.

Note that if we do this, and also want to protect against malicious caller modules we may have to use object capabilities which would add complexity.

cc: @dtribble to see if this would be a strict requirement for smart contracts, etc.

spec/app/ics-interchain-queries/README.md
angbrav: I have clarified this by adding: This means that, for every chain, the underlying consensus engine satisfies safety (e.g., the chain does not fork) and the execution of the state machine follows the described protocol.

spec/app/ics-interchain-queries/README.md
angbrav: Agree, fixed.

spec/app/ics-interchain-queries/README.md
angbrav: Fixed

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+A bounty is paid to incentivize relayers for participating in interchain queries.
+
+## Technical Specification
+
+### General Design 
+
+The Querying Chain MUST implement the Cross-chain Querying module, which allows the Querying Chain to query state at the Queried Chain. 
+
+Cross-chain querying relies on relayers operating between both chains. When a query request is received by the Querying Chain, the Cross-chain Querying module emits a `sendQuery` event. Relayers operating between the Querying and Queried chains must monitor the Querying chain for `sendQuery` events. Eventually, a relayer will retrieve the query request and execute it at the Queried Chain. The relayer then submits (on-chain) the result at the Querying Chain. The result is finally registered at the Querying Chain by the Cross-chain Querying module.
```
cwgoes, line 77: What do you mean, "execute it at the Queried Chain" - query the full node for proof information?

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,325 @@
...
+```typescript
+enum QueryResult {
+  SUCCESS,
+  FAILURE,
+  TIMEOUT,
+}
+```
+- A query that returns a value is marked as `SUCCESS`. This means that the query has been executed at the Queried Chain and there was a value associated to the queried path at the requested height.
+- A query that is executed but does not return a value is marked as `FAILURE`. This means that the query has been executed at the Queried Chain, but there was no value associated to the queried path at the requested height.
+- A query that timeout before a result is committed at the Querying Chain is marked as `TIMEOUT`.
```
cwgoes, line 117:
```suggestion
- A query that timed out before a result is committed at the Querying Chain is marked as `TIMEOUT`.
```

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+The Cross-chain Querying module stores query requests when it processes them. 
+
+A CrossChainQuery is a particular interface to represent query requests. A request is retrieved when its result is submitted.
+
+```typescript
+interface CrossChainQuery struct {
+    id: Identifier
+    path: CommitmentPath
+    timeoutHeight: Height
```
cwgoes, line 91: should there also be a timeout timestamp, or do we think that isn't useful here?

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,325 @@
...
+        provableStore.set(resultQueryPath(queryIdentifier), resultRecord)
+    }
+}
+```
+- **Precondition**
+  - There is a query request stored in the `privateStore` identified by `queryId`.
+- **Postcondition**
+  - If the query has indeed timeout, then
+    - the query request identified by `queryId` is deleted from the `privateStore`;
+    - the fact that the query has timeout is recorded in the `provableStore`.
```
crodriguezvega, line 325: In other specs we also have `History` and `Copyright` sections at the bottom.

cwgoes (changes requested):
Generally this looks fine to me, modulo a few minor comments. One other question I would have is around "callback" functionality - often, I expect the querying chain will want additional state machine logic to be executed when the query response is received. It seems to me like we could easily allow the `bounty` already in the query struct to pay for some additional execution - is that worth considering?

One concern I have with this bounty system is that it is vulnerable to front-running - any relayer who actually bothers to query the other chain can be front-run by the block proposer, who doesn't need to bother to query the other chain anymore. Have we considered this issue?

crodriguezvega (commented):
Nice spec, @angbrav!

I have a question:
- Will there be any version/metadata negotiation needed during channel handshake so that the querying chain understands in what format the received data is so that it can correctly un-marshall it? Something may like what we have for [ICS-27](https://github.com/cosmos/ibc/blob/master/spec/app/ics-027-interchain-accounts/README.md#metadata-negotiation).

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+A Queried Chain has to do no implementation work or add any module to enable cross-chain querying. By utilizing an RPC client on a relayer, this is possible.
+
+#### Modular
+
+Adding cross-chain querying should be as easy as implementing a module in your chain.
+
+#### Control Queried Data
+
+The Querying Chain should have ultimate control over how to handle queried data. Like querying for a certain query form/type.
```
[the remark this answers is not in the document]
angbrav: I did not write this, it was in the original draft. I am not sure what it means, I lean towards deleting it.

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+#### Incentivization
+
+A bounty is paid to incentivize relayers for participating in interchain queries.
```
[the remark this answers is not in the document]
angbrav: Yes. I have clarified it: fetching data from the Queried Chain and submitting it (together with proofs) to the Querying Chain.

angbrav:
> One other question I would have is around "callback" functionality - often, I expect the querying chain will want additional state machine logic to be executed when the query response is received. It seems to me like we could easily allow the `bounty` already in the query struct to pay for some additional execution - is that worth considering?

I think that if that's the common case, we should consider it. 

> One concern I have with this bounty system is that it is vulnerable to front-running - any relayer who actually bothers to query the other chain can be front-run by the block proposer, who doesn't need to bother to query the other chain anymore. Have we considered this issue?

I am not sure I fully understand the problem. Could you please elaborate a bit more? Is this referring to the fact that the proposer of the Querying Chain may never execute the transaction including the query result because it is not interested, i.e., the relayer never gets "paid"? I thought about this, and that's why one of the assumptions is that the Querying Chain is censorship-resistant, i.e., the Querying Chain cannot selectively omit transactions.

spec/app/ics-interchain-queries/README.md
angbrav: I have clarified that:

...execute it, i.e., fetch the data and generate the corresponding proofs, at the Queried Chain.

angbrav:
> * Will there be any version/metadata negotiation needed during channel handshake so that the querying chain understands in what format the received data is so that it can correctly un-marshall it? Something may like what we have for [ICS-27](https://github.com/cosmos/ibc/blob/master/spec/app/ics-027-interchain-accounts/README.md#metadata-negotiation).

The spec does not requires the existence of a channel between the Querying and Queried chain, so that's not possible.

spec/app/ics-interchain-queries/README.md
```diff
@@ -0,0 +1,325 @@
...
+A CrossChainQuery is a particular interface to represent query requests. A request is retrieved when its result is submitted.
+
+```typescript
+interface CrossChainQuery struct {
+    id: Identifier
+    path: CommitmentPath
+    timeoutHeight: Height
+    queryHeight: Height
+    clientId: Identifier
+    bounty: sdk.Coin
```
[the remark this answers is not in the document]
angbrav: I think it is a good idea. It is kind of weird to have cross-apps dependencies, isn't? We should find a way to define the fee interface somewhere else in the core specs. Since this is a relayer thingy, maybe the relayer should define it.

spec/app/ics-interchain-queries/README.md
cwgoes: Query timeouts based on a timestamp on the destination chain, instead of a height. I can imagine this could be useful for use-cases which need to coordinate based on a global clock (e.g. oracles)

cwgoes:
> I am not sure I fully understand the problem. Could you please elaborate a bit more? Is this referring to the fact that the proposer of the Querying Chain may never execute the transaction including the query result because it is not interested, i.e., the relayer never gets "paid"? I thought about this, and that's why one of the assumptions is that the Querying Chain is censorship-resistant, i.e., the Querying Chain cannot selectively omit transactions.

No, the proposer can just copy the results of the query from the transaction in the mempool and resubmit the query result in order to claim the fee for themselves, meaning that the relayer never receives it - and thus no relayer, in expectation of the proposer employing this strategy, has any reason to fetch the data at all (so I don't really think this incentive mechanism works, without an alteration such as designating a specific relayer).

spec/app/ics-interchain-queries/README.md
AdityaSripal: The timeouts here are semantically different than the timeouts in IBC applications. Since there isn't a way for the relayer in this case to prove that they couldn't query a particular key by the timeout height on the destination, this timeout is the timeout on the **source** chain, at which point the source chain will timeout the query.

Perhaps we should either name it something different to avoid confusion or do away with the concept

spec/app/ics-interchain-queries/README.md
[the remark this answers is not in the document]
AdityaSripal: The spec should not be dependent on SDK types

spec/app/ics-interchain-queries/README.md
cwgoes: Ah right right, I stand corrected. I still think this could be useful, but it's not that important.

michaelfig:
Could someone here please review the pubsub implementation I'm proposing at #759?  The TLDR is [the comment where I describe the features I'd need](https://github.com/cosmos/ibc/discussions/759#discussioncomment-2824351) from IBC queries to be able to implement pubsub as an application.

spec/app/ics-interchain-queries/README.md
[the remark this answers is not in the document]
schnetzlerjoe: > What does this mean, I don't quite follow - querying a specific location in state? What is a query form/type? Can we provide an example?

Sorry I wrote this part, To be clear, I am referring to what you are referencing above " I expect the querying chain will want additional state machine logic to be executed when the query response is received". The Querying chain should have full authority and the ability to handle any logic needed on query completion.

michaelfig:
For those who didn't get my edits above, I'm proposing two features to enable Querying Chains to consume immutable streams of values from the Queried chain:

- add a "not-equal" query limiter that requires the result not to match in order to satisfy the query
- add an "after" query limiter that can only satisfy the query with a result in some block after the specified blockheight
- combine them to get "not-equal, after"

Implementing these restrictions in relayers should be straightforward, but they can be assisted by standard "on-change" events, such as:
- `ibcpub.path=<store>:<key>` the path's data has changed
- `ibcpub.data=<base64-encoding, or 'null'>` the new contents of the path

These events would be advisory and for unproved clients; relayers would still request proof of them in the Queried chain's Merkle tree.

michaelfig:
Thought people may want to see how this relates to streaming: https://github.com/Agoric/agoric-sdk/blob/mfig-vstream/golang/cosmos/x/vstream/spec/01_concepts.md#forward-iteration-and-interchain-queries

AdityaSripal (commented):
Capability logic looks correct. Nice work!

AdityaSripal:
We should also add a relayer list field that can be ignored for now into the request struct

angbrav:
I have addressed all pending concerns (listed in the PR description) as discussed. This is ready for review.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+The Cross-chain Querying module stores query requests when it processes them. 
+
+A CrossChainQuery is a particular interface to represent query requests. A request is retrieved when its result is submitted.
+
+```typescript
+interface CrossChainQuery struct {
+    id: Identifier
+    path: CommitmentPath
+    localTimeoutHeight: Height
```
AdityaSripal, line 95: renaming++

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+```
+
+### Sub-protocols
+
+#### Query lifecycle
+
+1) When the Querying Chain receives a query request, it calls `CrossChainQueryRequest` of the Cross-chain Querying module. This function generates a unique identifier for the query, stores it in its `privateStore` and emits a `sendQuery` event. Query requests can be submitted by other IBC modules as transactions to the Querying Chain or simply executed as part of the `BeginBlock` and `EndBlock` logic.
+2) A correct relayer listening to `sendQuery` events from the Querying Chain will eventually pick the query request up and execute it at the Queried Chain. The result is then submitted (on-chain) to the Querying Chain.
+3) When the query result is committed at the Querying Chain, this calls the `CrossChainQueryResult` function of the Cross-chain Querying module.
+4) The `CrossChainQueryResult` first retrieves the query from the `privateStore` using the query's unique identifier. It then proceeds to verify the result using its local client. If it passes the verification, the function removes the query from the `privateStore` and stores the result in a public path.
```
AdityaSripal, line 175: Why does the result need to be in a public path?

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+The Querying Chain calls the `checkQueryTimeout` function to check whether a specific query has timed out. 
+
+> There are several alternatives on how to handle timeouts. For instance, the relayer could submit on-chain timeout notifications to the Querying Chain. Since the relayer is untrusted, for each of these notifications the Cross-chain Querying module of the Querying Chain MUST call the `checkQueryTimeout` to check if the query has indeed timed out. An alternative could be to make the Cross-chain Querying module responsible for checking  
+if any query has timed out by iterating over the ongoing queries at the beginning of a block and calling `checkQueryTimeout`. This is an implementation detail that this specification does not cover.
+
+We pass the relayer address just as in `CrossChainQueryResult` to allow for possible incentivization here as well.
+
+```typescript
+function checkQueryTimeout(
```
AdityaSripal, line 342: We should store state in such a way that we can efficiently iterate over existing queries in timeout order (for timestamp and height), though this may be left as details for implementation

AdityaSripal (approved):
Nice work!

I think supporting the streams work @michaelfig is important, but would like to merge this and see any changes for those in a separate PR so we can review them in isolation

spec/app/ics-031-crosschain-queries/README.md
angbrav: No need. Fixed.

spec/app/ics-031-crosschain-queries/README.md
angbrav: I think this is an implementation detail. I have added the following sentence: "In this case, ongoing queries should be stored indexed by `localTimeoutTimestamp` and `localTimeoutHeight` to allow iterating over them more efficiently"

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+```typescript
+function generateQueryIdentifier = () -> Identifier
+```
+
+### Sub-protocols
+
+#### Query lifecycle
+
+1) When the Querying Chain receives a query request, it calls `CrossChainQueryRequest` of the Cross-chain Querying module. This function generates a unique identifier for the query, stores it in its `privateStore` and emits a `sendQuery` event. Query requests can be submitted by other IBC modules as transactions to the Querying Chain or simply executed as part of the `BeginBlock` and `EndBlock` logic.
```
mpoke, line 172: > Query requests can be submitted by other IBC modules

Why only IBC modules?

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+2) A correct relayer listening to `sendQuery` events from the Querying Chain will eventually pick the query request up and execute it at the Queried Chain. The result is then submitted (on-chain) to the Querying Chain.
+3) When the query result is committed at the Querying Chain, this calls the `CrossChainQueryResult` function of the Cross-chain Querying module.
```
mpoke, line 174: > this calls the `CrossChainQueryResult` function

Name collision with the data structure for results / responses.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+  bounty: Fee,
+  ): [Identifier, CapabilityKey] {
+
+    // Check that there exists a client of the Queried Chain. The client will be used to verify the query result.
+    abortTransactionUnless(queryClientState(clientId) !== null)
+
+    // Sanity-check that localTimeoutHeight is 0 or greater than the current height, otherwise the query will always time out.
+    abortTransactionUnless(localTimeoutHeight === 0 || localTimeoutHeight > getCurrentHeight())
+    // Sanity-check that localTimeoutTimestamp is 0 or greater than the current timestamp, otherwise the query will always time out.
+    abortTransactionUnless(localTimeoutTimestamp === 0 || localTimeoutTimestamp > currentTimestamp())
```
mpoke, line 198: `localTimeoutTimestamp` is missing from function signature.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+    // Check that the relier executed the query at the requested height at the Queried Chain.
+    abortTransactionUnless(query.queryHeight !== proofHeight)
+
+    // Check that localTimeoutHeight is 0 or greater than the current height.
+    abortTransactionUnless(query.localTimeoutHeight === 0 || query.localTimeoutHeight > getCurrentHeight())
+    // Check that localTimeoutTimestamp is 0 or greater than the current timestamp.
+    abortTransactionUnless(query.localTimeoutTimestamp === 0 || query.localTimeoutTimestamp > currentTimestamp()) 
+
+
+    // Verify query result using the local light client of the Queried Chain. If success, then verify that the data is indeed the value associated with query.path at query.queryHeight at the Queried Chain. Otherwise, verify that query.path does not exist at query.queryHeight at the Queried Chain.
```
mpoke, line 261: Split on multiple lines

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+#### Timeouts
+
+Query requests have associated a `localTimeoutHeight` and a `localTimeoutTimestamp` field that specifies the height and timestamp limit at the Querying Chain after which a query is considered to have failed. 
+
+The Querying Chain calls the `checkQueryTimeout` function to check whether a specific query has timed out. 
+
+> There are several alternatives on how to handle timeouts. For instance, the relayer could submit on-chain timeout notifications to the Querying Chain. Since the relayer is untrusted, for each of these notifications the Cross-chain Querying module of the Querying Chain MUST call the `checkQueryTimeout` to check if the query has indeed timed out. An alternative could be to make the Cross-chain Querying module responsible for checking  
+if any query has timed out by iterating over the ongoing queries at the beginning of a block and calling `checkQueryTimeout`. In this case, ongoing queries should be stored indexed by `localTimeoutTimestamp` and `localTimeoutHeight` to allow iterating over them more efficiently. These are implementation details that this specification does not cover.
```
mpoke, line 337: nit: remove new line (i.e., there are two spaces at the end of the prev line)

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+  - There is a query result stored in the `privateStore` identified by `queryId`.
+  - The caller has the right to clean the query result
+- **Postcondition**
+  - The query result identified by `queryId` is deleted from the `privateStore`.
+
+#### Timeouts
+
+Query requests have associated a `localTimeoutHeight` and a `localTimeoutTimestamp` field that specifies the height and timestamp limit at the Querying Chain after which a query is considered to have failed. 
+
+The Querying Chain calls the `checkQueryTimeout` function to check whether a specific query has timed out.
```
mpoke, line 334: Not sure I understand how the "Querying Chain calls the `checkQueryTimeout` function". Is this a transaction? The pseudocode seems to contain several `abortTransactionUnless`.

mpoke (commented):
Great work @angbrav. See my comments below.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+## System Model and Properties
+
+### Assumptions
+
+- **Safe chains:** Both the Querying and Queried chains are safe. This means that, for every chain, the underlying consensus engine satisfies safety (e.g., the chain does not fork) and the execution of the state machine follows the described protocol.
+
+- **Live chains:** Both the Querying and Queried chains MUST be live, i.e., new blocks are eventually added to the chain.
+
+- **Censorship-resistant Querying Chain:**  The Querying Chain cannot selectively omit transactions.
```
cwgoes, line 49: what does this mean? not a very formal definition.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+The Querying Chain MUST implement the Cross-chain Querying module, which allows the Querying Chain to query state at the Queried Chain. 
+
+Cross-chain querying relies on relayers operating between both chains. When a query request is received by the Querying Chain, the Cross-chain Querying module emits a `sendQuery` event. Relayers operating between the Querying and Queried chains must monitor the Querying chain for `sendQuery` events. Eventually, a relayer will retrieve the query request and execute it, i.e., fetch the data and generate the corresponding proofs, at the Queried Chain. The relayer then submits (on-chain) the result at the Querying Chain. The result is finally registered at the Querying Chain by the Cross-chain Querying module. 
+
+A query request includes the height of the Queried Chain at which the query must be executed. The reason is that the keys being queried can have different values at different heights. Thus, a malicious relayer could choose to query a height that has a value that benefits it somehow. By letting the Querying Chain decide the height at which the query is executed, we can prevent relayers from affecting the result data.
```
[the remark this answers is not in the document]
cwgoes: worth noting, this still creates an opportunity for altering state on the queried chain if the height is in the future in order to change the results of the query

(not a relayer-specific problem, but would be a form of cross-chain MEV)

cwgoes (approved):
ACK, but see notes

spec/app/ics-031-crosschain-queries/README.md
angbrav: This means that all correctly submitted transactions are eventually committed. Unless one considers live chains to already guarantee this, I think this is required to guarantee that the query result submitted by a correct relayer is eventually processed.

cwgoes: Correctly submitted to whom? Any nodes can always selectively omit transactions, since you cannot prove that another party received a message. No blockchain can instantiate this property in the general case.

I agree that we need an assumption in order to guarantee that the query result submitted by a correct relayer is eventually processed - isn't that assumption stronger (processed within a time bound) if timeouts are involved, anyways? I think you could say something like "the relayer must submit and the querying chain must process a "QueryResult" transaction within the specified timeout bound in order for the query protocol to return results to the application" - would that do?

Raulnori77:
#

Raulnori77:
238b6b116f5dc73212019cdecccf564031707f934a7c0012ed8cf1a7bdef9fde.patch

spec/app/ics-031-crosschain-queries/README.md
angbrav: > I think you could say something like "the relayer must submit and the querying chain must process a "QueryResult" transaction within the specified timeout bound in order for the query protocol to return results to the application" - would that do?

Yes that would work.

angbrav: > Correctly submitted to whom?

To the Querying Chain.

> Any nodes can always selectively omit transactions, since you cannot prove that another party received a message. No blockchain can instantiate this property in the general case.

I think this is an interesting (open) research question. Non-reconfigurable Byz. consensus protocols are able to satisfy this property. PBFT for instance is censorship-resistant by broadcasting requests, setting timeouts per request, and forcing a view change if a timeout expires. When adding reconfiguration to the equation, things get more complicated, but I believe things may be solved by adding extra assumptions.

cwgoes: > I think this is an interesting (open) research question. Non-reconfigurable Byz. consensus protocols are able to satisfy this property. PBFT for instance is censorship-resistant by broadcasting requests, setting timeouts per request, and forcing a view change if a timeout expires. When adding reconfiguration to the equation, things get more complicated, but I believe things may be solved by adding extra assumptions.

How can a user with a transaction force a view change if a timeout expires? I thought timeouts were for proposals - and block proposers can always choose to omit transactions. Perhaps I don't follow, sorry. Under assumptions of bounded latency message delivery from a user authoring a transaction to a sufficient quorum of validators, and honesty of those validators w.r.t. including all received messages in blocks they propose in a particular order, and limited other transactions competing for block space, I could see some sort of transaction inclusion property holding, but these are far more strict assumptions than IBC in general makes, and should be spelled out in detail imo.

angbrav: >How can a user with a transaction force a view change if a timeout expires? I thought timeouts were for proposals - and block proposers can always choose to omit transactions. Perhaps I don't follow, sorry.

Let me elaborate a bit. This is roughly how it works. Clients broadcast their requests to all consensus nodes. When a node receives a request, it puts it into its FIFO queue and starts a timer. If the request's timer expires before the request is delivered/committed at a node, it attempts to advance to the next view. Therefore, if the leader of the current view is Byzantine and ignores a request, eventually all correct nodes will attempt to advance and force a leader change (actually we do not need all correct nodes but a majority). Note that if a correct node delivers/commits requests, then all correct ones eventually do. You can find a formal proof at https://arxiv.org/pdf/2202.06679.pdf: Figures 4 and 5 show the pseudocode of a variant of PBFT which includes this mechanism, and Section 5 proves a liveness property that includes censorship-resistance.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+### Desired Properties
+
+#### Permissionless
+
+A Querying Chain can query a chain and implement cross-chain querying without any approval from a third party or chain governance. Note that since there is no prior negotiation between chains, the Querying Chain cannot assume that queried data will be in an expected format.
+
+#### Minimal Queried Chain Work
+
+A Queried Chain has to do no implementation work or add any module to enable cross-chain querying. By utilizing an RPC client on a relayer, this is possible.
```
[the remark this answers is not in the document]
angbrav: Are there chains that do not provide query support? What's required to support queries? This is more for my own education :)

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+#### Modular
+
+Adding cross-chain querying should be as easy as implementing a module in your chain.
+
+#### Control Queried Data
+
+The Querying Chain should have ultimate control over how to handle queried data. Like querying for a certain query form/type.
```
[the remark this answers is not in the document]
angbrav: I've remove this property. See **https://github.com/cosmos/ibc/pull/735#discussion_r874671723**

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+#### Incentivization
+
+A bounty is paid to incentivize relayers for participating in interchain queries: fetching data from the Queried Chain and submitting it (together with proofs) to the Querying
```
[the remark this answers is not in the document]
angbrav: Fixed.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+## Technical Specification
+
+### General Design 
+
+The Querying Chain MUST implement the Cross-chain Querying module, which allows the Querying Chain to query state at the Queried Chain. 
+
+Cross-chain querying relies on relayers operating between both chains. When a query request is received by the Querying Chain, the Cross-chain Querying module emits a `sendQuery` event. Relayers operating between the Querying and Queried chains must monitor the Querying chain for `sendQuery` events. Eventually, a relayer will retrieve the query request and execute it, i.e., fetch the data and generate the corresponding proofs, at the Queried Chain. The relayer then submits (on-chain) the result at the Querying Chain. The result is finally registered at the Querying Chain by the Cross-chain Querying module.
```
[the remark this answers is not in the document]
angbrav: Creates a transaction with the result and submits it to the querying chain. I have removed the on-chain. Now it simply says "submits the result in a transaction to the querying chain"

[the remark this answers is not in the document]
angbrav: I have dealt similarly with other occurrences of on-chain

spec/app/ics-031-crosschain-queries/README.md
angbrav: I think I discussed this with Aditya. I believe the conclusion was that in most of the cases, cross-chain queries will be issued by IBC modules, as end users would directly query the queried chain. At least this is what I remember, but maybe @AdityaSripal can elaborate a bit more. In any case, I think that even if that's the common case, we should not constraint the specification, so I will fix the text.

spec/app/ics-031-crosschain-queries/README.md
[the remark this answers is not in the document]
mpoke: > Are there chains that do not provide query support?

Probably not, but just to be sure since it's a prerequisite of Interchain Queries.

spec/app/ics-031-crosschain-queries/README.md
angbrav: It depends. The spec discusses two alternatives: in one `checkQueryTimeout` is a transaction. I understand the confusion though. Ideally, this function (since it can be both a transaction or simply a function call from the module) would not include any `abortTransactionUnless`. Any suggestion on how to solve this? I can think of two solutions:
- remove `abortTransactionUnless` from `checkQueryTimeout` and use if-else conditions. Probably the best solution.
- commit to one of the discussed alternatives

spec/app/ics-031-crosschain-queries/README.md
cwgoes: Thanks! Makes sense (and very cool research).

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+## Synopsis
+
+This standard document specifies the data structures and state machine handling logic of the Cross-chain Querying module, which allows for cross-chain querying between IBC enabled chains.
+
+## Overview and Basic Concepts
+
+### Motivation
+
+Interchain Accounts (ICS-27) brings one of the most important features IBC offers, cross-chain transactions (on-chain). Limited in this functionality is the querying of state from one chain, on another chain. Adding cross-chain querying via the Cross-chain Querying module gives unlimited flexibility to chains to build IBC enabled protocols around Interchain Accounts and beyond.
```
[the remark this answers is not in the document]
angbrav: There is a new motivation now. I think it is better. Thanks for pointing it out :)

spec/app/ics-031-crosschain-queries/README.md
angbrav: I have opted for the second solution.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -0,0 +1,402 @@
...
+
+### Assumptions
+
+- **Safe chains:** Both the queryin and queried chains are safe. This means that, for every chain, the underlying consensus engine satisfies safety (e.g., the chain does not fork) and the execution of the state machine follows the described protocol.
+
+- **Live chains:** Both the querying and queried chains MUST be live, i.e., new blocks are eventually added to the chain.
+
+- **Censorship-resistant querying chain:**  The querying chain cannot selectively omit valid transactions.
+
+> For example, this means that if a relayer submits a valid transaction to the querying chain, the transaction is guaranteed to be eventually included in a committed block. Note that Tendermint does not currently guarantee this.
```
mpoke, line 51: Not sure whether Tendermint can ever guarantee this.

mpoke (approved):
Nice work @angbrav.

PR #741, opened 2022-05-18
Multi-Hop Channel Draft

AdityaSripal:


AdityaSripal:
This spec is quite complicated. The main issues I see are the following:

1. A connection is identified differently on both sides (this is an unavoidable property of IBC)
2. In order to maintain backwards compatibility and allow for a single channel to be serviced by multiple routes; we've squeezed multiple connection routes into the `connectionHops: [string]` field. This creates confusion since in some places the list of strings refers to multiple independent routes and in other places (the router functions) it is individual hops in a specific route. This clearly can and should be fixed to avoid confusion and improve consistency.

To minimize remaining work, we should merge the packet handling changes to this draft as-is. With the work from the product team in fully gathering requirements and use cases, we can revisit this spec to determine what we should do for it moving forward after those results come in.

Issue #744, opened 2022-05-19
ICS28: Send SlashAcks only for downtime slash requests


mpoke:
This issue was mentioned here https://github.com/cosmos/interchain-security/pull/97#discussion_r873802441

The fix entails the following changes to [onRecvSlashPacket](https://github.com/cosmos/ibc/blob/marius/ccv/spec/app/ics-028-cross-chain-validation/technical_specification.md#ccv-pcf-rcvslash1)
```typescript
if packet.data.downtime {
  // add validator to list of slash requests for chainId
  slashRequests[chainId].Append(packet.data.valAddress)
}
```

PR #745, opened 2022-05-19
ICS28: Send SlashAcks only for downtime slash requests

mpoke:
Closes https://github.com/cosmos/ibc/issues/744


sainoe (approved):
Looks good,!

Issue #739, opened 2022-05-17
ICS28: vsc will not mature if there are no updates from provider


danwt:
_a6bc8d14dc1fc45cdcd052e6da8c6765389b7d61_

It seems that if the early return is taken in consumer::EndBlock

```go
  if pendingChanges.IsEmpty() {
    // do nothing
    return []
  }
 ..
  if providerChannel != "" {
    UnbondMaturePackets()
  }
```

then the UnbondMaturePackets will never be called. If there are no changes for many blocks, the provider will never receive the maturity.

https://github.com/cosmos/ibc/blob/a6bc8d14dc1fc45cdcd052e6da8c6765389b7d61/spec/app/ics-028-cross-chain-validation/technical_specification.md?plain=1#L1723-L1751

danwt:
There's a good chance I've made a mistake in my understanding here, if so, please feel free to close the issue :)

mpoke:
@danwt Nice catch. Could you please open an issue also on the implementation side?

PR #747, opened 2022-05-19
ICS28: vsc will not mature if there are no updates from provider

mpoke:
Closes https://github.com/cosmos/ibc/issues/739


danwt:
I think this works.

danwt (approved):

PR #731, opened 2022-05-03
ICS4: Simplify sendPacket interface

mpoke:
Closes https://github.com/cosmos/ibc/issues/708

Before merging update the calls to sendPacket throughout the standard, i.e., 

- [x] [ICS 20](https://github.com/cosmos/ibc/blob/master/spec/app/ics-020-fungible-token-transfer/README.md#packet-relay), which entails changing the signature of `createOutgoingPacket` (already inconsistent with ICS-4, see https://github.com/cosmos/ibc/issues/709)
- [x] [ICS27](https://github.com/cosmos/ibc/blob/master/spec/app/ics-027-interchain-accounts/README.md#sendtx), which entail updating `SendTx` (already inconsistent with ICS-4)
- [x] [ICS 29](https://github.com/cosmos/ibc/blob/master/spec/app/ics-029-fee-payment/README.md#embedded-applications-calling-into-ics-4)
- [x] [ICS 30](https://github.com/cosmos/ibc/blob/master/spec/app/ics-030-middleware/README.md)

PR #753, opened 2022-05-20
ICS27: Simplify sendPacket interface

mpoke:
Makes ICS27 consistent with https://github.com/cosmos/ibc/pull/731

### DIFF
--- a/spec/app/ics-027-interchain-accounts/README.md
+++ b/spec/app/ics-027-interchain-accounts/README.md
@@ -63,7 +63,7 @@ on what middleware is enabled on either end of the channel. Thus it is recommend
 version automatically and allow for users to optionally enable additional middleware versioning.
 
 ```typescript
-function RegisterInterchainAccount(connectionId: string, owner: string, version: string) returns (error) {
+function RegisterInterchainAccount(connectionId: Identifier, owner: string, version: string) returns (error) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -273,7 +299,7 @@ icaPacketData = InterchainAccountPacketData{
 }
 
 // Sends the message to the host chain, where it will eventually be executed 
-SendTx(ownerAddress, portID, data, timeout)
+SendTx(ownerAddress, connectionId, portID, data, timeout)
```
mpoke, line 302: @AdityaSripal @colin-axner What is the `connectionId` here exactly?

spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -72,9 +72,35 @@ function RegisterInterchainAccount(connectionId: string, owner: string, version:
 `SendTx` is used to send an IBC packet containing instructions (messages) to an interchain account on a host chain for a given interchain account owner.
 
 ```typescript
-function SendTx(channelCapability: ChannelCapability, portId: string, icaPacketData: InterchainAccountPacketData, timeoutTimestamp uint64) returns (uint64, error) {
-    // A call to GetActiveChannel() checks if there is a currently active channel for this portId which also implies an interchain account has been registered using this port identifier
-    // if there are no errors CreateOutgoingPacket() is called with the given packet data and timeout and the IBC packet will be sent to the host chain on the active channel
+function SendTx(
+  capability: CapabilityKey, 
+  connectionId: string,
+  portId: string,
```
AdityaSripal, line 78: Match rest of spec by using `Identifier`

spec/app/ics-027-interchain-accounts/README.md
AdityaSripal: The `connectionId` here identifies the ICS3 connection that the channel is built on top of. ICA only allows one ICA channel per connection to allow for reopening logic

AdityaSripal (approved):

PR #758, opened 2022-05-23
Capitalize type name 'byte' in Rust code

junha1:
That's the convention for types in Rust.


ethanfrey:
Hmmm... I always saw `Vec<u8>`.

Unless you are talking about [`bytes::Bytes`](https://docs.rs/bytes/latest/bytes/) which is upper cased, but a rather low-level optimisation to include in a spec

junha1:
@ethanfrey Yeah I know `Vec<u8>` is better, but I was not sure if there was some significant intention to designate the type as 'byte', so I just kept that but just capitalizing it. Anyway I'll change to `u8`.

ethanfrey (approved):
This looks better

mpoke (approved):

Issue #673, opened 2022-03-03
ICS28: Setting CCV parameters


mpoke:
For CCV v1, starting a consumer chain requires a governance proposal to be passed on the provider chain. This proposal can be seen as a contract between the provider and consumer chains, i.e., "The provider chain agrees to provide security to the consumer chain in certain conditions". 

These conditions are described by some CCV parameters, i.e., 
- `UnbondingPeriod` - the period that must elapsed on the consumer chain before bonded tokens can be unlocked;
- `SlashingFraction`s - the percentages of the stake slashed for different infractions;
- `JailDuration`s - the period a validator is jailed for different infractions;
- `DistributionFraction` - the percentage of rewards accumulated on the consumer chain that are transferred to the provider chain (as payment for providing security). 

The first three parameters affect the state of the provider chain. The fourth parameter affects the earnings of the provider chain. The consumer chain should not be able to change them without the consent of the provider chain, i.e., a change should entail a renegotiation of the contract. Thus, these parameters should be part of the governance  proposal to spawn a new consumer chain and the consumer chain should not be able to modify them through a consumer governance proposal.

mpoke:
CCV will use the `SlashingFraction` and `JailDuration` set on the provider; the consumer will not be able to choose its own. See https://github.com/cosmos/ibc/issues/726

mpoke:
The consumer should also use the UnbondingPeriod set on the provider. See discussion on the implementation side https://github.com/cosmos/interchain-security/issues/80

mpoke:
Regarding the `DistributionFraction`, see discussion on the implementation side https://github.com/cosmos/interchain-security/issues/81

PR #771, opened 2022-06-20
ICS28: Add consumer UnbondingPeriod

mpoke:
The consumer chain unbonding period is computed from the provider chain unbonding period. 

Closes https://github.com/cosmos/ibc/issues/673


AdityaSripal (approved):
Approved though this seems like a detail of the current implementation

Issue #767, opened 2022-06-16
ICS28: Remove existing pending spawn/stop proposals when adding new ones for a given consumer chain ID


danwt:
_ See implementation mirror issue https://github.com/cosmos/interchain-security/issues/155 _

When a proposal gets handled it will be added to a `pending` list if `spawnTime` has not yet passed.

https://github.com/cosmos/ibc/blob/ad30596341b3f0901d9520b00f175a8b3c702ce5/spec/app/ics-028-cross-chain-validation/methods.md?plain=1#L351-L361

I'm suspicious of what might happen if we allow multiple entries in `pending` for a given chain. I cannot think of any situation where this would be beneficial. It could only lead to strange behavior.

I think we should ensure that `SetPendingClientInfo` only has at most one entry per chainID and updates the existing entry if it already exists instead of creating a new entry.

PR #775, opened 2022-06-23
ICS28: Add conditions for CreateConsumerClient and StopConsumerChain

mpoke:
Closes https://github.com/cosmos/ibc/issues/767

### DIFF
--- a/spec/app/ics-028-cross-chain-validation/methods.md
+++ b/spec/app/ics-028-cross-chain-validation/methods.md
@@ -408,6 +408,11 @@ function SpawnConsumerChainProposalHandler(p: SpawnConsumerChainProposal) {
 // PCF: Provider Chain Function
 // Utility method
 function CreateConsumerClient(p: SpawnConsumerChainProposal) {
+  // check that a client for this chain does not exist
+  if p.chainId IN chainToClient.Keys() {
+    return
+  }
+
   // create client state
   clientState = ClientState{
     chainId: p.chainId,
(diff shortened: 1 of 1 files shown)


danwt (approved):
Great!

spec/app/ics-028-cross-chain-validation/methods.md
```diff
@@ -440,10 +445,12 @@ function CreateConsumerClient(p: SpawnConsumerChainProposal) {
 - **Precondition** 
   - `currentTimestamp() > p.spawnTime`.
 - **Postcondition** 
-  - A client state is created with `chainId = p.chainId` and `unbondingPeriod` set to `ComputeConsumerUnbondingPeriod(stakingKeeper.UnbondingTime())`.
-  - A consensus state is created with `validatorSet` set to the validator set the provider chain own consensus state at current height.
-  - A client of the consumer chain is created and the client ID is added to `chainToClient`.
-  - `lockUnbondingOnTimeout[p.chainId]` is set to `p.lockUnbondingOnTimeout`.
+  - If a client for `p.chainId` already exists, the state is not changed.
```
AdityaSripal, line 448: Why would this ever be the case?

AdityaSripal (approved):
Just a question on if chainIDs don't exist in either function

spec/app/ics-028-cross-chain-validation/methods.md
mpoke: It's a sanity check. Let's say that there is already a consumer chain `C1` with `chainId`, with a client, connection and channel established. If a proposal passes to create another consumer chain `C2` with the same `chainId`, then this can break the CCV channel to `C1`.

mpoke: Most likely, such a proposal should not pass, but I'd prefer to not have to trust validators that they're checking the `chainId`s.

AdityaSripal: Ok sounds good :+1:, just wanted to make sure i wasn't missing something in my understanding

PR #789, opened 2022-07-07
ICS27: Update Callbacks

AdityaSripal:
Update based on #629


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -379,7 +379,7 @@ function onChanOpenInit(
   channelIdentifier: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  version: string) {
+  version: string) => (version: string, err: Error) {
```
crodriguezvega, line 382: This is a bit of a nit but in some function signatures we use `=>` and in others `:`. It would be good to have some kind of coding guidelines for the pseudocode language that we use in the specs. What do you think?

mpoke (approved):
LGTM

PR #790, opened 2022-07-07
ICS30: Update callbacks

AdityaSripal:
Update based on #629

### DIFF
--- a/spec/app/ics-030-middleware/README.md
+++ b/spec/app/ics-030-middleware/README.md
@@ -89,10 +89,17 @@ function onChanOpenInit(
   channelIdentifier: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  version: string) {
-    middlewareVersion, appVersion = splitMiddlewareVersion(version)
+  version: string) (version: string, err: Error) {
+    if version != "" {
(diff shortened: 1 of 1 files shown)


spec/app/ics-030-middleware/README.md
```diff
@@ -101,6 +108,12 @@ function onChanOpenInit(
         counterpartyChannelIdentifier,
         appVersion, // note we only pass app version here
     )
+    abortTransactionUnless(err != nil)
```
mpoke, line 111: If we call `abortTransactionUnless` here, is there a reason to return `Error`?

spec/app/ics-030-middleware/README.md
```diff
@@ -130,6 +144,12 @@ function OnChanOpenTry(
           cpAppVersion, // note we only pass counterparty app version here
           appVersion, // only pass app version
       )
+      abortTransactionUnless(err != nil)
```
mpoke, line 147: ditto

mpoke (approved):
See my comment below.

spec/app/ics-030-middleware/README.md
AdityaSripal: This is true, but I'm just trying to stay consistent with the ICS26 function signature. In practice, the abortTransaction process is carried out by returning an error at this level.

PR #791, opened 2022-07-07
ICS721: Update Callbacks

AdityaSripal:
Updates based on #629

### DIFF
--- a/spec/app/ics-721-nft-transfer/README.md
+++ b/spec/app/ics-721-nft-transfer/README.md
@@ -212,11 +212,13 @@ function onChanOpenInit(
   channelIdentifier: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  version: string) {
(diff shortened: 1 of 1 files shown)


mpoke (approved):
LGTM

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -212,11 +212,13 @@ function onChanOpenInit(
...
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  version: string) {
+  version: string) (version: string, err: Error) {
   // only unordered channels allowed
   abortTransactionUnless(order === UNORDERED)
   // assert that version is "ics721-1"
-  abortTransactionUnless(version === "ics721-1")
+  // or relayer passed in empty version
+  abortTransactionUnless(version === "ics721-1" || version === "")
```
crodriguezvega, line 220: We should communicate with the team implementing the NFT transfer app so that they take into account this change (if they don't already), right?

PR #798, opened 2022-07-19
META: Update README Table

AdityaSripal:
Used stages from https://github.com/cosmos/ibc/blob/master/meta/PROCESS.md

Added the links I know of. Please open additional PRs if I miss a completed implementation


mpoke (approved):
Nice work. I would be cool, once we add tagged releases to this repo, to point to exact implementation  releases.

Issue #764, opened 2022-06-01
`ClientState` is stored in the `privateStore` in the standard, but is provable in ibc-go


plafer:
ibc-go [returns a proof in `QueryClientStateResponse`](https://github.com/cosmos/ibc-go/blob/b9a95020d38384dd990fd68a09a76731ceda91c5/proto/ibc/core/client/v1/query.proto#L70-L77), suggesting that `ClientState` is stored in the `provableStore`, whereas the standard claims that `ClientState` is stored in `privateStore` (see ICS-24 and ICS-2).

The standard should probably be adjusted to remove this inconsistency.

plafer:
Moreover, ibc-go relies on a proven `ClientState` for its [`ConnOpenTry()`](https://github.com/cosmos/ibc-go/blob/e19336552f75277c12e98ad3ba6038e19fc267f4/modules/core/03-connection/keeper/handshake.go#L70) and [`ConnOpenAck()`](https://github.com/cosmos/ibc-go/blob/e19336552f75277c12e98ad3ba6038e19fc267f4/modules/core/03-connection/keeper/handshake.go#L202) handlers, whereas the standard doesn't.

AdityaSripal:
Indeed the clientstate must be stored in the public store. 

Also, definitely need to add the client verification from ibc-go into the standard. Not sure why this was missed.

Thanks @plafer for finding these! Are you interested in submitting the fixes?

plafer:
Yes sure I can do that!

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #802, opened 2022-07-26
Put `ClientState` in `provableStore` and use it in `connOpen{Try,Ack}`

plafer:
Closes: #764

I wasn't sure where to define the `validateSelfClient()` function. ibc-go puts it in the ICS-2 client keeper, with the Tendermint implementation. However, `validateSelfClient()` is a function that should be provided by the host, and so belongs in ICS-24 in my opinion. Since ICS-24 doesn't have any per-consensus-protocol sections, I opted for writing the generic declaration of `validateSelfClient()` in a new "Client state validation" section, and provide the Tendermint implementation as an example. I'm open to better ideas on how to organize this.

I faced another challenge. In [ibc-go](https://github.com/cosmos/ibc-go/blob/85893eebee1002f7ef74514b49c0981215405ceb/modules/core/02-client/keeper/keeper.go#L270), `validateSelfClient()` takes an `sdk.Context` as parameter and compares the appropriate fields against the fields of the `ClientState` passed in as argument. The most elegant way I found to abstract this use of the `sdk.Context` is to require the hosts to implement 

```typescript
type getHostClientState = (height: Height) => ClientState
```
which asks the host to "represent itself as a `ClientState`". In the case of the sdk, this would be implemented by building the `ClientState` from the appropriate fields in the `sdk.Context`. My only concern with this abstraction is that it might confuse users, as the `ClientState` is a light client abstraction, not a host chain one. Hopefully, the example provided is enough to clear up any confusion. Again, I'm open to better ways of doing it!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,4 +40,4 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ## [Unreleased]
 
 ### Improvements
-* \#764 Move `ClientState` to the `provableStore` and add `ClientState` validation in `connOpenTry` and `connOpenAck`
+* [\#802](https://github.com/cosmos/ibc/pull/802) Move `ClientState` to the `provableStore` and add `ClientState` validation in `connOpenTry` and `connOpenAck`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,3 +38,6 @@ Ref: https://keepachangelog.com/en/1.0.0/
 # Changelog
 
 ## [Unreleased]
+
+### Improvements
+* \#764 Move `ClientState` to the `provableStore` and add `ClientState` validation in `connOpenTry` and `connOpenAck`
--- a/spec/core/ics-002-client-semantics/README.md
+++ b/spec/core/ics-002-client-semantics/README.md
@@ -915,6 +915,8 @@ Jan 13, 2020 - Revisions for client type separation & path alterations
 
 Jan 26, 2020 - Addition of query interface
 
+Jul 27, 2022 - Addition of `verifyClientState` function, and move `ClientState` to the `provableStore`
+
 ## Copyright
 
 All content herein is licensed under [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0).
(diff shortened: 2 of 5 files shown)


spec/core/ics-002-client-semantics/README.md
```diff
@@ -315,6 +315,18 @@ type verifyClientConsensusState = (
   => boolean
 ```
 
+`verifyClientState` verifies a proof of the client state of the specified client stored on the target machine.
+
+```typescript
+type verifyClientState = (
```
AdityaSripal, line 321: Rather than adding this function to the specification, we should transition to using the generic verify membership methods. Though I suppose that can be done in a separate PR

AdityaSripal (commented):
Nice work! Agreed that `ValidateSelfClient` makes sense in 24-host

AdityaSripal (approved):

AdityaSripal:
Please add a changelog entry

plafer:
@AdityaSripal Done! [5a4e0c8](https://github.com/cosmos/ibc/pull/802/commits/5a4e0c8f706701f9960032adf31ff3ef7861951e)

spec/core/ics-002-client-semantics/README.md
plafer: If you would like to open an issue describing specifically what's to be done, I'd be happy to take a stab at it in another PR

mpoke:
> @AdityaSripal Done! [5a4e0c8](https://github.com/cosmos/ibc/pull/802/commits/5a4e0c8f706701f9960032adf31ff3ef7861951e)

I think he was referring to add an entry to https://github.com/cosmos/ibc/blob/main/CHANGELOG.md

mpoke (approved):
Nice work. In the future, please open a PR per modified standard. It helps keeping track of changes.

CHANGELOG.md
```diff
@@ -38,3 +38,6 @@ Ref: https://keepachangelog.com/en/1.0.0/
 # Changelog
 
 ## [Unreleased]
+
+### Improvements
+* \#764 Move `ClientState` to the `provableStore` and add `ClientState` validation in `connOpenTry` and `connOpenAck`
```
crodriguezvega, line 43: Just a nit. We prefer to add the link to the PR.

```suggestion
* [\#802](https://github.com/cosmos/ibc/pull/802) Move `ClientState` to the `provableStore` and add `ClientState` validation in `connOpenTry` and `connOpenAck`
```

PR #806, opened 2022-07-27
ICS26: Add previous version to upgrade callbacks

AdityaSripal:


mpoke (approved):

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -86,7 +83,8 @@ function onChanUpgradeTry(
   channelIdentifier: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  counterpartyVersion: string) => (version: string, err: Error) {
+  counterpartyVersion: string,
+  previousVersion: string) => (version: string, err: Error) {
```
damiannolan, line 87: Should this be `previousCounterpartyVersion`?

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -40,12 +40,7 @@ interface ModuleUpgradeCallbacks {
 are valid and perform any custom `UpgradeInit` logic.
 It may return an error if the chosen parameters are invalid 
 in which case the upgrade handshake is aborted.
-If the provided version string is empty, `onChanUpgradeInit` should return 
-a default version string or an error if the provided version is invalid.
-Note that if the upgrade provides an empty string, this is an indication to upgrade
-to the default version which MAY be a new default from when the channel was first initiated.
-If there is no default version string for the application,
-it should return an error if provided version is empty string.
+The callback is provided both the previous version of the channel and the new proposed version. It may perform the necessary logic and state changes necessary to upgrade the channel from the previous version to the new version. If upgrading the application from the previous version to the new version is not supported, it must return an error.
```
damiannolan, line 43: So including the previous channel version is intended such that an application cannot jump versions??
For example: `ics20-1` cannot be upgraded to `ics20-3`, is that correct?

spec/core/ics-026-routing-module/UPGRADES.md
AdityaSripal: No, `counterpartyVersion` is the new version the counterparty is proposing to upgrade to

`previousVersion` is the current version of our chain before the upgrade. Maybe we should rename to current version?

spec/core/ics-026-routing-module/UPGRADES.md
AdityaSripal: No, certainly some applications could choose to make that restriction

But perhaps `ics20-1` -> ics20-3` requires state changes A

But `ics20-2` -> `ics20-3` require a different sequence of state changes B.

Knowing what version you are upgrading from gives you the information you need to decide if the upgrade is allowed and if allowed, what changes to make.

THis information could be stored in the app state itself, but we pass it in here to be explicit

spec/core/ics-026-routing-module/UPGRADES.md
damiannolan: Ah okay, thanks, sounds good to me!

damiannolan: I think `previousVersion` makes sense for naming, but maybe we could have `previousVersion` come before `counterpartyVersion` in the list of args? Would it makes more sense for the "self" to be stated before the counterparty. It doesn't make too much of a difference to me either way

damiannolan (approved):

spec/core/ics-026-routing-module/UPGRADES.md
```diff
@@ -59,7 +54,8 @@ function onChanUpgradeInit(
   channelIdentifier: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
-  version: string) => (version: string, err: Error) {
+  version: string,
+  previousVersion: string) => (version: string, err: Error) {
```
damiannolan, line 58: Thoughts on reordering these so that `previousVersion` comes before `version`?

PR #808, opened 2022-07-28
ICS4: Fix Sequence Paths

AdityaSripal:
Fix sequence paths in ICS4 to align with ICS24 and implementation


AdityaSripal:
cc: @plafer to make sure ibc-rs is using the right key format as defined in ICS24

colin-axner (approved):

mpoke (approved):

plafer:
Double checked that ibc-rs is using the proper paths, thanks for the heads up!

Issue #773, opened 2022-06-21
Update ICS-2 with the changes made by implementation in 02-client-refactor


AdityaSripal:
https://github.com/orgs/cosmos/projects/7/views/14

Issue #694, opened 2022-03-21
ICS2: Combine Header and Misbehaviour under ClientMessage interface


colin-axner:
In ibc-go, we have decided to refactor the 02-client submodule along with the interfaces defined within the `ClientState`. One of the major changes we are doing is merging the Misbehaviour, Header types under a new interface `ClientMessage`. This allows light client implementations more flexibility to create other types that allow for client updates such as batch updates. 

One motivator for this decision is described in #693 

##### references
[discussion leading to design](https://github.com/cosmos/ibc-go/issues/284)
[ibc-go issue](https://github.com/cosmos/ibc-go/issues/875)

Issue #693, opened 2022-03-21
ICS2: Split checkValidityAndUpdateState into 4 functions


colin-axner:
In ibc-go, we have decided to refactor the 02-client submodule along with the interfaces defined within the `ClientState`. One of the decisions made was to split the `checkValidityAndUpdateState` into 4 functions:

- `VerifyClientMessage`
- `CheckForMisbehaviour`
- `UpdateStateOnMisbehaviour`
- `UpdateState`

Part of the reasoning for this change was that `checkValidityAndUpdateState` began to become overloaded with functionality as it was required to account for misbehaviour detection as well, which may not be apparent to light client developers, so a more explicit API requiring misbehaviour detection seems ideal. Since the entry point to updating a IBC client now accounts for misbehaviour detection, it made sense to combine these interfaces so any sort of update (misbehaviour, normal updates, batch updates) could use the same API

##### references
[discussion leading to design](https://github.com/cosmos/ibc-go/issues/284)
[ibc-go issue](https://github.com/cosmos/ibc-go/issues/668)

Issue #689, opened 2022-03-17
ICS2: Add getTimestampAtHeight function to client state interface


colin-axner:
In [ibc-go](https://github.com/cosmos/ibc-go/issues/876#issue-1126087300) we are rewriting the `getTimestampAtHeight` functionality. Currently it is housed in ICS 003, it's functionality is to obtain the consensus state at the proof height and check its timestamp value. The problem with this is the solo machine doesn't store consensus states at each height it has had a consensus state. This is because the solo machine only maintains the latest consensus state. 

I propose the spec to do the same changes (or devise another way to get timeouts to work with the solo machine). This would be appreciated as some clients may have not so apparent requirements to fulfill this function. For example, the solo machine should return the timestamp of the latest consensus state. If the height passed in is > solomachine.Sequence + 1, it should return an error.

PR #813, opened 2022-08-04
ICS2: 02 client refactor

AdityaSripal:
Closes: #773
Closes: #689 
Closes: #693 
Closes: #694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,14 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
+### API-Breaking
+
+- [\#813](https://github.com/cosmos/ibc/pull/813) Breaks up `checkValidityAndUpdateState` into `verifyClientMessage` and `UpdateState`
+- [\#813](https://github.com/cosmos/ibc/pull/813) Breaks up `checkMisbehaviourAndUpdateState` into `checkForMisbehaviour` and `UpdateStateOnMisbehaviour`
+- [\#813](https://github.com/cosmos/ibc/pull/813) Removes `Header` and `Misbehaviour` interfaces for generic `ClientMessage` interface
+- [\#813](https://github.com/cosmos/ibc/pull/813) Removes specific verify functions from ClientState interface in exchange for generic `verifyMembership` and `verifyNonMembership` methods
+- [\#813](https://github.com/cosmos/ibc/pull/813) Adds `getTimeoutAtHeight` method to ClientState interface.
+
 ### Bug Fixes
 
 - [\#808](https://github.com/cosmos/ibc/pull/808) Fix channel sequence paths in ICS4
--- a/spec/core/ics-002-client-semantics/README.md
+++ b/spec/core/ics-002-client-semantics/README.md
@@ -152,6 +152,31 @@ State machines implementing the IBC protocol are expected to respect these clien
 
 ### Data Structures
 
+#### Height
+
+`Height` is an opaque data structure defined by a client type.
+It must form a partially ordered set & provide operations for comparison.
+
+```typescript
+type Height
+```
+
+```typescript
+enum Ord {
+  LT
+  EQ
+  GT
+}
+
+type compare = (h1: Height, h2: Height) => Ord
+```
+
+A height is either `LT` (less than), `EQ` (equal to), or `GT` (greater than) another height.
+
+`>=`, `>`, `===`, `<`, `<=` are defined through the rest of this specification as aliases to `compare`.
+
+There must also be a zero-element for a height type, referred to as `0`, which is less than all non-zero heights.
+
 #### ConsensusState
 
 `ConsensusState` is an opaque data structure defined by a client type, used by the validity predicate to
@@ -180,86 +205,6 @@ The `ConsensusState` MUST define a `getTimestamp()` method which returns the tim
 type getTimestamp = ConsensusState => uint64
 ```
 
-#### Header
-
-A `Header` is an opaque data structure defined by a client type which provides information to update a `ConsensusState`.
-Headers can be submitted to an associated client to update the stored `ConsensusState`. They likely contain a height, a proof,
-a commitment root, and possibly updates to the validity predicate.
-
-```typescript
-type Header = bytes
-```
-
-#### Validity predicate
-
-A validity predicate is an opaque function defined by a client type to verify `Header`s depending on the current `ConsensusState`.
-Using the validity predicate SHOULD be far more computationally efficient than replaying the full consensus algorithm
-for the given parent `Header` and the list of network messages.
-
-The validity predicate & client state update logic are combined into a single `checkValidityAndUpdateState` type, which is defined as
-
-```typescript
-type checkValidityAndUpdateState = (Header) => Void
-```
-
-`checkValidityAndUpdateState` MUST throw an exception if the provided header was not valid.
-
-If the provided header was valid, the client MUST also mutate internal state to store
-now-finalised consensus roots and update any necessary signature authority tracking (e.g.
-changes to the validator set) for future calls to the validity predicate.
-
-Clients MAY have time-sensitive validity predicates, such that if no header is provided for a period of time
-(e.g. an unbonding period of three weeks) it will no longer be possible to update the client.
-In this case, a permissioned entity such as a chain governance system or trusted multi-signature MAY be allowed
-to intervene to unfreeze a frozen client & provide a new correct header.
-
-#### Misbehaviour predicate
-
-A misbehaviour predicate is an opaque function defined by a client type, used to check if data
-constitutes a violation of the consensus protocol. This might be two signed headers
-with different state roots but the same height, a signed header containing invalid
-state transitions, or other proof of malfeasance as defined by the consensus algorithm.
-
-The misbehaviour predicate & client state update logic are combined into a single `checkMisbehaviourAndUpdateState` type, which is defined as
-
-```typescript
-type checkMisbehaviourAndUpdateState = (bytes) => Void
-```
-
-`checkMisbehaviourAndUpdateState` MUST throw an exception if the provided proof of misbehaviour was not valid.
-
-If misbehaviour was valid, the client MUST also mutate internal state to mark appropriate heights which
-were previously considered valid as invalid, according to the nature of the misbehaviour.
-
-Once misbehaviour is detected, clients SHOULD be frozen so that no future updates can be submitted.
-A permissioned entity such as a chain governance system or trusted multi-signature MAY be allowed
-to intervene to unfreeze a frozen client & provide a new correct header.
-
-#### Height
-
-`Height` is an opaque data structure defined by a client type.
-It must form a partially ordered set & provide operations for comparison.
-
-```typescript
-type Height
-```
-
-```typescript
-enum Ord {
-  LT
-  EQ
-  GT
-}
-
-type compare = (h1: Height, h2: Height) => Ord
-```
-
-A height is either `LT` (less than), `EQ` (equal to), or `GT` (greater than) another height.
-
-`>=`, `>`, `===`, `<`, `<=` are defined through the rest of this specification as aliases to `compare`.
-
-There must also be a zero-element for a height type, referred to as `0`, which is less than all non-zero heights.
-
 #### ClientState
 
 `ClientState` is an opaque data structure defined by a client type.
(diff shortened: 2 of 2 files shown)


AdityaSripal (commented):
This will probably be easier to review in split mode

spec/core/ics-002-client-semantics/README.md
```diff
@@ -196,95 +255,64 @@ A validity predicate is an opaque function defined by a client type to verify `H
...
+`verifyClientMessage` must be called before this function, and `checkForMisbehaviour` must return false before this function is called.
 
-There must also be a zero-element for a height type, referred to as `0`, which is less than all non-zero heights.
+The client MUST also mutate internal state to store
+now-finalised consensus roots and update any necessary signature authority tracking (e.g.
+changes to the validator set) for future calls to the validity predicate.
 
-#### ClientState
+Clients MAY have time-sensitive validity predicates, such that if no ClientMessage is provided for a period of time
+(e.g. an unbonding period of three weeks) it will no longer be possible to update the client.
```
mpoke, line 296:
```suggestion
(e.g. an unbonding period of three weeks) it will no longer be possible to update the client, i.e., the client is being frozen. 
```

spec/core/ics-002-client-semantics/README.md
```diff
@@ -196,95 +255,64 @@ A validity predicate is an opaque function defined by a client type to verify `H
...
-type ClientState = bytes
+type UpdateStateOnMisbehaviour = (ClientMessage) => Void
 ```
 
-Client types MUST define a method to initialise a client state with a provided consensus state, writing to internal state as appropriate.
-
-```typescript
-type initialise = (consensusState: ConsensusState) => ClientState
-```
+`verifyClientMessage` must be called before this function, and `checkForMisbehaviour` must return `true` before this function is called.
```
mpoke, line 308: This is confusion: Does `UpdateStateOnMisbehaviour` require both `verifyClientMessage` and `checkForMisbehaviour` to be called?

spec/core/ics-002-client-semantics/README.md
```diff
@@ -180,14 +205,48 @@ The `ConsensusState` MUST define a `getTimestamp()` method which returns the tim
 type getTimestamp = ConsensusState => uint64
 ```
 
-#### Header
```
mpoke, line 183: Removing `Header` should be made consistent throughout the entire standard (e.g., see the Update section).

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,29 +569,43 @@ to allow governance mechanisms to perform these actions
...
 function updateClient(
   id: Identifier,
-  header: Header) {
-    clientType = provableStore.get(clientTypePath(id))
-    abortTransactionUnless(clientType !== null)
-    clientState = provableStore.get(clientStatePath(id))
-    abortTransactionUnless(clientState !== null)
-    clientType.checkValidityAndUpdateState(clientState, header)
+  clientMessage: ClientMessage) {
+    if err := clientState.VerifyClientMessage(clientMessage); err != nil {
```
mpoke, line 573: How is `clientState` retrieved? 
```
clientState = provableStore.get(clientStatePath(id))
abortTransactionUnless(clientState !== null)
```

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,29 +569,43 @@ to allow governance mechanisms to perform these actions
...
+        return err
```
mpoke, line 574: `updateClient` has no return value. Shouldn't it call `abortTransactionUnless` instead?

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,29 +569,43 @@ to allow governance mechanisms to perform these actions
...
+    foundMisbehaviour := clientState.CheckForMisbehaviour(clientMessage)
+    if foundMisbehaviour {
+        clientState.UpdateStateOnMisbehaviour(header)
+        // emit misbehaviour event
+        return 
+    }
+    
+    clientState.UpdateState(clientMessage) // expects no-op on duplicate header
+    // emit update event
+    return
```
mpoke, line 586:
```suggestion
    if foundMisbehaviour {
        clientState.UpdateStateOnMisbehaviour(header)
        // emit misbehaviour event
    }
    else {    
        clientState.UpdateState(clientMessage) // expects no-op on duplicate header
        // emit update event
    }
```

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,29 +569,43 @@ to allow governance mechanisms to perform these actions
...
-If the client detects proof of misbehaviour, the client can be alerted, possibly invalidating
+If a relayer wishes to explicitly submit misbehaviour. They may alert the client to the misbehaviour directly, possibly invalidating
 previously valid state roots & preventing future updates.
 
 ```typescript
 function submitMisbehaviourToClient(
   id: Identifier,
-  misbehaviour: bytes) {
+  clientMessage: ClientMessage) {
     clientType = provableStore.get(clientTypePath(id))
```
mpoke, line 599: `clientType` not used.

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,29 +569,43 @@ to allow governance mechanisms to perform these actions
...
+    clientState.UpdateState(clientMessage) // expects no-op on duplicate header
+    // emit update event
+    return
 }
 ```
 
 #### Misbehaviour
 
-If the client detects proof of misbehaviour, the client can be alerted, possibly invalidating
+If a relayer wishes to explicitly submit misbehaviour. They may alert the client to the misbehaviour directly, possibly invalidating
```
mpoke, line 592:
```suggestion
A relayer may alert the client to the misbehaviour directly, possibly invalidating
```

spec/core/ics-002-client-semantics/README.md
mpoke: Also, `VerifyClientMessage` does not return an error.

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,29 +569,43 @@ to allow governance mechanisms to perform these actions
...
+  clientMessage: ClientMessage) {
     clientType = provableStore.get(clientTypePath(id))
     abortTransactionUnless(clientType !== null)
     clientState = provableStore.get(clientStatePath(id))
     abortTransactionUnless(clientState !== null)
-    clientType.checkMisbehaviourAndUpdateState(clientState, misbehaviour)
+    // authenticate client message
+    clientState.verifyClientMessage(clientMessage)
+    // check that client message is valid instance of misbehaviour
+    clientState.checkForMisbehaviour(clientMessage)
```
mpoke, line 606: The return value of `checkForMisbehaviour` is not checked.

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,29 +569,43 @@ to allow governance mechanisms to perform these actions
...
 ```
 
 #### Misbehaviour
 
-If the client detects proof of misbehaviour, the client can be alerted, possibly invalidating
+If a relayer wishes to explicitly submit misbehaviour. They may alert the client to the misbehaviour directly, possibly invalidating
 previously valid state roots & preventing future updates.
 
 ```typescript
 function submitMisbehaviourToClient(
```
mpoke, line 596: Why do we still need `submitMisbehaviourToClient` given the new meaning of `updateClient`?

mpoke (changes requested):

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,29 +569,43 @@ to allow governance mechanisms to perform these actions
...
     abortTransactionUnless(clientState !== null)
-    clientType.checkMisbehaviourAndUpdateState(clientState, misbehaviour)
+    // authenticate client message
+    clientState.verifyClientMessage(clientMessage)
+    // check that client message is valid instance of misbehaviour
+    clientState.checkForMisbehaviour(clientMessage)
+    // update state based on misbehaviour
+    clientState.UpdateStateOnMisbehaviour(misbehaviour)
 }
 ```
```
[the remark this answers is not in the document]
AdityaSripal: agreed

spec/core/ics-002-client-semantics/README.md
```diff
@@ -180,14 +205,48 @@ The `ConsensusState` MUST define a `getTimestamp()` method which returns the tim
...
+```typescript
+type latestClientHeight = (
+  clientState: ClientState)
+  => Height
+```
+
+Client types MUST define a method on the client state to fetch the timestamp at a given height
+
+```typescript
+type getTimestampAtHeight = (
```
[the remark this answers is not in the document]
AdityaSripal: cc: @colin-axner , this was necessary to accomodate the special case of solomachines

CHANGELOG.md
```diff
@@ -39,6 +39,14 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ## [Unreleased]
 
+### API-Breaking
+
+- [\#813](https://github.com/cosmos/ibc/pull/813) Breaks up `checkValidityAndUpdateState` into `verifyClientMessage` and `UpdateState`
+- [\#813](https://github.com/cosmos/ibc/pull/813) Breaks up `checkMisbehaviourAndUpdateState` into `checkForMisbehaviour` and `UpdateStateOnMisbehaviour`
+- [\#813](https://github.com/cosmos/ibc/pull/813) Removes `Header` and `Misbehaviour` interfaces for generic `ClientMessage` interface
```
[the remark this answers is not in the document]
AdityaSripal: https://github.com/cosmos/ibc-go/pull/1846

spec/core/ics-002-client-semantics/README.md
```diff
@@ -196,95 +255,64 @@ A validity predicate is an opaque function defined by a client type to verify `H
...
 constitutes a violation of the consensus protocol. This might be two signed headers
 with different state roots but the same height, a signed header containing invalid
 state transitions, or other proof of malfeasance as defined by the consensus algorithm.
 
-The misbehaviour predicate & client state update logic are combined into a single `checkMisbehaviourAndUpdateState` type, which is defined as
+The misbehaviour predicate is defined as
 
 ```typescript
-type checkMisbehaviourAndUpdateState = (bytes) => Void
+type checkForMisbehaviour = (ClientMessage) => bool
```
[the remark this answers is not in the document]
AdityaSripal: good point, `verifyClientMessage` is for authenticating the message is from the blockchain itself. In the tendermint case, this would be verifying the headers were signed by valid validator sets.

`checkForMisbehaviour` checks if the client message constitutes misbehaviour. This could be explicit, if the client message contains two headers check that they are conflicting.

Or implicit, if the clientmessage contains one header but it conflicts with a consensus state that already exists on chain.

`verifyHeaderAndUpdateState` used to verify the header, check for implicit misbehaviour, and then update state.

We split these up into explicit functions so light client developers did not miss implementing a critical piece.

See ADR for more details

spec/core/ics-002-client-semantics/README.md
```diff
@@ -278,148 +227,113 @@ Client types MUST define a method to initialise a client state with a provided c
...
-  proof: CommitmentProof,
-  clientIdentifier: Identifier,
-  clientState: ClientState)
-  => boolean
-```
+#### Misbehaviour predicate
 
-`verifyConnectionState` verifies a proof of the connection state of the specified connection end stored on the target state machine.
+A misbehaviour predicate is an opaque function defined by a client type, used to check if a ClientMessage
+constitutes a violation of the consensus protocol. For example, if the state machine is a blockchain; this might be two signed headers
```
mpoke, line 273: nit
```suggestion
constitutes a violation of the consensus protocol. For example, if the state machine is a blockchain, this might be two signed headers
```

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,228 +573,47 @@ to allow governance mechanisms to perform these actions
 ```typescript
 function updateClient(
   id: Identifier,
-  header: Header) {
-    clientType = provableStore.get(clientTypePath(id))
-    abortTransactionUnless(clientType !== null)
-    clientState = provableStore.get(clientStatePath(id))
-    abortTransactionUnless(clientState !== null)
-    clientType.checkValidityAndUpdateState(clientState, header)
+  clientState: ClientState,
```
mpoke, line 576: I think we should retrieve the clientState from the store here instead of passing it as an argument, i.e.,
```
clientState = provableStore.get(clientStatePath(id))
abortTransactionUnless(clientState !== null)
```

mpoke (changes requested):
Looks good. See my comments below. Also, there are three comments re. `submitMisbehaviourToClient` that were not addressed yet (i.e., https://github.com/cosmos/ibc/pull/813#discussion_r944276890, https://github.com/cosmos/ibc/pull/813#discussion_r944272990, https://github.com/cosmos/ibc/pull/813#discussion_r944276067).

mpoke:
Should "State verification" and "Query interface" sections be subsections of the "Data Structures" section?

spec/core/ics-002-client-semantics/README.md
```diff
@@ -660,228 +573,47 @@ to allow governance mechanisms to perform these actions
...
-    h2 = misbehaviour.h2
-    abortTransactionUnless(clientState.pastPublicKeys.contains(h1.publicKey))
-    abortTransactionUnless(h1.sequence === h2.sequence)
-    abortTransactionUnless(h1.commitmentRoot !== h2.commitmentRoot || h1.publicKey !== h2.publicKey)
-    abortTransactionUnless(h1.publicKey.verify(h1.signature))
-    abortTransactionUnless(h2.publicKey.verify(h2.signature))
-    clientState.frozen = true
-}
-```
+Please see the ibc-go implementations of light clients for examples of how to implement your own: https://github.com/cosmos/ibc-go/blob/main/modules/light-clients
```
mpoke, line 616: We should move this under the other "Example Implementation" section, i.e., 
```
## Forwards Compatibility

New client types can be added by IBC implementations at-will as long as they conform to this interface.

## Example Implementation

Coming soon.

## Other Implementations

Coming soon.
```

spec/core/ics-002-client-semantics/README.md
```diff
@@ -278,148 +227,113 @@ Client types MUST define a method to initialise a client state with a provided c
...
+
+#### State verification
+
+Client types must define functions to authenticate internal state of the state machine which the client tracks.
+Internal implementation details may differ (for example, a loopback client could simply read directly from the state and require no proofs).
+
+- The `delayPeriodTime` is passed to the verification functions for packet-related proofs in order to allow packets to specify a period of time which must pass after a consensus state is added before it can be used for packet-related verification.
+- The `delayPeriodBlocks` is passed to the verification functions for packet-related proofs in order to allow packets to specify a period of blocks which must pass after a consensus state is added before it can be used for packet-related verification.
+
+##### Required functions
```
mpoke, line 335: I think we can remove this line since all functions are required.

spec/core/ics-002-client-semantics/README.md
AdityaSripal: `submitMisbehaviourToClient` is for explicit misbehaviour handling. THe relayer knows the message it is sending will shutdown the client

In UpdateClient, it is possible for innocent attempts to add a new header to actually be proof of misbehaviour which we catch and handle.

In the future we could possibly deprecate `submitMisbehaviourToClient` but we won't for now to minimize disruptions to relayers and light client implementers

spec/core/ics-002-client-semantics/README.md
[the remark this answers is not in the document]
colin-axner: Clients which only have a latest consensus state would only have a latest timestamp and thus might not have a timestamp based on a consensus state. The current spec didn't work for solo machines as solo machines don't store consensus states (the consensus state is kept within the client state)

colin-axner (approved):
LGTM. Thanks for updating

PR #835, opened 2022-09-02
ICS28: Adds precision to wording for sumUnbonding definition

danwt:
A suggestion

### DIFF
--- a/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
+++ b/spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
@@ -120,7 +120,7 @@ CCV provides the following system properties.
   - `hc'` is the smallest height on `cc` that satisfies `ts(hc') >= ts(hc) + UnbondingPeriod`, i.e., `val` cannot completely unbond on `cc` before `hc'`;   
   - `hp` is the largest height on the provider chain that satisfies `hp << hc`, i.e., `Power(pc,hp,val) = Power(cc,hc,val)`, where `pc` is the provider chain;
   - `hp'` is the smallest height on the provider chain that satisfies `hc' << hp'`, i.e., `val` cannot completely unbond on the provider chain before `hp'`;
-  - `sumUnbonding(hp, h, val)` is the sum of all tokens of `val` that are unbonding on the provider at height `h` which began unbonding not earlier than at `hp + 1`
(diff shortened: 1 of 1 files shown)


mpoke:
I find the original definition more precise, i.e., 
> `sumUnbonding(hp, h, val)` is the sum of all tokens `val` start unbonding on the provider at all heights `hu`, such that `hp < hu <= h`;

It's concise and specifies exactly the unbondings of interest.

danwt:
It's not accurate though because a unbonding is no longer relevant if the validator rebonds. This case would be counted with that definition because 'sum of all tokens val start unbonding on the provider at all heights hu' would include it

spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
```diff
@@ -120,7 +120,7 @@ CCV provides the following system properties.
   - `hc'` is the smallest height on `cc` that satisfies `ts(hc') >= ts(hc) + UnbondingPeriod`, i.e., `val` cannot completely unbond on `cc` before `hc'`;   
   - `hp` is the largest height on the provider chain that satisfies `hp << hc`, i.e., `Power(pc,hp,val) = Power(cc,hc,val)`, where `pc` is the provider chain;
   - `hp'` is the smallest height on the provider chain that satisfies `hc' << hp'`, i.e., `val` cannot completely unbond on the provider chain before `hp'`;
-  - `sumUnbonding(hp, h, val)` is the sum of all tokens `val` start unbonding on the provider at all heights `hu`, such that `hp < hu <= h`;
+  - `sumUnbonding(hp, h, val)` is the sum of all tokens of `val` that are unbonding on the provider at height `h` which began unbonding not earlier than at `hp + 1`
```
mpoke, line 123:
```suggestion
  - `sumUnbonding(hp, h, val)` is the sum of all tokens of `val` that start unbonding on the provider at all heights `hu` and are still unbonding at height `h`, such that `hp < hu <= h`
```

mpoke:
> It's not accurate though because a unbonding is no longer relevant if the validator rebonds. This case would be counted with that definition because 'sum of all tokens val start unbonding on the provider at all heights hu' would include it

That's true. Nice catch.

spec/app/ics-028-cross-chain-validation/system_model_and_properties.md
danwt: Nice :)

mpoke (approved):

Issue #584, opened 2021-06-27
ICS20 (Revision): Add memo string to FungibleTokenPacketData


thorchain-admin:
---
ics: 20
title: Memo String for FungibleTokenPacketData
stage: Proposed
category: IBC/APP
kind: instantiation
author: devs@thorchain.org
created: July 2021
requires: [20](https://github.com/cosmos/ibc/blob/master/spec/app/ics-020-fungible-token-transfer/README.md)
---

## Synopsis

Allow users to add transactional intent to FungibleTokenPacketData in the form of a 180-bytes string "memo".

### Motivation

It's not sufficient to just transfer tokens between users on different chains. Increasingly, cross-chain composability will require modules to begin transferring tokens between each other, and they will need the ability to pass transactional intent (ie, call-data). 

THORChain is an example of this. Users must be able to use IBC to send assets *as well* as state what to do with them. THORChain currently [uses on-chain memos](https://docs.thorchain.org/developers/transaction-memos) to enable cross-chain composability (ie, to do swaps between Bitcoin, Ethereum and more). Nomic also indicate that [on-chain commitments](https://gist.github.com/mappum/da11e37f4e90891642a52621594d03f6#simple-op_return-commitment-scheme) (memos) is a solution for their pegged Bitcoin implementation. 

Various work-arounds include sending to separate modules, or concatenating the memo on the end of the destination address string, but this is very unwieldy. 

The best solution is to simply modify ICS20 to add an optional memo string parameter to the packet data. 

## Technical Specification

Simply modify the interface to add a memo string. 

```
interface FungibleTokenPacketData {
  denomination: string
  amount: uint256
  sender: string
  receiver: string
  memo:string
}
```

It can be limited to 180-bytes. 

## Backwards Compatibility

Since it is optional, clients that don't implement the memo can simply drop it. It may not be necessary to add a new version (ie `ics20-2`

## Forwards Compatibility

180-bytes memo will allow the establishment of cross-chain call data. For example, THORChain use this standard:

```
function:param1:param2:param3:param4
```

The receiving module parses the string and separates the parameters. It then delegates it to the appropriate function handler to execute. 

## Example Implementation

```
interface FungibleTokenPacketData {
  denomination: atom
  amount: 100000000
  sender: cosmos1tf38ulmsrpe3x0r4glqc68pstufv8p6d5jqcs2tf
  receiver: thor1pstufv8p6d5jqcs2tf38ulmsrpe3x0r4glqc68
[...]

PR #842, opened 2022-09-29
Add metadata field to transfer packet data

AdityaSripal:
Supports adding metadata field to the transfer packet data in a backwards compatible way

closes: https://github.com/cosmos/ibc/issues/584
ref: https://github.com/cosmos/ibc/discussions/727


spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -42,9 +42,12 @@ interface FungibleTokenPacketData {
   amount: uint256
   sender: string
   receiver: string
+  metadata: bytes
```
colin-axner, line 45: I think it'd be nice to have some reasoning for using bytes instead of string given that I think historically string is used over bytes. This is the only difference I can find between the two:

> As described, a string must use UTF-8 character encoding. A string cannot exceed 2GB.

> As described, bytes can store custom data types, up to 2GB in size.

[source](https://developers.google.com/protocol-buffers/docs/encoding)

Do we anticipate any not UTF-8 character encodings to cause problems? I think we already account for this in ibc-go for [Any's](https://github.com/cosmos/ibc-go/blob/7dcfc2c8e8f3ac35ee2590d6c9e0b2776a46f78a/modules/core/02-client/keeper/events.go#L36) and [packet data](https://github.com/cosmos/ibc-go/blob/7dcfc2c8e8f3ac35ee2590d6c9e0b2776a46f78a/modules/core/04-channel/keeper/events.go#L135) - but probably worth noting for other implementations as this is a subtle issue

colin-axner (approved):
Concept ACK, should there be any instruction/suggestion on what the field should be used for (or perhaps not used for)?

I guess there is no length validation on this field as well, which seems okay to me

spec/app/ics-020-fungible-token-transfer/README.md
AdityaSripal: I don't think so because the metadata is completely used outside of transfer application, any middleware or external user that uses metadata and expects UTF-8 encoding can enforce it on their end.

Otoh, since we are allowing arbitrary data to be parsed on the level above transfer, I don't want to make arbitrary limitations in the transfer level and block a potential usecase

crodriguezvega: In interchain accounts we have a [string `memo` in the `InterchainAccountPacketData`](https://github.com/cosmos/ibc/blob/main/spec/app/ics-027-interchain-accounts/README.md#packet-data). Would it make sense to rename it also to `metadata` and make it `bytes`?

colin-axner:
@dtribble mentioned on the IBC call that this is likely something that might be useful for non transfer applications. Has any thought been given to a generalized approach?

ethanfrey:
So, there is a new field with no semantics attached to its use?

This will quickly lead to many incompatible implementations unless some guidance is given.

ethanfrey:
BTW "Supports adding metadata field to the transfer packet data in a backwards compatible way" is a bit of a misnomer, since if you send the new packet to an old implementation, it will have a parse error with "unknown field 'metadata'". I guess it means old versions could send packets to the new version, just be interpreted as an empty metadata field.

I have tried to add optional fields to ics20 on my contract implementation and had it refused by the standard transfer module for this very reason

ethanfrey:
> can we start with documenting the use cases of the initial users of this addition?

My understanding is they wanted to embed ICA inside ICS20 or something like that... which definitely needs a spec if you want to have compatible implementations

Issue #823, opened 2022-08-23
ICS 031: Extending cross-chain queries to support IBC Casting


angbrav:
This issue sketches how to extend [cross-chain queries](https://github.com/cosmos/ibc/tree/main/spec/app/ics-031-crosschain-queries) to support [IBC casting](https://github.com/cosmos/ibc/discussions/759).

A few definitions for context:
- Querying chain: The chain that is interested in getting data from another chain (queried chain). The querying chain is the chain that implements the Cross-chain Queries module.
- Queried chain: The chain whose state is being queried. 

Goal: Maintain the nice properties of cross-chain queries, such as permissionless and minimal queried chain work.

Extensions:
- Currently, queries are `ONE-SHOT`. We need a new type of query: `PERSISTENT` queries. 
- A querying chain can issue persistent queries only on a special key. The set of special keys can be modified by transactions at the queried chain.
- When a querying chain issues a persistent query, it emits an event. This is the same for one-shot queries.
- Every time the queried chain updates a special key, it emits an event. 
- Between a querying and a queried chain, relayers keep track of the persistent queries issued by the querying chain. Thus, every time the queried chain emits an event for a key for which the querying chain has issued a persistent query, the relayer submits a query result transaction to the querying chain.
- When the querying chain receives an update for a key `k`, it should be able to verify that it is not missing any update in between. Assume that the queried chain updates `k` to `v'`  at height `h'`. Consider also that `k` was previously updated at `h` to value `v`. To enable such verification:
	- The queried chain stores the latest value for a given key, together with the previous value and height at which it was updated. Thus, it stores under a public path that `k`'s current value is `v'` and that its previous value was `v` which was set at `h`.
	- Assume that the querying chain receives an update for `k` and that the previous update it received was `v''` at `h''`. The querying chain then verifies using its local client of the queried chain that `v''=v` and `h''=h'`. 
	- If the verification passes, then the querying chain accepts the update and updates its private store accordingly.

AdityaSripal:
> together with the previous value and height at which it was updated

Does it need to store both? Or only the previous height of last update. Seems like checking the previous value is unnecessary, unless we want to prevent sending packets if the publisher updates key with exact same value multiple times

angbrav:
> > together with the previous value and height at which it was updated
> 
> Does it need to store both? Or only the previous height of last update. Seems like checking the previous value is unnecessary, unless we want to prevent sending packets if the publisher updates key with exact same value multiple times

[...]

PR #843, opened 2022-09-30
Extend cross-chain queries to support IBC pub/sub

angbrav:
This PR extends cross-chain queries to support IBC pub/sub (#759). The extensions are described in https://github.com/cosmos/ibc/issues/823.

### DIFF
--- a/spec/app/ics-031-crosschain-queries/README.md
+++ b/spec/app/ics-031-crosschain-queries/README.md
@@ -26,6 +26,10 @@ We expect on-chain applications to depend on reads from other chains, e.g., a pa
 
 `Queried chain`: The chain whose state is being queried. The queried chain gets queried via a relayer utilizing its RPC client which is then submitted back to the querying chain.
 
+`Oneshot query`: A oneshot query is short-living query. The query's caller at the querying chain issues a oneshot query when it is interested in querying the state of the queried chain at a particular height. Once the caller retrieves the result, the query's lifecycle ends.
+
+`Persistent query`: A persistent query is a long-living query. The query's caller at the querying chain issues a persistent query when it is interested in receiving an update every time the queried state is updated at the queried chain.
+
 `Cross-chain Queries Module`: The module that implements the cross-chain querying protocol. Only the querying chain integrates it.
 
 `Height` and client-related functions are as defined in ICS 2.
@@ -42,7 +46,7 @@ We expect on-chain applications to depend on reads from other chains, e.g., a pa
 
 ### Assumptions
 
-- **Safe chains:** Both the queryin and queried chains are safe. This means that, for every chain, the underlying consensus engine satisfies safety (e.g., the chain does not fork) and the execution of the state machine follows the described protocol.
+- **Safe chains:** Both the querying and queried chains are safe. This means that, for every chain, the underlying consensus engine satisfies safety (e.g., the chain does not fork) and the execution of the state machine follows the described protocol.
 
 - **Live chains:** Both the querying and queried chains MUST be live, i.e., new blocks are eventually added to the chain.
 
@@ -69,9 +73,13 @@ specification requires additional assumptions: both the querying chain and at le
 
 The querying chain can query a chain without permission from the latter and implement cross-chain querying without any approval from a third party or chain governance. Note that since there is no prior negotiation between chains, the querying chain cannot assume that queried data will be in an expected format.
 
-#### Minimal queried chain Work
+#### Minimal queried chain work
 
 Any chain that provides query support can act as a queried chain, requiring no implementation work or any extra module. This is possible by utilizing an RPC client on a relayer.
+
+#### Persistent query updates are applied in order at the querying chain
+
+When a querying chain receives an persistent query update, it should be able to verify that it is not missing any update in between the received one and the last update it accepted.
 
 #### Modular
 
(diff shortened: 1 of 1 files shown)


spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -299,14 +359,27 @@ function CrossChainQueryResponse(
     // Delete the query from the local, private store.
     privateStore.delete(queryPath(queryId))
 
+    // Reassign query's heightUpdate and value in case the query is of type PERSISTENT and the
+    // result was SUCCESS
+    heightUpdate = null
+    value = data
+    if (query.type === PERSISTENT && data !== null) {
```
AdityaSripal, line 366: We also need to check that the previous value and height equal the previous resultRecord

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -133,18 +154,24 @@ enum QueryResult {
...
+    udpateHeight: Height
     data: []byte
 }
 ```
 
 - The `id` field uniquely identifies the query at the querying chain.
+- The `type` field specifies the type of query.
+- The `path` field is the path queried at the queried chain.
 - The `result` field indicates whether the query was correctly executed at the queried chain and if the queried path exists.
+- The `udpateHeight` field indicates the height at which the queried chain updated the queried path with `data`.
```
AdityaSripal, line 174:
```suggestion
- The `updateHeight` field indicates the height at which the queried chain updated the queried path with `data`.
```

AdityaSripal (changes requested):

ethanfrey:
Maybe a bit out of scope, but I think important: how do we account for fees.

ICQ had some reward paid to the relayer IIRC. ICS29 allows for paying relayer fees on the packets we send.

If I make a subscription, and 1000s of packets come over, how do we account for fees? All free? Deduct from the subscriber for each contract? Have some "account" we need to fill up to pay for subscriptions and they stop when not paid???

I think we should consider relayer incentives as first class citizens in all new protocols we design.

AdityaSripal:
I think the benefit of client queries is the efficiency. No need for packet roundtrips for each query. No need for having to send a packet to each individual querier. This makes a lot of sense for high demand queries (e.g. price feeds, oracle data, CNS, etc.)

However, client queries does introduce a brand new primitive that relayers and applications will have to adopt (building on top of the client layer instead of channels). We need to also rebuild certain concepts like relayer incentives on top as @ethanfrey mentioned.

I think with the right abstractions we can get the best of both worlds. Ideally we get the efficiency of client queries while still offering the same channel interface that relayers and applications are already used to. We then also get the higher-level primitives for free (ICS-29, any desired middleware).

My suggestion is to introduce a new channel type: `Broadcast channel`

It is effectively a one-to-many unidirectional channel.

This will allow a single write of `SendPacket` on a "publishing" chain to be received on an arbitrary number of subscribers.

It also preserves the same semantics that relayers and applications are already used to (though we get rid of acknowledgments and timeouts, the handshake will also be simplified)

Broadcast channels i think will also be a more broadly useful primitive beyond ibc queries

Will discuss the following with @angbrav  next week

ethanfrey:
> My suggestion is to introduce a new channel type: Broadcast channel
>
> It is effectively a one-to-many unidirectional channel.
>
> This will allow a single write of SendPacket on a "publishing" chain to be received on an arbitrary number of subscribers.
>
> It also preserves the same semantics that relayers and applications are already used to (though we get rid of acknowledgments and timeouts, the handshake will also be simplified)
>
> Broadcast channels i think will also be a more broadly useful primitive beyond ibc queries

I agree with the direction of making this more generally useful primitive.

I am not sure we can hide it under the channel interface (which assumes a handshake... can we have 5 different handshakes on different connections to one channel?) and the fee middleware would be hard to extend as well.

I do agree it needs to be as generalised and reusable as the channel model to adopt into core ibc, allowing many different applications to be built using pub-sub, which we haven't yet envisioned, and showing how to add middleware-like functionality, routing etc, to pubsub.

Whether this is a new channel type. Or a new primitive on top of Connection that is the same level as channel (between Connection and app level) is an open question to me. But I agree strongly with the general direction you are pushing for.

spec/app/ics-031-crosschain-queries/README.md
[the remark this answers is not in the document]
angbrav: That's correct.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -316,6 +389,52 @@ function CrossChainQueryResponse(
...
+                                                  proof,
+                                                  resultRecord.path,
+                                                  data))
+    
+    resultRecord.updateHeight = linkedlistValue.height
+    resultRecord.data = linkedlistValue.value
+
+    // Store the result in the local, private store.
+    privateStore.set(resultQueryPath(queryIdentifier), resultRecord)
+
```
crodriguezvega, line 428: Remove empty line

```suggestion
```

crodriguezvega (commented):
Thank you, @angbrav!

I left a bunch of comments for typos or consistency fixes.

Besides that, just mentioning as suggestion maybe the following change:

```diff
-function resultQueryPath(id: Identifier): Path {
+function queryResultPath(id: Identifier): Path {
-   return "queriesresult/{id}"
+   return "queries/{id}/result"
}
```

And I have a couple of questions about the linked list:
- If a querying chain wants to subscribe to a stream from a queried chain, the queried chain must first create this linked list on its side, right?
- If the linked list needs to be created on the queried chain, does that mean that a value that would be queried with a one shot query or streamed needs to be stored twice in the chain? I mean: it needs to be stored in the regular KV store of the chain and besides it also needs to be stored in the linked list to keep track of the previous values it had. Is that correct?

angbrav:
Thank you all for the comments. I've talked to @AdityaSripal and we will work on making this more generally useful primitive.

crodriguezvega:
Agreed with @angbrav to close this PR since the design will take the alternative direction described in #889.

Issue #603, opened 2021-09-16
ICS27: Interchain Accounts allows packets to be received on controller side


colin-axner:
Giving the spec a quick glance, I don't see a check enforcing that the receiving port address is the host chain port. This means a malicious ics27 module (host chain) could trick the controller chain into accidentally trying to execute a transaction. I'm not sure if it is exploitable in any way, but there should probably be a check enforcing that controller chain ports cannot receive packets

crodriguezvega:
> there should probably be a check enforcing that controller chain ports cannot receive packets

@colin-axner, is then the action for this issue to update the spec to include the same kind of logic that ibc-go has on [the `OnRecvPacket` callback of the controller chain](https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-interchain-accounts/controller/ibc_middleware.go#L135-L145)?

colin-axner:
Yes I think so. Will defer to @AdityaSripal on how this should be specified/indicated in the spec

PR #845, opened 2022-10-03
ICS27: add explanation about expected behavior for controller modules `onRecvPacket` callback

crodriguezvega:
closes: #603

### DIFF
--- a/spec/app/ics-027-interchain-accounts/README.md
+++ b/spec/app/ics-027-interchain-accounts/README.md
@@ -528,7 +528,7 @@ function onChanCloseConfirm(
 
 ```typescript
 // Called on Host Chain by Relayer
-function OnRecvPacket(packet Packet) {
+function onRecvPacket(packet Packet) {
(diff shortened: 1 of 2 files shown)


spec/app/ics-027-interchain-accounts/README.md
```diff
@@ -574,6 +576,8 @@ function onTimeoutPacket(packet: Packet) {
 }
 ```
 
+Note that interchain accounts controller modules should not execute any logic upon packet receipt, i.e. the `OnRecvPacket` callback should not be called, and in case it is called, it should simply return an error acknowledgement.
```
mpoke, line 579: I think it would be clearer to just add the OnRecvPacket pseudocode with the error acknowledgement returned.

mpoke (changes requested):

colin-axner (approved):
Thanks @crodriguezvega

mpoke (approved):
LGTM

PR #862, opened 2022-10-14
fix: change `verifyClientMessage` return type to bool

### LINKED ISSUE #837: ICS2: `ClientState.verifyClientMessage` should return a boolean
Is there a reason why `ClientState.verifyClientMessage()` throws an exception as opposed to returning a boolean? I would have expected it to return a boolean and be called inside a `abortTransactionUnless()` in the handlers. For me, it interrupts the flow to remember that this function works differently from the other verification functions in the spec.

Also as a nit, ICS-2 mentions that `verifyClientMessage()` is a validity **predicate** (function which returns a boolean), which is inconsistent with the current signature.

crodriguezvega:
closes: #837

### DIFF
--- a/spec/client/ics-007-tendermint-client/README.md
+++ b/spec/client/ics-007-tendermint-client/README.md
@@ -206,13 +206,12 @@ Tendermint client validity checking uses the bisection algorithm described in th
 
 ```typescript
 function verifyClientMessage(
-  clientMsg: ClientMessage) {
+  clientMsg: ClientMessage) => bool {
     switch typeof(clientMsg) {
       case Header:
-        verifyHeader(clientMsg)
+        return verifyHeader(clientMsg)
       case Misbehaviour:
-        verifyHeader(clientMsg.h1)
-        verifyHeader(clientMsg.h2)
+        return verifyHeader(clientMsg.h1) && verifyHeader(clientMsg.h2)
     }
 }
 ```
@@ -220,30 +219,40 @@ function verifyClientMessage(
 Verify validity of regular update to the Tendermint client
 
 ```typescript
-function verifyHeader(header: Header) {
+function verifyHeader(header: Header) => bool {
     clientState = get("clients/{header.identifier}/clientState")
-    // assert trusting period has not yet passed
-    assert(currentTimestamp() - clientState.latestTimestamp < clientState.trustingPeriod)
-    // assert header timestamp is less than trust period in the future. This should be resolved with an intermediate header.
-    assert(header.timestamp - clientState.latestTimeStamp < trustingPeriod)
+    // trusting period must have not yet passed
+    if clientState.latestTimestamp + clientState.trustingPeriod > currentTimestamp() {
+      return false
+    }
+    // header timestamp must be less than or equal to the trusting period in the future. This should be resolved with an intermediate header.
+    if clientState.latestTimeStamp + clientState.trustingPeriod > header.timestamp {
+      return false
+    }
     // trusted height revision must be the same as header revision
     // if revisions are different, use upgrade client instead
-    // trusted height must be less than header height
-    assert(header.GetHeight().revisionNumber == header.trustedHeight.revisionNumber)
-    assert(header.GetHeight().revisionHeight > header.trustedHeight.revisionHeight)
+    if header.GetHeight().revisionNumber != header.trustedHeight.revisionNumber {
+      return false
+    }
+    // trusted height must be less than or equal to header height
+    if header.GetHeight().revisionHeight <= header.trustedHeight.revisionHeight {
+      return false
+    }
     // fetch the consensus state at the trusted height
     consensusState = get("clients/{header.identifier}/consensusStates/{header.trustedHeight}")
-    // assert that header's trusted validator set hashes to consensus state's validator hash
-    assert(hash(header.trustedValidatorSet) == consensusState.nextValidatorsHash)
+    // header's trusted validator set must hash to consensus state's validator hash
+    if header.trustedValidatorSet) != consensusState.nextValidatorsHash {
+      return false
+    }
 
     // call the tendermint client's `verify` function
-    assert(tmClient.verify(
+    return tmClient.verify(
       header.trustedValidatorSet,
       clientState.latestHeight,
       clientState.trustingPeriod,
       clientState.maxClockDrift,
       header.TendermintSignedHeader,
-    ))
+    )
 }
 ```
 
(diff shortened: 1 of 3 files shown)


spec/client/ics-007-tendermint-client/README.md
```diff
@@ -206,44 +206,53 @@ Tendermint client validity checking uses the bisection algorithm described in th
...
 ```typescript
-function verifyHeader(header: Header) {
+function verifyHeader(header: Header) => bool {
     clientState = get("clients/{header.identifier}/clientState")
-    // assert trusting period has not yet passed
-    assert(currentTimestamp() - clientState.latestTimestamp < clientState.trustingPeriod)
-    // assert header timestamp is less than trust period in the future. This should be resolved with an intermediate header.
-    assert(header.timestamp - clientState.latestTimeStamp < trustingPeriod)
+    // trusting period must have not yet passed
+    if clientState.latestTimestamp + clientState.trustingPeriod > currentTimestamp() {
```
crodriguezvega, line 225: I wrote the conditions for the trusting period checks in a way that was easier for me to wrap my head around, but I would appreciate someone double checking they are fine.

spec/client/ics-007-tendermint-client/README.md
```diff
@@ -206,44 +206,53 @@ Tendermint client validity checking uses the bisection algorithm described in th
...
+      return false
+    }
+    // header timestamp must be less than or equal to the trusting period in the future. This should be resolved with an intermediate header.
+    if clientState.latestTimeStamp + clientState.trustingPeriod > header.timestamp {
```
mpoke, line 229: Cannot find this check in the implementation.

spec/client/ics-007-tendermint-client/README.md
```diff
@@ -206,44 +206,53 @@ Tendermint client validity checking uses the bisection algorithm described in th
...
+    // trusted height must be less than or equal to header height
+    if header.GetHeight().revisionHeight <= header.trustedHeight.revisionHeight {
+      return false
+    }
     // fetch the consensus state at the trusted height
     consensusState = get("clients/{header.identifier}/consensusStates/{header.trustedHeight}")
-    // assert that header's trusted validator set hashes to consensus state's validator hash
-    assert(hash(header.trustedValidatorSet) == consensusState.nextValidatorsHash)
+    // header's trusted validator set must hash to consensus state's validator hash
+    if header.trustedValidatorSet) != consensusState.nextValidatorsHash {
```
mpoke, line 244:
```suggestion
    if header.trustedValidatorSet != consensusState.nextValidatorsHash {
```

mpoke (changes requested):

spec/core/ics-002-client-semantics/README.md
```diff
@@ -262,7 +262,7 @@ for the given parent `ClientMessage` and the list of network messages.
 The validity predicate is defined as:
 
 ```typescript
-type VerifyClientMessage = (ClientMessage) => Void
+type VerifyClientMessage = (ClientMessage) => bool
 ```
 
 `VerifyClientMessage` MUST throw an exception if the provided ClientMessage was not valid.
```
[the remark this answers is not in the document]
crodriguezvega: I was also not completely sure about this line, but I left it like this because I saw that in line 283 we currently have `checkForMisbehaviour MUST throw an exception if the provided proof of misbehaviour was not valid.` and `checkForMisbehaviour` is also a validity predicate... Maybe both statements are inconsistent then?

spec/client/ics-007-tendermint-client/README.md
crodriguezvega: Good point. I also cannot find it...

spec/core/ics-002-client-semantics/README.md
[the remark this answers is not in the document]
mpoke: @AdityaSripal may have a better understanding of this.

spec/client/ics-007-tendermint-client/README.md
AdityaSripal: So the tendermint light client verification function actually does the following check:

`clientState.latestTimestamp + clientState.trustingPeriod <= header.Timestamp + clientState.MaxClockDrift`

By the following checks:

https://github.com/tendermint/tendermint/blob/af2981a2f75500bc02412bad590fd26228cf5bc3/light/verifier.go#L207

https://github.com/tendermint/tendermint/blob/af2981a2f75500bc02412bad590fd26228cf5bc3/light/verifier.go#L176

crodriguezvega: The check 

`currentTimestamp() >= clientState.latestTimestamp + clientState.trustingPeriod`

is done [here](https://github.com/tendermint/tendermint/blob/af2981a2f75500bc02412bad590fd26228cf5bc3/light/verifier.go#L46).

But looks like this check

`header.timestamp >= clientState.latestTimeStamp + clientState.trustingPeriod`

is not in the implementation.

And the implementation has these two checks, that we don't seem to have in the spec:

https://github.com/tendermint/tendermint/blob/af2981a2f75500bc02412bad590fd26228cf5bc3/light/verifier.go#L170 (which I think would translate to `header.timestamp <= clientState.latestTimestamp`)

https://github.com/tendermint/tendermint/blob/af2981a2f75500bc02412bad590fd26228cf5bc3/light/verifier.go#L176 (which I think would translate to `header.timestamp >= currentTimestamp() + clientState.maxClockDrift`)

AdityaSripal: > https://github.com/tendermint/tendermint/blob/af2981a2f75500bc02412bad590fd26228cf5bc3/light/verifier.go#L170 (which I think would translate to header.timestamp <= clientState.latestTimestamp)

No because the trusted consensus state is not necessarily the latest one.

My point on the two links above.

The first checks the following:

`now < trustedConsState.timestamp + trustingPeriod`

And the second checks:

`now > header.timestamp + maxClockDrift`

Putting these together yields:

`header.timeStamp + maxClockDrift < trustedConsensusState.timestamp + clientState.trustingPeriod`

The highest trusted consensus state is the client state's latest timestamp, so we can replace this in the future update case with:

`header.timeStamp + maxClockDrift < clientState.latesttimestamp + clientState.TrustingPeriod`

Which would require the check:

`header.timestamp >= consensusState.timestamp  + clientState.TrustingPeriod - clientState.maxClockDrift`

So i would recommend replacing the line with the above conditional check

crodriguezvega:
I will close this for now. After discussing with @AdityaSripal he thinks that it's better if we don't return a boolean in `verifyClientMessage`.

PR #863, opened 2022-10-17
Create Atomic Swap and Interchain Swap

liangping:
Here's two ics draft which defined how to implement atomic swap and AMM swap based on IBC protocol.


crodriguezvega:
Thanks for submitting these specs, @liangping! We will plan the review process and try to give some initial feedback in the coming weeks. We will try to keep you updated on the timelines that you can expect from us.

In the meantime, would it be possible for you to actually create two PRs, one for each spec? That way we can decouple one spec from another (in case one of them is ready to be merge before the other).

liangping:
@crodriguezvega Ok, let me close this one, will add new ones

nullmames:
Closed with reference to the following new PR's
ICS101 Interchain Swap - https://github.com/cosmos/ibc/pull/866
ICS100 Atomic Swap - https://github.com/cosmos/ibc/pull/867

Issue #536, opened 2021-02-09
ICS4: question on send packet with channel not opened


ancazamfir:
## Summary
The send packet handler checks if a channel is not closed (*), however I believe it will only be successful if the channel is in open state. Trying to send a fungible token transfer packet over a channel in init state returns this error:
```
log=Log(\"failed to execute message; message index: 0: packet failed basic validation: invalid destination channel ID: identifier cannot be blank: invalid identifier\")"
```
With the latest specs/ code the destination channel identifier is only known once in open state.

(*) same in the ics02 pseudocode `function sendPacket(packet: Packet) {..` but there is also this comment before "Checks that the channel & connection are open to send packets".
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

cwgoes:
cc @AdityaSripal has this been fixed since?

cwgoes:
The spec comment is clearly wrong and should change. If optimistic sends are desired, the `ibc-go` implementation will need to change as well. cc @dtribble is Agoric testing this?

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #865, opened 2022-10-19
fix: update comment 

crodriguezvega:
closes: #536 

In #536 there is a broader discussion about optimistic sends but I think the only actionable item at the moment in that issue is to update this comment in ICS04. I propose we open a different issue to discuss how to fix optimistic sends in general.

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -533,7 +533,7 @@ Calling modules MUST execute application logic atomically in conjunction with ca
 
 The IBC handler performs the following steps in order:
 
-- Checks that the channel in not closed to send packets
+- Checks that the channel is not closed to send packets
 - Checks that the calling module owns the sending port (see [ICS 5](../ics-005-port-allocation))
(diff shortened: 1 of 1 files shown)


spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -551,7 +551,7 @@ function sendPacket(
   data: bytes) {
     channel = provableStore.get(channelPath(sourcePort, sourceChannel))
 
-    // check that the channel & connection are open to send packets; 
+    // check that the channel is not closed to send packets; 
     // note: optimistic sends are permitted once the handshake has started
```
mpoke, line 555: IMO, this comment is also misleading. If the handshake only did INIT, then optimistic sends do not work.

mpoke (approved):
Thanks @crodriguezvega. Only one comment.

AdityaSripal (approved):

spec/core/ics-004-channel-and-packet-semantics/README.md
crodriguezvega: I removed the comment.

PR #866, opened 2022-10-19
ICS101 Interchain Swap

liangping:
ICS-101 Interchain Swaps enables chains their own token pricing mechanism and exchange protocol via IBC transactions.  Each chain can thus play a role in a fully decentralised exchange network.

if you have any opinions, be free to contact with @liangping @egunawan85


liangping:
We are going to close this one and resubmit a new one later

PR #868, opened 2022-10-24
Remove deprecated comment from ICS4 and ICS26 READMEs

### LINKED ISSUE #859: ICS4: `chanOpenTry`'s `version` field should not be marked deprecated
The `version` argument to `chanOpenTry()` is currently marked deprecated, but then is used in the implementation. Therefore it is not deprecated.

I believe the original intention was to signal that the `version` field in the datagram `MsgChanOpenTry` is deprecated. However, the `version` argument in `chanOpenTry()` is not coming from the datagram: `handleChanOpenTry()` [in ICS26](https://github.com/cosmos/ibc/blob/main/spec/core/ics-026-routing-module/README.md#channel-lifecycle-management) properly sends the `version` field coming from `onChanOpenTry`, as opposed to coming from the datagram.

Just a nit, but still confused me.

rrrliu:
Fixes #859

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -334,7 +334,7 @@ function chanOpenTry(
   counterpartyChosenChannelIdentifer: Identifier,
   counterpartyPortIdentifier: Identifier,
   counterpartyChannelIdentifier: Identifier,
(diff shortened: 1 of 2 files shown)


AdityaSripal (approved):
Thanks @rrrliu !

mpoke (approved):

PR #871, opened 2022-10-26
ICS27: remove counterparty port check on host

AdityaSripal:


colin-axner (approved):

crodriguezvega:
Should we have added a changelog entry for this change?

PR #873, opened 2022-10-28
ICS-3: update `Version` field

### LINKED ISSUE #809: Connection `version` still defined with `string`: inconsistent with ibc-go
The connection version in the standard is inconsistent with what's being used in ibc-go. Specifically, in the standard, a version is an opaque `string`, whereas in ibc-go, it is as follows:

```proto
message Version {
  option (gogoproto.goproto_getters) = false;

  // unique version identifier
  string identifier = 1;
  // list of features compatible with the specified identifier
  repeated string features = 2;
}
```

My understanding is that the standard fell behind and should be updated to reflect what's in ibc-go.

Additionally, the values for currently standard identifier/features should be added to the standard:

+ identifier: `"1"` to specify IBC 1.0.0
+ features: `"ORDER_UNORDERED"` and `"ORDER_ORDERED"` to specify unordered and ordered channels, respectively.

The signature of helper functions `getCompatibleVersions` and `pickVersion` should be adjusted accordingly.

I'm happy to open a PR with the fix if we agree that the standard should indeed be fixed as described.

plafer:
Closes #809

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 
 ### API-Breaking
 
+- [\#873](https://github.com/cosmos/ibc/pull/873) Defines the connection `Version` datatype and adjusts the signature of `getCompatibleVersions()` and `pickVersion()` accordingly
 - [\#813](https://github.com/cosmos/ibc/pull/813) Breaks up `checkValidityAndUpdateState` into `verifyClientMessage` and `UpdateState`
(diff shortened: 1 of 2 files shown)


plafer:
The CI claims that https://blog.cosmos.network/eli5-what-is-ibc-a212f518715f is dead, but it works when I try it.

AdityaSripal (approved):
Thanks @plafer !

mpoke (approved):
Nice work @plafer.

PR #877, opened 2022-11-08
ICS20: Replace Metadata with Memo

AdityaSripal:
Continuation of #842 to bring it in line with ibc-go implementation.

See comment here: https://github.com/cosmos/ibc/pull/842#issuecomment-1270600590

We have changed the new field to be `memo string`. This is to encourage non-semantic use for now. Of course, there are users intending to make use of the memo in the state machine (e.g. osmosis). We don't want to necessarily enforce a particular way of using the memo statefully at the moment, but will monitor the usecases and provide guidance as necessary.

The ultimate solution will be to generalize this beyond transfer packets so that we can pass data into higher-level middleware from the packet in a standardized way. This is intended to serve as a stopgap solution in the meantime, and when the final design is out, it can be downgraded to merely being used for off-chain usecases


mpoke (approved):
LGTM. Could you please update the Changelog.

colin-axner (approved):
Thanks for updating the spec!

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -42,13 +42,13 @@ interface FungibleTokenPacketData {
   amount: uint256
   sender: string
   receiver: string
-  metadata: bytes
+  memo: string
 }
 ```
 
-Note: Since earlier versions of this specification did not include a `metadata` field, implementations must ensure that the new packet data is still compatible with chains that expect the old packet data. A legacy implementation MUST be able to unmarshal a new packet data with a `nil` metadata into the legacy `FungibleTokenPacketData` struct. Similarly, an implementation supporting `metadata` must be able to unmarshal a legacy packet data into the current struct with the metadata field set to `nil`.
+Note: Since earlier versions of this specification did not include a `memo` field, implementations must ensure that the new packet data is still compatible with chains that expect the old packet data. A legacy implementation MUST be able to unmarshal a new packet data with an empty string memo into the legacy `FungibleTokenPacketData` struct. Similarly, an implementation supporting `memo` must be able to unmarshal a legacy packet data into the current struct with the memo field set to the empty string.
```
crodriguezvega, line 49:
```suggestion
Note: Since earlier versions of this specification did not include a `memo` field, implementations must ensure that the new packet data is still compatible with chains that expect the old packet data. A legacy implementation MUST be able to unmarshal a new packet data with an empty string memo into the legacy `FungibleTokenPacketData` struct. Similarly, an implementation supporting `memo` must be able to unmarshal a legacy packet data into the current struct with the `memo` field set to the empty string.
```

Issue #530, opened 2020-11-24
ICS3: Inconsistencies in the spec and code of Connection Handshake


istoilkovska:
Surfaced from Informal Systems IBC Audit of cosmos-sdk hash  cosmos/cosmos-sdk@fe8a891.

- [ ] In `connOpenInit()` and `connOpenTry()`, there is no error handling in the [spec](https://github.com/cosmos/ics/tree/master/spec/ics-003-connection-semantics#opening-handshake) when calling `addConnectionToClient()`. In the [code](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/keeper.go#L158), the transaction is aborted if the client does not exist.

- [ ] In the code of `connOpenTry()`, but not in the spec:
    - the client parameters are [validated](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L87) (by calling `ValidateSelfClient()`)
    - getting a consensus state from a given height [aborts on error](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L93)
    - there is [error handling](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L133) when picking a version
    - the [client state is verified](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L150), in addition to the client consensus state 

- [ ] in the code of `connOpenAck()`, but not in the spec:
    - there is [error handling](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L133) when getting the connection end by ID from the provable store
    - the client parameters are [validated](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L133) (by calling `ValidateSelfClient`)
[...]

PR #893, opened 2022-12-03
ICS3: fix inconsistencies in the spec and code of connection handshake

crodriguezvega:
This PR addresses the comments of #530 that were still relevant. I will copy those here:

- In connOpenInit() and connOpenTry(), there is no error handling in the [spec](https://github.com/cosmos/ics/tree/master/spec/ics-003-connection-semantics#opening-handshake) when calling addConnectionToClient(). In the [code](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/keeper.go#L158), the transaction is aborted if the client does not exist.

- In the code of connOpenTry(), but not in the spec: 
  - getting a consensus state from a given height [aborts on error](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L93).

- In the code of connOpenAck(), but not in the spec: 
  - there is [error handling](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L133) when getting the connection end by ID from the provable store.
  - getting a consensus state from a given height [aborts on error](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L133).

- In the code of ConnOpenConfirm(), but not in the spec: there is [error handling](https://github.com/cosmos/cosmos-sdk/blob/fe8a891f11dcefe708e43d53549beec808f98687/x/ibc/core/03-connection/keeper/handshake.go#L133) when getting the connection end by ID from the provable store.

Besides that I also fixed some inconsistencies in the usage of the `versions` field in the connection end.

Closes: #530 
Closes: #894

### DIFF
--- a/spec/core/ics-003-connection-semantics/README.md
+++ b/spec/core/ics-003-connection-semantics/README.md
@@ -125,7 +125,6 @@ function clientConnectionsPath(clientIdentifier: Identifier): Path {
 function addConnectionToClient(
   clientIdentifier: Identifier,
   connectionIdentifier: Identifier) {
-    abortTransactionUnless(queryClientState(clientIdentifier) !== null)
     conns = privateStore.get(clientConnectionsPath(clientIdentifier))
     conns.add(connectionIdentifier)
     privateStore.set(clientConnectionsPath(clientIdentifier), conns)
@@ -344,8 +343,12 @@ function connOpenInit(
   version: string,
   delayPeriodTime: uint64,
   delayPeriodBlocks: uint64) {
+    // generate a new identifier
     identifier = generateIdentifier()
+
+    abortTransactionUnless(queryClientState(clientIdentifier) !== null)
     abortTransactionUnless(provableStore.get(connectionPath(identifier)) == null)
+    
     state = INIT
     if version != "" {
       // manually selected version must be one we can support
@@ -383,6 +386,7 @@ function connOpenTry(
     // generate a new identifier
     identifier = generateIdentifier()
     
+    abortTransactionUnless(queryClientState(clientIdentifier) !== null)
     abortTransactionUnless(validateSelfClient(clientState))
     abortTransactionUnless(consensusHeight < getCurrentHeight())
     expectedConsensusState = getConsensusState(consensusHeight, hostConsensusStateProof)
--- a/spec/core/ics-003-connection-semantics/README.md
+++ b/spec/core/ics-003-connection-semantics/README.md
@@ -125,6 +125,7 @@ function clientConnectionsPath(clientIdentifier: Identifier): Path {
 function addConnectionToClient(
   clientIdentifier: Identifier,
   connectionIdentifier: Identifier) {
+    abortTransactionUnless(queryClientState(clientIdentifier) !== null)
     conns = privateStore.get(clientConnectionsPath(clientIdentifier))
     conns.add(connectionIdentifier)
     privateStore.set(clientConnectionsPath(clientIdentifier), conns)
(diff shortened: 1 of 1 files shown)


spec/core/ics-003-connection-semantics/README.md
```diff
@@ -125,6 +125,7 @@ function clientConnectionsPath(clientIdentifier: Identifier): Path {
 function addConnectionToClient(
   clientIdentifier: Identifier,
   connectionIdentifier: Identifier) {
+    abortTransactionUnless(queryClientState(clientIdentifier) !== null)
```
plafer, line 128: IMO this is less clear than if it was a statement in the handlers' body directly ([as we do in ibc-rs](https://github.com/cosmos/ibc-rs/blob/1cbd1d93a87cd1967c943897d199f3274ee84259/crates/ibc/src/core/ics03_connection/handler/conn_open_init.rs#L24)). I also don't like that the check occurs after the `provableStore.set(connectionPath(identifier), connection)` line (e.g. in `ConnOpenTry`). Although I think we state somewhere that changes are reverted when a transaction is aborted, it's clearer to make all checks before state change statements.

crodriguezvega: I see your point, thanks. I think I will defer to @AdityaSripal to decision to make such change. I would  also do it in a different PR that would close #894

AdityaSripal (approved):

spec/core/ics-003-connection-semantics/README.md
AdityaSripal: Don't understand the context here, please fill me in on slack @crodriguezvega

crodriguezvega: Moved the statement to the body of `connOpenInit` and `connOpenTry`.

spec/core/ics-003-connection-semantics/README.md
```diff
@@ -356,7 +360,7 @@ function connOpenInit(
     connection = ConnectionEnd{state, "", counterpartyPrefix,
       clientIdentifier, counterpartyClientIdentifier, versions, delayPeriodTime, delayPeriodBlocks}
     provableStore.set(connectionPath(identifier), connection)
-    addConnectionToClient(clientIdentifier, identifier)
+    addConnectionToClient(clientIdentifier, identifier) // aborts transaction if no client state is found
```
plafer, line 363:
```suggestion
    addConnectionToClient(clientIdentifier, identifier)
```
removed comment as `addConnectionToClient` no longer aborts the transaction

spec/core/ics-003-connection-semantics/README.md
```diff
@@ -382,24 +386,26 @@ function connOpenTry(
...
-                               clientIdentifier, counterpartyClientIdentifier, version, delayPeriodTime, delayPeriodBlocks}
+                               clientIdentifier, counterpartyClientIdentifier, [version], delayPeriodTime, delayPeriodBlocks}
     abortTransactionUnless(connection.verifyConnectionState(proofHeight, proofInit, counterpartyConnectionIdentifier, expectedConnectionEnd))
     abortTransactionUnless(connection.verifyClientState(proofHeight, proofClient, clientState))
     abortTransactionUnless(connection.verifyClientConsensusState(
       proofHeight, proofConsensus, counterpartyClientIdentifier, consensusHeight, expectedConsensusState))
 
     provableStore.set(connectionPath(identifier), connection)
-    addConnectionToClient(clientIdentifier, identifier)
+    addConnectionToClient(clientIdentifier, identifier) // aborts transaction if no client state is found
```
plafer, line 408:
```suggestion
    addConnectionToClient(clientIdentifier, identifier)
```

plafer (approved):
Looks much better! Just need to remove the now incorrect comments

PR #897, opened 2022-12-07
ICS31: fixes and consistency suggestions

crodriguezvega:


spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -163,17 +163,17 @@ function queryPath(id: Identifier): Path {
...
 ```
 
 ### Helper functions
 
-The querying chain MUST implement a function `generateQueryIdentifier`, which generates a unique query identifier:
+The querying chain MUST implement a function `generateIdentifier`, which generates a unique query identifier:
 
 ```typescript
-function generateQueryIdentifier = () -> Identifier
+function generateIdentifier = () -> Identifier
```
angbrav, line 176: I thought it would be an issue for this to clash with `generateIdentifier` of ICS04, but I have seen that ICS03 and ICS04'sa `generateIdentifier` are already clashing.

spec/app/ics-031-crosschain-queries/README.md
```diff
@@ -163,17 +163,17 @@ function queryPath(id: Identifier): Path {
 The result query path is a private path that stores the result of completed queries.
 
 ```typescript
-function resultQueryPath(id: Identifier): Path {
-    return "queriesresult/{id}"
+function queryResultPath(id: Identifier): Path {
+    return "queries/{id}/result"
```
angbrav, line 167: Isn't weird that the path ends with `result` instead of a function argument {}?

angbrav (commented):
I left a couple of comments, the rest LGTM

AdityaSripal (approved):

spec/app/ics-031-crosschain-queries/README.md
crodriguezvega: Yes, in ibc-go we actually use the names [`GenerateConnectionIdentifier`](https://github.com/cosmos/ibc-go/blob/main/modules/core/03-connection/keeper/keeper.go#L56) and [`GenerateChannelIdentifier`](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/keeper.go#L60), but since they are both in different keepers we could have used the `GenerateIdentifier` function name. My suggestion with this renaming was just to keep it consistent with the other specs...

spec/app/ics-031-crosschain-queries/README.md
crodriguezvega: To me it doesn't look strange, and we also follow a similar pattern in [other paths](https://github.com/cosmos/ibc/blob/main/spec/core/ics-024-host-requirements/README.md#path-space).

spec/app/ics-031-crosschain-queries/README.md
angbrav: Yes, I understand. I believe it would be better to use different names throughout the spec (even when they are in different keepers). I guess this could be addressed if needed in a different PR. Resolving.

angbrav (approved):

PR #899, opened 2022-12-14
ICS721: Add classData and memo fields to PacketData

haifengxi:
Putting class/collection/contract level metadata into its own field `classData` instead of squeezing everything into `tokenData` is a better design that is called for by Stargaze/Ark/Juno teams who are relying on WASM contracts to implement and support ICS721.

As for the `memo` field, we are borrowing it from ICS20, there have been thorough discussions about why this field is introduced and how it's supposed to be used (please see #584 #842 #877).

### DIFF
--- a/spec/app/ics-721-nft-transfer/README.md
+++ b/spec/app/ics-721-nft-transfer/README.md
@@ -67,8 +67,8 @@ Both `tokenData` entries and `classData` MUST be Base64 encoded strings which SH
 
 ```json
 {
-  "key1" : { "value":"...", "mime":"..."},
-  "key2" : { "value":"...", "mime":"..."},
+  "key1" : { "value":"...", "mime":"..." },
+  "key2" : { "value":"...", "mime":"..." },
   ...
 }
 ```
@@ -82,7 +82,7 @@ An example of `classData` content (raw JSON before being Base64 encoded) is show
 ```json
 {
   "opensea:name" : { "value":"Crypto Creatures" },
-  "opensea:image" : { "value":"...(Base64 encoded media binary)", "mime":"image/png"},
+  "opensea:image" : { "value":"...(Base64 encoded media binary)", "mime":"image/png" },
   "opensea:seller_fee_basis_points" : { "value":"100" }
 }
 ```
--- a/spec/app/ics-721-nft-transfer/README.md
+++ b/spec/app/ics-721-nft-transfer/README.md
@@ -7,7 +7,7 @@ requires: 25, 26
 kind: instantiation
 author: Haifeng Xi <haifeng@bianjie.ai>
 created: 2021-11-10
-modified: 2022-12-14
+modified: 2022-12-15
 ---
 
 > This standard document follows the same design principles of [ICS 20](../ics-020-fungible-token-transfer) and inherits most of its content therefrom, while replacing `bank` module based asset tracking logic with that of the `nft` module.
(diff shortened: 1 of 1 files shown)


spec/app/ics-721-nft-transfer/README.md
```diff
@@ -35,29 +35,35 @@ The IBC handler interface & IBC routing module interface are as defined in [ICS
...
+
+Each `tokenId` has another corresponding entry in `tokenData` which, if not null, contains opaque on-chain application data associated with the token.
 
-Each `tokenId` has another corresponding entry in `tokenData` which, if present, contains some opaque application data associated with the token (e.g., royalty parameters).
+The `memo` field is not used within the token transfer, however, it may be used either for external off-chain users (i.e. exchanges) or for middleware wrapping transfer that can parse and execute custom logic on the basis of the passed in memo. If the memo is intended to be parsed and interpreted by higher-level middleware, then these middlewares are advised to namespace their additions to the memo string so that they do not overwrite each other. Chains should ensure that there is some length limit on the entire packet data to ensure that the packet does not become a DOS vector. However, these do not need to be protocol-defined limits. If the receiver cannot accept a packet because of length limitations, this will lead to a timeout on the sender side.
```
dreamer-zq, line 66:
```
If the receiver cannot accept a packet because of length limitations, this will lead to a timeout on the sender side.
```
If there is an error in the destination chain `ReceivePacket`, the wrong Ack should be returned instead of timeout

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -110,28 +116,24 @@ The sub-protocols described herein should be implemented in a "non-fungible toke
 The NFT asset tracking module should implement the following functions:
 
 ```typescript
-function SaveClass(classId: string, classUri: string) {
+function SaveClass(classId: string, classUri: string, classData: string) {
```
dreamer-zq, line 119: 1. I think `SaveClass` is better named `SaveOrUpdateClass`
2. `classData` defines the []byte type above, and it is also the []byte type here

spec/app/ics-721-nft-transfer/README.md
0xekez: might be nice to specify here as well that `memo` is associated with the transfer, not the token. something like:

> the `memo` field is associated with the token transfer, not the token.

0xekez: for CosmWasm contracts, we actually can't do this because too long of a memo field could cause a out-of-gas error during deserialization of the packet, before we can run any custom logic. for example:

```
let packet = from_binary(packet_data);
if packet.memo.is_too_long() {
  ACK_FAIL
}
```

could run out of gas during the `from_binary` step. requiring that we return an ACK-FAIL then makes it almost impossible for this to be correctly implemented in CW. the spec being aspirational is fine (we don't necessarily have to pander to the limitations of every execution environment) though. is there a specific advantage of a ACK-FAIL vs a timeout here? ex: does the SDK ics721 implementation handle those differently?

dreamer-zq: I thought about it, if this verification is verified before the transaction is executed, it will indeed not return Fail Ack, thank you for correcting

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -35,29 +35,35 @@ The IBC handler interface & IBC routing module interface are as defined in [ICS
...
 ### Data Structures
 
-Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the class id, class uri, token id's, token uri's, sender address, and receiver address.
+Only one packet data type is required: `NonFungibleTokenPacketData`, which specifies the class id, class uri, class data, token id array, token uri array, token data array, sending address, and receiving address.
 
 ```typescript
 interface NonFungibleTokenPacketData {
   classId: string
   classUri: string
+  classData: byte[]
```
0xekez, line 44: how should this be serialized? if I have `classData` that is `AAA` does that serialize to `[65, 65, 65]`?

spec/app/ics-721-nft-transfer/README.md
haifengxi: Resolved.

0xekez:
some thoughts from doing an [implementation](https://github.com/public-awesome/ics721/compare/zeke/on-chain-metadata?expand=1) of these changes today:

1. make `memo` optional so as not to require it in every TXn.
2. change the type of `tokenData` to `string[]` and `classData` to `string` to keep the encoding flexible and simple.

humanalgorithm:
As discussed previously: 
If we send json with string data and then specify the way it was encoded, we would avoid the need for chains to keep track of how other chains send their data. The sending chain could just put a "mime" field for each key/value and then the receiving chain could just apply the encoding provided to decode the given value. Something like this: 
```
{ 
   
   "key1" : {"value": "0xbef67", "mime": bin/64"},
   "key2" : {"value": "my_value", "mime": bin/string"},
   ...
}
```

spec/app/ics-721-nft-transfer/README.md
haifengxi: So we'll leave the length limit check of `memo` out of the ics721 protocol.

spec/app/ics-721-nft-transfer/README.md
haifengxi: Resolved in https://github.com/cosmos/ibc/pull/899/commits/a6a2b330fb0ca1d74cea6067c9c8a66d0f969238

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -35,29 +35,59 @@ The IBC handler interface & IBC routing module interface are as defined in [ICS
...
 
-`classUri` is optional, but will be extremely beneficial for cross-chain interoperability with NFT marketplaces like OpenSea, where [class/collection metadata](https://docs.opensea.io/docs/contract-level-metadata) can be added for better user experience.
+`classUri` is an optional field which, if present and not empty, contains off-chain [class metadata](https://docs.opensea.io/docs/contract-level-metadata) that could be extremely beneficial for cross-chain interoperability with NFT marketplaces like OpenSea.
 
-`tokenIds` uniquely identifies some tokens of the given class that are being transferred. In the case of an ERC-1155 compliant smart contract, for example, a `tokenId` could be a string representation of the bottom 128 bits of the token ID.
+`classData`is an optional field which, if present and not empty, contains on-chain class metadata such as royalty related parameters.
```
0xekez, line 58:
```suggestion
`classData`is an optional field which, if present, contains on-chain class metadata such as royalty related parameters.
```

spec/app/ics-721-nft-transfer/README.md
```diff
@@ -35,29 +35,59 @@ The IBC handler interface & IBC routing module interface are as defined in [ICS
...
 
-Each `tokenId` has a corresponding entry in `tokenUris` which, if present, refers to an off-chain resource that is typically an immutable JSON file containing the token's metadata.
+`tokenIds` array is an optional field which, if present, SHOULD have a size greater than zero and contain non-empty entries that uniquely identify tokens (of the given class) that are being transferred. In the case of an ERC-1155 compliant smart contract, for example, a `tokenId` could be a string representation of the bottom 128 bits of the token ID.
```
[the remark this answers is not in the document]
haifengxi: Actually, one can construct a packet that only contains class related fields causing a downstream IBC class to be created or updated.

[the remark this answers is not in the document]
haifengxi: Fixed.

AdityaSripal (approved):

dreamer-zq (approved):

zhangyelong:
Hi admins, the Game of NFTs testnet requires this PR to be merged, it'll be great if we can get this done sooner, thanks a lot 💯 💯 💯 

@cwgoes @colin-axner

mpoke (approved):
LGTM

PR #907, opened 2023-01-10
Updates to ICS 02, 06, 07 and 26

crodriguezvega:
Several updates that I discussed with @AdityaSripal:

- Changed signatures of `verifyMembership` and `verifyNonMembership` in ICS 02 to return an error instead of a boolean. Pseudocode for these functions has also been updated in ICS 06 and ICS 07.
- Used `provableStore.get(...)` and `provableStore.set(...)` in ICS 06 and ICS 07 to keep consistency with the other specs.
- Updated `createClient` function in ICS 02 and `initialise` in ICS 06 and 07 to reflect the [latest changes in ibc-go](https://github.com/cosmos/ibc-go/pull/2936).
- Updated `checkForMisbehaviour` in ICS 07 to extend check of conflicting header and to add check for valid misbehavior of BFT time violation.

### DIFF
--- a/spec/client/ics-007-tendermint-client/README.md
+++ b/spec/client/ics-007-tendermint-client/README.md
@@ -148,7 +148,7 @@ func (Header) GetHeight() {
 
 Header implements `ClientMessage` interface.
 
-### Misbehaviour
+### `Misbehaviour`
  
 The `Misbehaviour` type is used for detecting misbehaviour and freezing the client - to prevent further packet flow - if applicable.
 Tendermint client `Misbehaviour` consists of two headers at the same height both of which the light client would have considered valid.
@@ -169,15 +169,13 @@ Tendermint client initialisation requires a (subjectively chosen) latest consens
 
 ```typescript
 function initialise(
-  chainID: string, consensusState: ConsensusState,
+  identifier: Identifier, chainID: string, consensusState: ConsensusState,
   trustLevel: Fraction, height: Height, trustingPeriod: uint64, unbondingPeriod: uint64,
   upgradePath: []string, maxClockDrift: uint64, proofSpecs: []ProofSpec): ClientState {
     assert(trustingPeriod < unbondingPeriod)
     assert(height > 0)
     assert(trustLevel >= 1/3 && trustLevel <= 1)
-    // implementations may define a identifier generation function
-    identifier = generateClientIdentifier()
-    set("clients/{identifier}/consensusStates/{height}", consensusState)
+    provableStore.set("clients/{identifier}/consensusStates/{height}", consensusState)
     return ClientState{
       chainID,
       trustLevel,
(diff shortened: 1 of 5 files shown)


spec/client/ics-007-tendermint-client/README.md
```diff
@@ -278,22 +281,28 @@ function checkForMisbehaviour(
...
-        // assert that the heights are the same
-        assert(misbehaviour.h1.height === misbehaviour.h2.height)
-        // assert that the commitments are different
-        assert(misbehaviour.h1.commitmentRoot !== misbehaviour.h2.commitmentRoot)
+        // if heights are equal check that this is valid misbehaviour of a fork
+        if (misbehaviour.h1.height === misbehaviour.h2.height && misbehaviour.h1.commitmentRoot !== misbehaviour.h2.commitmentRoot) {
+          return true
+        }
+        // otherwise if heights are unequal check that this is valid misbehavior of BFT time violation
+        if (misbehaviour.h1.timestamp <= misbehaviour.h2.timestamp) {
```
crodriguezvega, line 289: This check assumes that `h1.height > h2.height`, but I am not sure if we should also check the other way around (i.e. `h2.height > h1.height && h2.timestamp <= misbehaviour.h1.timestamp`). [In ibc-go we also seem to assume that `h1.height > h2.height`...](https://github.com/cosmos/ibc-go/blob/7d60efa07b7d564e5d4f531a0398d3f670fc973f/modules/light-clients/07-tendermint/misbehaviour_handle.go#L89).

AdityaSripal (approved):

spec/client/ics-007-tendermint-client/README.md
```diff
@@ -278,22 +281,31 @@ function checkForMisbehaviour(
           return true
         }
       case Misbehaviour:
-        // assert that the heights are the same
-        assert(misbehaviour.h1.height === misbehaviour.h2.height)
-        // assert that the commitments are different
-        assert(misbehaviour.h1.commitmentRoot !== misbehaviour.h2.commitmentRoot)
+        if (misbehaviour.h1.height < misbehaviour.h2.height) {
+          return false
+        }
```
crodriguezvega, line 286: @AdityaSripal I added this check to [follow ibc-go](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/misbehaviour.go#L87) as we discussed. If this is ok, then I'll merge.

spec/client/ics-007-tendermint-client/README.md
crodriguezvega: See [comment](https://github.com/cosmos/ibc/pull/907/files#r1089430849).

AdityaSripal (approved):

angbrav (approved):
LGTM. The only question I have is whether there is an `Error` type define or not and whether we should return something more informative instead of `error`. I am referring to `verifyNonMembership` for instance.

crodriguezvega:
> LGTM. The only question I have is whether there is an `Error` type define or not and whether we should return something more informative instead of `error`. I am referring to `verifyNonMembership` for instance.

I had a look and `Error` is not defined anywhere in the specs. But not sure if it should be, though. I could see it more as an implementation detail: each implementation can decide what information to add in the returned errors... If you still think that it should be specified, I would propose to work on it in a separate PR, since it affects all of the specs and not necessarily only this PR.

I will merge this PR for now.

angbrav:
> I had a look and `Error` is not defined anywhere in the specs. But not sure if it should be, though. I could see it more as an implementation detail: each implementation can decide what information to add in the returned errors... If you still think that it should be specified, I would propose to work on it in a separate PR, since it affects all of the specs and not necessarily only this PR.

I think readers could benefit from minimal formatting for Errors. For instance, if errors have a string field that gives a hint of what the error is about. This will help to navigate specs and understand safety checks.

PR #909, opened 2023-01-14
writeup change in relayer strategy on timeout receipt channels

kaampegu:
deleted once the acknowledgement is relayed), and if so relay the acknowledgement to
the source chain.


crodriguezvega (approved):
Thanks, @kaampegu!

PR #914, opened 2023-01-18
ICS7: add nextValidatorsHash instead of validatorSet to ConsensusState

bluele:


crodriguezvega (approved):
Thanks for fixing this, @bluele!

mpoke (approved):

PR #916, opened 2023-01-19
ICS20: Specify Memo and Define Best Practices

AdityaSripal:
Specifies how to use memo if it is to be interpreted and executed upon by the state machine

### DIFF
--- a/spec/app/ics-020-fungible-token-transfer/README.md
+++ b/spec/app/ics-020-fungible-token-transfer/README.md
@@ -46,9 +46,6 @@ interface FungibleTokenPacketData {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -328,6 +325,31 @@ function onTimeoutPacketClose(packet: Packet) {
 }
 ```
 
+### Using the Memo Field
+
+Note: Since earlier versions of this specification did not include a `memo` field, implementations must ensure that the new packet data is still compatible with chains that expect the old packet data. A legacy implementation MUST be able to unmarshal a new packet data with an empty string memo into the legacy `FungibleTokenPacketData` struct. Similarly, an implementation supporting `memo` must be able to unmarshal a legacy packet data into the current struct with the `memo` field set to the empty string.
```
crodriguezvega, line 330: I realize now that the PR that added the `memo` field (well, actually the PR that renamed `metadata` to `memo`) was merged without an entry in the `History` section. Should we add one for that PR (i.e. for November 11, 2022) so that then we can mention here that versions of the specification earlier than this date did not include the `memo` field? Then at least it's clear from what version the `memo` field was introduced.

AdityaSripal: Yes, thanks Carlos

crodriguezvega (approved):

colin-axner (approved):
This looks great, thank you!

PR #926, opened 2023-02-10
ICS29: update escrow amount

crodriguezvega:
Thanks to @pr0n00gler for raising this discrepancy. The spec says that the amount `max(receiveFee+ackFee, timeoutFee)` is escrowed, but [ibc-go escrows the total amount of all fees](https://github.com/cosmos/ibc-go/blob/8f0cfb0682bb6dc6cf311d4644b300041e277866/modules/apps/29-fee/keeper/escrow.go#L27).

### DIFF
--- a/spec/app/ics-029-fee-payment/README.md
+++ b/spec/app/ics-029-fee-payment/README.md
@@ -164,7 +164,7 @@ function RegisterCounterpartyPayee(relayer: string, counterPartyAddress: string)
 // for the packet. In the case where the previous amount is zero, the provided 
 // fees are the initial escrow amount.
 // They may set a separate receiveFee, ackFee, and timeoutFee to be paid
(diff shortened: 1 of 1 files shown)


colin-axner (approved):

AdityaSripal:
I don't think this change is necessary. Actually the spec is the most efficient, and it is intended that the implementation does the same. The implementation isn't wrong, it is doing what the spec says, its just escrowing a little extra

We initially just escrowed everything for simplicity, but I think we should eventually be more capital efficient. especially once channel upgradability makes this more widely used

crodriguezvega:
Thanks, @AdityaSripal, for the context! I will close this then.

colin-axner:
I think it could be worth noting an alternative format. In the situation that the recv, ack, timeout fee are all in different denominations, the formula stated does not work as neither can be considered a maximum

PR #928, opened 2023-02-17
update codeowners

crodriguezvega:
As requested by @mpoke, this change should keep him as owner for core and clients, but owner only of ics28 for apps.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,4 +11,4 @@
 
 /spec/apps @adityasripal @cwgoes @colin-axner @angbrav
 
-/spec/apps/ics-028-cross-chain-validation @mpoke
+/spec/apps/ics-028-cross-chain-validation @mpoke @adityasripal @cwgoes @angbrav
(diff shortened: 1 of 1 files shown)


colin-axner:
github code owners order by last declaration. Is the intention to remove @mpoke from other IBC apps, or to make @mpoke the sole owner of ICS28 or both? 

With this change, @mpoke would be the only owner of ICS 28, as opposed to only removing ownership from other apps

crodriguezvega:
> github code owners order by last declaration. Is the intention to remove @mpoke from other IBC apps, or to make @mpoke the sole owner of ICS28 or both?
> 
> With this change, @mpoke would be the only owner of ICS 28, as opposed to only removing ownership from other apps

Thank you, @colin-axner. I want to remove @mpoke from the other apps, but keep him an owner for ICS28. Then I need to swap the last two lines, right?

colin-axner:
I think swapping the last two lines would result in the same behaviour. I would recommend adding all global owners to the ICS 28 section. This was the ICS 28 ownership remains unchanged, but @mpoke is removed from ownership of other apps. 

You can keep me off the owners of ICS 28, I have limited context and don't work on the implementation

AdityaSripal (approved):

PR #931, opened 2023-02-24
ICS31: small change in path to follow consistency

crodriguezvega:
This is a tiny change to make the path more consistent with the format used in [ICS24](https://github.com/cosmos/ibc/blob/main/spec/core/ics-024-host-requirements/README.md#path-space).


angbrav (approved):
LGTM! It would be nice to spell out path format conventions. I am always confused about what's the proper way to define them.

crodriguezvega:
> LGTM! It would be nice to spell out path format conventions. I am always confused about what's the proper way to define them.

Yes, I agree with that!

Issue #737, opened 2022-05-13
ICS20: support base denoms with slashes


plafer:
ibc-go [now supports base denoms with slashes](https://github.com/cosmos/ibc-go/pull/978). We should probably update the standard so that other implementations can also support [chains that require it](https://github.com/cosmos/ibc-go/issues/894).

ibc-rs is planning on supporting base denoms with slashes as well ([PR](https://github.com/informalsystems/ibc-rs/pull/1989)).

crodriguezvega:
Thanks for opening this issue, @plafer.

When I opened the PR to add the support for base denoms with slashes, I thought that it was a bug on ibc-go that base denoms were not supported, since they are valid denoms for the Cosmos SDK. But back then I didn't realize (re-reading the spec now) that the [spec said](https://github.com/cosmos/ibc/blob/master/spec/app/ics-020-fungible-token-transfer/README.md#data-structures):

```
If {denom} contains /, then it must also be in the ics20 form which indicates that this token has a multi-hop record. Note that this requires that the / (slash character) is prohibited in non-IBC token denomination names.
```

So does the above means that ibc-go should not have added support for slashes in base denoms? Looks like we should have agreed on a spec change first before implementing the support in the implementation...

ethanfrey:
@sunnya96
@valardragon
 this may be relevant to the token factory osmosis inplemented

colin-axner:
I think this is only an issue if the chain implementing ics20 stores the received denominations using the full path (ie it doesn't differetiate between a minted token with the denom `{ics20Port}/{ics20Channel}/{denom}` and a IBC token received with the denom `{ics20Port}/{ics20Channel}/{denom}`). Since ibc-go uses the `ibc/{hash}` representation for internal tracking, a newly minted token in the form `{ics20Port}/{ics20Channel}/{denom}` will be treated differently than the actual token received via IBC with that same path

If my assumption is correct, I'd be in favor of changing the spec to indicate this. I think it also might be useful to recommend the `ibc/{hash}` approach as best practice, but maybe there are good reasons not to do this?

cc @AdityaSripal

PR #935, opened 2023-03-02
ICS20: Update Spec to support slashes in base denomination

AdityaSripal:
closes: https://github.com/cosmos/ibc/issues/737


spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -49,7 +49,7 @@ interface FungibleTokenPacketData {
...
+The ics20 token denominations are represented the form `{ics20Port}/{ics20Channel}/{denom}`, where `ics20Port` and `ics20Channel` are an ics20 port and channel on the current chain for which the funds exist. The prefixed port and channel pair indicate which channel the funds were previously sent through. Implementations are responsible for correctly parsing the IBC trace information from the base denomination. Since the channel identifiers are all auto-generated in the form: `channel-n`. ibc-go uses this fact to correctly parse out the IBC trace information from the base denom which may have slashes, but will not have a prefix: `transfer/channel-n`. If this assumption is broken, the trace information will be parsed incorrectly. Thus chains must make sure that base denominations do not have the ability to create arbitrary prefixes that can mock the ICS20 logic.
```
colin-axner, line 52:
```suggestion
The ics20 token denominations are represented the form `{ics20Port}/{ics20Channel}/{denom}`, where `ics20Port` and `ics20Channel` are an ics20 port and channel on the current chain for which the funds exist. The prefixed port and channel pair indicate which channel the funds were previously sent through. Implementations are responsible for correctly parsing the IBC trace information from the base denomination. Since the channel identifiers are all auto-generated in the form: `channel-n`. ibc-go uses this fact to correctly parse out the IBC trace information from the base denom which may have slashes, but will not have a substring: `transfer-port-name/channel-n`. If this assumption is broken, the trace information will be parsed incorrectly. Thus chains must make sure that base denominations do not have the ability to create arbitrary prefixes that can mock the ICS20 logic.
```
Main point of confusion is that it isn't about the prefix but the substring when splitting the slashes evenly. See [here](https://github.com/cosmos/ibc-go/blob/main/modules/apps/transfer/types/trace.go#L97). Okay with leaving `transfer` instead of `transfer-port-name`, just a technicality that on ibc-go it could be a different value and still parse fine

colin-axner (approved):
LGTM, left a suggestion for clarity

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -49,7 +49,7 @@ interface FungibleTokenPacketData {
...
+The ICS 20 token denominations are represented by the form `{ics20Port}/{ics20Channel}/{denom}`, where `ics20Port` and `ics20Channel` are an ICS 20 port and channel on the current chain for which the funds exist. The prefixed port and channel pair indicate which channel the funds were previously sent through. Implementations are responsible for correctly parsing the IBC trace information from the base denomination. Since the channel identifiers are all auto-generated in the form: `channel-n`. ibc-go uses this fact to correctly parse out the IBC trace information from the base denom which may have slashes, but will not have a substring: `transfer-port-name/channel-n`. If this assumption is broken, the trace information will be parsed incorrectly. Thus chains must make sure that base denominations do not have the ability to create arbitrary prefixes that can mock the ICS 20 logic.
```
crodriguezvega, line 52:
```suggestion
The ICS 20 token denominations are represented by the form `{ics20Port}/{ics20Channel}/{denom}`, where `ics20Port` and `ics20Channel` are an ICS 20 port and channel on the current chain for which the funds exist. The prefixed port and channel pair indicate which channel the funds were previously sent through. Implementations are responsible for correctly parsing the IBC trace information from the base denomination. The way the reference ICS 20 implementation in ibc-go handles this is by taking advantage of the fact that it automatically generates channel identifiers with the format `channel-{n}`, where `n` is a integer greater or equal than 0. It can then correctly parse out the IBC trace information from the base denom which may have slashes, but will not have a substring of the form `{transfer-port-name}/channel-{n}`. If this assumption is broken, the trace information will be parsed incorrectly (i.e. part of the base denom will be misinterpreted as trace information). Thus chains must make sure that base denominations do not have the ability to create arbitrary prefixes that can mock the ICS 20 logic.
```

crodriguezvega (commented):
Thanks, @AdityaSripal! I suggested a couple of changes for clarification. Since the spec doesn't enforce the `channel-{n}` format for channel identifiers I thought it would be worth  to be explicit that this is ibc-go's decision.

PR #936, opened 2023-03-02
ICS09: loopback client spec

crodriguezvega:
Bringing back the [loopback client spec](https://github.com/cosmos/ibc/pull/878) with updates for 02-client refactor and adding the ideas from the special [localhost connection identifier PR](https://github.com/cosmos/ibc/pull/879).


spec/client/ics-009-loopback-cilent/README.md
```diff
@@ -0,0 +1,244 @@
...
+  revisionHeight: uint64
+}
+```
+
+### Header
+
+The header includes the identifier of the light client and the current block height of the local ledger.
+
+```typescript
+interface Header {
```
crodriguezvega, line 74: I know that we don't have a `Header` type in ibc-go for the localhost client, but I thought that it was nice as an abstraction, so that we don't need to specify a `GetSelfHeight` function that is used in `UpdateState`. Happy to change this though, if folks think otherwise.

spec/client/ics-009-loopback-cilent/README.md
```diff
@@ -0,0 +1,247 @@
...
+    }
+}
+```
+
+### Validity predicate
+
+Loopback client validity checking makes sure the height is greater than zero. If the provided header is valid, the client state update logic is executed.
+
+```typescript
+function verifyClientMessage(
```
crodriguezvega, line 117: I know that we don't use this in ibc-go, but again, I think from an abstraction point of view there's no reason why it could be done. It was ibc-go's decision to not implement localhost client this way.

AdityaSripal: I would also remove this because it would be very inefficient for relayers to update in this way.

We should explicitly have this spec no-op or error on these methods. And have some small section explaining that the implementation should run some per-block logic to update the localhost client state height. You do not need to introduce EndBlocker or GetSelfHeight explicitly. It is up to the specific execution environment how it will be implemented

spec/client/ics-009-loopback-cilent/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+`UpdateStateOnMisbehaviour` is unsupported by the loopback client and performs a no-op.
+
+```typescript
+function updateStateOnMisbehaviour(clientMsg: clientMessage) { }
+```
+
+### State verification functions
+
+State verification functions simply need to read state from the local ledger and compare with the bytes stored under the standardized key paths. The loopback client needs read-only access to the complete IBC store of the local ledger, and not only to its own client identifier-prefixed store.
```
AdityaSripal, line 170:
```suggestion
State verification functions simply need to read state from the local ledger and compare with the bytes stored under the standardized key paths. The loopback client needs read-only access to the **entire IBC store** of the local ledger, and not only to its own client identifier-prefixed store.
```

AdityaSripal: Boldened and changed word for clarity

AdityaSripal (changes requested):
Thanks @crodriguezvega 

I think we should explicitly remove notions of header and submitting headers since they will not be done for any realistic implementation.

Instead attach a small section explaining that localhost client should be automatically updated

AdityaSripal (approved):
THanks for this @crodriguezvega !

Issue #934, opened 2023-03-01
ICS18: some possible mistakes


michwqy:
I found some places that made me difficult to understand while reading ICS18.
```Go
function pendingDatagrams(chain: Chain, counterparty: Chain): List<Set<Datagram>> {
  ...
  counterpartyHeight = counterparty.latestHeight()
  ...
  for (const localEnd of connections) {
    ...
    else if (localEnd.state === INIT && remoteEnd.state === TRYOPEN)
      // Handshake has started on the other end (2 steps done), relay `connOpenAck` to the local end
      localDatagrams.push(ConnOpenAck{
        identifier: localEnd.identifier,
        version: remoteEnd.version,
        proofTry: remoteEnd.proof(),
        proofConsensus: remoteEnd.client.consensusState.proof(),
        proofHeight: remoteEnd.client.height,
        consensusHeight: remoteEnd.client.height,
      })
  ...
  for (const localEnd of channels) {
    ...
    else if (localEnd.state === INIT && remoteEnd.state === TRYOPEN)
      localDatagrams.push(ChanOpenAck{
        portIdentifier: localEnd.portIdentifier,
        channelIdentifier: localEnd.channelIdentifier,
        version: remoteEnd.version,
        proofTry: remoteEnd.proof(),
        proofHeight: localEnd.client.height,
      })
    ...
````
I'm not sure, but I think `proofHeight` in `ConnOpenAck` and `ChanOpenAck` should be `counterpartyHeight`. Appreciate for any help.

crodriguezvega:
Thanks for identifying this error, @michwqy!

Yes, I think you're right that it should be `counterpartyHeight` for both datagrams, since `proofHeight` is the height of the counterparty for which the local light client has has a commitment root for proving the proofs.

Would you like to open a PR to fix it?

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #938, opened 2023-03-07
Fix some mistakes in ICS18

michwqy:
As #934 said, I found some mistakes about the ```proofHeight``` in ICS18.

closes: #934

### DIFF
--- a/spec/relayer/ics-018-relayer-algorithms/README.md
+++ b/spec/relayer/ics-018-relayer-algorithms/README.md
@@ -152,7 +152,7 @@ function pendingDatagrams(chain: Chain, counterparty: Chain): List<Set<Datagram>
         version: remoteEnd.version,
         proofTry: remoteEnd.proof(),
         proofConsensus: remoteEnd.client.consensusState.proof(),
-        proofHeight: remoteEnd.client.height,
+        proofHeight: counterpartyHeight,
         consensusHeight: remoteEnd.client.height,
       })
     else if (localEnd.state === OPEN && remoteEnd.state === TRYOPEN)
@@ -193,7 +193,7 @@ function pendingDatagrams(chain: Chain, counterparty: Chain): List<Set<Datagram>
         channelIdentifier: localEnd.channelIdentifier,
         version: remoteEnd.version,
         proofTry: remoteEnd.proof(),
-        proofHeight: localEnd.client.height,
+        proofHeight: counterpartyHeight,
       })
     else if (localEnd.state === OPEN && remoteEnd.state === TRYOPEN)
       // Handshake has confirmed locally (3 steps done), relay `chanOpenConfirm` to the remote end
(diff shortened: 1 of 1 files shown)


angbrav (commented):
I'd like to understand this better before accepting the changes. As far as I understand `proofHeight` in those datagrams does not need to be of the latest chain height but the height at which `proofTry` was generated. The question now is for which height `remoteEnd.proof()` generates its proofs. Cannot really figure out the latter, any idea where to look for it?

On a separate note, this spec needs work, I think it is not complete and has some inconsistencies with other standards.

crodriguezvega:
Thanks for taking a look, @angbrav!

> I'd like to understand this better before accepting the changes. As far as I understand proofHeight in those datagrams does not need to be of the latest chain height but the height at which proofTry was generated. The question now is for which height remoteEnd.proof() generates its proofs. Cannot really figure out the latter, any idea where to look for it?

Correct: `proofHeight` does not need to be the latest height of the counterparty, but the height at which `proofTry` can be verified.

I agree the spec is underspecified and needs some work, but for this particular problem and considering the current state of the spec I think `proofHeight` should be `counterpartyHeight`, which is found at the top of this pseudocode block:

```
counterpartyHeight = counterparty.latestHeight()
client = chain.queryClientConsensusState(counterparty)
if client.height < counterpartyHeight {
  header = counterparty.latestHeader()
  localDatagrams.push(ClientUpdate{counterparty, header})
}
```

So here the relayer queries the latest height of the counterparty and it updates the light client on the chain if the client height is behind the counterparty height. So that means that the light client for the counterparty will always be on the latest height of the counterparty. And that's why I think that the value for `proofHeight` should be `counterpartyHeight`, because that's the height of the counterparty at which the light client is.

> This spec needs work, I think it is not complete and has inconsistencies with other standards.

Agree! Definitely many things are not well specified and I completely understand your confusion.

angbrav:
Thanks @crodriguezvega, I think it is clearer now. It would be nice to specify those `proof` functions properly but I guess that should be a different PR. Do you know if relayers use their clients to call the functions in https://github.com/cosmos/ibc/tree/main/spec/core/ics-002-client-semantics#proof-construction to construct their proofs?

angbrav (approved):

crodriguezvega:
> It would be nice to specify those proof functions properly but I guess that should be a different PR.

I agree. Let's try to improve all these things when ICS18 gets a facelift. 💇‍♀️ 

> Do you know if relayers use their clients to call the functions in https://github.com/cosmos/ibc/tree/main/spec/core/ics-002-client-semantics#proof-construction to construct their proofs?

I think they don't use their light clients, but (at least for CometBFT) they just make ABCI queries to the RPC endpoint. For example, [here's Hermes code to query and prove the client state](https://github.com/informalsystems/hermes/blob/v1.0.0/relayer/src/chain/cosmos.rs#L778).

PR #939, opened 2023-03-09
[DRAFT] Extend 02-client w/ Conditional Clients 

notbdu:
Modular blockchains break up a logical blockchain into many constituent parts. To accurately represent these chains and also to account for various types of shared security primitives that are coming up, we need to introduce dependencies between clients.

This is a DRAFT spec change. Opening for discussion.


spec/core/ics-002-client-semantics/README.md
```diff
@@ -67,6 +68,22 @@ appropriate gas metering mechanism to charge for compute and storage. On a host
...
 could be provided as executable WASM functions when the client instance is created.
 
+Conditional clients can express dependencies on existing client instances. The idea is here is that modular blockchains change
+the way a logical blockchain can be expressed. Since a logical blockchain can be broken up into its constituent parts, 
+a single client can become multiple clients. This results in greater state layer diversity for IBC.
+
+For example, we could have a different client for each of the following functional layers of the blockchain stack:
+* Data availability
+* Sequencing or transaction ordering
+* Execution
```
AdityaSripal, line 78: For IBC, we only really care about execution. ie the `app state` of the blockchain.

In the modular world, in order to have a client following execution, do we need a client of the other layers?

notbdu: The issue from my POV is that to prove execution - you also need to prove sequencing (TO) & data availability (DA) etc. 

For example, in the optimistic rollup (ORU) on Celestia case in the simple case you could have:
- A consensus proof from a Celestia client (TMLC) for proof of DA and TO
- A fraud proof from an ORU client (optimistic-LC) for proof of execution
  - ORU client depends on Celestia client
  - Accepts headers that and starts a fraud window only when the header block has been persisted to the Celestia network

Also, with new shared security primitives like babylon chain (checkpointing on BTC) - you can have dependencies on a babylon client for proof of security/checkpoint.

spec/core/ics-002-client-semantics/README.md
```diff
@@ -607,6 +661,39 @@ function submitMisbehaviourToClient(
...
+    clear(bonded)
+}
+```
+
+#### Bonding
+
+A slashable bond may be placed on a client to economically guarantee correct behaviour and/or incentivize reporting of misbehaviour.
+
+```typescript
+function bondToken(
```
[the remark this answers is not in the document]
notbdu: We could do that 😄 ! Listing out some pros/cons off the top of my head.

Cons of bonding logic within light clients:
- lose standard RPC calling interface at the client level
- clients would need to expose a gRPC server with an arbitrary calling interface (can try to encourage a calling standard)
- no calling interface from `app` protocols (e.g. transfer and bond in single packet/tx)
  - maybe not a huge concern because they can always transfer first and then call RPC to bond in two separate tx
- reward distribution from utilization will be more complicated (not in the scope of this spec)
  - this issue should be more clearer when looking at client update incentivization

Pros of bonding logic within light clients:
- less complexity on 02-client side
- not every client requires a bond to be secure
- give light client writer more freedom

As a hybrid approach, slashing conditions can be pushed down into the client implementation but bond management could be handled within the client module?

spec/core/ics-002-client-semantics/README.md
adiraviraj: in this specific case that you're talking about for ORUs on Celestia, by proving execution are you not also proving DA and TO? since if the data was not available and the txs were not ordered properly (which is verified on the client side), the rollup nodes would reject those txs anyway?

nevertheless, i think proving DA would be necessary since each Celestia rollup can in theory have their own fork choice rule (i.e. 'Sovereign rollups')

notbdu: If you don't prove DA before initiating the fraud window you can get the following:
- Accept header optimistically w/o ensuring data is available
- Start fraud window
- Data is withheld so no fraud prover can actually generate a fraud proof
- Invalid header gets acceptable after fraud window period

spec/core/ics-002-client-semantics/README.md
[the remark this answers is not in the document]
notbdu: A different idea:
- Add `IBCClientMiddleware` for Client lifecycle methods
- Allow `IBCClientMiddleware` to register gRPC service/query routes

spec/core/ics-002-client-semantics/README.md
```diff
@@ -536,12 +583,13 @@ Calling `createClient` with the client type and initial consensus state creates
 ```typescript
 function createClient(
```
AdityaSripal, line 584: This function in general is out of date. Since we take in a ClientState now.

I think if we refactored this to follow what ibc-go is doing this can be achieved.

Since the clientDependencies will be fields on the ClientState

AdityaSripal (commented):
This seems quite plausible to me. The main thing is that the client handler at the 02-layer will need to authorize access to the clientstore(s) that our client is dependent on. I think we would ideally only give read-only access? And in that case, I don't think any permissioning from the dependencies is required.

It would be helpful if there's a more concrete example/use to work off of here

AdityaSripal:
Happy to see the changes are clean and minimal!

crodriguezvega (commented):
Thank you, @notbdu. I think it would be good to update your fork with the latest from main, since there have been some updates in ICS 02 since you opened the PR.

And just a naming suggestion: instead of conditional clients, what dependent clients?

Issue #906, opened 2023-01-06
ICS-4: `chanOpenTry` still stores a `ChannelEnd` with an incorrect `version`


plafer:
In `chanOpenTry`, the following lines are incorrect:

```js
    channel = ChannelEnd{TRYOPEN, order, counterpartyPortIdentifier,
                         counterpartyChannelIdentifier, connectionHops, version}
    provableStore.set(channelPath(portIdentifier, channelIdentifier), channel)
```
The `version` field is deprecated since the removal of crossing hellos, and so we shouldn't store it as part of our `ChannelEnd`.

Seems like this was forgotten to be updated in #629.

Related: #851

crodriguezvega:
Thanks, @plafer. I have a question:

> The version field is deprecated since the removal of crossing hellos, and so we shouldn't store it as part of our ChannelEnd.

As you pointed out in [this issue](https://github.com/cosmos/ibc/issues/859), The `version` field in the datagram `MsgChanOpenTry` datagram is deprecated, but not the `version` passed to the `chanOpenTry` handler, right?

Given that in `handleChanOpenTry` in ICS26 the module `onChanOpenTry` callback is called first and then the `chanOpenTry` handler function is called with the `version` returned by the callback, why is the `version` used to construct the `ChannelEnd` in `chanOpenTry` incorrect? [In ibc-go we do store the version as part of the channel end](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/handshake.go#L195).

plafer:
Ah yes, you're right; I probably got mixed up by reading the standard and ibc-go together, and forgot that the standard's ICS-26 calls `onChanOpenTry` first.

So IIUC, the error in the spec is now that `onChanOpenTry` is called with `datagram.channelIdentifier`, whereas it should be called with the `channelIdentifier` returned by `chanOpenTry`, right? Then, the correct way to do it would be to write the `ChannelEnd` to storage in ICS-26, as is done in ibc-go, right?

crodriguezvega:
> So IIUC, the error in the spec is now that onChanOpenTry is called with datagram.channelIdentifier, whereas it should be called with the channelIdentifier returned by chanOpenTry, right? Then, the correct way to do it would be to write the ChannelEnd to storage in ICS-26, as is done in ibc-go, right?

Where you say `channelIdentifier`, is that a typo and you actually meant `version`?

[...]

Issue #851, opened 2022-10-07
ICS26: order of module and handler calls different to ibc-go implementation


crodriguezvega:
While going through [this issue in ibc-go](https://github.com/cosmos/ibc-go/issues/25) I was able to verify the claim of the first bullet point regarding the differences between the spec and the implementation in the order of calling the module and the handler functions. 

For [channel lifecycle](https://github.com/cosmos/ibc/blob/main/spec/core/ics-026-routing-module/README.md#channel-lifecycle-management) (except for `ChanCloseInit` and `ChanCloseConfirm`) and [packet relay](https://github.com/cosmos/ibc/blob/main/spec/core/ics-026-routing-module/README.md#packet-relay) the order of the calls in the spec is the opposite to that in the ibc-go implementation: the spec calls first the callback on the module and then calls the handler function, while in the implementation is the other way around.

Does the spec need to be updated to match the order used in the implementation?

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #946, opened 2023-03-28
fix: correct order of function calling in ICS26 / write channel end with version from application callback

crodriguezvega:
closes: #851
closes: #906


spec/core/ics-026-routing-module/README.md
```diff
@@ -610,14 +632,14 @@ interface PacketRecv {
...
-      acknowledgement
-    )
+  module = lookupModule(datagram.packet.destPort)
+  handler.recvPacket(
+    datagram.packet,
+    datagram.proof,
+    datagram.proofHeight,
+    acknowledgement
+  )
+  acknowledgement = module.onRecvPacket(datagram.packet)
```
AdityaSripal, line 642: acknowledgement is nil on the recvPacket call.

We need to use WriteAcknowledgement after the callback

AdityaSripal (changes requested):
Looks great! One change necessary

spec/core/ics-026-routing-module/README.md
crodriguezvega: Fixed! Thank you, @AdityaSripal!

AdityaSripal (approved):

PR #955, opened 2023-04-03
add version compatibility column

crodriguezvega:
The PR adds a column to the table on the top of each spec for an optional list of implementations' versions that are compatible with the spec. 

This is [an example](https://github.com/cosmos/ibc/blob/carlos/add-version-compatibility-column/spec/core/ics-002-client-semantics/README.md) of how it looks. Unfortunately the table is too width and needs a scroll bar.


AdityaSripal (approved):
Thanks @crodriguezvega !

Godssentralen:
takk

Issue #856, opened 2022-10-13
Validate `ClientState` and set `clientStatePath` on client creation


hu55a1n1:
The create client handler [spec](https://github.com/cosmos/ibc/tree/main/spec/core/ics-002-client-semantics#create) is currently not accepting a `ClientState` as input and is not setting the `clientStatePath`. 
The handler must additionally validate the received `ClientState` (for e.g. using [ClientType.initialize()](https://github.com/cosmos/ibc/blob/main/spec/client/ics-007-tendermint-client/README.md#client-initialisation)).

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #810, opened 2022-07-29
ICS-2: `ClientType` is not defined and confusing


plafer:
The type `ClientType` is not clearly defined in ICS-2, and seems to play 2 different roles.

On the one hand, it is the type of the value stored at `"clients/{identifier}/clientType"` in the `provableStore` (see ICS-24 table). Given how it's meant to be used, I would expect it to be a well-defined type (as opposed to an opaque data structure). For example, in ibc-go, `ClientType` is a string.

On the other hand, the reader can gather that it is an object that is expected to have the methods such as `initialise()`, `checkValidityAndUpdateState()`, and `checkMisbehaviourAndUpdateState()`. I was able to figure that out by looking at how it is used in the code snippets, and then link it to the text that says

> Client types MUST define a method to initialise a client state with a provided consensus state, writing to internal state as appropriate. [the `initialise` method]

It wasn't immediately clear to me that the quote above referred to the `ClientType` type.

Given this, what is `ClientType` supposed to be: the former or the latter?

Finally, we should also ensure that the `ClientType` value `"07-tendermint"` that ibc-go uses to identify the Tendermint client is documented somewhere.

PR #956, opened 2023-04-04
ICS 02: remove concept of client type

crodriguezvega:
Remove deprecated concept of `ClientType`

closes: https://github.com/cosmos/ibc/issues/810
closes: https://github.com/cosmos/ibc/issues/856

### DIFF
--- a/spec/core/ics-002-client-semantics/README.md
+++ b/spec/core/ics-002-client-semantics/README.md
@@ -221,10 +221,10 @@ but they must expose this common set of query functions to the IBC handler.
 type ClientState = bytes
 ```
 
-Client types MUST define a method to initialise a client state with the provided client identifier and consensus state, writing to internal state as appropriate.
+Client types MUST define a method to initialise a client state with the provided client identifier, client state and consensus state, writing to internal state as appropriate.
 
 ```typescript
-type initialise = (identifier: Identifier, consensusState: ConsensusState) => ClientState
+type initialise = (identifier: Identifier, clientState: ClientState, consensusState: ConsensusState) => Void
 ```
 
 Client types MUST define a method to fetch the current height (height of the most recent validated state update).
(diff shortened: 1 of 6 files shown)


spec/core/ics-002-client-semantics/README.md
```diff
@@ -552,15 +552,12 @@ logical correctness.
...
-    identifier = generateClientIdentifier()
-    abortTransactionUnless(provableStore.get(clientStatePath(identifier)) === null)
-    abortSystemUnless(provableStore.get(clientTypePath(identifier)) === null)
-    clientType.initialise(identifier, consensusState)
-    provableStore.set(clientTypePath(identifier), clientType)
+function createClient(consensusState: ConsensfusState) {
+  // implementations may define a identifier generation function
+  identifier = generateClientIdentifier()
+  abortTransactionUnless(provableStore.get(clientStatePath(identifier)) === null)
+  clientState = initialise(identifier, consensusState)
```
crodriguezvega, line 559: @AdityaSripal I did this differently to what we discussed: I thought leaving the `initialise` function as it was (i.e. creating the client state) was easier and involved fewer changes. Let me know what you think.

AdityaSripal: Diverges from the implementation:

https://github.com/cosmos/ibc-go/blob/main/modules/core/02-client/keeper/client.go#L17

AdityaSripal (changes requested):
Thanks for taking this on @crodriguezvega , I think we should follow implementation here

spec/core/ics-002-client-semantics/README.md
crodriguezvega: Yes, it does diverge... Ok, I will make it comply with implementation, but I will need to do some changes in ICS 06, 07 and 09 as well.

AdityaSripal (approved):

PR #967, opened 2023-04-24
Upgradability fixes

AdityaSripal:
Refactor according to @colin-axner suggestions

- Store upgrade in separate path and keep original channel in state until channel is OPEN
- Pause new packets and flush in-flight packets before channel can reopen

----

Happy path handshake is done

Cancel/Timeout still needs to change

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -140,7 +140,7 @@ interface ErrorReceipt {
 
 #### Upgrade Channel Path
 
-The chain must store the proposed upgrade until the upgrade handshake successfully. This must be stored in the publice store. It may be deleted once the upgrade is successful.
+The chain must store the proposed upgrade upon initiating an upgrade. The proposed upgrade must be stored in the public store. It may be deleted once the upgrade is successful or has been aborted.
 
 ```typescript
 function channelUpgradePath(portIdentifier: Identifier, channelIdentifier: Identifier): Path {
@@ -148,7 +148,7 @@ function channelUpgradePath(portIdentifier: Identifier, channelIdentifier: Ident
  }
 ```
 
-The upgrade path has an associated verification membership added to the connection interface so that a counterparty may verify that chain has stored the provided counterparty upgrade.
+The upgrade path has an associated verification of membership method added to the connection interface so that a counterparty may verify that chain has stored the provided counterparty upgrade.
 
 ```typescript
 // Connection VerifyChannelUpgrade method
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -656,6 +656,8 @@ function chanUpgradeAck(
 }
 ```
 
+`chanUpgradeOpen` may only be called once both sides have moved to FLUSHCOMPLETE. If there exists unprocessed packets in the queue when the handshake goes into `FLUSHING` mode, then the packet handlers must move the channelEnd to `FLUSHCOMPLETE` once the last packet on the channelEnd has been processed.
+
 ```typescript
 function chanUpgradeOpen(
     portIdentifier: Identifier,
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -39,7 +39,7 @@ enum ChannelState {
 
 - The chain that is proposing the upgrade should set the channel state from `OPEN` to `INITUPGRADE`
 - The counterparty chain that accepts the upgrade should set the channel state from `OPEN` to `TRYUPGRADE`
-- Once the initiating chain verifies the counterparty is in `TRYUPGRADE`, it must move to `ACKUPGRADE` in the case where there still exist in-flight packets on **both ends** or complete the upgrade and move to `OPEN`
+- Once the initiating chain verifies the counterparty is in `TRYUPGRADE`, it must move to `ACKUPGRADE` unless all in-flight packets are already flushed on both ends, in which case it must move directly to `OPEN`.
 - The `TRYUPGRADE` chain must prove the counterparty is in `ACKUPGRADE` or completed the upgrade in `OPEN` AND have no in-flight packets on **both ends** before it can complete the upgrade and move to `OPEN`.
 - The `ACKUPGRADE` chain may OPEN once in-flight packets on **both ends** have been flushed.
 
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -588,56 +588,89 @@ function chanUpgradeAck(
 function chanUpgradeConfirm(
     portIdentifier: Identifier,
     channelIdentifier: Identifier,
-    counterpartyChannel: ChannelEnd,
+    counterpartyChannelState: ChannelState,
     proofChannel: CommitmentProof,
     proofUpgradeError: CommitmentProof,
     proofUpgradeSequence: CommitmentProof,
     proofHeight: Height,
 ) {
-     // if packet commitments are not empty then abort the transaction
+    // if packet commitments are not empty then abort the transaction
     abortTransactionUnless(pendingInflightPackets(portIdentifier, channelIdentifier))
 
     // current channel is in TRYUPGRADE
     currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
-    abortTransactionUnless(channel.state == BLOCKUPGRADE)
+    abortTransactionUnless(channel.state == TRYUPGRADE)
 
-    // counterparty must be in OPEN state
-    abortTransactionUnless(counterpartyChannel.State == OPEN)
+    connection = getConnection(currentChannel.connectionIdentifier)
+    counterpartyHops = getCounterpartyHops(connection)
 
-    // get current sequence
-    sequence = provableStore.get(channelUpgradeSequencePath(portIdentifier, channelIdentifier))
+    // counterparty must be in OPEN or BLOCKUPGRADE state
+    if counterpartyChannelState == OPEN {
+        // get upgrade since counterparty should have upgraded to these parameters
+        upgrade = provableStore.get(channelUpgradePath(portIdentifier, channelIdentifier))
 
-    // get underlying connection from the original channel for proof verification
-    originalChannel = provableStore.get(channelRestorePath(portIdentifier, channelIdentifier))
-    connection = getConnection(originalChannel.connectionIdentifier)
+        counterpartyChannel = ChannelEnd{
+            state: OPEN,
+            ordering: upgrade.fields.ordering,
+            counterpartyPortIdentifier: portIdentifier,
+            counterpartyChannelIdentifier: channelIdentifier,
+            connectionHops: upgrade.fields.connectionHops,
+            version: upgrade.fields.version,
+            sequence: currentChannel.sequence,
+        }
+    } else if counterpartyChannelState == BLOCKUPGRADE {
+        counterpartyChannel = ChannelEnd{
+            state: BLOCKUPGRADE,
+            ordering: currentChannel.ordering,
+            counterpartyPortIdentifier: portIdentifier,
+            counterpartyChannelIdentifier: channelIdentifier,
+            connectionHops: counterpartyHops,
+            version: currentChannel.version,
+            sequence: currentChannel.sequence,
+        }
+    } else {
+        abortTransactionUnless(false)
+    }
 
-    // verify proofs of counterparty state
     abortTransactionUnless(verifyChannelState(connection, proofHeight, proofChannel, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, counterpartyChannel))
-    // verify counterparty did not abort upgrade handshake by writing upgrade error
-    // must have absent value at upgradeError path at the current sequence
-    abortTransactionUnless(verifyUpgradeChannelErrorAbsence(connection, proofHeight, proofUpgradeError, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier))
 
-    // verify that the counterparty sequence is the same as the current sequence to ensure that the proofs were
-    // retrieved from the current upgrade attempt
-    // since all proofs are retrieved from same proof height, and there can not be multiple upgrade states in the store for a given
-    // channel at the same time
-    abortTransactionUnless(verifyUpgradeSequence(connection, proofHeight, proofUpgradeSequence, currentChannel.counterpartyPortIdentifier,
-    currentChannel.counterpartyChannelIdentifier, sequence))
+    // move channel to OPEN and adopt upgrade parameters
+    openChannelHandshake(portIdentifier, channelIdentifier)
 
     // call modules onChanUpgradeConfirm callback
     module = lookupModule(portIdentifier)
     // confirm callback must not return error since counterparty successfully upgraded
-    module.onChanUpgradeConfirm(
+    module.onChanUpgradeOpen(
         portIdentifer,
         channelIdentifier
     )
-    
-    // upgrade is complete
-    // set channel to OPEN and remove unnecessary state
-    currentChannel.state = OPEN
-    provableStore.set(channelPath(portIdentifier, channelIdentifier), currentChannel)
-    provableStore.delete(channelUpgradeTimeoutPath(portIdentifier, channelIdentifier))
-    privateStore.delete(channelRestorePath(portIdentifier, channelIdentifier))
+}
+```
+
+```typescript
+// chanUpgradeOpen can be called unilaterally on the channelEnd that is at `BLOCKUPGRADE`
+// once the packets are flushed
+function chanUpgradeOpen(
+    portIdentifier: Identifier,
+    channelIdentifier: Identifier,
+) {
+    // current channel is in BLOCKUPGRADE
+    currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+    abortTransactionUnless(channel.state == BLOCKUPGRADE)
+
+    // if packet commitments are not empty then abort the transaction
+    abortTransactionUnless(pendingInflightPackets(portIdentifier, channelIdentifier))
+
+    // move channel to OPEN and adopt upgrade parameters
+    openChannelHandshake(portIdentifier, channelIdentifier)
+
+    // call modules onChanUpgradeConfirm callback
+    module = lookupModule(portIdentifier)
+    // confirm callback must not return error since counterparty successfully upgraded
+    module.onChanUpgradeOpen(
+        portIdentifer,
+        channelIdentifier
+    )
 }
 ```
 
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -312,18 +312,11 @@ function startFlushUpgradeHandshake(
     if counterpartyUpgradeSequence != channel.upgradeSequence {
         // error on the higher sequence so that both chains move to a fresh sequence
         maxSequence = max(counterpartyUpgradeSequence, channel.upgradeSequence)
-        errorReceipt = ErrorReceipt{
-            sequence: maxSequence,
-            errorMsg: ""
-        }
-        provableStore.set(channelUpgradeErrorPath(portIdentifier, channelIdentifier), errorReceipt)
-        provableStore.set(channelUpgradeSequencePath(portIdentifier, channelIdentifier), maxSequence)
-        return
-    }
-
-    // timeouts on upgrade should be the same
-    if upgrade.timeout != counterpartyUpgrade.timeout {
+        currentChannel.UpgradeSequence = maxSequence
+        provableStore.set(channelPath(portIdentifier, channelIdentifier), currentChannel)
+        
         restoreChannel(portIdentifier, channelIdentifier)
+        return
     }
 
     // proposed ordering must be the same as the counterparty proposed ordering
@@ -462,15 +455,21 @@ function chanUpgradeInit(
     // call modules onChanUpgradeInit callback
     module = lookupModule(portIdentifier)
     version, err = module.onChanUpgradeInit(
-        proposedUpgrade.fields.ordering,
-        proposedUpgrade.fields.connectionHops,
         portIdentifier,
         channelIdentifier,
+        proposedUpgrade.fields.ordering,
+        proposedUpgrade.fields.connectionHops,
         upgradeSequence,
         proposedUpgrade.fields.version
     )
     // abort transaction if callback returned error
     abortTransactionUnless(err != nil)
+
+    // replace channel version with the version returned by application
+    // in case it was modified
+    currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+    currentChannel.version = version
+    provableStore.set(channelPath(portIdentifier, channelIdentifier), currentChannel)
 }
 ```
 
@@ -518,8 +517,8 @@ function chanUpgradeTry(
         // so that both channel ends are using the same sequence for the current upgrade
         // initUpgradeChannelHandshake will increment the sequence so after that call
         // both sides will have the same upgradeSequence
-        if counterpartyUpgradeSequence > channel.upgradeSequence {
-            channel.upgradeSequence = counterpartyUpgradeSequence - 1
+        if counterpartyUpgradeSequence > currentChannel.upgradeSequence {
+            currentChannel.upgradeSequence = counterpartyUpgradeSequence - 1
         }
 
         initUpgradeChannelHandshake(portIdentifier, channelIdentifier, upgradeFields, counterpartyUpgrade.timeout)
@@ -542,6 +541,7 @@ function chanUpgradeTry(
         connectionHops: counterpartyHops,
         version: currentChannel.version,
         sequence: counterpartyUpgradeSequence,
+        flushStatus: NOTINFLUSH,
     }
 
     // call startFlushUpgrade handshake to move channel from INITUPGRADE to TRYUPGRADE and start flushing
@@ -650,7 +650,7 @@ function chanUpgradeAck(
 
     // if both sides have already flushed then open the upgrade handshake immediately
     if  currentChannel.state == FLUSHCOMPLETE && counterpartyFlushStatus == FLUSHCOMPLETE {
-        openChannelHandshake(portIdentifier, channelIdentifier)
+        openUpgradelHandshake(portIdentifier, channelIdentifier)
         module.onChanUpgradeOpen(portIdentifier, channelIdentifier)
     }
 }
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -15,7 +15,7 @@ As new features get added to IBC, chains may wish to take advantage of new chann
 - The channel upgrade protocol is atomic, i.e., 
   - either it is unsuccessful and then the channel MUST fall-back to the original channel parameters; 
   - or it is successful and then both channel ends MUST adopt the new channel parameters and the applications must process packet data appropriately.
-- Packets sent under the previously negotiated parameters must be processed under the previously negotiated parameters, packets sent under the newly negotiated parameters must be processed under the newly negotiated parameters.
+- Packets sent under the previously negotiated parameters must be processed under the previously negotiated parameters, packets sent under the newly negotiated parameters must be processed under the newly negotiated parameters. Thus, in-flight packets sent before upgrade handshake is complete will be processed according to the original parameters.
 - The channel upgrade protocol MUST NOT modify the channel identifiers.
 
 ## Technical Specification
@@ -87,7 +87,7 @@ interface UpgradeFields {
 
 MAY BE MODIFIED:
 - `version`: The version MAY be modified by the upgrade protocol. The same version negotiation that happens in the initial channel handshake can be employed for the upgrade handshake.
-- `ordering`: The ordering MAY be modified by the upgrade protocol.
+- `ordering`: The ordering MAY be modified by the upgrade protocol so long as the new ordering is supported by underlying connection.
 - `connectionHops`: The connectionHops MAY be modified by the upgrade protocol.
 
 MUST NOT BE MODIFIED:
@@ -122,7 +122,7 @@ interface Upgrade {
 }
 ```
 
-The upgrade contains the proposed upgrade for the channel end on the executing chain, the timeout for the upgrade attempt, and the last packet send sequence for the channel. The `lastPacketSent` allows the counterparty to know which packets need to be flushed before the channel can reopen with the newly negotiated parameters.
+The upgrade contains the proposed upgrade for the channel end on the executing chain, the timeout for the upgrade attempt, and the last packet send sequence for the channel. The `lastPacketSent` allows the counterparty to know which packets need to be flushed before the channel can reopen with the newly negotiated parameters. Any packet sent to the channel end with a packet sequence above the `lastPacketSent` will be rejected until the upgrade is complete.
 
 #### `ErrorReceipt`
 
@@ -138,7 +138,7 @@ interface ErrorReceipt {
 
 ### Store Paths
 
-#### Upgrade Channel Path
+#### Channel Upgrade Path
 
 The chain must store the proposed upgrade upon initiating an upgrade. The proposed upgrade must be stored in the provable store. It may be deleted once the upgrade is successful or has been aborted.
 
@@ -171,7 +171,7 @@ function verifyChannelUpgrade(
 The chain must store the counterparty's last packet sequence on `startFlushUpgradeHandshake`. This will be stored in the `counterpartyLastPacketSequence` path on the private store.
 
 ```typescript
-function channelCounterpartyLastPacketSequencePath(portIdentifier: Identifier, channelIdentifier: ChannelIdentifier): Path {
+function channelCounterpartyLastPacketSequencePath(portIdentifier: Identifier, channelIdentifier: Identifier): Path {
     return "channelUpgrades/counterpartyLastPacketSequence/ports/{portIdentifier}/channels/{channelIdentifier}"
 }
 ```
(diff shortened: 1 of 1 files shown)


spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -350,94 +374,57 @@ Access control on counterparty should inform choice of timeout values, i.e. time
...
         if counterpartyUpgradeSequence > channel.upgradeSequence {
-            channel.upgradeSequence = counterpartyUpgradeSequence
+            channel.upgradeSequence = counterpartyUpgradeSequence - 1
         }
 
-        // this is first message in upgrade handshake on this chain so we must store original channel in restore channel path
-        // in case we need to restore channel later.
-        privateStore.set(channelRestorePath(portIdentifier, channelIdentifier), currentChannel)
-        provableStore.set(channelPath(portIdentifier, channelIdentifier), channel)
+        initUpgradeChannelHandshake(portIdentifier, channelIdentifier, upgradeFields, counterpartyUpgrade.timeout)
```
charleenfei, line 418: don't we need to write the upgrade init in order for anything else to happen other than the channel upgrade being set? if we aren't writing the channel state, then we need to manually set or we are only incrementing the sequence, which seems like then going thru the whole handshake is overkill?

see comment here: https://github.com/cosmos/ibc-go/pull/3478/files#r1175181578

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -488,69 +490,57 @@ NOTE: It is up to individual implementations how they will provide access-contro
...
+
+    // construct counterpartyChannel from existing information and provided
+    // counterpartyUpgradeSequence
+    counterpartyChannel = ChannelEnd{
+        state: TRYUPGRADE,
+        ordering: currentChannel.ordering,
+        counterpartyPortIdentifier: portIdentifier,
+        counterpartyChannelIdentifier: channelIdentifier,
+        connectionHops: counterpartyHops,
+        version: currentChannel.version,
```
charleenfei, line 513: why is this using the current channel version? wouldnt this be incorrect on asymmetrical versions ie: interchain accounts?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
AdityaSripal: Both INIT and TRY handlers will use the `initUpgradeHandshake` utility function to execute common logic if the message is the first message the executing chain has seen for the given upgrade attempt.

See utility function here: https://github.com/cosmos/ibc/pull/967/files#diff-69d7c4ac1d4cdeeabcc757d3a34e85945ebc5613a034a52ebd0769cd8a0fc560R214

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -32,12 +33,12 @@ enum ChannelState {
   TRYOPEN,
   OPEN,
   INITUPGRADE,
-  TRYUPGRADE,
+  BLOCKUPGRADE,
```
colin-axner, line 36: In general I like the `BLOCK` state naming, but I could see it being mildly confusing (block upgrade, sounds like you are halting the upgrade, which is partially true). Just curious if there's any alternatives we should consider?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -49,16 +50,27 @@ interface ChannelEnd {
   counterpartyChannelIdentifier: Identifier
   connectionHops: [Identifier]
   version: string
+  upgradeSequence: uint64
```
colin-axner, line 53: This should be updated outside of the upgrades spec as well?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -95,42 +121,42 @@ interface ErrorReceipt {
 
 ### Store Paths
 
-#### Upgrade Sequence Path
+#### Upgrade Channel Path
 
-The upgrade sequence path is a public path that stores the current sequence of the upgrade attempt. The sequence will increment with each attempted upgrade on the given channel. The sequence will be used to ensure that different error receipts referring to different upgrade attempts do not interfere with each other.
+The chain must store the proposed upgrade until the upgrade handshake successfully. This must be stored in the publice store. It may be deleted once the upgrade is successful.
```
colin-axner, line 126:
```suggestion
The chain must store the proposed upgrade upon initiating an upgrade. The proposed upgrade must be stored in the public store. It may be deleted once the upgrade is successful or has been aborted.
```

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -95,42 +121,42 @@ interface ErrorReceipt {
...
-function channelUpgradeSequencePath(portIdentifier: Identifier, channelIdentifier: Identifier) Path {
-    return "channelUpgrades/upgradeSequence/ports/{portIdentifier}/channels/{channelIdentifier}"
-}
+function channelUpgradePath(portIdentifier: Identifier, channelIdentifier: Identifier): Path {
+    return "channelUpgrades/upgrades/ports/{portIdentifier}/channels/{channelIdentifier}"
+ }
 ```
 
-The upgrade sequence MUST also have a verification method so that chains can prove the upgrade sequence on the counterparty for the given channel upgrade.
+The upgrade path has an associated verification membership added to the connection interface so that a counterparty may verify that chain has stored the provided counterparty upgrade.
```
colin-axner, line 134:
```suggestion
The upgrade path has an associated verification of membership method added to the connection interface so that a counterparty may verify that chain has stored the provided counterparty upgrade.
```
any wording is fine, but "verification membership" didn't read well to me

colin-axner (commented):
partial review, left off on verifyChannelUpgrade

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
chatton: Do we need to change this enum at all? I feel like it could remain as TRYUPGRADE (consistent with other handshakes) but the spec can outline that being in this state blocks new packets.

colin-axner: Semantically the upgrade steps are different then the opening handshake. It's no longer 4 steps, it is either 5 or 6:
chainA, chainB
init
block, block
open, open

or 
init, init
block, block
open, open

It is required both sides enter into the block state, whereas in the opening handshake you can skip over try (init -> open)

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
AdityaSripal: Will do once this doc stabilizes

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
charleenfei: is this still the latest version? Or are we removing the BLOCKUPGRADE state?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -176,88 +219,189 @@ function verifyChannelUpgradeErrorAbsence(
...
+
+### Utility Functions
 
-The upgrade timeout path is a public path set by the upgrade initiator to determine when the TRY step should timeout. It stores the `timeoutHeight` and `timeoutTimestamp` by which point the counterparty must have progressed to the TRY step. The TRY step will prove the timeout values set by the initiating chain and ensure the timeout has not passed. Or in the case of a timeout, in which case counterparty proves that the timeout has passed on its chain and restores the channel.
+`initUpgradeChannelHandshake` is a subprotocol that will initialize the channel end for the upgrade handshake. It will verify the upgrade parameters and set the channel state to INITUPGRADE and block sendpackets. During this time; `receivePacket`, `acknowledgePacket` and `timeoutPacket` will still be allowed and processed according to the original channel parameters. The new proposed upgrade will be stored in the public store for counterparty verification.
```
damiannolan, line 228: wording nits/suggestions... maybe we should use validate instead of verify for the initial proposed upgrade params?

```suggestion
`initUpgradeChannelHandshake` is a sub-protocol that will initialize the channel end for the upgrade handshake. It will validate the upgrade parameters and set the channel state to INITUPGRADE, blocking `sendPacket` from processing outbound packets on the channel end. During this time; `receivePacket`, `acknowledgePacket` and `timeoutPacket` will still be allowed and processed according to the original channel parameters. The new proposed upgrade will be stored in the provable store for counterparty verification.
```

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -176,88 +219,189 @@ function verifyChannelUpgradeErrorAbsence(
...
+    // channel upgrade to be accepted
+    proposedConnection = provableStore.Get(connectionPath(proposedFields.connectionHops[0])
+    abortTransactionUnless(proposedConnection != null && proposedConnection.state == OPEN)
+
+    // either timeout height or timestamp must be non-zero
+    abortTransactionUnless(timeout.timeoutHeight != 0 || timeout.timeoutTimestamp != 0)
+
+    // get last packet sent on channel and set it in the upgrade struct
+    // last packet sent is the nextSequenceSend on the channel minus 1
+    lastPacketSendSequence = publicStore.get(nextSequenceSendPath(portIdentifier, channelIdentifier)) - 1
```
damiannolan, line 254:
```suggestion
    lastPacketSendSequence = provableStore.get(nextSequenceSendPath(portIdentifier, channelIdentifier)) - 1
```

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -176,88 +219,189 @@ function verifyChannelUpgradeErrorAbsence(
...
+    upgrade = Upgrade{
+        fields: proposedFields,
+        timeout: timeout,
+        lastPacketSent: lastPacketSendSequence,
+    }
+
+    // store upgrade in public store for counterparty proof verification
+    publicStore.set(channelUpgradePath(portIdentifier, channelIdentifier), upgrade)
```
damiannolan, line 262:
```suggestion
    provableStore.set(channelUpgradePath(portIdentifier, channelIdentifier), upgrade)
```

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -176,88 +219,189 @@ function verifyChannelUpgradeErrorAbsence(
...
+
+    currentChannel.sequence = currentChannel.sequence + 1
+    currentChannel.state = INITUPGRADE
+    publicStore.set(channelPath(portIdentifier, channelIdentifier), channel)
```
damiannolan, line 266: shouldn't these all be "provableStore"? 

```suggestion
    provableStore.set(channelPath(portIdentifier, channelIdentifier), channel)
```

damiannolan (commented):
Nice work! I left a bunch of nits/suggestions and some questions. Will continue to review in more detail tomorrow

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -49,16 +55,39 @@ interface ChannelEnd {
...
+    version: string
+    ordering: ChannelOrder
+    connectionHops: [Identifier]
+}
+```
 
 MAY BE MODIFIED:
 - `version`: The version MAY be modified by the upgrade protocol. The same version negotiation that happens in the initial channel handshake can be employed for the upgrade handshake.
-- `ordering`: The ordering MAY be modified by the upgrade protocol. However, it MUST be the case that the previous ordering is a valid subset of the new ordering. Thus, the only supported change is from stricter ordering rules to less strict ordering. For example, switching from ORDERED to UNORDERED is supported, switching from UNORDERED to ORDERED is **unsupported**.
+- `ordering`: The ordering MAY be modified by the upgrade protocol.
```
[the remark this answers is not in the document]
AdityaSripal: Correct, the ordering must still be in the supported set of the underlying connection

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -81,6 +110,20 @@ interface UpgradeTimeout {
 
 At least one of the `timeoutHeight` or `timeoutTimestamp` MUST be non-zero.
```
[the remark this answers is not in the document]
AdityaSripal: This already exists on line 102, hidden since its not in diffs

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -95,42 +138,42 @@ interface ErrorReceipt {
...
+    path = applyPrefix(connection.counterpartyPrefix, channelUpgradePath(counterpartyPortIdentifier, counterpartyChannelIdentifier))
+    return verifyMembership(clientState, height, 0, 0, proof, path, upgrade)
 }
 ```
 
-#### Restore Channel Path
+#### CounterpartyLastPacketSequence Path
 
-The chain must store the previous channel end so that it may restore it if the upgrade handshake fails. This may be stored in the private store.
+The chain must store the counterparty's last packet sequence on `startFlushUpgradeHandshake`. This will be stored in the `counterpartyLastPacketSequence` path on the private store/
```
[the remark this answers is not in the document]
AdityaSripal: Does not need to be provable to counterparty. This is a moot point for current SDK since all storage is provable, but they are apparently changing this, in which case we can move any storage set in private store in the specs to this non-provable store in the sdk

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -49,16 +55,39 @@ interface ChannelEnd {
...
+enum FlushStatus {
+    NOTINFLUSH
+    FLUSHING
+    FLUSHCOMPLETE
+}
+```
+
+FlushStatus will be in `NOTINFLUSH` state when the channel is not in an upgrade handshake. It will be in `FLUSHING` mode when the channel end is flushing in-flight packets. The FlushStatus will change to `FLUSHCOMPLETE` once there are no in-flight packets left and the channelEnd is ready to move to OPEN.
+
+All other parameters will remain the same during the upgrade handshake until the upgrade handshake completes. When the channel is reset to `OPEN` on a successful upgrade handshake, the fields on the channel end will be switched over to the `UpgradeFields` specified in the upgrade.
```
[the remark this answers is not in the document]
AdityaSripal: Yea wasn't sure exactly how we wanted to handle this. I created 3 states: 

FLUSHING: flushing (during handshake)
FLUSHCOMPLETE: flushing complete (during handshake)
NOTINFLUSH: state when we are not in upgrade handshake

This NOTINFLUSH state is the default enum (0 in the case of golang). I thought it was weird to call it `UNINITIALIZED` since it will be a state we return to after a complete upgrade handshake.

I could also have done 2 states:

`FLUSHED`
`FLUSHING`

Then while packets are processing during normal channel execution, the flush state will be `FLUSHED`; it will only be FLUSHING while we are paused on upgrade handshake to flush packets.

However, I thought this might be confusing so I opted against it.

cc: @colin-axner

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
AdityaSripal: No, we realized the above had an issue. I've added two new states to the ChannelState enum: `TRYUPGRADE` and `ACKUPGRADE`.

And then a separate `FLUSHSTATUS`. Please implement only the ChannelState enum changes for now.

We will deal with FLUSHSTATUS in a future PR

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -81,6 +110,20 @@ interface UpgradeTimeout {
...
+
+```typescript
+interface Upgrade {
+    fields: UpgradeFields
+    timeout: UpgradeTimeout
+    lastPacketSent: uint64
+}
+```
+
+The upgrade contains the proposed upgrade for the channel end on the executing chain, the timeout for the upgrade attempt, and the last packet send sequence for the channel. The `lastPacketSent` allows the counterparty to know which packets need to be flushed before the channel can reopen with the newly negotiated parameters.
```
crodriguezvega, line 125: Could you elaborate a bit more how `lastPacketSent` should be used in the counterparty? I am wondering now if this is needed, since a chain can take care of flushing all the packets that it has sent, but doesn't need to worry about the packets sent by the counterparty. If the chain doesn't allow sending new packets, it should not be possible for the counterparty to receive packets with a sequence higher than `lastPacketSent`, right?

crodriguezvega (commented):
Thanks, @AdityaSripal. I left a bunch of comments, some of them nits or inconsistencies (I stopped adding more after I realised that the fine details still need to be worked out 😅).

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -350,94 +442,57 @@ Access control on counterparty should inform choice of timeout values, i.e. time
...
+            channel.upgradeSequence = counterpartyUpgradeSequence - 1
         }
 
-        // this is first message in upgrade handshake on this chain so we must store original channel in restore channel path
-        // in case we need to restore channel later.
-        privateStore.set(channelRestorePath(portIdentifier, channelIdentifier), currentChannel)
-        provableStore.set(channelPath(portIdentifier, channelIdentifier), channel)
+        initUpgradeChannelHandshake(portIdentifier, channelIdentifier, upgradeFields, counterpartyUpgrade.timeout)
+    } else if currentChannel.state == INITUPGRADE {
+        existingUpgrade = publicStore.get(channelUpgradePath)
```
charleenfei, line 488: do we expect the upgrade sequence to be fastforwarded in this case as well if the counterparty upgrade sequence is higher in crossing hellos?

damiannolan:
Regarding ordering and now being able to freely change regardless of "strictness", I think we need to still add the pseudo code for verifying the ordering is supported by the underlying connection. see https://github.com/cosmos/ibc-go/issues/3635#issuecomment-1559195828

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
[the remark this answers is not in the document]
AdityaSripal: Open to changing with consensus. cc: @colin-axner

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -350,94 +442,57 @@ Access control on counterparty should inform choice of timeout values, i.e. time
...
-    // proposedConnection must exist and be in OPEN state
-    // connectionHops can change in a channelUpgrade, however both sides must still be each other's counterparty.
-    proposedConnection = provableStore.Get(connectionPath(proposedUpgradeChannel.ConnectionHops[0])
-    abortTransactionUnless(proposedConnection != null && proposedConnection.state == OPEN)
-    abortTransactionUnless(counterpartyChannel.connectionHops[0] == proposedConnection.counterpartyConnectionIdentifier)
+    // create upgrade fields for this chain from counterparty upgrade and relayer-provided information
+    // version may be mutated by application callback
+    upgradeFields = Upgrade{
+        ordering: counterpartyUpgrade.fields.ordering,
+        connectionHops: proposedConnectionHops,
```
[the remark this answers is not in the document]
AdityaSripal: The connectionHops are not the same on both sides since each references the connection with their own local identifier

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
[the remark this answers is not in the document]
colin-axner: I actually like the 3 enums states @AdityaSripal proposed here. To me, it makes sense that `FLUSHCOMPLETE` is a very specific state, it tells the caller that the channel has no inflight packets, while `NOTINFLUSH` or `NOTFLUSHING` just means it isn't blocking on flushing in-flight packets. 

What's the benefit to reducing the number of states?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -176,88 +219,193 @@ function verifyChannelUpgradeErrorAbsence(
...
+    }
+
+    // connectionHops can change in a channelUpgrade, however both sides must still be each other's counterparty.
+    proposedConnection = provableStore.get(connectionPath(proposedUpgradeFields.connectionHops[0])
+    if (proposedConnection == null || proposedConnection.state != OPEN) {
+        restoreChannel(portIdentifier, channelIdentifier)
+    }
+    if (counterpartyUpgrade.fields.connectionHops[0] != proposedConnection.counterpartyConnectionIdentifier) {
+        restoreChannel(portIdentifier, channelIdentifier)
+    }
```
colin-axner, line 311: We should add this check?
```
if (channel.UpgradeSequence != counterpartyChannel.UpgradeSequence) {
    restoreChannel(portIdentifier, channelIdentifier)
}
```

that is, the general gist of this function:
```
    // verify fields
    
    // return error receipts for any upgrade fields that are incompatible
    
    // update self state to start flushing
```

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -176,88 +219,193 @@ function verifyChannelUpgradeErrorAbsence(
...
+    proofHeight: Height
 ) {
-    clientState = queryClientState(connection.clientIdentifier)
-    path = applyPrefix(connection.counterpartyPrefix, channelUpgradeTimeoutPath(counterpartyPortIdentifier, counterpartyChannelIdentifier))
-    return verifyMembership(clientState, height, 0, 0, proof, path, upgradeTimeout)
+    abortTransactionUnless(channelState == TRYUPGRADE || channelState == ACKUPGRADE)
+
+    // current channel must be OPEN
+    currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+    abortTransactionUnless(channel.state == OPEN)
```
damiannolan, line 289: In `chanUpgradeAck` on line 574 we lookup the current channel and do the following:

```
currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
abortTransactionUnless(currentChannel.state == INITUPGRADE || currentChannel.state == TRYUPGRADE)
```

And then we call `startFlushUpgradeHandshake` which makes this assertion.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -488,123 +563,155 @@ NOTE: It is up to individual implementations how they will provide access-contro
...
+    if  currentChannel.state == FLUSHCOMPLETE && counterpartyFlushStatus == FLUSHCOMPLETE {
+        openChannelHandshake(portIdentifier, channelIdentifier)
+        module.onChanUpgradeOpen(portIdentifier, channelIdentifier)
+    }
 }
 ```
 
 ```typescript
-function chanUpgradeConfirm(
+function chanUpgradeOpen(
```
damiannolan, line 641: Is there good reason to stray from the `Confirm` naming convention? I find it a bit more natural, as open is kind of close to `Init` in some ways

charleenfei: I guess since the channel state is changing to `OPEN` if the in flight packets have been flushed, it might be confusing to have the function called `Confirm` when the end state will actually `OPEN`?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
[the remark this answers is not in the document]
damiannolan: Was discussing with @colin-axner this morning and I had suggested also using a `flushing` boolean value. Either works I guess, I feel like I'd lean slightly towards the boolean.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -350,96 +446,79 @@ Access control on counterparty should inform choice of timeout values, i.e. time
...
+        counterpartyPortIdentifier: portIdentifier,
+        counterpartyChannelIdentifier: channelIdentifier,
+        connectionHops: counterpartyHops,
+        version: currentChannel.version,
+        sequence: counterpartyUpgradeSequence,
     }
 
+    // call startFlushUpgrade handshake to move channel from INITUPGRADE to TRYUPGRADE and start flushing
+    // upgrade is blocked on this channelEnd from progressing until flush completes on both ends
+    startFlushUpgradeHandshake(portIdentifier, channelIdentifier, upgradeFields, counterpartyChannel, counterpartyUpgrade, TRYUPGRADE, proofChannel, proofUpgrade, proofHeight)
```
charleenfei, line 514: at this point in in the code when `startFlushUpgradeHandshake` is called, in the non-crossing hellos scenario our own `channel.State` would still be in `UPGRADEINIT`. wouldn't this mean that the `startFlush...` subprotocol would fail on [its first check? ](https://github.com/cosmos/ibc/pull/967/files#diff-69d7c4ac1d4cdeeabcc757d3a34e85945ebc5613a034a52ebd0769cd8a0fc560R285)

charleenfei: ah, nevermind, it's the passed in `ChannelState` param and not our own channel state that is being checked. maybe a clearer naming, something like `NextChannelState`

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -488,123 +563,155 @@ NOTE: It is up to individual implementations how they will provide access-contro
...
 
-    // counterparty must be in OPEN state
-    abortTransactionUnless(counterpartyChannel.State == OPEN)
-
-    // get current sequence
-    sequence = provableStore.get(channelUpgradeSequencePath(portIdentifier, channelIdentifier))
+    // currentChannel must be in TRYUPGRADE or ACKUPGRADE and have completed flushing
+    currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+    abortTransactionUnless(currentChannel.state == TRYUPGRADE || currentChannel.state == ACKUPGRADE)
+    abortTransactionUnless(currentChannel.flushStatus == FLUSHCOMPLETE)
```
charleenfei, line 654: is there another way to set our `currentChannel.flushStatus` to `FLUSHCOMPLETE` besides this code snippet in `StartFlushingUpgradeHandshake`? 

```
    // if there are no in-flight packets on our end, we can automatically go to FLUSHCOMPLETE
    if pendingInflightPackets(portIdentifier, channelIdentifier) == nil {
        currentChannel.flushState = FLUSHCOMPLETE
    }
   ```

we call this in `Ack` and `Try` but we can't recall either of them once our channel state is in `UPGRADEACK`.

File: spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md [hunk 23]
```diff
@@ -488,123 +563,155 @@ NOTE: It is up to individual implementations how they will provide access-contro
...
+    // if no error, agree on final version
+    upgrade.version = counterpartyUpgrade.version
+    provableStore.set(channelUpgradePath(portIdentifier, channelIdentifier), upgrade)
+
+    // refresh channel
+    currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+
+    // if both sides have already flushed then open the upgrade handshake immediately
+    if  currentChannel.state == FLUSHCOMPLETE && counterpartyFlushStatus == FLUSHCOMPLETE {
+        openChannelHandshake(portIdentifier, channelIdentifier)
```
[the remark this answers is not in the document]
AdityaSripal: this is the openUpgradeHandshake subprotocol

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -488,123 +559,154 @@ NOTE: It is up to individual implementations how they will provide access-contro
...
-    // this means that the ordering must be the same and 
-    // any future introduced fields that must be compatible
-    // should also be checked
-    if counterpartyChannel.ordering != proposedUpgradeChannel.ordering {
-        restoreChannel(portIdentifier, channelIdentifier)
-        return
-    }
+    // prove counterparty and move our own state to ACKUPGRADE and start flushing
+    // upgrade is blocked on this channelEnd from progressing until flush completes on both ends
+    startFlushUpgradeHandshake(portIdentifier, channelIdentifier, upgrade.fields, counterpartyChannel, counterpartyUpgrade, ACKUPGRADE, proofChannel, proofUpgrade, proofHeight)
```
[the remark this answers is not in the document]
AdityaSripal: Doing this directly became very verbose in the spec and args list, so I keep as-is and simply verify the timeout in the flushing subprotocol.

I do remove the sequence however since we already know this information by this point

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
AdityaSripal: This will be done in the packet handlers in a future spec PR. this pr is purely focused on upgrade handshake changes to scope the changes.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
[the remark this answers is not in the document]
AdityaSripal: Let's discuss this synchronously

AdityaSripal:
> @colin-axner and I were chatting this morning and I suggested the possibility of enforcing both parties to perform the INIT step of the upgrade handshake. This may actually clean up/improve some of the handshake handler logic or reduce some code a little bit. From a ux perspective (or how upgrades will be used in practice) I think it might feel more natural for both chains to have to carry out the INIT step.

Let's discuss the pros and cons of this synchronously

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
charleenfei: that sounds good to me!

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
charleenfei: is it important in this spec that these handlers are at least mentioned? otherwise the logic for the onUpgradeOpen seems imcomplete.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -176,88 +219,189 @@ function verifyChannelUpgradeErrorAbsence(
...
+    proofHeight: Height
 ) {
-    clientState = queryClientState(connection.clientIdentifier)
-    path = applyPrefix(connection.counterpartyPrefix, channelUpgradeTimeoutPath(counterpartyPortIdentifier, counterpartyChannelIdentifier))
-    return verifyMembership(clientState, height, 0, 0, proof, path, upgradeTimeout)
+    abortTransactionUnless(channelState == TRYUPGRADE || channelState == ACKUPGRADE)
+
+    // get underlying connection for proof verification
+    connection = getConnection(currentChannel.connectionIdentifier)
+    counterpartyHops = getCounterpartyHops(connection)
```
charleenfei, line 304: why do we need this? we are receiving the `counterpartyChannel` already as a param

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -639,14 +766,15 @@ function cancelChannelUpgrade(
     abortTransactionUnless(verifyChannelUpgradeError(connection, proofHeight, proofUpgradeError, currentChannel.counterpartyPortIdentifier, currentChannel.counterpartyChannelIdentifier, errorReceipt))
 
     // cancel upgrade
-    // and restore original conneciton
+    // and restore original channel
     // delete unnecessary state
-    originalChannel = privateStore.get(channelRestorePath(portIdentifier, channelIdentifier))
+    currentChannel.state = OPEN
+    currentChannel.flushStatus = NOTINFLUSH
     provableStore.set(channelPath(portIdentifier, channelIdentifier), originalChannel)
```
colin-axner, line 773: should we fast forward our upgrade sequence upon canceling an upgrade if the error receipt is > our upgrade sequence?

colin-axner (approved):
Wahoo! Excellent work! :tada: 

Finally made it through all the changes! Leaving my approval as I think most everything can be addressed in followups. My main outstanding points of discussion:
- timeouts needing to be equal?
- enforcing init on both chains
- bool vs enum for flushing status

One side note, the upgrade sequence is somewhat of a tricky field since it changes often and it needs to be consistent for proofs. I think it'd be very valuable to add a section discussing when it changes, when it is fast forwarded and after what state it must not be modified, maybe also scenarios when sequences go out of sync. There's also a subtle check in canceling upgrades that the error receipt is >= current sequence. If it was just equality the handshake would break down, so I think that's also worth a direct mention outside of the pseudo-code

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -95,42 +138,42 @@ interface ErrorReceipt {
...
+#### CounterpartyLastPacketSequence Path
 
-The chain must store the previous channel end so that it may restore it if the upgrade handshake fails. This may be stored in the private store.
+The chain must store the counterparty's last packet sequence on `startFlushUpgradeHandshake`. This will be stored in the `counterpartyLastPacketSequence` path on the private store/
 
 ```typescript
-function channelRestorePath(portIdentifier: Identifier, channelIdentifier: Identifier): Path {
-    return "channelUpgrades/restore/ports/{portIdentifier}/channels/{channelIdentifier}"
- }
+function channelCounterpartyLastPacketSequencePath(portIdentifier: Identifier, channelIdentifier: ChannelIdentifier): Path {
```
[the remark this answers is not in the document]
colin-axner: Just noting in ibc-go the current usage is [`latest`](https://github.com/cosmos/ibc-go/blob/04-channel-upgrades/proto/ibc/core/channel/v1/upgrade.proto#L17). I'm happy with whatever but we should probably be consistent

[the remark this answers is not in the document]
colin-axner: previous discussion: https://github.com/cosmos/ibc-go/pull/3451#discussion_r1166434611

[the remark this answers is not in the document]
chatton: I think I now also prefer `last` here after thinking some more about it.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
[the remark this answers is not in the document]
DimitrisJim: ah, the boolean vs enum battle appears once again

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -350,111 +482,75 @@ Access control on counterparty should inform choice of timeout values, i.e. time
...
+    // if OPEN, then initialize handshake with upgradeFields
+    // otherwise, assert that the upgrade fields are the same for crossing-hellos case
+    if currentChannel.state == OPEN {
         // if the counterparty sequence is greater than the current sequence, we fast forward to the counterparty sequence
         // so that both channel ends are using the same sequence for the current upgrade
+        // initUpgradeChannelHandshake will increment the sequence so after that call
+        // both sides will have the same upgradeSequence
         if counterpartyUpgradeSequence > channel.upgradeSequence {
-            channel.upgradeSequence = counterpartyUpgradeSequence
+            channel.upgradeSequence = counterpartyUpgradeSequence - 1
```
DimitrisJim, line 523: should be `currentChannel` here and on line 522, right?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -176,88 +219,229 @@ function verifyChannelUpgradeErrorAbsence(
...
+
+    currentChannel.state = desiredChannelState
+    currentChannel.flushState = FLUSHING
+
+    // if there are no in-flight packets on our end, we can automatically go to FLUSHCOMPLETE
+    if pendingInflightPackets(portIdentifier, channelIdentifier) == nil {
+        currentChannel.flushState = FLUSHCOMPLETE
+    }
+
+    publicStore.set(channelPath(portIdentifier, channelIdentifier), channel)
```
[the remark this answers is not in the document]
DimitrisJim: A suggestion on a suggestion:

```suggestion
    provableStore.set(channelPath(portIdentifier, channelIdentifier), currentChannel)
```

there's a couple more references to `publicStore` in other places.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -488,123 +584,154 @@ NOTE: It is up to individual implementations how they will provide access-contro
...
+    // construct counterpartyChannel from existing information and provided
+    // flushStatus
+    counterpartyChannel = ChannelEnd{
+        state: TRYUPGRADE,
+        ordering: currentChannel.ordering,
+        counterpartyPortIdentifier: portIdentifier,
+        counterpartyChannelIdentifier: channelIdentifier,
+        connectionHops: counterpartyHops,
+        version: currentChannel.version,
+        sequence: channel.sequence,
```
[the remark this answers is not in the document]
DimitrisJim: what does channel refer to here?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -488,123 +584,154 @@ NOTE: It is up to individual implementations how they will provide access-contro
 function chanUpgradeAck(
     portIdentifier: Identifier,
     channelIdentifier: Identifier,
-    counterpartyChannel: ChannelEnd,
+    counterpartyFlushStatus: FlushStatus,
+    counterpartyUpgrade: Upgrade,
```
DimitrisJim, line 588: is there any benefit to passing the FlushStatus separately here?

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -462,15 +455,21 @@ function chanUpgradeInit(
...
         proposedUpgrade.fields.version
     )
     // abort transaction if callback returned error
     abortTransactionUnless(err != nil)
+
+    // replace channel version with the version returned by application
+    // in case it was modified
+    currentChannel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+    currentChannel.version = version
+    provableStore.set(channelPath(portIdentifier, channelIdentifier), currentChannel)
```
colin-axner, line 472: I believe you should setting the upgrade, not the channel

PR #972, opened 2023-05-12
ics 06: putting sequence in `ClientState`

crodriguezvega:
[The `sequence` is in the `ClientState` in ibc-go](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/06-solomachine/client_state.go#L24).


DimitrisJim (approved):
looked through the source and lgtm

PR #974, opened 2023-05-26
Fix mistyped variable name

thpani:
ICS 20, 100, and 721 contain a mistyped variable name in the pseudocode definition of `onAcknowledgePacket`:
The formal parameter name is `acknowledgement`, but in the function body it is referred to by `ack`.

Since this is a small fix, I did not open an issue or update the ICS metadata. Please lmk if I should do so.

&nbsp;
&nbsp;
_Found while writing the [ICS-20 spec in Quint](https://github.com/informalsystems/quint/tree/main/examples/cosmos/ics20), simply by copy-pasting._


crodriguezvega (approved):
Thanks for this fix, @thpani!

PR #976, opened 2023-05-30
#draft #wip data/ics-1001 just for discussion, not for merge

dzmitry-lahoda:


crodriguezvega:
Hi @dzmitry-lahoda, thanks for opening this PR. However, this spec seems to be still in the early stages and it's probably hard to give any meaningful feedback at the moment. If you want to discuss an idea I think it would make more sense to open an issue first. What do you think?

crodriguezvega:
Since this PR is still in draft and seems to be stale, I am going to close it for now. Feel free to re-open when you have a version we can review, even though ideally it would be better to discuss your idea first in an issue. Thank you!

Issue #965, opened 2023-04-19
ICS04: Something confusing about function timeoutPacket and timeoutOnClose


michwqy:
According to 
```javascript
function recvPacket(
  packet: OpaquePacket,
  proof: CommitmentProof,
  proofHeight: Height,
  relayer: string): Packet {
...
    switch channel.order {
...
      case ORDERED_ALLOW_TIMEOUT:
        if (getConsensusHeight() >= packet.timeoutHeight && packet.timeoutHeight != 0) || (currentTimestamp() >= packet.timeoutTimestamp && packet.timeoutTimestamp != 0) {
          nextSequenceRecv = nextSequenceRecv + 1
          provableStore.set(nextSequenceRecvPath(packet.destPort, packet.destChannel), nextSequenceRecv)
          provableStore.set(
          packetReceiptPath(packet.destPort, packet.destChannel, packet.sequence),
            TIMEOUT_RECEIPT
          )
        }
        return;
...
    }
...
    switch channel.order {
...
      case ORDERED_ALLOW_TIMEOUT:
        nextSequenceRecv = nextSequenceRecv + 1
        provableStore.set(nextSequenceRecvPath(packet.destPort, packet.destChannel), nextSequenceRecv)
        break;
...
}
```
As for ```packet```, there are three scenarios (assume that all previous packets were successfully received).

- Function ```recvPacket``` has not been called. So ```nextSequenceRecv  == packet.sequence``` whether timeout height/timestamp has passed on the counterparty chain or not.
- Function ```recvPacket``` has been called and timeout height/timestamp has passed on the counterparty chain. So ```nextSequenceRecv  == packet.sequence +1 ``` and ```verifyPacketReceipt(...packet.sequence, TIMEOUT_RECEIPT)) == True```
- Function ```recvPacket``` has been called and timeout height/timestamp has not passed on the counterparty chain. So ```nextSequenceRecv  == packet.sequence +1 ``` and ```verifyPacketReceipt(...packet.sequence, TIMEOUT_RECEIPT)) == False```

So in ```ORDERED_ALLOW_TIMEOUT``` channel, a module can claim a packet is timeout if 
```javascript
nextSequenceRecv  == packet.sequence
```
 means function ```recvPacket``` has not been called.
or 
```javascript
(nextSequenceRecv  == packet.sequence + 1) &&  (verifyPacketReceipt(...packet.sequence, TIMEOUT_RECEIPT)) == True)
``` 
[...]

PR #977, opened 2023-05-30
ICS 04: fix timeout conditions for ordered (without/with allow timeout) channels

crodriguezvega:
closes: #965


spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -1073,26 +1070,49 @@ function timeoutOnClose(
...
+          proof,
+          packet.destPort,
+          packet.destChannel,
+          packet.sequence
+        ))
+        break;
+
+      // NOTE: For ORDERED_ALLOW_TIMEOUT, the relayer must first attempt the receive on the destination chain
+      // before the timeout receipt can be written and subsequently proven on the sender chain in timeoutPacket
+      case ORDERED_ALLOW_TIMEOUT:
```
crodriguezvega, line 1100: I split `ORDERED` and `ORDERED_ALLOW_TIMEOUT` because conditions are different.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -989,10 +989,7 @@ function timeoutPacket(
       // NOTE: For ORDERED_ALLOW_TIMEOUT, the relayer must first attempt the receive on the destination chain
       // before the timeout receipt can be written and subsequently proven on the sender chain in timeoutPacket
       case ORDERED_ALLOW_TIMEOUT:
-        // ordered channel: check that packet has not been received
-        // only allow timeout on next sequence so all sequences before the timed out packet are processed (received/timed out)
-        // before this packet times out
-        abortTransactionUnless(nextSequenceRecv == packet.sequence)
+        abortTransactionUnless(nextSequenceRecv > packet.sequence)
```
crodriguezvega, line 992: For both packet timeouts and for timeout on close the `nextSequenceRecv` on the counterparty would have been incremented by one before the timeout receipt is written.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -964,7 +964,7 @@ function timeoutPacket(
         // ordered channel: check that packet has not been received
         // only allow timeout on next sequence so all sequences before the timed out packet are processed (received/timed out)
         // before this packet times out
-        abortTransactionUnless(nextSequenceRecv == packet.sequence)
+        abortTransactionUnless(nextSequenceRecv <= packet.sequence)
```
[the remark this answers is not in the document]
AdityaSripal: I think its clearer to write

`abortTransactionUnless(packet.Sequence >= nextSequenceRecv)`

spec/core/ics-004-channel-and-packet-semantics/README.md
AdityaSripal: I believe this was correct as it was

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -1073,26 +1070,49 @@ function timeoutOnClose(
...
+          packet.destPort,
+          packet.destChannel,
+          packet.sequence
+        ))
+        break;
+
+      // NOTE: For ORDERED_ALLOW_TIMEOUT, the relayer must first attempt the receive on the destination chain
+      // before the timeout receipt can be written and subsequently proven on the sender chain in timeoutPacket
+      case ORDERED_ALLOW_TIMEOUT:
+        abortTransactionUnless(nextSequenceRecv > packet.sequence)
```
AdityaSripal, line 1101: This should be an equals

AdityaSripal (changes requested):

spec/core/ics-004-channel-and-packet-semantics/README.md
crodriguezvega: The reason for changing it is that in line 709 inside `recvPacket` the `nextSequenceRecv` is incremented by 1 before the timeout receipt is written:

```
nextSequenceRecv = nextSequenceRecv + 1
provableStore.set(nextSequenceRecvPath(packet.destPort, packet.destChannel), nextSequenceRecv)
```

So if I relayer tries to timeout the packet, the `nextSequenceRecv` in the timeout message will be > than `packet.sequence`, won't it?

Same thing for the condition checked in `timeoutOnClose`.

Is this reasoning correct, @AdityaSripal?

crodriguezvega: @AdityaSripal for timing out a packet if the channel is open, should the condition still be that the relayer must attempt to receive the packet (as the comment above this line says)? If that's the case, then I guess it's ok to not support the situation where `packet.sequence` is >= `nextSequenceRecv`.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -1073,26 +1070,54 @@ function timeoutOnClose(
...
+        break;
+
+      case ORDERED_ALLOW_TIMEOUT:
+        // if packet.sequence >= nextSequenceRecv, then the relayer has not attempted
+        // to receive the packet on the destination chain (e.g. because the channel is already closed).
+        // In this situation it is not needed to verify the presence of a timeout receipt.
+        // Otherwise, if packet.sequence < nextSequenceRecv, then the relayer has attempted
+        // to receive the packet on the destination chain, and nextSequenceRecv has been incremented.
+        // In this situation, verify the presence of timeout receipt. 
+        if packet.sequence < nextSequenceRecv {
```
crodriguezvega, line 1105: If `packet.sequence >= nextSequenceRecv`, then timeout, otherwise verify there is a timeout receipt.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -1239,12 +1238,9 @@ function timeoutPacket(
       // NOTE: For ORDERED_ALLOW_TIMEOUT, the relayer must first attempt the receive on the destination chain
       // before the timeout receipt can be written and subsequently proven on the sender chain in timeoutPacket
       case ORDERED_ALLOW_TIMEOUT:
-        // ordered channel: check that packet has not been received
-        // only allow timeout on next sequence so all sequences before the timed out packet are processed (received/timed out)
-        // before this packet times out
-        abortTransactionUnless(nextSequenceRecv == packet.sequence)
+        abortTransactionUnless(packet.sequence < nextSequenceRecv)
```
crodriguezvega, line 1241: @AdityaSripal if a channel is open and the relayer must first attempt to receive the packet on destination chain, I guess it is ok to assume that `nextSequenceRecv` will have been incremented and therefore is > `packet.sequence`?

AdityaSripal: No we will actually want the equals here because I think it makes sense to enforce that timeout happens in order as well.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -1190,8 +1190,7 @@ function timeoutPacket(
         // ordered channel: check that packet has not been received
         // only allow timeout on next sequence so all sequences before the timed out packet are processed (received/timed out)
         // before this packet times out
-        abortTransactionUnless(nextSequenceRecv == packet.sequence)
-
+        abortTransactionUnless(packet.sequence >= nextSequenceRecv)
```
AdityaSripal, line 1193: Again I think this should remain as is. It would enforce that the only packet that gets timed out is the first in line

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -1368,50 +1364,78 @@ function timeoutOnClose(
...
+        if packet.sequence < nextSequenceRecv {
+          abortTransactionUnless(connection.verifyPacketReceipt(
+            proofHeight,
+            proof,
+            packet.destPort,
+            packet.destChannel,
+            packet.sequence
+            TIMEOUT_RECEIPT,
+          ))
+        }
```
AdityaSripal, line 1432: where is the else clause?

AdityaSripal (changes requested):
Thanks Carlos, I think its still important to ensure ordering in the timeout logic. Perhaps we should even add this to timeoutOnClose. Also I believe the else clause in the ordered_allow_timeout case is missing.

spec/core/ics-004-channel-and-packet-semantics/README.md
crodriguezvega: Should the condition be then `abortTransactionUnless(packet.sequence == nextSequenceRecv - 1)` since `nextSequenceRecv` will have been incremented when attempting to receive packet with `packet.sequence`?

AdityaSripal (approved):
Thank you @crodriguezvega !

PR #982, opened 2023-06-12
ics04: Channel Upgrade Packet Flushing Logic

AdityaSripal:


spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -879,6 +883,11 @@ function acknowledgePacket(
     // delete our commitment so we can't "acknowledge" again
     provableStore.delete(packetCommitmentPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
 
+    // if there are no in-flight packets on our end, we can automatically go to FLUSHCOMPLETE
+    if channel.flushStatus === FLUSHING && pendingInflightPackets(portIdentifier, channelIdentifier) == nil {
```
crodriguezvega, line 887:
```suggestion
    if channel.flushStatus === FLUSHING && pendingInflightPackets(packet.sourcePort, packet.sourceChannel) == nil {
```

Same for the others, right?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -88,6 +90,8 @@ interface ChannelEnd {
 - The `connectionHops` stores the list of connection identifiers, in order, along which packets sent on this channel will travel. At the moment this list must be of length 1. In the future multi-hop channels may be supported.
 - The `version` string stores an opaque channel version, which is agreed upon during the handshake. This can determine module-level configuration such as which packet encoding is used for the channel. This version is not used by the core IBC protocol. If the version string contains structured metadata for the application to parse and interpret, then it is considered best practice to encode all metadata in a JSON struct and include the marshalled string in the version field.
 
+See the [upgrade spec](./UPGRADES.md) for details on `sequence` and `flushStatus`.
+
```
crodriguezvega, line 94: Should we also add in the Definitions section above something like "`pendingInFlightPackets` is as defined in the [upgrade spec](./UPGRADES.md#utility-functions)"?

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -589,7 +593,7 @@ function sendPacket(
 
     // check that the channel is not closed to send packets; 
     abortTransactionUnless(channel !== null)
-    abortTransactionUnless(channel.state !== CLOSED)
+    abortTransactionUnless(channel.state !== CLOSED && channel.flushStatus === NOTINFLUSH)
```
DimitrisJim, line 596: this was written initially to allow for optimistic sends right? (That's why in spec we do allow `channel.State` to be in states other than `OPEN`, excluding `CLOSED`). 

Currently, in the implementation, we err if the state [isn't `OPEN`](https://github.com/cosmos/ibc-go/blob/b60617df557264abf23b376255f4375bcd8d27eb/modules/core/04-channel/keeper/packet.go#LL36C2-L36C2), wouldn't this cover us during an upgrade? Could we ever find ourselves in a situation where `channel.State == OPEN` and `channel.FlushStatus` is anything but `NOTINFLUSH`?

DimitrisJim: would be nice to add either way to play it as defensively as possible.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -653,7 +657,7 @@ function recvPacket(
 
     channel = provableStore.get(channelPath(packet.destPort, packet.destChannel))
     abortTransactionUnless(channel !== null)
-    abortTransactionUnless(channel.state === OPEN)
+    abortTransactionUnless(channel.state === OPEN || channel.flushStatus === FLUSHING)
```
crodriguezvega, line 660: If there is a counterparty `lastPacketSent` stored, should we also abort transaction in `recvPacket` unless `packet.sequence` is <= `lastPacketSent`? At the moment I think `lastPacketSent` is stored in `startFlushUpgradeHandshake` and deleted when the upgrade completes/cancels/timeouts, but it is not used anywhere, right?

chatton: Is this check required at all? We want this check to exist on `recvPacket` to process packets while flushing, but we don't want to be able to send more packets when we are flushing.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -879,6 +883,11 @@ function acknowledgePacket(
     // delete our commitment so we can't "acknowledge" again
     provableStore.delete(packetCommitmentPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
 
+    // if there are no in-flight packets on our end, we can automatically go to FLUSHCOMPLETE
+    if channel.flushStatus === FLUSHING && pendingInflightPackets(portIdentifier, channelIdentifier) == nil {
+        channel.flushState = FLUSHCOMPLETE
```
colin-axner, line 888: we need to set the updated channel in state as well

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -1025,6 +1034,11 @@ function timeoutPacket(
       provableStore.set(nextSequenceAckPath(packet.srcPort, packet.srcChannel), nextSequenceAck)
     }
 
+    // if there are no in-flight packets on our end, we can automatically go to FLUSHCOMPLETE
+    if channel.flushStatus === FLUSHING && pendingInflightPackets(portIdentifier, channelIdentifier) == nil {
+        channel.flushState = FLUSHCOMPLETE
+    }
```
colin-axner, line 1040: I would move this to be next to packet deletion (as the logic is related). In addition, when ORDERED channels close, they should set flushing to `NOTINFLUSH` so that the counterparty can safely close as well. I think if the flushing is in `FLUSHING` or `FLUSHCOMPLETE` then the counterparty channel close will fail on proof validation (since it expects the counterparty to be in `NOTINFLUSH`

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -1099,6 +1113,11 @@ function timeoutOnClose(
     // delete our commitment
     provableStore.delete(packetCommitmentPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
 
+    // if there are no in-flight packets on our end, we can automatically go to FLUSHCOMPLETE
+    if channel.flushStatus === FLUSHING && pendingInflightPackets(portIdentifier, channelIdentifier) == nil {
+        channel.flushState = FLUSHCOMPLETE
+    }
+
```
colin-axner, line 1120: Is this necessary? The counterparty channel is closed. Might be good to spend a little time thinking about the cases when a channel closes during an upgrade

colin-axner (commented):
ACK, but see concerns around channels closing during an upgrade (would like to fix ordered channel closing logic before merge)

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -75,6 +77,8 @@ interface ChannelEnd {
   counterpartyChannelIdentifier: Identifier
   connectionHops: [Identifier]
   version: string
+  sequence: uint64
+  flushState: FlushStatus
```
damiannolan, line 81: Is this intended to be `flushState` as the field name, or should it also be `flushStatus`?

colin-axner (approved):
Thank you @AdityaSripal!

PR #983, opened 2023-06-12
remove capability authentication checks in core handlers that are routed from ICS 26

### LINKED ISSUE #886: Remove port/channel capability authentication from ics 026 routed handlers
## Summary

Remove all port/channel capability authentication checks in core handlers which are routed by ICS 026. 

These checks do not add additional protection. [ICS 026](https://github.com/cosmos/ibc/tree/main/spec/core/ics-026-routing-module#channel-lifecycle-management) will use the port or channel identifier to obtain the capability provided to the core handler. Then the core handler will verify the capability using the same ICS 026 module. The circular logic became evident during a code walk through.
 
#### Code example
Here is the call stack when this occurs in ibc-go. 
1. Initiate `MsgChanOpenInit`
2. Obtain [`portCap`](https://github.com/cosmos/ibc-go/blob/main/modules/core/keeper/msg_server.go#L170)
       - which calls [`capabilityKeeper.LookupModules`](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/keeper/keeper.go#L76)
       - which calls [`GetCapability`](https://github.com/cosmos/cosmos-sdk/blob/main/x/capability/keeper/keeper.go#L443)
3. Authenticate [`portCap`](https://github.com/cosmos/ibc-go/blob/main/modules/core/04-channel/keeper/handshake.go#L53)
       - which calls [`capabilityKeeper.AuthenticateCapability`](https://github.com/cosmos/ibc-go/blob/main/modules/core/05-port/keeper/keeper.go#L71)
       - which calls [`GetCapabilityName`](https://github.com/cosmos/cosmos-sdk/blob/main/x/capability/keeper/keeper.go#L287)
       - which performs the opposite lookup as [`GetCapability`](https://github.com/cosmos/cosmos-sdk/blob/main/x/capability/keeper/keeper.go#L366-#L407)
       
Thus, within practice, we are obtaining the capability via the reverse lookup (using IBC scoped keeper) and authenticating the capability using the forward lookup (using the same scoped keeper). Capability authentication should only occur when the capability provided is by another module (SendPacket for example) 
       
#### Spec change example:
Remove the following check from [`chanOpenInit`](https://github.com/cosmos/ibc/tree/main/spec/core/ics-004-channel-and-packet-semantics#opening-handshake) 

```go
abortTransactionUnless(authenticateCapability(portPath(portIdentifier), portCapability))
```

Note, I believe the spec isn't specifying how `portCapability` is obtained

crodriguezvega:
closes: #886 

I have also updated all the `onChanOpenInit` and `onChanOpenTry` callbacks to receive the channel capability as an argument and adapted the implementation of the callbacks to claim the channel capability where required.


colin-axner:
Is this r4r?

crodriguezvega:
I will close this PR, but keep the branch for now. I might be able to reuse it for https://github.com/cosmos/ibc/issues/1133.

PR #988, opened 2023-06-21
Add Token Forwarding to ICS20

AdityaSripal:


spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -54,6 +55,18 @@ The ICS 20 token denominations are represented by the form `{ics20Port}/{ics20Ch
...
 
+The memo is not parsed by the ICS20 application and may be used for external users or higher-level middleware and smart contracts. If the memo is intended to be used within the state machine, it must be in JSON dictionary format with the relevant information keyed for the intended user.
```
AdityaSripal, line 58: TODO: Add Link to reserved keywords. cc: @womensrights

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -54,6 +55,18 @@ The ICS 20 token denominations are represented by the form `{ics20Port}/{ics20Ch
...
+- `channels:origin/transfer/channel-4`: This forwardingPath is similar to the above, but it will first unwind the tokens to their origin chain before executing the logic described above.
+- `channels:origin`: This will simply unwind the tokens to their origin chain and send to the intended receiver.
+- `chains:cosmoshub/osmosis/juno`: On each receiving chain the implementation must first consult its local chain registry to resolve the chainID to a source port and source channel, then it will forward the tokens along that channel. Once the forwardingPath is empty, it will send to specified receiver which must be a valid address at the final destination.
+- `chains:origin/juno/osmosis`: Similar to the previous logic, it will first unwind to the origin before resolving the first chainID and forwarding along the returned channel to forward along the rest of the path.
+- `chains:origin`: The tokens will be unwound to the origin and sent to intended receiver. Note this does not require any read from a chain registry.
```
AdityaSripal, line 68: The logic is slightly more verbose having `origin` come after `channels` and `chains`. However, I think its nicer from a UX standpoint

spec/app/ics-020-fungible-token-transfer/README.md
womensrights: the memo keys are [here](https://github.com/cosmos/chain-registry/blob/master/_memo_keys/ICS20_memo_keys.json)

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -300,6 +487,18 @@ function onAcknowledgePacket(
 function onTimeoutPacket(packet: Packet) {
   // the packet timed-out, so refund the tokens
   refundTokens(packet)
+  // if this packet was forwarded from a previously received packet
+  // then we need to send notification of timeout in reverse
+  reversePacket = store.get(packetForwardPath(packet.sourcePort, packet.sourceChannel, packet.Sequence))
+  if reversePacket != nil {
+    ack = FungibleTokenPacketAcknowledgement{false, "forwarding packet timed out"}
+    // prefix with forwardingPath so we know which chain it timed out on
+    ack.error = prefix(ack.error, reversePacket.sourcePort + "/" + reversePacket.sourceChannel)
```
crodriguezvega, line 497: Just for my understanding: if packet has to go chain A -> chain B -> chain C, the packet stored in `packetForwardPath` will be the packet sent from chain A to chain B, isn't it? Then in a potential time out between chain B and chain C, the `reversePacket.sourcePort` will be source port on chain A and `reversePacket.sourceChannel` will be source channel on chain A? Is that really what we want here?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -54,6 +55,18 @@ The ICS 20 token denominations are represented by the form `{ics20Port}/{ics20Ch
...
+
+Example valid forwardingPaths:
+- `channels:transfer/channel-3/transfer/channel-2/transfer/channel-300`: On each receiving chain, the implementation will pop a (srcPortID, srcChannelID) from the front of the list, resolve to a channel, and then forward the tokens along. Once the forwardingPath is empty, it will send to specified receiver which must be a valid address at the final destination.
```
crodriguezvega: If I understand it correctly, in this case `forwardingPath` should contain the source port and source channel for the chain between the origin and the destination, right? so if I want to send a packet chain A -> chain B -> chain C, then `forwardingPath` will only include the source port and source channel on chain B. Is that correct? If that's the case, maybe it would be good to mention it explicitly.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -256,29 +396,63 @@ function sendFungibleTokens(
...
   prefix = "{packet.sourcePort}/{packet.sourceChannel}/"
   // we are the source if the packets were prefixed by the sending chain
   source = data.denom.slice(0, len(prefix)) === prefix
+  // if we are forwarding the tokens along, we will move the coins to the forwardEscrowAddress to the source chain
+  // otherwise send directy to receiver defined by packet
+  if forwardingPath != "" {
+    // pass in denomination before receive to getNextHop
+    // i.e. denomination in packet
+    forwardPort, forwardChannel = getNextHop(forwardingPath, packet.denomination)
+    receiver = forwardEscrowAddress[forwardChannel]
```
crodriguezvega: Couldn't we just use here `channelEscrowAddresses`? Since the `forwardChannel` is the source channel for the packet that needs to be sent to the next hop, we could use the same escrow address as for any regular packet sent from this chain on that channel, right?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -256,29 +396,63 @@ function sendFungibleTokens(
...
   FungibleTokenPacketAcknowledgement ack = FungibleTokenPacketAcknowledgement{true, null}
   prefix = "{packet.sourcePort}/{packet.sourceChannel}/"
   // we are the source if the packets were prefixed by the sending chain
   source = data.denom.slice(0, len(prefix)) === prefix
+  // if we are forwarding the tokens along, we will move the coins to the forwardEscrowAddress to the source chain
+  // otherwise send directy to receiver defined by packet
+  if forwardingPath != "" {
+    // pass in denomination before receive to getNextHop
+    // i.e. denomination in packet
+    forwardPort, forwardChannel = getNextHop(forwardingPath, packet.denomination)
```
crodriguezvega, line 411: Assuming we are unwinding a packet that went chain A -> chain B -> chain C and `forwardingPath` is `channels:origin` and denomination on chain C is `transfer/channel-3/transfer/channel-1/denom` (assuming `channel-3` is the destination channel on chain C and `channel-1` is the destination channel on chain B when the packet was sent from A to C), if the packet is sent from chain C and received on chain B, then `getNextHop` will return `transfer` and `channel-3`, right? But what we really want is `transfer` and `channel-1`, isn't it?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -54,6 +55,19 @@ The ICS 20 token denominations are represented by the form `{ics20Port}/{ics20Ch
...
+- `channels:origin/transfer/channel-4`: This `forwardingPath` is similar to the above, but it will first unwind the tokens to their origin chain before executing the logic described above.
+- `channels:origin`: This will simply unwind the tokens to their origin chain and send to the intended receiver.
+- `chains:cosmoshub/osmosis/juno`: On each receiving chain the implementation must first consult its local chain registry to resolve the chainID to a source port and source channel, then it will forward the tokens along that channel. Once the forwardingPath is empty, it will send to specified receiver which must be a valid address at the final destination.
+- `chains:origin/juno/osmosis`: Similar to the previous logic, it will first unwind to the origin before resolving the first chainID and forwarding along the returned channel to forward along the rest of the path.
```
nicolaslara, line 68: Sanity check: We don't have any guarantees here that the chains recognized as `juno` and `osmosis` are the same ones recognized as such in the source chain, right? ie.: what origin considers to be "juno" may not the the same as what the sender chain considers as juno.

On the osmosis registries we are considering validating the channels by sending a packet through a list of channels that gets forwarded back to osmosis, and seeing that it returns with the expected path.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -54,6 +55,19 @@ The ICS 20 token denominations are represented by the form `{ics20Port}/{ics20Ch
...
+
+Example valid `forwardingPath`s:
+
+- `channels:transfer/channel-3/transfer/channel-2/transfer/channel-300`: On each receiving chain, the implementation will pop a (srcPortID, srcChannelID) from the front of the list, resolve to a channel, and then forward the tokens along. Once the `forwardingPath` is empty, it will send to specified receiver which must be a valid address at the final destination.
```
nicolaslara, line 64: is there a guarantee at the protocol level that ports and channel names cannot contain `/`? This may matter if someone can register a port/channel on one of the receiving chains and craft it to highjack packets expected to be forwarded further. 

IIRC, ibc-go does not do greedy matching but instead splits the path by `/` and matches the pairs, but double checking.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -71,11 +84,127 @@ interface FungibleTokenPacketError {
...
+
+`ics20Prefix` returns the ICS20 trace encoded in the denomination. It does this by checking if the second element of the slash-split denomination is prefixed by the special keyword "channel-". If this is true, then the two elements comprise a (portID, channelID) tuple.
+
+```typescript
+function ics20Prefix(denomination: string): (prefix: string) {
+  elems = split(denomination, "/")
+  ics20Prefix = ""
+  if len(elems) > 2 {
+    for i = 1; i < len(elems); i + 2 {
+      if hasPrefix(elems[i], "channel-") {
```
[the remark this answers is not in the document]
nicolaslara: mentioning this is IBC go specific sounds good. I can see value in migrating to different channel names in the future (e.g.: using chain names as prefixes `transfer/osmosis-0`)

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -44,6 +44,7 @@ interface FungibleTokenPacketData {
   sender: string
   receiver: string
   memo: string
+  forwardingPath: string
```
nicolaslara, line 47: I keep thinking that forwarding should be combined with execution (i.e.: ibc-hooks/adr-8) so we can define a more generic declarative model of computation on top of ibc. The simplest solution is to allow a callback to exist on each hop as long as it returns something that can be processed on the next hop (new packet, short-circuit, or maybe even a modified path). This would be very similar to a functional `fold` or, more generally, a combination of monads. 

A more advanced version of this may a allow for fan-in/out, but it may be better to leave that for the future as I think the fold model covers 90% of the use cases. 

Does it make sense to try to integrate those directly on this spec? I've been meaning to write a document explaining this in more detail but haven't gotten around to it.

spec/app/ics-020-fungible-token-transfer/README.md
AdityaSripal: Yes slashes are not permitted in portID or channelID: https://github.com/cosmos/ibc-go/blob/main/modules/core/24-host/validate.go#L44

spec/app/ics-020-fungible-token-transfer/README.md
AdityaSripal: This is correct. We could theoretically resolve in a single place like a global CNS, but this would require the global CNS to hold the relative identifiers of all chains to all chains.

Using this feature as it is currently specified would require the assumption that all chains hold the same social consensus

crodriguezvega:
Superseded by https://github.com/cosmos/ibc/pull/1090

PR #996, opened 2023-07-05
add commas after interface's methods

0xstepit:
This PR is a followup of #993 to add missing commas inside the interface.


crodriguezvega:
Actually, it should be the opposite: we usually don't add commas at the end of each field in the interfaces. 😓

0xstepit:
Now is without commas @crodriguezvega

crodriguezvega (approved):
Thank you, @0xstepit. :)

PR #997, opened 2023-07-05
ICS-022 multi token transfer

dzmitry-lahoda:


crodriguezvega:
Thanks for the PR, @dzmitry-lahoda. It should not be needed to have a separate ICS application to support transferring  multiple coins in the same packet data. It should be possible to update ICS 20 packet data to hold multiple coins instead of only one. But this requires bumping the application to v2, so we will probably try to gather what other features would need a breaking change of the application, so that we can bundle all of them together in the same upgrade. We will probably work on this towards the end of the year, or next year.

I will close then this PR, since the changes some be done in ICS 20.

PR #998, opened 2023-07-05
#draft ICS-200 batch application middleware

dzmitry-lahoda:
NEAR over IBC


crodriguezvega:
@dzmitry-lahoda I think I will close this PR, since it has been stale for 3 months and looks like it's still in draft.

PR #1003, opened 2023-07-12
Fix: Update description of MisbehaviourPredicate function

nadeemb53:


crodriguezvega (approved):
Thank you for the fix, @nadeemb53.

PR #1005, opened 2023-07-13
nit: Fix misnamed sequence into upgradeSequence

DimitrisJim:


crodriguezvega (approved):
Thanks, @DimitrisJim!

PR #1007, opened 2023-07-25
ICS27: add support for unordered channels/implement upgrade callbacks

### LINKED ISSUE #650: ICS27: Should verify that the previous active channel has the same metadata as the new active channel
In `OnChanOpenInit` on the controller chain, if a previous active channel was set, but has since been closed, when reopening a new active channel, a check should be added to ensure the metadata is equal to the previous version. Otherwise relayers could change which connections are used for the active channel

crodriguezvega:
It additionally...

closes #650


AdityaSripal (approved):
Nice work Carlos!

Additionally in a future PR, we can think about how to open an UNORDERED channel that can be used by multiple owners. THis should be possible by allowing a new ICA port without an owner address, and putting the address instead in the packet data

Issue #1004, opened 2023-07-13
ICS 04 (upgrades): Use counterparty connection hops when verifying channel state in UpgradeOpen.


DimitrisJim:
Specifically, when the counterparty is in `OPEN`, we should be using the counterparty connection hops based on the upgraded connection end. Relevant PR https://github.com/cosmos/ibc-go/pull/4074

Might be addressed in spec via #999 if we decide to store the full counterparty upgrade.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1008, opened 2023-07-30
ICS 04 (upgrades): fix counterparty connection hops in `chanUpgradeOpen`

crodriguezvega:
closes: #1004


DimitrisJim (approved):
looks right to me!

PR #1009, opened 2023-08-07
ics-004: fix packetReceiptPath and packetCommitmentPath

avahowell:
These two paths were out-of-sync with the current IBC implementation, as deployed throughout the ecosystem.


crodriguezvega (approved):
Good catch. Thank you, @avahowell!

Issue #797, opened 2022-07-13
Add base denomination field to ICS20 packet


colin-axner:
There has been strong community desire to allow for base denominations with slashes. However, supporting base denominations with slashes makes parsing the base denomination from a string of several IBC hops non-trivial without a clear splitting point (neither the port identifier nor the channel identifier are predictable in all circumstances atm). Thus it would be useful if the existing field `denom` in the `FungibleTokenPacketData` was split into `base_denom` and `denom` (would prefer a different name than `denom`, will leave up to others to think of a suitable name as non comes to my mind, maybe `record`, `trace`, `hops`)

Since this is a proposed change to the packet and thus requires a new ICS20 version to handle proper encoding/decoding of the packet, I suggest bundling this change with other changes (such as adding the memo field)

PR #1020, opened 2023-09-20
Draft Spec for Multi-Denom Packets ICS20 v2

AdityaSripal:
closes: https://github.com/cosmos/ibc/issues/797


spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -45,14 +45,26 @@ interface FungibleTokenPacketData {
...
 
+interface FungibleTokenPacketDataV2 {
+  tokens: []Token
+  sender: string
+  receiver: string
+  memo: string
+}
+
+interface Token {
+  denomination: string
```
chatton, line 57: is it an intentional decision to change this from `denom` (V1) to `denomination`?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -353,6 +469,13 @@ Example:
...
+
+As mentioned above, user interfaces that display the denomination without any trace information then they **must** make a decision to trust a canonical path for that denomination. Otherwise, user interfaces must still somehow display the trace information to the end user. This is because the metadata contains human-readable information that is not directly verified by the state machine. Thus, a malicious sender chain can fool user interfaces if they choose to send metadata that is the same as some well-known token issued by a different chain. However, given that a certain chain and denomination is trusted; the metadata can make it easier for displays to represent the token without relying on out-of-chain information.
```
crodriguezvega, line 476:
```suggestion
As mentioned above, user interfaces that display the denomination without any trace information **must** make a decision to trust a canonical path for that denomination. Otherwise, user interfaces must still somehow display the trace information to the end user. This is because the metadata contains human-readable information that is not directly verified by the state machine. Thus, a malicious sender chain can fool user interfaces if they choose to send metadata that is the same as some well-known token issued by a different chain. However, given that a certain chain and denomination is trusted; the metadata can make it easier for displays to represent the token without relying on out-of-chain information.
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -210,33 +253,47 @@ In plain English, between chains `A` and `B`:
...
+      // we are the source if the denomination is not prefixed
+      source = token.trace[0] !== prefix
+      onChainDenom = constructOnChainDenom(token.trace, token.denomination)
+      // check if we've already sent metadata for this token on this channel
+      isMetadataSent = getSentMetadataFlag(sourcePort, sourceChannel, onChainDenom)
+      // if it hasn't been sent, then we will set the token metadata in the token
+      // so receiver can store it
+      // otherwise we leave metadata field blank to reduce redundancy
+      if !isMetadataSent {
+        metadata = getMetadata(onChainDenom)
```
crodriguezvega, line 281: Should we address the situation where metadata changes and needs to be sent again?

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega: > This is because the metadata contains human-readable information that is not directly verified by the state machine

Would it be possible to verify the metadata? I see that you have the `getMetadata` function in `sendFungibleTokens`. Could this function retrieve the metadata for the token based on on-chain state?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -210,33 +253,47 @@ In plain English, between chains `A` and `B`:
...
+        bank.TransferCoins(sender, escrowAccount, onChainDenom, token.amount)
+      } else {
+        // receiver is source chain, burn vouchers
+        bank.BurnCoins(sender, onChainDenom, token.amount)
+      }
     }
 
     // create FungibleTokenPacket data
-    data = FungibleTokenPacketData{denomination, amount, sender, receiver}
+    data = FungibleTokenPacketData{tokens, sender, receiver, memo}
```
crodriguezvega, line 296: Should there be here some kind of conditional logic based on the channel version?

```typescript
channel = provableStore.get(channelPath(sourcePort, sourceChannel))
if channel.version === "ics20-1" {
  data = FungibleTokenPacketData{tokens[0].denomination, tokens[0].amount, sender, receiver, memo}
} else if channel.version === "ics20-2" {
  data = FungibleTokenPacketDataV2{tokens, sender, receiver, memo}
}
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -289,8 +377,20 @@ function onAcknowledgePacket(
...
+    // set metadata flag for any metadata we sent in the packet
+    // so we don't need to resend metadata for that token on this channel
+    // in the next sendPacket
+    FungibleTokenPacketDataV2 data = packet.data
+    for token in data.tokens {
+      if !isEmpty(token.metadata) {
+        onChainDenom = constructOnChainDenom(token.trace, token.denomination)
+        setMetadataFlag(packet.sourcePort, packet.sourceChannel, onChainDenom)
+      }
+    }
```
crodriguezvega, line 392: Should this be also wrapped in an if statement like `if channel.version === "ics20-2"`?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -256,27 +313,58 @@ function sendFungibleTokens(
...
+        ack = FungibleTokenPacketAcknowledgement{false, "mint coins failed"}
+        // break out of for loop on first error
+        break
+    }
+    // set metadata for the onchain denomination if token metadata was sent in the packet
+    // it is recommended that metadata is only set for tokens that come in with no trace
+    // i.e. directly from the source chain of the token
+    if !isEmpty(token.metadata) && isEmpty(token.trace) {
+      setMetadata(onChainDenom, token.metadata)
+    }
```
crodriguezvega, line 367: Should this also be done only in case `channel.version === "ics20-2"`?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -210,33 +253,47 @@ In plain English, between chains `A` and `B`:
   an acknowledgement of failure is preferable to aborting the transaction since it more easily enables the sending chain
   to take appropriate action based on the nature of the failure.
 
+Note: `constructOnChainDenom` is a helper function that will construct the local on-chain denomination for the bridged token. It **must** encode the trace and base denomination to ensure that tokens coming over different paths are not treated as fungible. The original trace and denomination must be retrievable by the state machine so that they can be passed in their original forms when constructing a new IBC path for the bridged token. The ibc-go implementation handles this by creating a local denomination: `hash(trace+base_denom)`.
+
```
crodriguezvega, line 257: Maybe we could also add a notes for functions `getSentMetadataFlag` and `getMetadata`.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -210,33 +253,47 @@ In plain English, between chains `A` and `B`:
...
+      prefix = "{sourcePort}/{sourceChannel}/"
+      // we are the source if the denomination is not prefixed
+      source = token.trace[0] !== prefix
+      onChainDenom = constructOnChainDenom(token.trace, token.denomination)
+      // check if we've already sent metadata for this token on this channel
+      isMetadataSent = getSentMetadataFlag(sourcePort, sourceChannel, onChainDenom)
+      // if it hasn't been sent, then we will set the token metadata in the token
+      // so receiver can store it
+      // otherwise we leave metadata field blank to reduce redundancy
+      if !isMetadataSent {
```
crodriguezvega, line 280: Maybe to make sure that metadata is only sent for source denoms (without trace) we could check `!isMetadataSent && isEmpty(token.trace)`? Or even to save gas do this?

```typescript
if isEmpty(token.trace) {
  isMetadataSent = getSentMetadataFlag(sourcePort, sourceChannel, onChainDenom)
  // if it hasn't been sent, then we will set the token metadata in the token
  // so receiver can store it
  // otherwise we leave metadata field blank to reduce redundancy
  if !isMetadataSent {
    metadata = getMetadata(onChainDenom)
    token.metadata = metadata
  }
}

**Review by crodriguezvega — COMMENTED** [review]
Nice work. Thank you, @AdityaSripal!

File: spec/app/ics-020-fungible-token-transfer/README.md (hunk 3 above)
Comment (AdityaSripal) [reply spec/app/ics-020-fungible-token-transfer/README.md:281]: This is handled automatically on the receiving side since we always overwrite. I didn't put it in the sender spec since its an edge case that might show up very rarely for specific chains. It doesn't need to be explicitly spelled out here in the protocol in my view

File: spec/app/ics-020-fungible-token-transfer/README.md [hunk 9]
```diff
@@ -134,10 +173,14 @@ function onChanOpenInit(
   // assert that version is "ics20-1" or empty
   // if empty, we return the default transfer version to core IBC
   // as the version for this channel
-  abortTransactionUnless(version === "ics20-1" || version === "")
+  abortTransactionUnless(version === "ics20-2" || version === "ics20-1" || version === "")
```
sangier, line 176: Line 160: "- The version string is `ics20-1`" 
Should be changed accordingly to line 176: 
"-The version string is `ics20-1` or `ics20-2`

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -153,12 +196,11 @@ function onChanOpenTry(
   // only unordered channels allowed
   abortTransactionUnless(order === UNORDERED)
   // assert that version is "ics20-1"
```
sangier, line 198:
```suggestion
  // assert that version is "ics20-1" or "ics20-2" 
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -134,10 +173,14 @@ function onChanOpenInit(
   // assert that version is "ics20-1" or empty
```
sangier, line 173:
```suggestion
  // assert that version is "ics20-1" or "ics20-2" or empty
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -153,12 +196,11 @@ function onChanOpenTry(
...
   // assert that version is "ics20-1"
-  abortTransactionUnless(counterpartyVersion === "ics20-1")
+  abortTransactionUnless(counterpartyVersion === "ics20-1" || counterpartyVersion === "ics20-2")
   // allocate an escrow address
   channelEscrowAddresses[channelIdentifier] = newAddress()
-  // return version that this chain will use given the
-  // counterparty version
-  return "ics20-1", nil
+  // return the same version as counterparty version so long as we support it
+  return counterpartyVersion, nil
```
sangier, line 203: Probably I am missing something, but how do we know here if we support the counterparty version?

spec/app/ics-020-fungible-token-transfer/README.md
sangier: The explanation of these functions is at line 490 under the populating token metadata section.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -406,6 +542,8 @@ July 27, 2020 - Re-addition of source field
 
 Nov 11, 2022 - Addition of a memo field
 
+Sep 22, 2023 - Support for multi-token packets
```
sangier, line 545: Should we update line 11 (ICS header, modified field) accordingly?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -210,33 +253,54 @@ In plain English, between chains `A` and `B`:
...
+      isMetadataSent = getSentMetadataFlag(sourcePort, sourceChannel, onChainDenom)
+      // if it hasn't been sent, then we will set the token metadata in the token
+      // so receiver can store it
+      // otherwise we leave metadata field blank to reduce redundancy
+      if !isMetadataSent {
+        metadata = getMetadata(onChainDenom)
+        token.metadata = metadata
+      }
+      // create FungibleTokenPacket data
+      data = FungibleTokenPacketDataV2{tokens, sender, receiver, memo}
```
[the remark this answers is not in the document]
sangier: Noted that consideration over the possible DOS based on the packet length are present at line 463: "Chains should ensure that there is some length limit on the entire packet data to ensure that the packet does not become a DOS vector. However, these do not need to be protocol-defined limits."

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -169,8 +211,9 @@ function onChanOpenAck(
   counterpartyChannelIdentifier: Identifier,
   counterpartyVersion: string) {
   // port has already been validated
-  // assert that counterparty selected version is "ics20-1"
-  abortTransactionUnless(counterpartyVersion === "ics20-1")
+  // assert that counterparty selected version is the same as our version
+  channel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+  abortTransactionUnless(counterpartyVersion === channel.version)
```
chatton, line 216: @AdityaSripal I believe this approach does not work with middleware. For example, if I have fee middleware enabled with `ics20-2`, the `counterpartyVersion` will be the unwrapped string `ics20-2`, while the `channel.version` will be a json object which includes both the fee/transfer app versions.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -210,33 +253,54 @@ In plain English, between chains `A` and `B`:
...
+        bank.BurnCoins(sender, onChainDenom, token.amount)
+      }
     }
 
-    // create FungibleTokenPacket data
-    data = FungibleTokenPacketData{denomination, amount, sender, receiver}
+    channel = provableStore.get(channelPath(sourcePort, sourceChannel))
+    if channel.version === "ics20-1" {
+      abortTransactionUnless(len(tokens) == 1)
+      data = FungibleTokenPacketData{tokens[0].denom, tokens[0].amount, sender, receiver, memo}
```
chatton, line 290: I don't think this is correct, 

`tokens[0].denom` is not the same as `FungibleTokenPacketData.Denom`, the former is something like `atom` while the latter is `transfer/channel-0/transfer/channel-1/atom`.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -210,33 +253,54 @@ In plain English, between chains `A` and `B`:
...
+      } else {
+        // receiver is source chain, burn vouchers
+        bank.BurnCoins(sender, onChainDenom, token.amount)
+      }
     }
 
-    // create FungibleTokenPacket data
-    data = FungibleTokenPacketData{denomination, amount, sender, receiver}
+    channel = provableStore.get(channelPath(sourcePort, sourceChannel))
+    if channel.version === "ics20-1" {
```
chatton, line 288: related to the other point I made, we can't actually use channel version to determine which application version we're using. One issue came up. Let's say we have chainA and chainB, chain A only supports ics20-1, while chainB supports ics20-2. The channel opened is application version ics20-1.

Chain B receives a MsgTransfer which contains muiltiple denoms to transfer. In the case of ics20-1, we should reject the message unless the array contains a single element. 

Right now, I don't think we have a way to correctly determine which application version we currently are on as there is no additional state outside of the channel version.

Does it make sense to add an additional state entry specifying the application version within the channel module that can be used for this purpose? (if empty assume ics20-1)

spec/app/ics-020-fungible-token-transfer/README.md
chatton: Discussed on slack, we can access the version via the `ics4Wrapper`

spec/app/ics-020-fungible-token-transfer/README.md
chatton: can access via ics4Wrapper, no need for additional state

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -131,13 +170,17 @@ function onChanOpenInit(
...
-  // assert that version is "ics20-1" or empty
+  // assert that version is "ics20-1" or "ics20-2" or empty
   // if empty, we return the default transfer version to core IBC
   // as the version for this channel
-  abortTransactionUnless(version === "ics20-1" || version === "")
+  abortTransactionUnless(version === "ics20-2" || version === "ics20-1" || version === "")
   // allocate an escrow address
   channelEscrowAddresses[channelIdentifier] = newAddress()
-  return "ics20-1", nil
+  if version === "ics20-1" {
```
sangier, line 179:
```suggestion
  if version === "" {
  // default to latest supported version
   return "ics20-2", nil
  }
  // If the version is not empty and is among those supported, we return the version
  return version, nil 
}
```
An informational thing about this block of code: in the future if we add "ics20-3", then we will need to add another if to handle the case where version is "ics20-2". While this way, we would only change line 173,176, and the default value without adding a new if.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -45,14 +45,53 @@ interface FungibleTokenPacketData {
...
+A sending chain may be acting as a source or sink zone. When a chain is sending tokens across a port and channel which are not equal to the last prefixed port and channel pair, it is acting as a source zone. When tokens are sent from a source zone, the destination port and channel will be prepended to the trace (once the tokens are received) adding another hop to a tokens record. When a chain is sending tokens across a port and channel which are equal to the last prefixed port and channel pair, it is acting as a sink zone. When tokens are sent from a sink zone, the first element of the trace, which was the last port and channel pair added to the trace is removed (once the tokens are received), undoing the last hop in the tokens record. A more complete explanation is [present in the ibc-go implementation](https://github.com/cosmos/ibc-go/blob/457095517b7832c42ecf13571fee1e550fec02d0/modules/apps/transfer/keeper/relay.go#L18-L49).
```
sangier, line 77: Is my understanding of Source/Sink Zones functioning correct? Shall we add an image describing the process? 

Screenshot 2024-02-07 alle 13 31 26

sangier: Based on the [adr-001](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md) the diagram appears to be correct.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -45,14 +45,53 @@ interface FungibleTokenPacketData {
...
+  denomUnits: []DenomUnit
+}
+
+interface DenomUnit {
+  denom: string
+  exponent: uint256
+}
 ```
 
+As tokens are sent across chains using the ICS 20 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `trace` field in the denomination.
```
sangier, line 73: As tokens are sent across chains using the ICS 20 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `trace` field and will be used in tandem with the base denomination, by the `constructOnChainDenom` helper function, to construct the local on-chain denomination for the bridged token. 
```suggestion
As tokens are sent across chains using the ICS 20 protocol, they begin to accrue a record of channels for which they have been transferred across. This information is encoded into the `trace` field and will be used in tandem with the base denomination, by the `constructOnChainDenom` helper function, to construct the local on-chain denomination for the bridged token.
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -256,27 +320,58 @@ function sendFungibleTokens(
...
+      token = Token{
+        denom: denom
+        trace: trace
+        amount: packet.amount
+      }
+      tokens = []Token{token}
+  } else if channel.version === "ics20-2" {
+    FungibleTokenPacketDataV2 data = packet.data
+    tokens = data.tokens
+  }
```
chatton, line 336: should be add an `abortUnless(channel.version == "ics20-1" || channel.version == "ics20-2")?

sangier: Doesn't `onRecvPacket` operate within the context of an already established IBC channel? If so, then the `channel.version` through which these packets are received should have been previously set up and agreed upon to be either `ics20-1` or `ics20-2`.

chatton: we have those assumptions, but I think it would still be preferable to be explicit in showing these are the only 2 possible states we can be in.

Also a similar point that this shouldn't be referencing `channel.version` since `channel.version` is not going to be the same as `application version` due to the structure of the version string when dealing with middleware.

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega: It might be good to be explicit and mention as well somewhere in the spec (maybe where the `Token` type is introduced) that `token.denom` is the base denomination, since in ibc-go [the denomination in the `sdk.Coins` that are passed to `sendTransfer`](https://github.com/cosmos/ibc-go/blob/749caee7f7a8ca902e94167d630fabe61d3e1e67/modules/apps/transfer/keeper/relay.go#L60) might be the IBC denomination ([from which we derive the trace and base denom information](https://github.com/cosmos/ibc-go/blob/749caee7f7a8ca902e94167d630fabe61d3e1e67/modules/apps/transfer/keeper/relay.go#L98)). Just to avoid confusion if someone reads the spec and compares with the code.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -256,27 +320,58 @@ function sendFungibleTokens(
 
 ```typescript
 function onRecvPacket(packet: Packet) {
-  FungibleTokenPacketData data = packet.data
+  channel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+  if channel.version === "ics20-1" {
+      FungibleTokenPacketData data = packet.data
```
crodriguezvega, line 325: Just to be explicit that the packet data is json marshalled and needs to be unanarshaled.

```suggestion
      FungibleTokenPacketData data = UnmarshalJSON(packet.data)
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -256,27 +320,58 @@ function sendFungibleTokens(
...
+      trace, denom = parseICS20V1Denom(data.denom)
+      token = Token{
+        denom: denom
+        trace: trace
+        amount: packet.amount
+      }
+      tokens = []Token{token}
+  } else if channel.version === "ics20-2" {
+    FungibleTokenPacketDataV2 data = packet.data
```
crodriguezvega, line 334:
```suggestion
    FungibleTokenPacketDataV2 data = UnmarshalJSON(packet.data)
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -289,8 +384,23 @@ function onAcknowledgePacket(
...
-  if (!acknowledgement.success)
+  if !(acknowledgement.success) {
     refundTokens(packet)
+  } else {
+    channel = provableStore.get(channelPath(sourcePort, sourceChannel))
+    if channel.version === "ics20-2" {
+      // set metadata flag for any metadata we sent in the packet
+      // so we don't need to resend metadata for that token on this channel
+      // in the next sendPacket
+      FungibleTokenPacketDataV2 data = packet.data
```
crodriguezvega, line 395:
```suggestion
      FungibleTokenPacketDataV2 data = UnmarshalJSON(packet.data)
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -307,17 +417,33 @@ function onTimeoutPacket(packet: Packet) {
 
 ```typescript
 function refundTokens(packet: Packet) {
-  FungibleTokenPacketData data = packet.data
+  channel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
+  if channel.version === "ics20-1" {
+      FungibleTokenPacketData data = packet.data
```
crodriguezvega, line 422:
```suggestion
      FungibleTokenPacketData data = UnmarshalJSON(packet.data)
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -307,17 +417,33 @@ function onTimeoutPacket(packet: Packet) {
...
+      trace, denom = parseICS20V1Denom(data.denom)
+      token = Token{
+        denom: denom
+        trace: trace
+        amount: packet.amount
+      }
+      tokens = []Token{token}
+  } else if channel.version === "ics20-2" {
+    FungibleTokenPacketDataV2 data = packet.data
```
crodriguezvega, line 431:
```suggestion
    FungibleTokenPacketDataV2 data = UnmarshalJSON(packet.data)
```

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega: Sounds good. And we should also probably clear the `version compatibility` field until this is released in ibc-go.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -316,17 +387,33 @@ function onTimeoutPacket(packet: Packet) {
 
 ```typescript
 function refundTokens(packet: Packet) {
-  FungibleTokenPacketData data = packet.data
+  channel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
```
crodriguezvega, line 390: Should we also do here `transferVersion = getAppVersion(channel.version)`?

sangier (approved):

spec/app/ics-020-fungible-token-transfer/README.md
AdityaSripal: it's done below

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega: I address this [here](https://github.com/cosmos/ibc/pull/1020/commits/d49a90bbb6ed28c7fcda99cb27ef209886ebad0b).

crodriguezvega (approved):
Thank you for writing the spec, @AdityaSripal. And thank you for the review, @chatton and @sangier.

Issue #1024, opened 2023-10-13
ICS 04(upgrades): consider removing last packet sent


crodriguezvega:
Already in consideration by @AdityaSripal and @colin-axner, but describing here a problem I noticed:

A problem with last packet sent is that it cannot be used in `recvPacket` of the counterparty until the chain has stored the value in its private store. For example: if chain B executes `chanUpgradeTry` and moves into `FLUSHING` state, according to the spec chain B can receive packets that have a sequence <= counterparty's last packet sent:

```typescript
counterpartyLastPacketSent = privateStore.get(channelCounterpartyLastPacketSequencePath(packet.destPort, packet.destChannel)
abortTransactionUnless(channel.state === OPEN || (channel.state === FLUSHING && packet.sequence <= counterpartyLastPacketSent))   
``` 

But chain B will not have the value of the counterparty's (chain A) last packet sent until it executes `chanUpgradeConfirm` (this is the fix suggested in #1023).

If this issue is implemented then #1015 and #1023 will not be needed and can be closed.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #1023, opened 2023-10-13
ICS 04(upgrades): missing setting counterparty's last packet sent


crodriguezvega:
In `chanUpgradeAck` and `chanUpgradeConfirm` we miss setting the counterparty's last packet sent in the private store. Something like this:

```typescript
// store counterparty's sequence number of last packet sent
privateStore.set(
  channelCounterpartyLastPacketSequencePath(portIdentifier, channelIdentifier), 
  counterpartyUpgrade.lastPacketSent
)
```

This is needed so that in `recvPacket` it is possible to check that we receive packets that have been sent before the last packet sent by the counterparty:

```typescript
counterpartyLastPacketSent = privateStore.get(channelCounterpartyLastPacketSequencePath(packet.destPort, packet.destChannel)
abortTransactionUnless(channel.state === OPEN || (channel.state === FLUSHING && packet.sequence <= counterpartyLastPacketSent))
```

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

Issue #1015, opened 2023-08-28
ICS004(channel-upgrades): change counterparty `LastSequenceSend` to `NextSequenceSend`


damiannolan:
In the current channel upgradability spec, we introduce a new concept of `LastSequenceSend` / `LatestSequenceSend` in order to provide a "watermark" or checkpoint for packet flushing while within the upgrade handshake. 

This can easily be changed to `NextSequenceSend` which is already a well-known concept within ibc.

Conditional logic in `recvPacket` would change like so:

Spec: 

```diff
- abortTransactionUnless(channel.state === OPEN || (channel.state === FLUSHING && packet.sequence <= counterpartyLastPacketSent))
+ abortTransactionUnless(channel.state === OPEN || (channel.state === FLUSHING && packet.sequence < counterpartyNextSequenceSend))
```

ibc-go example:

```diff
- if packet.GetSequence() > counterpartyUpgrade.LatestSequenceSend {
+ if packet.GetSequence() >= counterpartyUpgrade.NextSequenceSend {
	return errorsmod.Wrapf(
		types.ErrInvalidPacket,
		"failed to receive packet, cannot flush packet at sequence greater than counterparty last sequence send (%d) > (%d)", packet.GetSequence(), counterpartyUpgrade.LatestSequenceSend,
	)
}
```

I would consider this an improvement as we can reuse existing ibc concepts without introducing more concepts around send/recv sequencing.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1026, opened 2023-10-24
ICS004: Remove last packet sent

AdityaSripal:
closes: #1024
closes: #1015
closes: #1023

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -827,7 +827,6 @@ function recvPacket(
 
     channel = provableStore.get(channelPath(packet.destPort, packet.destChannel))
     abortTransactionUnless(channel !== null)
-    counterpartyLastPacketSent = privateStore.get(channelCounterpartyLastPacketSequencePath(packet.destPort, packet.destChannel)
     abortTransactionUnless(channel.state === OPEN || (channel.state === FLUSHING))
     abortTransactionUnless(authenticateCapability(channelCapabilityPath(packet.destPort, packet.destChannel), capability))
     abortTransactionUnless(packet.sourcePort === channel.counterpartyPortIdentifier)
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -828,7 +828,7 @@ function recvPacket(
     channel = provableStore.get(channelPath(packet.destPort, packet.destChannel))
     abortTransactionUnless(channel !== null)
     counterpartyLastPacketSent = privateStore.get(channelCounterpartyLastPacketSequencePath(packet.destPort, packet.destChannel)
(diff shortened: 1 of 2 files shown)


spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -828,7 +828,7 @@ function recvPacket(
     channel = provableStore.get(channelPath(packet.destPort, packet.destChannel))
     abortTransactionUnless(channel !== null)
     counterpartyLastPacketSent = privateStore.get(channelCounterpartyLastPacketSequencePath(packet.destPort, packet.destChannel)
```
crodriguezvega, line 830: This line should be removed as well, right?

```suggestion
```

crodriguezvega (approved):
Nice! 🧹

sangier (commented):
If I understand correctly, the core idea here is that we can remove the `lastPacketSent` field because the following statements remain true:

1. 'Any packet sent to the channel with a packet sequence number greater than the `lastPacketSent` will be rejected until the upgrade is complete.'
   - This statement remains valid because, even without the `lastPacketSent` field, the transition to the `FLUSHING` state during the `startFlushUpgradeHandshake` operation effectively blocks the `sendPacket` function. As a result, no additional packets can be sent over this channel during the upgrade process.
   - Considering that we were storing the `lastPacketSent` when we were moving to FLUSHING, the system without `lastPacketSent` achieve the same result. 

2. We still have a way to check if there are pending inflight packets.
   - `pendingInflightPacket` does not depend on `lastPacketSent`. Indeed is designed to monitor the pending in-flight packet verifying the presence/absence of packet commitment given that packet commitments are deleted only when the packet lifecycle is complete. Therefore, if a packet commitment exists on the sender's chain, it indicates that the packet lifecycle is still incomplete.

Am I missing something? If all the above is correct, IMO, removing `lastPacketSent` is safe.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -312,10 +301,7 @@ function startFlushUpgradeHandshake(
   // either timeout height or timestamp must be non-zero
```
sangier, line 301: In the comment al line 287, we should remove "and the las packet sent".

crodriguezvega: Yes, good catch!

damiannolan (approved):
Also closes #1015 as it won't be required after this PR 👍 
Thanks @AdityaSripal ❤️

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -348,7 +334,6 @@ function openUpgradeHandshake(
```
sangier, line 334: Typo at line 312: "`openUpgradeHandshake` will open the channel and switch the existing channel parameters to the newly agreed-upon **uprade** channel fields."

Fix: "`openUpgradeHandshake` will open the channel and switch the existing channel parameters to the newly agreed-upon upgrade channel fields."

PR #1036, opened 2023-11-16
Small consistency updates for channel upgrade spec

chatton:
While re-reviewing the channel upgrades spec to regain context, I noticed a few small inconsistencies in code style in the pseudo code and a few suggestions for different wording in several cases.

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -222,7 +222,7 @@ function initUpgradeHandshake(
   // If channel already has an upgrade but isn't in FLUSHING,
   // then this will override the previous upgrade attempt
   channel = provableStore.get(channelPath(portIdentifier, channelIdentifier))
-  abortTransactionUnless(channel.state == OPEN)
+  abortTransactionUnless(channel.state === OPEN)
 
   // new channel version must be nonempty
   abortTransactionUnless(proposedUpgradeFields.Version !== "")
@@ -256,10 +256,10 @@ function isCompatibleUpgradeFields(
   proposedUpgradeFields: UpgradeFields,
   counterpartyUpgradeFields: UpgradeFields,
 ): boolean {
-  if (proposedUpgradeFields.ordering !== counterpartyUpgradeFields.ordering) {
+  if (proposedUpgradeFields.ordering != counterpartyUpgradeFields.ordering) {
     return false
   }
-  if (proposedUpgradeFields.version !== counterpartyUpgradeFields.version) {
+  if (proposedUpgradeFields.version != counterpartyUpgradeFields.version) {
     return false
   }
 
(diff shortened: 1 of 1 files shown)


sangier (approved):
Thanks @chatton ! I just left an additional comment. Apart from that, LGTM!

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -962,14 +962,14 @@ function timeoutChannelUpgrade(
...
-        counterpartyHops == counterpartyChannel.connectionHops {
+    if ((upgrade.fields.version === counterpartyChannel.version) &&
+        (upgrade.fields.order === counterpartyChannel.order) &&
+        (counterpartyHops === counterpartyChannel.connectionHops)) {
           // counterparty has already successfully upgraded so we cannot timeout
           abortTransactionUnless(false)
     }
   }
-  abortTransactionUnless(counterpartyChannel.upgradeSequence >== channel.upgradeSequence)
+  abortTransactionUnless(counterpartyChannel.upgradeSequence >= channel.upgradeSequence)
```
sangier, line 972: Another consideration for the comment at line 999 :
"Both parties must not complete the upgrade handshake if the counterparty upgrade timeout has already passed. **Even if both sides could have successfully moved to FLUSHCOMPLETE**. This will prevent the channel ends from reaching incompatible states." 

Note that the `timeoutChannelUpgrade` cannot be called if we reach the state where both chain are in FLUSHING_COMPLETE because  `abortTransactionUnless(counterpartyChannel.state !== FLUSHCOMPLETE)` will bring to failure. That means that in this state we cannot send a `ChanUpgradeTimeout` datagram.  

Thus to ensure we can still timeout in this state, we should add a check in `ChanUpgradeOpen` checking if CounterPartyTimeoutHasExpired. If yes we could write an error, error that the counterparty chain can use to restore its channelEnd.

crodriguezvega (commented):
Thank you, @chatton. I think I wouldn't change the `==`/`!=` to `===`/`!==` for the regular if statements. In the rest of the specs the usage of `===`/`!==` is (mostly) limited to the boolean statements inside `assert` or `abortTransactionUnless`. In most of the if statements `==`/!=` are used, which is also closer to typescript, since that's the programming language we kinda use for the pseudocode.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
crodriguezvega: Please correct me if I'm wrong: we check the `counterpartyUpgradeTimeout` in `timeoutPacket` and `acknowledgePacket` and if the timeout has passed, then we execute `restoreChannel`, which should write an error receipt that can be used to cancel the upgrade on the counterparty. So in this case we wouldn't need to call `timeoutChannelUpgrade`, but we would call `cancelChannelUpgrade` instead?

And one question that was wondering after reading your comment: would it be possible for both channel ends to move to `FLUSH_COMPLETE` and then detect that the `counterpartyUpgradeTimeout` has passed? Asking because if the channel end moves to `FLUSH_COMPLETE` then we don't check the `counterpartyUpgradeTimeout` anymore in `timeoutPacket` and `acknowledgePacket`, right?

sangier: In my understanding: 
1. We only send a `timeoutChannelUpgrade` datagram if the timeout is detected in our end. Otherwise we write the error and then send a `cancelChannelUpgrade`. Thus, your statement is correct.
2. Once we reach the (`FLUSH_COMPLETE`,`FLUSH_COMPLETE`) the upgrade is mostly finalised and we only need to move to `OPEN`. At this point we don't care anymore about timeouts because all packets have been flushed.  

I also open this [PR](https://github.com/cosmos/ibc/pull/1038) to clarify the spec comment in **bold** mentioned in the previous message. As per this fix, we don't need to to ensure we can still timeout in this state. Thus no modification to `ChanUpgradeOpen` is required.

crodriguezvega (approved):
Thank you, @chatton!

PR #1038, opened 2023-11-20
specs comments details clarification

sangier:
Opening this PR, after discussion with Aditya, to clarify some comments in the protocol specs.

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -38,10 +38,10 @@ enum ChannelState {
 
 - In `ChanUpgradeInit`, the initializing chain that is proposing the upgrade should store the channel upgrade.
 - The counterparty chain executing `ChanUpgradeTry` that accepts the upgrade should store the channel upgrade, set the channel state from `OPEN` to `FLUSHING`, and start the flushing timer by storing an upgrade timeout.
-- Once the initiating chain verifies the counterparty is in `FLUSHING`, it must also move to `FLUSHING` unless all in-flight packets are already flushed on both ends, in which case it must move directly to `FLUSHCOMPLETE`. The initator will also store the counterparty timeout to ensure it does not move to `FLUSHCOMPLETE` after the counterparty timeout has passed.
-- The counterparty chain must prove that the initiator is  also in `FLUSHING` or completed flushing in `FLUSHCOMPLETE`. The counterparty will store the initiator timeout to ensure it does not move to `FLUSHCOMPLETE` after the initiator timeout has passed.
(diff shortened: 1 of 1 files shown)


spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -38,10 +38,10 @@ enum ChannelState {
...
+- The counterparty chain must prove that the initiator is also in `FLUSHING` or completed flushing in `FLUSHCOMPLETE`. The counterparty will store the initiator timeout to ensure it does not move to `FLUSHCOMPLETE` after the initiator timeout has passed.
 
-`FLUSHING` is a "blocking" states in that they will prevent the upgrade handshake from proceeding until the in-flight packets on both channel ends are flushed. Once both sides have moved to `FLUSHCOMPLETE`, a relayer can prove this on both ends with `ChanUpgradeOpen` to open the channel on both sides with the new parameters.
+`FLUSHING` is a "blocking" state that will prevent the upgrade handshake from proceeding until the in-flight packets on both channel ends are flushed. Once both sides have moved to `FLUSHCOMPLETE`, a relayer can prove this on both ends with `ChanUpgradeOpen` to open the channel on both sides with the new parameters.
```
crodriguezvega, line 44: >`FLUSHING` is a "blocking" state that will prevent the upgrade handshake from proceeding until the in-flight packets on both channel ends are flushed

Should we refine this sentence a bit more, since one channel end can move to `FLUSHCOMPLETE` independently from the other channel end, right?

crodriguezvega (approved):
Looks good, @sangier. Left just one comment.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
sangier: Yes I agree. What about: 

"FLUSHING` is a "blocking" state that prevents a channel end from advancing to `FLUSHCOMPLETE`, halting the upgrade handshake process, until the in-flight packets on its channel end are flushed. Once both sides have moved to `FLUSHCOMPLETE`, a relayer can prove this on both ends with `ChanUpgradeOpen` to open the channel on both sides with the new parameters."

AdityaSripal: I like your suggestion @sangier

AdityaSripal (changes requested):
Hi @sangier ,

The TRY step should not be able to go to FLUSHCOMPLETE. Both sides don't need to have finished flushing packets, but both sides need to have moved to the flushing state.

This is because if we get one side on FLUSHCOMPLETE, and the other on OPEN before even starting upgrade handshake we will reach problems completing handshake if the OPEN side rejects the upgrade or aborts later

AdityaSripal (approved):

PR #1040, opened 2023-11-21
fix: set upgrade timeout after `startFlushUpgradeHandshake in `chanUpgradeAck`

crodriguezvega:
Found this small bug when run the happy path quint test.

cc @AdityaSripal @sangier

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -647,16 +647,16 @@ function chanUpgradeAck(
     )
   )
 
-  upgrade = provableStore.get(channelUpgradePath(portIdentifier, channelIdentifier))
+  existingUpgrade = provableStore.get(channelUpgradePath(portIdentifier, channelIdentifier))
 
   // optimistically accept version that TRY chain proposes and pass this to callback for confirmation.
   // in the crossing hello case, we do not modify version that our TRY call returned and instead 
   // enforce that both TRY calls returned the same version
   if (channel.state == OPEN) {
-    upgrade.fields.version == counterpartyUpgrade.fields.version
+    existingUpgrade.fields.version == counterpartyUpgrade.fields.version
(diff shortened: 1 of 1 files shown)


spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -666,6 +666,10 @@ function chanUpgradeAck(
     // if we are already at flushing, then no state changes occur
     // upgrade is blocked on this channelEnd from progressing until flush completes on both ends
```
sangier, line 667:
```suggestion
    // upgrade is blocked on this channelEnd from progressing until flush completes on its end
```

sangier (approved):
Actually this is really important. Since we are overriding the upgrade at line 708 in `chanUpgradeAck`, without the proposed change we are violating the invariant that the upgrade.timeout MUST be set when we move to `FLUSHING` or directly to `FLUSHING_COMPLETE`.  Indeed we would be storing the timeout during the `startFlushUpgradeHandshake` (which contains the timeout value) and then override this value with an upgrade instance that does not contain the timeout at line 708.  

Really good catch! Thank you!

sangier:
Even though, what if instead, we just remove line 708?

crodriguezvega:
> Actually this is really important. Since we are overriding the upgrade at line 708 in `chanUpgradeAck`, without the proposed change we are violating the invariant that the upgrade.timeout MUST be set when we move to `FLUSHING` or directly to `FLUSHING_COMPLETE`. Indeed we would be storing the upgrade during the `startFlushUpgradeHandshake` (which contains the timeout value) and then override this value with an upgrade instance which contains the final agreed version but that does not contain the timeout at line 708.
> 
> Really good catch! Thank you!

Yes, correct! I checked ibc-go and we are good there. :)

AdityaSripal (approved):
Thanks for the catch @crodriguezvega ! Really happy that the formal verification work is paying off!

PR #1043, opened 2023-11-28
update `FungibleTokenPacketData` marshalling to match rest of spec

noot:
The `FungibleTokenPacketData` marshalling in the pseudocode was using protobuf, but the rest of the spec says "Note that both the `FungibleTokenPacketData` as well as `FungibleTokenPacketAcknowledgement` must be JSON-encoded (not Protobuf encoded) when they serialized into packet data." this is also what the ibc-go implementation does.


crodriguezvega (approved):
Good catch; thank you, @noot!

PR #1051, opened 2023-12-13
ics4-add active client check in sendpacket

### LINKED ISSUE #555: 04-channel sendPacket should add a check to ensure client is not frozen
Currently the spec assumes that all the client Verify functions ensure that the client is not frozen. The problem is that the `sendPacket` function does not call any verify functions. This allows packets to be sent with frozen clients. In the token transfer case, if the packet is never received, the tokens sent are lost since a timeout can never be processed. 

We check for this situation in the [code](https://github.com/cosmos/ibc-go/blob/372eaa42a9082bcbeaca344fc41151ab95f8b5cd/modules/core/04-channel/keeper/packet.go#L70). Given that the spec makes a reference to client freezing in `02-client`, I think it is ok to add this check into sendPacket. 

I also would like this to check to match the [proposed change](https://github.com/cosmos/ibc-go/issues/98) in `ibc-go` which would prevent users from sending funds on expired clients

sangier:
Closes: #555

Add: 
- `Status` function (in helper function section) to retrieve the client status 
- Check that client is active in `sendPacket`

### DIFF
--- a/spec/client/ics-008-wasm-client/README.md
+++ b/spec/client/ics-008-wasm-client/README.md
@@ -201,6 +201,28 @@ function verifyClientMessage(clientMsg: ClientMessage) {
 }
 ```
 
+### Retrieve client status 
+
+Return the status of the wasm client. Status can be either `Active`, `Expired`, `Unknown` or `Frozen`.
+
+```typescript
+interface StatusMsg {}
+```
+
+```typescript
+// returns the status of a client
+function Status(clientIdentifier: Identifier, clientState: clientState): Status {
+  payload = StatusMsg{}
+
+  // retrieve client identifier-prefixed store
+  clientStore = provableStore.prefixStore("clients/{clientIdentifier}")
+
+  // use underlying wasm contract to retrieve the status
+  result = callContract(clientStore, clientState, marshalJSON(payload))
+  return Status(result.status)
+} 
+```
+
 ### Misbehaviour predicate
 
 Function `checkForMisbehaviour` will check if an update contains evidence of misbehaviour. Wasm client misbehaviour checking determines whether or not two conflicting headers at the same height would have convinced the light client.
--- a/spec/core/ics-002-client-semantics/README.md
+++ b/spec/core/ics-002-client-semantics/README.md
@@ -337,6 +337,16 @@ Once misbehaviour is detected, clients SHOULD be frozen so that no future update
 A permissioned entity such as a chain governance system or trusted multi-signature MAY be allowed
 to intervene to unfreeze a frozen client & provide a new correct ClientMessage which updates the client to a valid state.
 
+#### Retrieve Client Status 
+
+Status is an opaque function defined by a client type to retrieve the current clientState. Status can be either Active, Expired, Unknown or Frozen. 
+
+The function is defined as:
+
+```typescript
+type Status = (Client) => Void
+```
+
 #### `CommitmentProof`
 
 `CommitmentProof` is an opaque data structure defined by a client type in accordance with [ICS 23](../ics-023-vector-commitments).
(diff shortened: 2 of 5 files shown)


spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -704,6 +704,29 @@ function getCounterPartyHops(proof: CommitmentProof | MultihopProof, lastConnect
...
+    return Unknown
+  }
+  // Check if Expired
+  let expirationTime := consState.Timestamp + client.TrustingPeriod
+  if (expirationTime <== now){
+    return Expired 
+  }
+
+  return Active
+}
```
crodriguezvega, line 727: Since this implementation of `Status` is specific to Tendermint, I think I would move it to ICS 07.

spec/core/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -763,8 +787,14 @@ function sendPacket(
     // check that the channel must be OPEN to send packets;
     abortTransactionUnless(channel !== null)
     abortTransactionUnless(channel.state === OPEN)
+
     connection = provableStore.get(connectionPath(channel.connectionHops[0]))
     abortTransactionUnless(connection !== null)
+    
+    client = provableStore.get(clientStatePath(connection.clientIdenfier))
```
crodriguezvega, line 794: You could do here `clientState = queryClientState(connection.clientIdenfier)` (`queryClientState` is defined in ICS 02)

crodriguezvega (commented):
Thank you, @sangier. I think you could add the `Status` function to ICS 02, together with the other functions defined for client state ([e.g. `verifyClientMessage`, `updateState`, etc](https://github.com/cosmos/ibc/blob/main/spec/core/ics-002-client-semantics/README.md#validity-predicate)).

spec/core/ics-004-channel-and-packet-semantics/README.md
sangier: Gotcha, thank you @crodriguezvega! Please, verify if the latest commit [c1637a9](https://github.com/cosmos/ibc/pull/1051/commits/c1637a98afe3220a86a9a443dbd8adba2cfdcac6) address all of your comments.

spec/core/ics-002-client-semantics/README.md
```diff
@@ -337,6 +337,16 @@ Once misbehaviour is detected, clients SHOULD be frozen so that no future update
...
 to intervene to unfreeze a frozen client & provide a new correct ClientMessage which updates the client to a valid state.
 
+#### Retrieve Client Status 
+
+Status is an opaque function defined by a client type to retrieve the current clientState. Status can be either Active, Expired, Unknown or Frozen. 
+
+The function is defined as:
+
+```typescript
+type Status = (Client) => Void
```
AdityaSripal, line 347: SHouldn't the return value not be Void? Since it is returning a status

sangier: Uh, definitively should not be void. @AdityaSripal would be correct to return `bytes` or shall we define a `type Status = bytes` in the Data Structure section and return `Status` itself?

spec/client/ics-008-wasm-client/README.md
```diff
@@ -201,6 +201,28 @@ function verifyClientMessage(clientMsg: ClientMessage) {
 }
 ```
 
+### Retrieve client status 
+
+Return the status of the wasm client. Status can be either `Active`, `Expired`, `Unknown` or `Frozen`.
```
sangier, line 206: Do we need the `Unknown` status for the wasm-client? Where it is used?

spec/core/ics-002-client-semantics/README.md
```diff
@@ -337,6 +337,25 @@ Once misbehaviour is detected, clients SHOULD be frozen so that no future update
 A permissioned entity such as a chain governance system or trusted multi-signature MAY be allowed
 to intervene to unfreeze a frozen client & provide a new correct ClientMessage which updates the client to a valid state.
 
+#### Retrieve client status 
+
+Status is an opaque function defined by a client type to retrieve the current clientState. Status can be either `Active`, `Expired`, `Unknown` or `Frozen`.
```
sangier, line 342: similarly here, If `Unknown` status is not needed in the wasm-client we can remove it from client-semantics too.

swift1337:
Closing due to inactivity

PR #1057, opened 2023-12-14
ics001: add ics header field description

sangier:
The `kind` field is used in all the specs header, however it is not described in ics-001.


crodriguezvega (approved):
Thanks for the improvement, @sangier!

PR #1058, opened 2023-12-14
specify units of `timeoutTimestamp` in ICS20

noot:
Previously the units for `timeoutTimestamp` were not specified and it's unclear whether they were unix seconds or nanoseconds.
This specifies that the units are unix nanoseconds as per the Go implementation: https://github.com/cosmos/ibc-go/blob/d0cf5b64dc8756206dad6c922e3530535ab78b79/proto/ibc/applications/transfer/v1/tx.proto#L47


crodriguezvega (approved):
Lovely, thank you very much!

AdityaSripal (approved):

sangier (approved):

Issue #957, opened 2023-04-05
ICS 04: Rename UpgradeTimeout to Timeout


colin-axner:
I'd propose renaming [`UpgradeTimeout`](https://github.com/cosmos/ibc/blob/main/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md#upgradetimeout) to `Timeout`:

```
interface Timeout {
    height: Height
    timestamp: uint64
}
```

The semantics of `UpgradeTimeout` are identical to the timeouts used in the packet lifecycle, the only difference is that there exists a variable type that encapsulates both. This seems like a general type that would be reused in any feature that requires timeouts. Was the intention to avoid confusion to why the packet timeout does not use this type?

chandiniv1:
Hello @colin-axner 
can I work on this issue?

crodriguezvega:
> Hello @colin-axner
> can I work on this issue?

Yes, that's fine. Thank you, @chandiniv1.

gjermundgaraba:
This issue was mistakenly reopened during some integration work, so closing again now.

PR #1064, opened 2023-12-21
rename UpgradeTimeout to Timeout

chandiniv1:
closes: #957

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -300,13 +300,13 @@ function startFlushUpgradeHandshake(
 
   channel.state = FLUSHING
 
-  timeout = getTimeout(channel.portIdentifier, channel.channelIdentifier)
+  upgradeTimeout = getUpgradeTimeout(channel.portIdentifier, channel.channelIdentifier)
   // either timeout height or timestamp must be non-zero
-  abortTransactionUnless(timeout.timeoutHeight != 0 || timeout.timeoutTimestamp != 0)
+  abortTransactionUnless(upgradeTimeout.timeoutHeight != 0 || upgradeTimeout.timeoutTimestamp != 0)
 
   nextSequenceSend = provableStore.get(nextSequenceSendPath(portIdentifier, channelIdentifier))
 
-  upgrade.timeout = timeout
+  upgrade.timeout = upgradeTimeout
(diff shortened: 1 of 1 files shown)


spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -106,7 +106,7 @@ The upgrade type will represent a particular upgrade attempt on a channel end.
 ```typescript
 interface Upgrade {
   fields: UpgradeFields
-  timeout: UpgradeTimeout
+  timeout: Timeout
```
crodriguezvega, line 109: I think these changes are good.

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -421,17 +421,17 @@ function pendingInflightPacketSequences(
...
 ```
 
-`getUpgradeTimeout` will return the upgrade timeout specified for the given channel. This may be a chain-wide parameter, or it can be a parameter chosen per channel. This is an implementation-level detail, so only the function signature is specified here. Note this should retrieve some stored timeout delta for the channel and add it to the current height and time to get the absolute timeout values.
+`getTimeout` will return the upgrade timeout specified for the given channel. This may be a chain-wide parameter, or it can be a parameter chosen per channel. This is an implementation-level detail, so only the function signature is specified here. Note this should retrieve some stored timeout delta for the channel and add it to the current height and time to get the absolute timeout values.
 
 ```typescript
-// getUpgradeTimeout
-function getUpgradeTimeout(portIdentifier: string, channelIdentifier: string) UpgradeTimeout {
+// getTimeout
+function getTimeout(portIdentifier: string, channelIdentifier: string) Timeout {
```
crodriguezvega, line 428: And the return type here is also good. But Every other change I would revert, since it makes sense still to have the word `upgrade` in the variable and functions names.

crodriguezvega (changes requested):
Thanks, @chandiniv1

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
chandiniv1: @crodriguezvega I have reverted the changes

AdityaSripal (approved):

crodriguezvega (approved):

PR #1068, opened 2024-01-18
fix: update spec organisation for CTA

charleenfei:
I went thru a few different configurations and ultimately think this is the easiest way to ask for the CTA. If people integrate just these Transport/Core standards, they should be able to interoperate w IBC regardless of what their underlying client/relayer impl looks like, this is why I didn't add those in. 

We want people to contribute to specific client/relayer impl if they wish as well, but those are not essential for interoperability, as the 02-client, 03-connection, 04-channel etc etc specs are.


crodriguezvega (approved):
Thanks, @charleenfei!

sangier (approved):

AdityaSripal (approved):

PR #1069, opened 2024-01-18
ICS4: Prevent new UpgradeInits from causing issues on handshakes

AdityaSripal:
- When cancelling the sequences should match if we are in FLUSH_COMPLETE (there is still a problem if one side times out, INITs, and cancel at the same time).
- documents that a new upgrade should only be started if the previous upgrade completed on both sides.


spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -462,6 +464,8 @@ A relayer may then submit a `ChanUpgradeCancel` datagram to the counterparty. Up
 If a chain does not reach `FLUSHCOMPLETE` within the counterparty specified timeout, then it MUST NOT move to `FLUSHCOMPLETE` and should instead abort the upgrade. A relayer may submit a proof of this to the counterparty chain in a `ChanUpgradeTimeout` datagram so that counterparty cancels the upgrade and restores its original channel as well.
 
 ```typescript
+// Channel Ends on both sides **must** be OPEN before this function is called
+// It is the responsibility of the authorized Upgrador to ensure this is the case
```
sangier, line 468:
```suggestion
// It is the responsibility of the authorized Upgrader to ensure this is the case
```

sangier (approved):
Thanks @AdityaSripal !!

crodriguezvega (approved):

PR #1072, opened 2024-01-19
Add diagram to channel upgradability spec

adiraviraj:
this PR adds a diagram to the `UPGRADES.md` file in `ics-004-channel-and-packet-semantics` to depict the happy path of a channel upgrade.

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -442,6 +442,8 @@ A successful protocol execution flows as follows (note that all calls are made t
 | Relayer   | `ChanUpgradeAck`     | A                | (OPEN, FLUSHING)                   | (FLUSHING/FLUSHCOMPLETE, FLUSHING)                    |
 | Relayer   | `ChanUpgradeConfirm` | B                | (FLUSHING/FLUSHCOMPLETE, FLUSHING) | (FLUSHING/FLUSHCOMPLETE, FLUSHING/FLUSHCOMPLETE/OPEN) |
 
+![Channel Upgrade Flow](channel-upgrade-flow.png)
+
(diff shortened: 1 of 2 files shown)


DimitrisJim (approved):
looks dope, one thing I'd note is that `ChanUpgradeOpen` step could be done on both chains (guess conditionally based on if UpgradeConfirm moved one end to OPEN).

adiraviraj:
> one thing I'd note is that `ChanUpgradeOpen` step could be done on both chains (guess conditionally based on if UpgradeConfirm moved one end to OPEN).

@DimitrisJim yea i was going by the case where B is already in OPEN post `ChanUpgradeConfirm`. but i can edit the diagram if needed to show `ChanUpgradeOpen` on both ends.

DimitrisJim:
whatever the spec gang requests :muscle:

crodriguezvega:
Thank you, @adiraviraj. For clarification, I think we could add some text (either above or below the diagram) explaining that this is a possible upgrade flow and that in this particular sample on step 7 after `ChanUpgradeConfirm` the channel end opens with the new parameters. You could also mention that where several channel states are listed it means that those are possible channel states after executing that step of the handshake.

adiraviraj:
@crodriguezvega i've added the the text your requested to give additional context on the diagram.

crodriguezvega (approved):
Very cool. Thank you, @adiraviraj!

sangier (approved):
Thanks @adiraviraj for the great example diagram!!

AdityaSripal:
Thanks @adiraviraj ! There is no onChanUpgradeConfirm callback, so to correctly depict the situation I would call `onChanUpgradeOpen` to chainB's app on ChanUpgradeConfirm

adiraviraj:
thank you @AdityaSripal. i've made the requested change to the diagram.

AdityaSripal (approved):

PR #1073, opened 2024-01-19
Remove double call to restore callback.

DimitrisJim:
this is called in `restoreChannel` so appears this wasn't removed during a refactor.


crodriguezvega (approved):
Good catch!

sangier (approved):
Thanks @DimitrisJim!

PR #1074, opened 2024-01-22
fix: allow recv packet when channel state is flush complete

crodriguezvega:
Found during review of https://github.com/cosmos/ibc-go/pull/5668.


sangier (approved):
Thanks @crodriguezvega! LGTM, since in any case we still have this defensive check to avoid new packet processing `abortTransactionUnless(counterpartyUpgrade.nextSequenceSend == 0 || packet.sequence < counterpartyUpgrade.nextSequenceSend)`

AdityaSripal (approved):

PR #1077, opened 2024-01-25
ICS4: fix off-by-one error in non-crossing-hello case

AdityaSripal:
If on TRY, the sequence is the same but the upgrade doesn't exist. Then the counterparty upgrade is stale so we must write upgrade error for that sequence


sangier (approved):
Thanks @AdityaSripal!

crodriguezvega (approved):

crodriguezvega (approved):

PR #1082, opened 2024-02-08
Add diagram to ICS-20

sangier:
Adding a diagram to visually represent the example `A -> B -> C -> A -> C -> B -> A`. 

The diagram is based on my understanding of [ICS-20](https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer) and the explanations given in the [ibc-go implementation](https://github.com/cosmos/ibc-go/blob/457095517b7832c42ecf13571fee1e550fec02d0/modules/apps/transfer/keeper/relay.go#L18-L49) and the [adr-001](https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md).


crodriguezvega (approved):
Good idea to add the diagram!

PR #1090, opened 2024-03-28
ICS20-2: Add Token Forwarding ability to ICS20-2

AdityaSripal:
Allows users to atomically route their tokens through a series of paths with a single packet.

Desired properties:
- Atomicity: Tokens either get fully sent to final destination chain and receiver or they return to original sender chain/sender. This means that on error or timeout, the intermediate chains must revert state such that there is no difference between their current state and their state prior to the original send (modulo unrelated state changes)
- Correctness: The token denoms and all the escrow accounts across the path must be the exact same as it would be if the tokens travelled through each hop with a separate packet

Special thanks to the @strangelove-ventures team for working on the PFM middleware which serves as a precursor to this feature

### DIFF
--- a/spec/app/ics-020-fungible-token-transfer/README.md
+++ b/spec/app/ics-020-fungible-token-transfer/README.md
@@ -51,7 +51,6 @@ interface FungibleTokenPacketDataV2 {
   sender: string
   receiver: string
   memo: string
-  unwind: boolean
   forwardingPath: []string
 }
 
@@ -244,7 +243,6 @@ function sendFungibleTokens(
   sender: string,
   receiver: string,
   memo: string,
-  unwind: boolean,
   forwardingPath: []string,
   sourcePort: string,
   sourceChannel: string,
@@ -275,18 +273,15 @@ function sendFungibleTokens(
   if transferVersion == "ics20-1" {
     abortTransactionUnless(len(tokens) == 1)
     token = tokens[0]
+    // abort if forwardingPath defined
+    abortTransactionUnless(forwardingPath == nil)
     // create v1 denom of the form: port1/channel1/port2/channel2/port3/channel3/denom
     v1Denom = constructOnChainDenom(token.trace, token.denom)
-    // v1 packet data does not support unwind and forwardingPath fields
+    // v1 packet data does not support forwardingPath fields
     data = FungibleTokenPacketData{v1Denom, token.amount, sender, receiver, memo}
   } else if transferVersion == "ics20-2" {
-    // unwinding and forwarding features disabled for multi-token packets
-    if len(tokens) > 1 {
-      abortTransactionUnless(unwind == false)
-      abortTransactionUnless(forwardingPath == nil)
-    }
     // create FungibleTokenPacket data
-    data = FungibleTokenPacketDataV2{tokens, sender, receiver, memo, unwind, forwardingPath}
+    data = FungibleTokenPacketDataV2{tokens, sender, receiver, memo, forwardingPath}
   } else {
     // should never be reached as transfer version must be negotiated to be either
     // ics20-1 or ics20-2 during channel handshake
@@ -330,35 +325,32 @@ function onRecvPacket(packet: Packet) {
   } else if transferVersion == "ics20-2" {
     FungibleTokenPacketDataV2 data = UnmarshalJSON(packet.data)
     tokens = data.tokens
-    // unwinding and forwarding features disabled for multi-token packets
-    if len(tokens) > 1 {
-      if unwind || forwardingPath != nil {
-        return FungibleTokenPacketAcknowledgement{false, "cannot set unwinding and forwarding fields if more than 1 token sent"}
-      }
-    }
     // if we need to forward the tokens onward
     // temporarily send to the channel escrow address of the intended receiver
-    if unwind == true || len(forwardingPath) > 0 {
+    if len(forwardingPath) > 0 {
       receiver = channelForwardingAddresses[packet.destinationChannel]
     }
-
   } else {
     // should never be reached as transfer version must be negotiated to be either
     // ics20-1 or ics20-2 during channel handshake
     abortTransactionUnless(false)
   }
 
+  assert(packet.sender !== "")
+  assert(receiver !== "")
+    
   // construct default acknowledgement of success
   FungibleTokenPacketAcknowledgement ack = FungibleTokenPacketAcknowledgement{true, null}
+
   prefix = "{packet.sourcePort}/{packet.sourceChannel}/"
+  receivedTokens = []Token
   for token in tokens {
     assert(token.denom !== "")
     assert(token.amount > 0)
-    assert(token.sender !== "")
-    assert(token.receiver !== "")
-      
+    
     // we are the source if the packets were prefixed by the sending chain
     source = token.trace[0] == prefix
+    var onChainTrace []string
     if source {
       // since we are receiving back to source we remove the prefix from the trace
       onChainTrace = token.trace[1:]
@@ -376,8 +368,8 @@ function onRecvPacket(packet: Packet) {
     } else {
       // since we are receiving to a new sink zone we prepend the prefix to the trace
       prefix = "{packet.destPort}/{packet.destChannel}/"
-      newTrace = append([]string{prefix}, token.trace...)
-      onChainDenom = constructOnChainDenom(newTrace, token.denom)
+      onChainTrace = append([]string{prefix}, token.trace...)
+      onChainDenom = constructOnChainDenom(onChainTrace, token.denom)
       // sender was source, mint vouchers to receiver (assumed to fail if balance insufficient)
       err = bank.MintCoins(receiver, onChainDenom, token.amount)
       if (err !== nil) {
@@ -386,59 +378,33 @@ function onRecvPacket(packet: Packet) {
         break
       }
     }
+
+    // add the received token to the received tokens list
+    recvToken = Token{
+      denom: token.denom,
+      trace: onChainTrace,
+      amount: token.amount,
+    }
+    receivedTokens = append(receivedTokens, recvToken)
   }
 
-  // if there is an error ack return immediately and do not unwind or forward further
+  // if there is an error ack return immediately and do not forward further
   if !ack.Success() {
     return ack
   }
 
-  // for v2 we will implement the unwinding and forwarding if the receive at this step is successful
-  if unwind == true {
-    // check if we need to continue unwinding by seeing if the trace is more than 1 hop long, since the packet data should have already been unwound by one hop by receiving to this chain
-    token = tokens[0]
-    if len(token.trace) > 1 {
-      nextHop = token.trace
-      nextPort, nextChannel = split(nextHop)
-      unwind = len(token.trace) > 2
-      nextToken = Token{
-        denom: token.denom,
-        trace: token.trace[1:],
-        amount: token.amount,
-      }
-      nextSequence = sendFungibleTokens(
-        []Token{nextToken},
-        receiver, // current receiver is the new sender
-        data.receiver, // retain original destination receiver
-        memo,
-        unwind,
-        forwardingPath,
-        nextPort,
-        nextChannel,
-        Height{},
-        currentTime() + DefaultHopTimeoutPeriod,
-      )
-      // store packet for future sending ack
-      store.set(packetForwardPath(nextPort, nextChannel, nextPacketSequence), packet)
-      // use async ack until we get successful acknowledgement from further down the line.
-      return nil
-    }
-  }
-  // if unwinding is complete, then start forwarding
+  // if acknowledgement is successful and forwarding path set
+  // then start forwarding
   if forwardingPath != nil {
-    nextPort, nextChannel = split(forwardingPath[0])
-    nextTrace = traceFromDenom(onChainDenom)
-    nextToken = Token{
-      denom: token.denom,
-      trace: nextTrace,
-      amount: token.amount,
-    }
+    nextPort, nextChannel = split(forwardingPath[0], "/")
+    // send the tokens we received above to the next port and channel
+    // on the forwarding path
+    // and reduce the forwardingPath by the first element
     nextSequence = sendFungibleTokens(
-      tokens: []Token{nextToken},
+      tokens: receivedTokens,
       sender: receiver,
       receiver: data.receiver,
       memo,
-      false,
       forwardingPath[1:],
       nextPort,
       nextChannel,
@@ -578,32 +544,33 @@ function refundTokens(packet: Packet) {
 // to ensure atomic packet forwarding
 function revertInFlightChanges(sentPacket: Packet, receivedPacket: Packet) {
   // the token on our chain is the token in the sentPacket
-  token = sentPacket.tokens[0]
-  // check if the packet we sent out was sending as source or not
-  // in this case we escrowed the outgoing tokens
-  if isSource(sentPacket.sourcePort, sentPacket.sourceChannel, token) {
-    // check if the packet we received was a source token for our chain
-    if isSource(receivedPacket.destinationPort, receivedPacket.desinationChannel, token) {
-      // receive sent tokens from the received escrow to the forward escrow account
-      // so we must send the tokens back from the forward escrow to the original received escrow account
-      forwardEscrow = channelEscrowAddresses[sentPacket.sourceChannel]
-      reverseEscrow = channelEscrowAddresses[packet.destChannel]
-      bank.TransferCoins(forwardEscrow, reverseEscrow, token.denom, token.amount)
+  for token in sentPacket.tokens {
+    // check if the packet we sent out was sending as source or not
+    // in this case we escrowed the outgoing tokens
+    if isSource(sentPacket.sourcePort, sentPacket.sourceChannel, token) {
+      // check if the packet we received was a source token for our chain
+      if isSource(receivedPacket.destinationPort, receivedPacket.desinationChannel, token) {
+        // receive sent tokens from the received escrow to the forward escrow account
+        // so we must send the tokens back from the forward escrow to the original received escrow account
+        forwardEscrow = channelEscrowAddresses[sentPacket.sourceChannel]
+        reverseEscrow = channelEscrowAddresses[packet.destChannel]
+        bank.TransferCoins(forwardEscrow, reverseEscrow, token.denom, token.amount)
+      } else {
+        // receive minted vouchers and sent to the forward escrow account
+        // so we must remove the vouchers from the forward escrow account and burn them
+        bank.BurnCoins(forwardEscrow, token.denom, token.amount)
+      }
     } else {
-      // receive minted vouchers and sent to the forward escrow account
-      // so we must remove the vouchers from the forward escrow account and burn them
-      bank.BurnCoins(forwardEscrow, token.denom, token.amount)
+      // in this case we burned the vouchers of the outgoing packets
+      // check if the packet we received was a source token for our chain
+      // in this case, the tokens were unescrowed from the reverse escrow account
+      if isSource(receivedPacket.destinationPort, receivedPacket.desinationChannel, token) {
+        // in this case we must mint the burned vouchers and send them back to the escrow account
+        bank.MintCoins(reverseEscrow, token.denom, token.amount)
+      }
+      // if it wasn't a source token on receive, then we simply had minted vouchers and burned them in the receive.
+      // So no state changes were made, and thus no reversion is necessary
     }
-  } else {
-    // in this case we burned the vouchers of the outgoing packets
-    // check if the packet we received was a source token for our chain
-    // in this case, the tokens were unescrowed from the reverse escrow account
-    if isSource(receivedPacket.destinationPort, receivedPacket.desinationChannel, token) {
-      // in this case we must mint the burned vouchers and send them back to the escrow account
-      bank.MintCoins(reverseEscrow, token.denom, token.amount)
-    }
-    // if it wasn't a source token on receive, then we simply had minted vouchers and burned them in the receive.
-    // So no state changes were made, and thus no reversion is necessary
   }
 }
 ```
--- a/spec/app/ics-020-fungible-token-transfer/README.md
+++ b/spec/app/ics-020-fungible-token-transfer/README.md
@@ -51,6 +51,8 @@ interface FungibleTokenPacketDataV2 {
   sender: string
   receiver: string
   memo: string
+  unwind: boolean
+  forwardingPath: []string
 }
 
 interface Token {
@@ -92,6 +94,7 @@ The fungible token transfer bridge module tracks escrow addresses associated wit
 ```typescript
 interface ModuleState {
   channelEscrowAddresses: Map<Identifier, string>
+  channelForwardingAddresses: Map<Identifier, string>
 }
 ```
 
@@ -241,6 +244,8 @@ function sendFungibleTokens(
   sender: string,
   receiver: string,
   memo: string,
+  unwind: boolean,
+  forwardingPath: []string,
   sourcePort: string,
   sourceChannel: string,
   timeoutHeight: Height,
@@ -269,11 +274,19 @@ function sendFungibleTokens(
   transferVersion = getAppVersion(channel.version)
   if transferVersion == "ics20-1" {
     abortTransactionUnless(len(tokens) == 1)
-    v1Denom = tokens[0].trace + tokens[0].denom)
-    data = FungibleTokenPacketData{v1Denom, tokens[0].amount, sender, receiver, memo}
+    token = tokens[0]
+    // create v1 denom of the form: port1/channel1/port2/channel2/port3/channel3/denom
+    v1Denom = constructOnChainDenom(token.trace, token.denom)
+    // v1 packet data does not support unwind and forwardingPath fields
+    data = FungibleTokenPacketData{v1Denom, token.amount, sender, receiver, memo}
   } else if transferVersion == "ics20-2" {
+    // unwinding and forwarding features disabled for multi-token packets
+    if len(tokens) > 1 {
+      abortTransactionUnless(unwind == false)
+      abortTransactionUnless(forwardingPath == nil)
+    }
     // create FungibleTokenPacket data
-    data = FungibleTokenPacketDataV2{tokens, sender, receiver, memo}
+    data = FungibleTokenPacketDataV2{tokens, sender, receiver, memo, unwind, forwardingPath}
   } else {
     // should never be reached as transfer version must be negotiated to be either
     // ics20-1 or ics20-2 during channel handshake
@@ -302,6 +315,8 @@ function onRecvPacket(packet: Packet) {
   // getAppVersion returns the transfer version that is embedded in the channel version
   // as the channel version may contain additional app or middleware version(s)
   transferVersion = getAppVersion(channel.version)
+  var tokens []Token
+  var receiver string
   if transferVersion == "ics20-1" {
      FungibleTokenPacketData data = UnmarshalJSON(packet.data)
      trace, denom = parseICS20V1Denom(data.denom)
@@ -311,9 +326,22 @@ function onRecvPacket(packet: Packet) {
        amount: packet.amount
      }
      tokens = []Token{token}
+     receiver = packet.receiver
   } else if transferVersion == "ics20-2" {
     FungibleTokenPacketDataV2 data = UnmarshalJSON(packet.data)
     tokens = data.tokens
+    // unwinding and forwarding features disabled for multi-token packets
+    if len(tokens) > 1 {
+      if unwind || forwardingPath != nil {
+        return FungibleTokenPacketAcknowledgement{false, "cannot set unwinding and forwarding fields if more than 1 token sent"}
+      }
+    }
+    // if we need to forward the tokens onward
+    // temporarily send to the channel escrow address of the intended receiver
+    if unwind == true || len(forwardingPath) > 0 {
+      receiver = channelForwardingAddresses[packet.destinationChannel]
+    }
+
   } else {
     // should never be reached as transfer version must be negotiated to be either
     // ics20-1 or ics20-2 during channel handshake
@@ -339,7 +367,7 @@ function onRecvPacket(packet: Packet) {
       // determine escrow account
       escrowAccount = channelEscrowAddresses[packet.destChannel]
       // unescrow tokens to receiver (assumed to fail if balance insufficient)
-      err = bank.TransferCoins(escrowAccount, data.receiver, onChainDenom, token.amount)
+      err = bank.TransferCoins(escrowAccount, receiver, onChainDenom, token.amount)
       if (err != nil) {
         ack = FungibleTokenPacketAcknowledgement{false, "transfer coins failed"}
         // break out of for loop on first error
@@ -351,7 +379,7 @@ function onRecvPacket(packet: Packet) {
       newTrace = append([]string{prefix}, token.trace...)
       onChainDenom = constructOnChainDenom(newTrace, token.denom)
       // sender was source, mint vouchers to receiver (assumed to fail if balance insufficient)
-      err = bank.MintCoins(data.receiver, onChainDenom, token.amount)
+      err = bank.MintCoins(receiver, onChainDenom, token.amount)
       if (err !== nil) {
         ack = FungibleTokenPacketAcknowledgement{false, "mint coins failed"}
         // break out of for loop on first error
@@ -359,6 +387,70 @@ function onRecvPacket(packet: Packet) {
       }
     }
   }
+
+  // if there is an error ack return immediately and do not unwind or forward further
+  if !ack.Success() {
+    return ack
+  }
+
+  // for v2 we will implement the unwinding and forwarding if the receive at this step is successful
+  if unwind == true {
+    // check if we need to continue unwinding by seeing if the trace is more than 1 hop long, since the packet data should have already been unwound by one hop by receiving to this chain
+    token = tokens[0]
+    if len(token.trace) > 1 {
+      nextHop = token.trace
+      nextPort, nextChannel = split(nextHop)
+      unwind = len(token.trace) > 2
+      nextToken = Token{
+        denom: token.denom,
+        trace: token.trace[1:],
+        amount: token.amount,
+      }
+      nextSequence = sendFungibleTokens(
+        []Token{nextToken},
+        receiver, // current receiver is the new sender
+        data.receiver, // retain original destination receiver
+        memo,
+        unwind,
+        forwardingPath,
+        nextPort,
+        nextChannel,
+        Height{},
+        currentTime() + DefaultHopTimeoutPeriod,
+      )
+      // store packet for future sending ack
+      store.set(packetForwardPath(nextPort, nextChannel, nextPacketSequence), packet)
+      // use async ack until we get successful acknowledgement from further down the line.
+      return nil
+    }
+  }
+  // if unwinding is complete, then start forwarding
+  if forwardingPath != nil {
+    nextPort, nextChannel = split(forwardingPath[0])
+    nextTrace = traceFromDenom(onChainDenom)
+    nextToken = Token{
+      denom: token.denom,
+      trace: nextTrace,
+      amount: token.amount,
+    }
+    nextSequence = sendFungibleTokens(
+      tokens: []Token{nextToken},
+      sender: receiver,
+      receiver: data.receiver,
+      memo,
+      false,
+      forwardingPath[1:],
+      nextPort,
+      nextChannel,
+      Height{},
+      currentTime() + DefaultHopTimeoutPeriod,
+    )
+    // store packet for future sending ack
+    store.set(packetForwardPath(nextPort, nextChannel, nextPacketSequence), packet)
+    // use async ack until we get successful acknowledgement from further down the line.
+    return nil
+  }
+
   return ack
 }
 ```
@@ -373,6 +465,30 @@ function onAcknowledgePacket(
   if !(acknowledgement.success) {
     refundTokens(packet)
   }
+
+  // check if the packet was sent is from a previously forwarded packet
+  prevPacket = store.get(packetForwardPath(packet.sourcePort, packet.sourceChannel))
+
+  if prevPacket != nil {
+    if acknowledgement.success {
+      FungibleTokenPacketAcknowledgement ack = FungibleTokenPacketAcknowledgement{true, "forwarded packet succeeded"}
+      handler.writeAcknowledgement(
+        prevPacket,
+        ack,
+      )
+    } else {
+      // the forwarded packet has failed, thus the funds have been refunded to the forwarding address.
+      // we must revert the changes that came from successfully receiving the tokens on our chain
+      // before propogating the error acknowledgement back to original sender chain
+      revertInFlightChanges(packet, prevPacket)
+      // write error acknowledgement
+      FungibleTokenPacketAcknowledgement ack = FungibleTokenPacketAcknowledgement{false, "forwarded packet failed"}
+      handler.writeAcknowledgement(
+        prevPacket,
+        ack,
+      )
+  }
+
 }
 ```
 
@@ -382,6 +498,31 @@ function onAcknowledgePacket(
 function onTimeoutPacket(packet: Packet) {
   // the packet timed-out, so refund the tokens
   refundTokens(packet)
+
+  // check if the packet was sent is from a previously forwarded packet
+  prevPacket = store.get(packetForwardPath(packet.sourcePort, packet.sourceChannel))
+
+  if prevPacket != nil {
+    // the forwarded packet has failed, thus the funds have been refunded to the forwarding address.
+    // we must revert the changes that came from successfully receiving the tokens on our chain
+    // before propogating the error acknowledgement back to original sender chain
+    revertInFlightChanges(packet, prevPacket)
+    // write error acknowledgement
+    FungibleTokenPacketAcknowledgement ack = FungibleTokenPacketAcknowledgement{false, "forwarded packet failed"}
+    handler.writeAcknowledgement(
+      prevPacket,
+      ack,
+    )
+  }
+}
+```
+
+##### Helper functions
+
+```typescript
+function isSource(portId: string, channelId: string, token: Token) {
+  firstTrace = "{portId}/{channelId}"
+  return token.trace[0] == firstTrace
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jtieri:
One important aspect of PFM is the ability to compose more complex user experiences by integrating multiple middleware in the stack together, so that you can essentially receive an ICS-20 transfer, perform some action on the intermediate chain, and then forward tokens to the destination

Example: A user wants to send tokens from a source chain to an intermediate chain. On the intermediate chain the funds will be received via `OnRecvPacket`, now that the funds are in the user account the tokens will be swapped in a liquidity pool for a new asset. If the swap is successful, forward the tokens received from the swap to a different account on a destination chain

With middleware this flow would involve wiring up the stack like this:

```
channel.OnRecvPacket
swap.OnRecvPacket
forward.OnRecvPacket
transfer.OnRecvPacket
```

Then when a packet comes in for an ICS-20 transfer, the swap middleware would call into the underlying app and that call would fall through to the transfer module so that the appropriate bookkeeping could take place to get funds into the user controlled account. Control would then be passed back up to the swap middleware where the swap could take place, the middleware would mutate the `FungibleTokenPacketData` so that the token denom and amount refer to the token that was received from the swap. Then control is passed to the forward middleware to initiate the forward.

[Swap middleware](https://github.com/duality-labs/duality/tree/main/x/ibcswap)

Bringing token forwarding into the ICS-20 protocol is amazing to hear, but I think preserving this type of composability for developers to build these more complicated experiences is an important aspect; otherwise, it becomes a bit less appealing.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+        // so we must remove the vouchers from the forward escrow account and burn them
+        bank.BurnCoins(forwardEscrow, token.denom, token.amount)
+      }
+    } else {
+      // in this case we burned the vouchers of the outgoing packets
+      // check if the packet we received was a source token for our chain
+      // in this case, the tokens were unescrowed from the reverse escrow account
+      if isSource(receivedPacket.destinationPort, receivedPacket.desinationChannel, token) {
+        // in this case we must mint the burned vouchers and send them back to the escrow account
+        bank.MintCoins(reverseEscrow, token.denom, token.amount)
```
crodriguezvega, line 568: `reserveEscrow` is not defined in this scope.

crodriguezvega (commented):
Thank you for the quick turnaround with the changes, @AdityaSripal.

I did a first pass and left a few comments. I think my biggest gap of understanding at the moment is in the usage of `channelForwardingAddresses` (I think there might be some logic errors with that). 

To help with the reasoning I tried to make a coupe of diagrams. Both go through the scenario where and error ack is written on the destination chain, but in one of them the middle hop is the source of the token and in the other one it is not.

path-forwarding-error-ack

([excalidraw link](https://excalidraw.com/#json=csvnfNttmfu3c3Od9iAqE,Goy7F59kKQzDj6iHLx8sWQ))

The diagram is not completely correct, because I think the logic to transfer/burn token on the middle hop needs to be adjusted a bit.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -348,17 +368,55 @@ function onRecvPacket(packet: Packet) {
...
+      receiver: data.receiver,
+      memo,
+      forwardingPath[1:],
+      nextPort,
+      nextChannel,
+      Height{},
+      currentTime() + DefaultHopTimeoutPeriod,
+    )
+    // store packet for future sending ack
+    store.set(packetForwardPath(nextPort, nextChannel, nextPacketSequence), packet)
```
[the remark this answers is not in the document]
sangier: should we rename the store as a ProvableStore?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+// to ensure atomic packet forwarding
+function revertInFlightChanges(sentPacket: Packet, receivedPacket: Packet) {
+  // the token on our chain is the token in the sentPacket
+  for token in sentPacket.tokens {
+    // check if the packet we sent out was sending as source or not
+    // in this case we escrowed the outgoing tokens
+    if isSource(sentPacket.sourcePort, sentPacket.sourceChannel, token) {
+      // check if the packet we received was a source token for our chain
+      if isSource(receivedPacket.destinationPort, receivedPacket.desinationChannel, token) {
+        // receive sent tokens from the received escrow to the forward escrow account
```
sangier, line 552:
```suggestion
        // Previous Actions: 
                //1. Receive tokens from the receivedEscrowAccount 
                //2. Send token to forwardEscrowAccount
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+        // so we must send the tokens back from the forward escrow to the original received escrow account
```
sangier, line 553:
```suggestion
        // Reverse Actions: 
        //1. Send tokens back from forwardEscrowAccount to the original receivedEscrowAccount
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+        forwardEscrow = channelEscrowAddresses[sentPacket.sourceChannel]
+        reverseEscrow = channelEscrowAddresses[packet.destChannel]
+        bank.TransferCoins(forwardEscrow, reverseEscrow, token.denom, token.amount)
+      } else {
+        // receive minted vouchers and sent to the forward escrow account
+        // so we must remove the vouchers from the forward escrow account and burn them
+        bank.BurnCoins(forwardEscrow, token.denom, token.amount)
+      }
+    } else {
+      // in this case we burned the vouchers of the outgoing packets
```
sangier, line 563:
```suggestion
      // Previous Actions: 
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+    if isSource(sentPacket.sourcePort, sentPacket.sourceChannel, token) {
+      // check if the packet we received was a source token for our chain
+      if isSource(receivedPacket.destinationPort, receivedPacket.desinationChannel, token) {
+        // receive sent tokens from the received escrow to the forward escrow account
+        // so we must send the tokens back from the forward escrow to the original received escrow account
+        forwardEscrow = channelEscrowAddresses[sentPacket.sourceChannel]
+        reverseEscrow = channelEscrowAddresses[packet.destChannel]
+        bank.TransferCoins(forwardEscrow, reverseEscrow, token.denom, token.amount)
+      } else {
+        // receive minted vouchers and sent to the forward escrow account
```
sangier, line 558:
```suggestion
        // Previous Actions: 
                //1. Received minted vouchers 
                //2. Send Vouchers to forwardEscrow Account. 
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+        // so we must remove the vouchers from the forward escrow account and burn them
```
sangier, line 559:
```suggestion
        // Reverse Actions: 
        //1. Burn Vouchers from the forwardEscrowAccount
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+        bank.BurnCoins(forwardEscrow, token.denom, token.amount)
+      }
+    } else {
+      // in this case we burned the vouchers of the outgoing packets
+      // check if the packet we received was a source token for our chain
+      // in this case, the tokens were unescrowed from the reverse escrow account
+      if isSource(receivedPacket.destinationPort, receivedPacket.desinationChannel, token) {
+        // in this case we must mint the burned vouchers and send them back to the escrow account
```
sangier, line 567:
```suggestion
        // // Previous Actions: 
                //1. Burn outgoing packet Vouchers
                //2. Unescrow from reverseEscrowAccount
                //Reverse Actions: 
                //1. Mint Vouchers to reverseEscrowAccount
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+        bank.MintCoins(reverseEscrow, token.denom, token.amount)
+      }
+      // if it wasn't a source token on receive, then we simply had minted vouchers and burned them in the receive.
```
sangier, line 570:
```suggestion
      // if it wasn't a source token on receive: 
      //Previous Actions: 
            //1. Mint Vouchers 
            //2. Burn Vouchers in the receive
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +535,45 @@ function refundTokens(packet: Packet) {
...
+      // So no state changes were made, and thus no reversion is necessary
```
sangier, line 571:
```suggestion
      //Reverse Actions: 
      // None. No state changes occurred and thus no reversion is necessary
```

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -348,17 +368,55 @@ function onRecvPacket(packet: Packet) {
...
+  // then start forwarding
+  if forwardingPath != nil {
+    nextPort, nextChannel = split(forwardingPath[0], "/")
+    // send the tokens we received above to the next port and channel
+    // on the forwarding path
+    // and reduce the forwardingPath by the first element
+    nextSequence = sendFungibleTokens(
+      tokens: receivedTokens,
+      sender: receiver,
+      receiver: data.receiver,
```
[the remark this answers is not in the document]
sangier: Here receiver is a global variable which is either set at line 324 or 331. Additionally we verify receiver is not empty at line 340. I agree we should use directly receiver.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -373,6 +431,30 @@ function onAcknowledgePacket(
   if !(acknowledgement.success) {
     refundTokens(packet)
   }
+
+  // check if the packet was sent is from a previously forwarded packet
+  prevPacket = store.get(packetForwardPath(packet.sourcePort, packet.sourceChannel))
```
sangier, line 436: Same here, store should be provableStore?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -311,26 +321,36 @@ function onRecvPacket(packet: Packet) {
...
+      receiver = channelForwardingAddresses[packet.destinationChannel]
+    }
   } else {
     // should never be reached as transfer version must be negotiated to be either
     // ics20-1 or ics20-2 during channel handshake
     abortTransactionUnless(false)
   }
 
+  assert(packet.sender !== "")
+  assert(receiver !== "")
```
sangier, line 340: Note that this assertion may fail if we are in the case where `transferVersion == ics-20-2 && forwardingPath is nil` because we are not setting the `receiver` variable. If I got it correctly this would be the case where we are the destChain and we don't need to keep forwarding. Thus we should handle this scenario setting a proper receiver to allow lines 362 or 374 to execute properly.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -311,26 +321,36 @@ function onRecvPacket(packet: Packet) {
        amount: packet.amount
      }
      tokens = []Token{token}
+     receiver = packet.receiver
   } else if transferVersion == "ics20-2" {
     FungibleTokenPacketDataV2 data = UnmarshalJSON(packet.data)
     tokens = data.tokens
+    // if we need to forward the tokens onward
+    // temporarily send to the channel escrow address of the intended receiver
+    if len(forwardingPath) > 0 {
```
sangier, line 330: For consistency we should use `len(forwardingPath)>0` or `forwardingPath != nil` in both lines 330 and 398.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -92,6 +93,7 @@ The fungible token transfer bridge module tracks escrow addresses associated wit
 ```typescript
 interface ModuleState {
   channelEscrowAddresses: Map<Identifier, string>
+  channelForwardingAddresses: Map<Identifier, string>
```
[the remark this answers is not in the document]
sangier: I agree, we could use directly channelEscrowAddresses.  The channelEscrowAddresses mapping is populated during `onChainOpenTry` at line 181. Instead the channelForwardingAddresses is never set. Thus the read we do at line 331 may be illegal.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -311,26 +321,36 @@ function onRecvPacket(packet: Packet) {
...
      }
      tokens = []Token{token}
+     receiver = packet.receiver
   } else if transferVersion == "ics20-2" {
     FungibleTokenPacketDataV2 data = UnmarshalJSON(packet.data)
     tokens = data.tokens
+    // if we need to forward the tokens onward
+    // temporarily send to the channel escrow address of the intended receiver
+    if len(forwardingPath) > 0 {
+      receiver = channelForwardingAddresses[packet.destinationChannel]
```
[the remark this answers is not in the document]
sangier: Could be ok. If we remove the `channelForwardingAddresses` and use `channelEscrowAddresses` we could do directly: 

`receiver = channelEscrowAddresses[packet.destinationChannel]`

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -311,26 +321,36 @@ function onRecvPacket(packet: Packet) {
...
+    }
```
sangier, line 332:
```suggestion
    } else{
              receiver=data.receiver
              }
```

spec/app/ics-020-fungible-token-transfer/README.md
sangier, line 568:
```suggestion
        reverseEscrow = channelEscrowAddresses[receivedPacket.destChannel]
        bank.MintCoins(reverseEscrow, token.denom, token.amount)
```

spec/app/ics-020-fungible-token-transfer/README.md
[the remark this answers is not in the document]
crodriguezvega: I think it should be `privateStore` instead, since we don't need to prove anything against this path, right? The packet that is stored in this key is just auxiliary data to correctly execute refund logic, right?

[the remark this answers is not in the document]
sangier: Yes, you're right! The prevPacket is used to call the revertInFlightChanges which works locally in the calling chain.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -348,17 +368,55 @@ function onRecvPacket(packet: Packet) {
...
   }
+
+  // if acknowledgement is successful and forwarding path set
+  // then start forwarding
+  if forwardingPath != nil {
+    nextPort, nextChannel = split(forwardingPath[0], "/")
+    // send the tokens we received above to the next port and channel
+    // on the forwarding path
+    // and reduce the forwardingPath by the first element
+    nextSequence = sendFungibleTokens(
```
sangier, line 403:
```suggestion
    nextPacketSequence = sendFungibleTokens(
```

sangier (changes requested):
Hello guys and thank you for the great work! 

I have done a first review pass and left some comments and suggestions. 
Overall, I found a bit complicated reasoning by hand on certain specific scenarios, thus I agree having quint spec would be much helpful to ensure certain properties are respected. 

To analyse the function flows I derived with the help of GPT4 and this [app.code2flow](https://app.code2flow.com/manage/flowcharts) website some flowcharts diagrams that I repost here for reference in case they can be useful for others. 

SendFungibleTokens

onRecvPacket

OnAckPacket

OnTimeout

revertInFlightChanges

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -382,6 +464,31 @@ function onAcknowledgePacket(
 function onTimeoutPacket(packet: Packet) {
   // the packet timed-out, so refund the tokens
   refundTokens(packet)
```
crodriguezvega, line 466: I think we need some `if` statement is around `refundTokens`, otherwise if `prevPacket` is not `nil`, aren't we executing refund logic twice?

Alternatively, maybe there is a way in which we can execute `refundTokens` always and if `prevPacket` is not nil, then we execute additional logic in `revertInFlightChanges` (but this logic could be simplified from its current form).

spec/app/ics-020-fungible-token-transfer/README.md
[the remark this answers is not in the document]
crodriguezvega: I still think that this should be:

```
nextPort, nextChannel = split(forwardingPath[0], "/")
channelEscrowAddresses[nextChannel]
```

because `packet.destinationChannel` is the same as the source channel where tokens where sent from, and therefore `channelEscrowAddresses[packet.destinationChannel]` would be equal to the escrow address where the tokens were locked when they left the chain. 

I hope this makes sense!

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega, line 403: in this call to `sendFungibleTokens`, when the chain is source of the tokens, we do a bank transfer where sender and receiver are the same address (i.e. the escrow account on the source channel). Just mentioning, in case there's a way to avoid the extra bank transfer.

crodriguezvega (commented):
These diagrams helped me personally to reason through the flow of an error acknowledgement with a middle hop for the 4 scenarios that are handled in `revertInFlightChanges`.

path-forwarding-3
([excalidraw link](https://excalidraw.com/#json=yacTZW8SiqJV8iW-H96sK,E50IeXMVkVRM5p9n56-4jw))

When doing this exercise, I noticed a couple of things that I left as comments.

spec/app/ics-020-fungible-token-transfer/README.md
AdityaSripal: i don't believe so. 

So as an example, during a receive in an intermediate chain:

We might mint tokens to `receiverAddrA` and then sending from `receiverAddrA` to the forward escrow account.

The refundTokens will simply send from the forwardEscrowAccount to `receiverAddrA` whereas revertInflightChanges will burn the tokens in `receiverAddrA`'s account.

So the intermediate receive involves a **receive** and then a **send**.

refundTokens reverts the `send`

and `revertInflightPacket` reverts the receive.

Hopefully that's useful context to verify the logic of `revertInflightChanges`

spec/app/ics-020-fungible-token-transfer/README.md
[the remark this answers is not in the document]
AdityaSripal: Since we are using different addresses from the escrow this should be fine so long as we stay consistent. Do you see a bug in the logic, or just a preference?

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega: I think there are two cases on which executing logic on both `refundTokens` and `revertInflightPacket` will leave the middle chain in an undesirable state in case of an error acknowledgement.

### When middle hop is not source when sending and source when receiving

In `onRecvPacket` vouchers are transferred from escrow account to forwarding account, then in `sendFungibleTokens` vouchers in forwarding account are burnt. In `onAcknowledgePacket` in `refundTokens` vouchers are minted to forwarding account and in `revertInflightPacket` vouchers are minted on escrow account. End state: vouchers in forwarding account.

## When middle hop is not source when sending and not source when receiving

In `onRecvPacket` vouchers are minted to forwarding account, then in `sendFungibleTokens` vouchers in forwarding account are burnt. In `onAcknowledgePacket` in `refundTokens` vouchers are minted to forwarding account and in `revertInflightPacket` nothing is done. End state: vouchers in forwarding account.

I would appreciate if you can take a look at [the diagrams above](https://github.com/cosmos/ibc/pull/1090#pullrequestreview-1981084855), the bottom two diagrams show these two situations, in case I am missing something...

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -428,6 +577,45 @@ function refundTokens(packet: Packet) {
...
 ```
 
+```typescript
+// revertInFlightChanges reverts the receive packet and send packet
+// that occurs in the middle chains during a packet forwarding
+// If an error occurs further down the line, the state changes
+// on this chain must be reverted before sending back the error acknowledgement
+// to ensure atomic packet forwarding
+function revertInFlightChanges(sentPacket: Packet, receivedPacket: Packet) {
+  forwardEscrow = channelEscrowAddresses[sentPacket.sourceChannel]
```
crodriguezvega, line 587: Shouldn't this be `forwardEscrow = channelForwardingAddresses[receivedPacket.destChannel]` instead?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -51,12 +51,13 @@ interface FungibleTokenPacketDataV2 {
   sender: string
   receiver: string
   memo: string
+  forwardingPath: []string // a list of portID/channelID pairs determining where the tokens must be forwarded next
 }
 
 interface Token {
   denom: string // base denomination
   trace: []string
```
charleenfei, line 59: it would be helpful if maybe we clarified the naming here, between `trace` and `path` -- below we understand that traces are only the port and channel pairs, but in ie: the `DenomTrace` struct we understand that concept of trace as containing the path AND base denom.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -373,6 +473,30 @@ function onAcknowledgePacket(
...
+      // the forwarded packet has failed, thus the funds have been refunded to the forwarding address.
+      // we must revert the changes that came from successfully receiving the tokens on our chain
+      // before propogating the error acknowledgement back to original sender chain
+      revertInFlightChanges(packet, prevPacket)
+      // write error acknowledgement
+      FungibleTokenPacketAcknowledgement ack = FungibleTokenPacketAcknowledgement{false, "forwarded packet failed"}
+      handler.writeAcknowledgement(
+        prevPacket,
+        ack,
+      )
```
sangier, line 497:
```suggestion
      )}
```
We are missing a closing bracket.

spec/app/ics-020-fungible-token-transfer/README.md
charleenfei: there's also quite a lot of overlap between DenomTrace and Token, the only difference seems to be that the Token trace is structured a bit differently than the trace on DenomTrace (port/channel pairs v one long string), and that there's an amount field. would we maybe want to embed the DenomTrace struct into Token and then have those methods be available there?

sangier: Thanks @charleenfei for raising this concern. The approach you propose is pretty similar to the one that our colleagues at informal systems have used to model the ics20v1 in quint. I personally like it.

spec/app/ics-020-fungible-token-transfer/README.md
sangier: Hey guys, in this [PR](https://github.com/cosmos/quint/pull/3) we have the test of all the 4 scenarios depicted above regarding error acknowledgements. Assuming the translation from 'spec' to 'quint' is correct, the specifications have been proven accurate during the execution of error acknowledgments with no undesirable ending state detected. Reviews would be much appreciated. 

For reference I post here  
- [new testing scenario diagrams](https://excalidraw.com/#json=imWPrnjVUYGRX0ruIxjii,bWrrA-vuQx43Sd7Z-pSB2w), adapted to existing nomenclature present in the `ics20.qnt` specification.   
- [the state transition tables for every scenario](https://docs.google.com/spreadsheets/d/1XCI6aNBfqDYhdOpzg0tFxaXwWCPRLaRXgiuPzvePRjk/edit#gid=2106405052)

Potential next steps to ensure the correctness of this spec would be: 
- Adding timeout testing scenarios (even if, since reversion logic is correct, everything should be ok) 
- Add multidenom logic and tests.   

Again thanks @crodriguezvega and @AdityaSripal for the great work!

crodriguezvega (approved):
Did another review, and left a couple of small comments, but this looks good to me now. Thank you @AdityaSripal for the specification and @sangier for reviewing and all the work on the quint spec.

sangier (approved):
I have updated the [quint spec](https://github.com/cosmos/quint/pull/3) with multidenoms logic + tests and a couple of invariants. Note that timeout tests are skipped in quint as they would trigger the same refund/revert logic as the error acknowledgements tests. 

After testing spec looks good to me!

Thanks @AdityaSripal and @crodriguezvega for all the work!

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -106,8 +110,8 @@ FungibleTokenPacketDataV2 {
   ],
   sender: cosmosexampleaddr1,
   receiver: cosmosexampleaddr2,
-  memo: "exampleMemo",
-  forwardingPath: [{"transfer", "channel-7", ""}, {"transfer", "channel-13", "swap: {...}"}],
+  memo: "",
+  forwardingPath: {[{"transfer", "channel-7"}, {"transfer", "channel-13"}], , "swap: {...}"}, // provide hops in order and the memo intended for final hop
```
sangier, line 114: Is the double `,` close to swap in line 114 intentional?  Should it be `forwardingPath: {[{"transfer", "channel-7"}, {"transfer", "channel-13"}], "swap: {...}"}, // provide hops in order and the memo intended for final hop` ?

sangier (commented):
Hey guys, I have done another review pass for the latest changes and left some comments.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -370,7 +376,9 @@ function onRecvPacket(packet: Packet) {
     tokens = data.tokens
     // if we need to forward the tokens onward
     // overwrite the receiver to temporarily send to the channel escrow address of the intended receiver
-    if len(forwardingPath) > 0 {
+    if len(forwardingPath.hops) > 0 {
+      // memo must be empty
+      abortTransactionUnless(memo == "")
```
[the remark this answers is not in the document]
crodriguezvega: Yes, also `forwardingPath`.

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -285,12 +289,14 @@ function sendFungibleTokens(
...
   memo: string,
-  forwardingPath: []ForwardingInfo,
+  forwardingPath: ForwardingInfo,
   sourcePort: string,
   sourceChannel: string,
   timeoutHeight: Height,
   timeoutTimestamp: uint64, // in unix nanoseconds
 ): uint64 {
+  // memo and forwardingPath cannot both be non-empty
+  abortTransactionUnless(memo != "" && forwardingPath != nil)
```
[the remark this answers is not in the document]
crodriguezvega: Actually, couldn't we get rid of one of of the two memos? If the memo supposed to trigger an action on the final destination of the tokens, why cannot we have something like this?

```
interface FungibleTokenPacketDataV2 {
  tokens: []Token
   sender: string
   receiver: string
   memo: string
   forwardingPath: ForwardingInfo
}

interface ForwardingInfo {
   hops: []Hop
}
```

But then it would not be possible to trigger any action on intermediary hops. Is that desired behaviour?

crodriguezvega (commented):
I think I don't understand yet the reason for having two `memo` fields, one in `FungibleTokenPacketDataV2` and one in `ForwardingInfo`, if the `memo` is only going to trigger an action on the receiving chain...

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -348,17 +425,71 @@ function onRecvPacket(packet: Packet) {
...
+      finalReceiver, // receiver of next packet
+      memo,
+      nextForwardingPath,
+      forwardingPath.hops[0].portID,
+      forwardingPath.hops[0].channelID,
+      Height{},
+      currentTime() + DefaultHopTimeoutPeriod,
+    )
+    // store packet for future sending ack
+    privateStore.set(packetForwardPath(forwardingPath.hops[0].portID, forwardingPath.hops[0].channelID, nextPacketSequence), packet)
```
sangier, line 488: Note that we have an incompatibility in the number of parameters used with the `packetForwardPath` in the : 
- `OnRecvPacket`  :  we use `packetForwardPath(forwardingPath.hops[0].portID, forwardingPath.hops[0].channelID, nextPacketSequence)`. Here we include nextPacketSequence as a key for the path. 

- `onAcknowledgePacket`&& `onTimeoutPacket` :  we use `packetForwardPath(packet.sourcePort, packet.sourceChannel)` without using the nextPacketSequence as a discriminator for the key. 

The result is that the previous packet in `onAcknowledgePacket`&& `onTimeoutPacket` is always nil. 

Since, unless we store it somewhere, should be not possible to retrieve the nextPacketSequence in `onAcknowledgePacket` && `onTimeoutPacket`, a couple of possible solutions could be: 
1. remove nextPacketSequence in the key for packetForwardPath, 
2. remove nextPacketSequence in the key for packetForwardPath, but save it together with packet in store. 

Do you see a possible better approach?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -311,26 +370,44 @@ function onRecvPacket(packet: Packet) {
        amount: packet.amount
```
sangier, line 370: should this be data.amount instead?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -311,26 +370,44 @@ function onRecvPacket(packet: Packet) {
...
+    } else {
+      receiver = data.receiver
+    }
   } else {
     // should never be reached as transfer version must be negotiated to be either
     // ics20-1 or ics20-2 during channel handshake
     abortTransactionUnless(false)
   }
 
+  assert(packet.sender !== "")
```
sangier, line 396: Should this be data.sender?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -380,8 +534,33 @@ function onAcknowledgePacket(
...
+  // check if the packet was sent is from a previously forwarded packet
+  prevPacket = privateStore.get(packetForwardPath(packet.sourcePort, packet.sourceChannel))
+
+  if prevPacket != nil {
+    // the forwarded packet has failed, thus the funds have been refunded to the forwarding address.
+    // we must revert the changes that came from successfully receiving the tokens on our chain
+    // before propogating the error acknowledgement back to original sender chain
+    revertInFlightChanges(packet, prevPacket)
+    // write error acknowledgement
+    FungibleTokenPacketAcknowledgement ack = FungibleTokenPacketAcknowledgement{false, "forwarded packet failed"}
```
crodriguezvega, line 546: Probably this is fine, but just wanted to comment that in case of time out, the middle chains will propagate an error ack back to the sending chain. Would there be any implications for applications or middleware on the sending chain if the packet is error acknowledged instead of timed out?

crodriguezvega: Maybe change the message to make it more specific?

```suggestion
    FungibleTokenPacketAcknowledgement ack = FungibleTokenPacketAcknowledgement{false, "forwarded packet timed out"}
```

Vedat7711:
Cosmoss olanak adress Givvv *eth:0xFB4eBFD7B72934759E6414a32F882E5719f9F239*

eth:0xFB4eBFD7B72934759E6414a32F882E5719f9F239


Vedat afsar

14 May 2024 Sal 14:21 tarihinde Carlos Rodriguez ***@***.***>
şunu yazdı:

> ***@***.**** commented on this pull request.
>
> I think I don't understand yet the reason for having two memo fields, one
> in FungibleTokenPacketDataV2 and one in ForwardingInfo, if the memo is
> only going to trigger an action on the receiving chain...
> ------------------------------
>
> In spec/app/ics-020-fungible-token-transfer/README.md
> <https://github.com/cosmos/ibc/pull/1090#discussion_r1599814601>:
>
> >    sourcePort: string,
>    sourceChannel: string,
>    timeoutHeight: Height,
>    timeoutTimestamp: uint64, // in unix nanoseconds
>  ): uint64 {
> +  // memo and forwardingPath cannot both be non-empty
> +  abortTransactionUnless(memo != "" && forwardingPath != nil)
>
> Yes, I think you're right.
> ------------------------------
>
> In spec/app/ics-020-fungible-token-transfer/README.md
> <https://github.com/cosmos/ibc/pull/1090#discussion_r1599847206>:
>
> >    sourcePort: string,
>    sourceChannel: string,
>    timeoutHeight: Height,
>    timeoutTimestamp: uint64, // in unix nanoseconds
>  ): uint64 {
> +  // memo and forwardingPath cannot both be non-empty
> +  abortTransactionUnless(memo != "" && forwardingPath != nil)
>
> Actually, couldn't we get rid of one of of the two memos? If the memo
> supposed to trigger an action on the final destination of the tokens, why
> cannot we have something like this?
>
> interface FungibleTokenPacketDataV2 {
>   tokens: []Token
>    sender: string
>    receiver: string
>    memo: string
>    forwardingPath: ForwardingInfo
> }
>
> interface ForwardingInfo {
>    hops: []Hop
> }
>
> But then it would not be possible to trigger any action on intermediary
> hops. Is that desired behaviour?
> ------------------------------
>
> In spec/app/ics-020-fungible-token-transfer/README.md
> <https://github.com/cosmos/ibc/pull/1090#discussion_r1599819836>:
>
> > @@ -370,7 +376,9 @@ function onRecvPacket(packet: Packet) {
>      tokens = data.tokens
>      // if we need to forward the tokens onward
>      // overwrite the receiver to temporarily send to the channel escrow address of the intended receiver
> -    if len(forwardingPath) > 0 {
> +    if len(forwardingPath.hops) > 0 {
> +      // memo must be empty
> +      abortTransactionUnless(memo == "")
>
> Yes, also forwardingPath.
>

—*eth:0xFB4eBFD7B72934759E6414a32F882E5719f9F239*

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -369,9 +526,32 @@ function onRecvPacket(packet: Packet) {
...
   packet: Packet,
   acknowledgement: bytes) {
-  // if the transfer failed, refund the tokens
-  if !(acknowledgement.success) {
-    refundTokens(packet)
+  // check if the packet that was sent is from a previously forwarded packet
+  prevPacket = privateStore.get(packetForwardPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
+
+  if prevPacket != nil {
+    if acknowledgement.success {
```
gjermundgaraba, line 533: It' not really specified here or else where, as far as I can see, but is there any point in time where it is safe for the privateStore to remove these packets?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -348,17 +451,71 @@ function onRecvPacket(packet: Packet) {
...
+  if !ack.Success() {
+    return ack
+  }
+
+  // if acknowledgement is successful and forwarding path set
+  // then start forwarding
+  if len(forwardingPath.hops) > 0 {
+    //check that next channel supports token forwarding
+    channel = provableStore.get(channelPath(forwardingPath.hops[0].portID, forwardingPath.hops[0].channelID))
+    if channel.version != "ics20-2" && len(forwardingPath.hops) > 1 {
```
gjermundgaraba, line 484: Does all the previous minting / transfer need to be reverted here, or is it implicit?

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -380,8 +560,33 @@ function onAcknowledgePacket(
 
 ```typescript
 function onTimeoutPacket(packet: Packet) {
-  // the packet timed-out, so refund the tokens
-  refundTokens(packet)
+  // check if the packet sent is from a previously forwarded packet
+  prevPacket = privateStore.get(packetForwardPath(packet.sourcePort, packet.sourceChannel, packet.sequence))
```
gjermundgaraba, line 564: I have an observation on the naming here from reading this quite fresh: 
It took me quite a bit of time and mental gymnastic to get the packets straight in my head here (to be fair, this could just be because I am not so used to this).
In particular, I found it tricky to map all these names to the right context:
- Forwarded packet
- Previous packet
- Sent packet
- Received packet
- Packet (in the context of onTimeout, onAck, OnRecv)

At first I thought it was an error that the timeout packet was sent in to `revertInFlightChanges` as sentPacket, because "surely the forwarded packet, aka prevPacket was the one we sent". But after rereading, I realized that forwardedPacket was not actually received packet. So it is correct, of course, but with all these names it was not so easy to make a mental model of the different packets.

Not sure what the best way to deal with this is, but perhaps there is a way we could tweak the names to make it a bit easier to grasp on first read.

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega: Yes, it's kinda implicit: if an error ack is returned then the datagram handler should not commit any state changes. But I agree, that maybe it would be good to mention this explicitly in the specs (maybe in ICS26? I cannot remember from the top of my head if we already do that...

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega: This is good feedback; thanks, @gjermundgaraba. Maybe a diagram would also help; I can try to draw something in the next couple of days.

spec/app/ics-020-fungible-token-transfer/README.md
crodriguezvega: Good point! Maybe it's possible to delete the `prevPacket` after `OnAcknowledgePacket` or `OnTimeoutPacket` have executed. What do you think, @AdityaSripal?

spec/app/ics-020-fungible-token-transfer/README.md
AdityaSripal: Thank you for the feedback @gjermundgaraba . Will mull over better naming, but if you have some suggestions feel free to share!

spec/app/ics-020-fungible-token-transfer/README.md
AdityaSripal: Yes it is, added it in now

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -51,12 +51,23 @@ interface FungibleTokenPacketDataV2 {
   sender: string
   receiver: string
   memo: string
+  forwardingPath: ForwardingInfo // a struct containing the list of next hops, determining where the tokens must be forwarded next, and the memo for the final hop
+}
+
+interface ForwardingInfo {
+  hops: []Hop
+  memo: string,
+}
```
damiannolan, line 60: Personally, I feel like the memo in `FungibleTokenPacketData` should cover this, and that we should not include a memo field in this data structure.

damiannolan: There would also be overlap between `ForwardingInfo` / `Hop` and the `Trace` type which includes the port id and channel id tuple.

damiannolan: > Personally, I feel like the memo in FungibleTokenPacketData should cover this, and that we should not include a memo field in this data structure.

I understand the concern here is preventing unintentional middleware invocation from source chain or intermediary recving chain, but also thought about what if you can rather have a memo attached to each individual hop. 

Spoke about this with @colin-axner and fine to keep this structure as introducing this on a per-hop basis is opening up a big rabbit hole of potential issues.

spec/app/ics-020-fungible-token-transfer/README.md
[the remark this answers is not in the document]
damiannolan: note https://github.com/cosmos/cosmos-sdk/issues/7737#issuecomment-735671951

chatton (commented):
This is looking great! Had just a few small questions/comments

spec/app/ics-020-fungible-token-transfer/README.md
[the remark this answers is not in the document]
colin-axner: The forwarding address is mainly just used as pass through storage based on how ics20 is constructed in the non-forwarding case. I think the spec just needs to specify that this address should be an address the transfer module controls. I don't think it should reuse the escrow addresses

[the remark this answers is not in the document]
colin-axner: Unless I am missing something, I can't think of a reason why you need channel isolated forwarding addresses https://github.com/cosmos/ibc-go/issues/6561#issuecomment-2176403684

spec/app/ics-020-fungible-token-transfer/README.md
```diff
@@ -348,17 +451,71 @@ function onRecvPacket(packet: Packet) {
...
+      finalReceiver, // receiver of next packet
+      memo,
+      nextForwardingPath,
+      forwardingPath.hops[0].portID,
+      forwardingPath.hops[0].channelID,
+      Height{},
+      currentTime() + DefaultHopTimeoutPeriod,
+    )
+    // store packet for future sending ack
+    privateStore.set(packetForwardPath(forwardingPath.hops[0].portID, forwardingPath.hops[0].channelID, packetSequence), packet)
```
colin-axner, line 514: only store packetID, storing the other packet fields in unnecessary

colin-axner (approved):
Leaving my general approval. Spec needs updating based on latest implementation developments. Happy to see those applied in whatever fashion necessary

spec/app/ics-020-fungible-token-transfer/README.md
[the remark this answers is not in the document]
AdityaSripal: The reason I decided to split it off to a separate address is just an additional security blanket. The balance of the escrow addresses is security critical. If the balance ends up in an invalid state, that can be a very painful issue to unwind.

It should be completely fine if there's no bugs. But if there's a bug in forwarding, i figure having a separate account might make the impact a bit more isolated.

Of course it is receiving from and sending back to the escrow addresses, so if there's issues there it could still be a problem for escrow account balances.

I'm ok either way. I just figured to start out with a separate account for clarity and avoiding having the escrow account representing more than one thing semantically.

We can use the same account as well as @colin-axner suggests

spec/app/ics-020-fungible-token-transfer/README.md
AdityaSripal: Need to change the function signature of writeAcknowledgement which I didn't want to do in this PR. But agree we should probs do for release

AdityaSripal: But the acknowledgement handler needs the full packet. The relayers currently grab the packet from the writeAcknowledgement event emissions. So we would need to think a bit more about how to handle relaying in the async ack case.

PR #1093, opened 2024-04-16
IBC Eureka

AdityaSripal:
The following is a spec that intends to produce an alternative implementation of the IBC spec that is much easier to get to market while still maintaing the light-client backed security that IBC promises and maintains compatibility with the rich ecosystem of chains, light clients, and applications that already exist in the IBC ecosystem.

It works by taking a standard IBC light client and connecting it with as minimal complexity as possible to a standard IBC application.

Thus the connection and channel handshakes are completely stripped out (though the counterparty client verification done in a 2-step handshake that replaces the 8-step connection/channel handshake)

The client is connected to a a multi-application channel that can connect to any IBC application. Thus, a single light client and a single channel would enable all IBC app data to flow between two chains.

This also automatically gives us atomic IBC packets while still requiring very little changes to the IBC application.

Unsupported atm:

- ORDERED channel (the mulit-IBC channel is UNORDERED, probably will remain)
- Async acknowledgements (seems possible to fix, just more complex since we must maintain state in receiver until final ack is ready)
- Compatibility with traditional channels with a separate port and version. This seems much harder but still possible and arguably the most desirable of current unsupported features


AdityaSripal:
Note TODO, I need one more handshake message here, and I need to ensure both sides have the counterparty channel store Identifier and has verified this field on the other end as well

Vedat7711:
> TODO'ya dikkat edin, burada bir el sıkışma mesajına daha ihtiyacım var ve her iki tarafın da karşı taraf kanal deposu Tanımlayıcısına sahip olduğundan ve bu alanı diğer tarafta da doğruladığından emin olmam gerekiyor

UlAsin bana 0533 519 62 07 WhatsApp

srdtrk:
Also remove delay period for clients

hdevalence:
Would be great to align with PICA/ICAv2: https://hackmd.io/bqkQQ9AxR0KFR82APKXocQ

srdtrk:
We should look at if it's possible to remove timeout revision number and timeout revision height from the packet commitment. And only keep timeout timestamp.

spec/micro/README.md
```diff
@@ -0,0 +1,781 @@
...
+    // store commitment to the packet data & packet timeout
+    channelStore.set(
+      packetCommitmentPath(commitPort, sourceChannel, sequence),
+      hash(hash(data), timeoutHeight, timeoutTimestamp)
+    )
+
+    // increment the sequence. Thus there are monotonically increasing sequences for packet flow
+    // from sourcePort, sourceChannel pair
+    channelStore.set(nextSequenceSendPath(commitPort, sourceChannel), sequence+1)
+
```
srdtrk, line 193: This should be making an application callback I think. This is because:
1. Even if we want the packet lifecycle to begin in the IBC app. There is nothing wrong with the application receiving a callback. In fact, this is necessary to receive a reply in the Actor model of CosmWasm.
2. Might make multipacket lifecycle easier to implement

spec/micro/README.md
```diff
@@ -0,0 +1,770 @@
...
+- Applications must be mutually compatible with standard IBC applications.
+- Must be capable of being implemented in smart contract environments with resource constraints and high gas costs.
+
+### Specification
+
+### Light Clients
+
+The light client module can be implemented exactly as-is with regards to its functionality. It **must** have external endpoints for relayers (off-chain processes that have full-node access to other chains in the network) to initialize a client, update the client, and submit misbehaviour in case the trust model of the client is violated by the counterparty consensus mechanism (e.g. committing to different headers for the same height).
+
+The implementation of each of these endpoints will be specific to the particular consensus mechanism targetted. The choice of consensus algorithm itself is arbitrary, it may be a Proof-of-Stake algorithm like CometBFT, or a multisig of trusted authorities, or a rollup that relies on an additional underlying client in order to verify its consensus.
```
brapse, line 27: Probably requires single slot finality for the state proofs or some kind of finality gadget.

spec/micro/README.md
```diff
@@ -0,0 +1,638 @@
...
+): (Identifier, error)
+
+// once a client has been created, it can be referenced with the identifier and passed the header
+// to keep the client up-to-date. In most cases, this will cause a new consensus state derived from the header
+// to be stored in the client
+function updateClient(
+    clientId: Identifier,
+    header: bytes,
+): error
+
```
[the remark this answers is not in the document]
brapse: If the majority of the cases for micro-ibc would be specificially new connections, what would be the motivation for transitioning from an existing client to a micro-client if the work has already been done?

spec/micro/README.md
```diff
@@ -0,0 +1,638 @@
...
+// once a client has been created, relayers can submit misbehaviour that proves the counterparty chain
+// The light client must verify the misbehaviour using the trust model of the consensus mechanism
+// and execute some custom logic such as freezing the client from accepting future updates and proof verification.
+function submitMisbehaviour(
+    clientId: Identifier,
+    misbehaviour: bytes,
+): error
```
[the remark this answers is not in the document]
brapse: As in, terminate the client with the reason? I guess the question is how much misbehavior should be tolerated. If it's zero than perhaps it would make sense to change the client state vs accumulate evidence of behavior for a later action.

spec/micro/README.md
```diff
@@ -0,0 +1,638 @@
+# Micro IBC Architecture
+
+### Context
+
+The implementation of the entire IBC protocol as it currently stands is a large undertaking. While there exists ready-made implementations like ibc-go this is only deployable on the Cosmos-SDK. Similarly, there exists ibc-rs which is a library for chains to integrate. However, this requires the chain to be implemented in Rust, there still exists some non-trivial work to integrate the ibc-rs library into the target state machine, and certain limitations either in the state machine or in ibc-rs may prevent using the library for the target chain.
```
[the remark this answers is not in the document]
brapse: I would say the big advantage is that with this compressed version of the protocol the verification of the state is much more straight forward but slightly compressed into a single round trip which is i think a net win for implementors but probably leaks previous sub domains of the previous layers into a single macro domain but again is a net win.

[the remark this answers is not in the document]
sangier: What's written in the context, in part, are the consequences of the initial design choices that were made. I wonder if we shall point this out or keep it implicit.

spec/micro/README.md
```diff
@@ -0,0 +1,770 @@
...
+
+Writing an implementation from scratch is a problem many ecosystems face as a major barrier for IBC adoption.
+
+The goal of this document is to serve as a "micro-IBC" specification that will allow new ecosystems to implement a protocol that can communicate with fully implemented IBC chains using the same security assumptions.
```
sangier, line 9: I guess that another goal of this document is to revisit the initial design choices, learn from them, keep the core meat, and provide a smoother version of protocol that will be useful not only for new ecosystem, but even for cosmos itself.

spec/micro/README.md
sangier: I agree we could include some consideration around consensus finalities here.

spec/micro/README.md
```diff
@@ -0,0 +1,770 @@
...
+
+Thus, IBC lite will introduce a new message `ProvideCounterparty` that will associate the counterparty client of our chain with our client of the counterparty. Thus, if the `ProvideCounterparty` message is submitted to both sides correctly. Then both sides have mirrored <client,client> pairs that can be treated as channel identifiers. Assuming they are correct, the client on each side is unique and provides an authenticated stream of packet data between the two chains. If the `ProvideCounterparty` message submits the wrong clientID, this can lead to invalid behaviour; but this is equivalent to a relayer submitting an invalid client in place of a correct client for the desired chain. In the simplest case, we can rely on out-of-band social consensus to only send on valid <client, client> pairs that represent a connection between the desired chains of the user; just as we currently rely on out-of-band social consensus that a given clientID and channel built on top of it is the valid, canonical identifier of our desired chain.
+
+```typescript
+interface Counterparty {
+    channelId: Identifier
```
sangier, line 92: I am not entirely sure in this spec we should keep using channelIDs instead of clientIDs. Is this an implementation related detail to ensure compatibility with ibc-go? If yes, shouldn't we use clientIDs in the spec and, additionally, provide a comment stating that the implementation we will be using channelIDs?

spec/micro/README.md
```diff
@@ -0,0 +1,638 @@
...
+
+Properties of Channel:
+
+- Separates applications into dedicated 1-1 communication channels. This prevents applications from writing into each other's channels.
+- Allows applications to come to agreement on the application parameters (version negotiation). Ensures that each side can understand the other's communication and that they are running mutually compatible logic. This version negotiation is a multi-step process that allows the finalized version to differ substantially from the one initially proposed
+- Establishes the ordering of the channel
+- Establishes unique identifiers for the applications on either chain to use to reference each other when sending and receiving packets.
+- The application protocol can be continually upgraded over time by using the upgrade handshake which allows the same channel which may have accumulated state to use new mutually agreed upon application packet data format(s) and associated new logic.
+- Ensures exactly-once delivery of packet flow datagrams (Send, Receive, Acknowledge, Timeout)
+- Ensures valid packet flow (Send => Receive => Acknowledge) XOR (Send => Timeout)
```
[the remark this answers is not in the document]
sangier: Shall we add some diagrams too?

spec/micro/README.md
```diff
@@ -0,0 +1,770 @@
...
+
+All packet flow message (SendPacket, RecvPacket, and TimeoutPacket) are sent with the full packet. The packet contains both sender and receiver identifiers. Thus on packet flow messages sent to the receiver (RecvPacket), we use the receiver identifier in the packet to retrieve our local client and the source identifier to determine which path the sender stored the packet under. We can thus use our retrieved client to verify a key/value membership proof to validate that the packet was sent by the counterparty.
+
+Similarly, for packet flow messages sent to the sender (AcknowledgePacket, TimeoutPacket); the packet is provided again. This time, we use the sender identifier to retrieve the local client and the destination identifier to determine the key path that the receiver must have written to when it received the packet. We can thus use our retrieved client to verify a key/value membership proof to validate that the packet was sent by the counterparty. In the case of timeout, if the packet receipt wasn't written to the receipt path determined by the destination identifier this can be verified by our retrieved client using the key nonmembership proof.
+
+### Soundness
```
sangier, line 366: Note that, potentially, in the future we could enforce client identifiers global uniqueness. For now I agree that  the packet is committed key path is a strong guarantee.

spec/micro/README.md
```diff
@@ -0,0 +1,770 @@
...
+
+
+
+------
+
+IGNORE THE MULTICHANNEL WORK FOR NOW
```
sangier, line 386: Potentially we could address the multichannel work in a different PR 👍

faddat:
Hey I just wanted to say that I think that this is really really cool, and I'm going to give it a read and provide feedback if I come up with anything. 

Thank you!

spec/micro/README.md
```diff
@@ -0,0 +1,381 @@
...
+- Establishes the ordering of the channel
+- Establishes unique identifiers for the applications on either chain to use to reference each other when sending and receiving packets.
+- The application protocol can be continually upgraded over time by using the upgrade handshake which allows the same channel which may have accumulated state to use new mutually agreed upon application packet data format(s) and associated new logic.
+- Ensures exactly-once delivery of packet flow datagrams (Send, Receive, Acknowledge, Timeout)
+- Ensures valid packet flow (Send => Receive => Acknowledge) XOR (Send => Timeout)
+
+### Identifying Counterparties
+
+In core IBC, the connection and channel handshakes serve to validate the clients are valid clients of the counterparty, ensure the IBC version and application versions are mutually compatible, as well as providing unique identifiers for each side to refer to the counterparty.
```
alexanderbez, line 84: nit: just slightly easier read

```suggestion
In core IBC, the connection and channel handshakes serve to ensure the validity of counterparty clients, ensure the IBC and application versions are mutually compatible, as well as providing unique identifiers for each side to refer to the counterparty.
```

spec/micro/README.md
```diff
@@ -0,0 +1,381 @@
...
+
+Since we are removing handshakes in IBC lite, we must have a different way to provide the chain with knowledge of the counterparty. With a client, we can prove any key/value path on the counterparty. However, without knowing which identifier the counterparty uses when it sends messages to us; we cannot differentiate between messages sent from the counterparty to our chain vs messages sent from the counterparty with other chains. Most implementations will not be able to store the ICS-24 paths directly as a key in the global namespace; but will instead write to a reserved, prefixed keyspace so as not to conflict with other application state writes. Thus the counteparty information we must have includes both its identifier for our chain as well as the key prefix under which it will write the provable ICS-24 paths.
+
```
alexanderbez, line 87: I feel like adding a simple diagram of the above will aid significantly in understanding, especially given how critical this is to part of the design.

spec/micro/README.md
```diff
@@ -0,0 +1,381 @@
...
+  timeoutHeight: Height
+  timeoutTimestamp: uint64
+  sourcePort: Identifier // identifier of the application on sender
+  sourceChannel: Identifier // identifier of the client of destination on sender chain
+  destPort: Identifier // identifier of the application on destination
+  destChannel: Identifier // identifier of the client of sender on the destination chain
+}
+```
+
+Since the packets are addressed **directly** with the underlying light clients, there are **no** more handshakes necessary. Instead the packet sender must be capable of providing the correct <client, client> pair.
```
alexanderbez, line 140: I feel like it might be worth noting here, that there might be increased bandwidth for relayers or packet senders, with the added benefit of significantly reducing protocol complexity.

AdityaSripal:
Thank you everyone for the feedback so far. We are committed to going in this direction with the IBC protocol.

Given the scope of this change, it doesn't make sense to house everything in a single file and solicit feedback. We are going to make a new directory to house a new, vastly simplified spec that will form the basis of IBC v2 (codenamed Eureka). We're hoping to take this new approach and solve the issues that we found in maintaining the current spec.

Mainly we will ensure to keep a distinction between what is required by the protocol and what are design decisions being made by ibc-go. We will also work on versioning so that it is clearer for implementers to get a cohesive spec that they can rely on without having to reference ibc-go.

As such, we will break down this spec into multiple smaller issues and PRs and solicit feedback on them individually as they go in. Me and @sangier will be leading the efforts to get this merged.

For now, we will merge this PR into the folder it currently is with an EXPERIMENTAL label.

Once we have flushed out a more formal spec that will eventually replace the current one; we will delete this file. From that point, implementers can refer to the v2 folder in this repository to get a clear and concise understanding of how to implement the eureka protocol.

The v1 specs will be retained but eventually placed into maintainance mode.

Looking forward to collaborating with all of you on this exciting future direction!

PR #1098, opened 2024-04-27
Fix typos

subnetdev0:
In the README.md of ibc/blob/main/spec/app/ics-100-atomic-swap/README.md, 'constructs' misspelt as 'contructs', 'Packet' misspelt as 'Paket'.


crodriguezvega (approved):
Great, thanks, @subnetdev0!

PR #1104, opened 2024-05-14
upgrade checkout action to version using node20

cario-dev:
Node16 is deprecated and Actions using it will stop working 3rd of June
https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/


crodriguezvega (approved):
Thank you for the PR!

PR #1114, opened 2024-06-13
ICS04: fix a code comment in chanUpgradeTry in UPGRADES.md

siburu:
I may misunderstand something about crossing hello, but isn't this code comment wrong somehow?

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
@@ -571,16 +571,16 @@ function chanUpgradeTry(
   }
 
   // NON CROSSING HELLO CASE:
-  // if the counterparty sequence is less than the current sequence,
+  // if the counterparty sequence is less than or equal to the current sequence,
   // then either the counterparty chain is out-of-sync or the message
   // is out-of-sync and we write an error receipt with our sequence
   // so that the counterparty can abort their attempt and resync with our sequence.
   // When the next upgrade attempt is initiated, both sides will move to a fresh
   // never-before-seen sequence number
   // CROSSING HELLO CASE:
-  // if the counterparty sequence is less than or equal to the current sequence,
+  // if the counterparty sequence is less than the current sequence,
   // then either the counterparty chain is out-of-sync or the message
-  // is out-of-sync and we write an error receipt with our sequence
+  // is out-of-sync and we write an error receipt with our sequence minus one
   // so that the counterparty can update their sequence as well.
   // This will cause the outdated counterparty to upgrade the sequence
   // and abort their out-of-sync upgrade without aborting our own since


spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -571,16 +571,16 @@ function chanUpgradeTry(
   }
 
   // NON CROSSING HELLO CASE:
-  // if the counterparty sequence is less than the current sequence,
```
crodriguezvega, line 574: I think this was correct because in the non-crossing hello case the `counterpartyUpgradeSequence` should be `==` to `expectedUpgradeSequence` (which is `channel.UpgradeSequence + 1` on line 570).

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
```diff
@@ -571,16 +571,16 @@ function chanUpgradeTry(
...
+  // if the counterparty sequence is less than or equal to the current sequence,
   // then either the counterparty chain is out-of-sync or the message
   // is out-of-sync and we write an error receipt with our sequence
   // so that the counterparty can abort their attempt and resync with our sequence.
   // When the next upgrade attempt is initiated, both sides will move to a fresh
   // never-before-seen sequence number
   // CROSSING HELLO CASE:
-  // if the counterparty sequence is less than or equal to the current sequence,
+  // if the counterparty sequence is less than the current sequence,
```
crodriguezvega, line 581: I think this change is correct because `counterpartyUpgradeSequence` should be `==` to `expectedUpgradeSequence` in the crossing hello scenario.

crodriguezvega (commented):
Thanks, @siburu. I believe the comment for the non-crossing hello scenario was already correct, so should not be changed...

spec/core/ics-004-channel-and-packet-semantics/UPGRADES.md
siburu: @crodriguezvega Thank you for reviewing this!

> in the non-crossing hello case the counterpartyUpgradeSequence should be == to expectedUpgradeSequence (which is channel.UpgradeSequence + 1 on line 570)

Yes, but doesn't `the current sequence` mean not `expectedUpgradeSequence` but `channel.UpgradeSequence`?

If so, I think that:
`counterpartyUpgradeSequence < expectedUpgradeSequence`
<=> `the counterparty sequence is less than (the current sequence + 1)`
<=> `the counterparty sequence is less than or equal to the current sequence`

crodriguezvega: Sorry for taking so long to come back to this... Yes, I think you're right! Thank you for explaining and correcting me. 🙏

crodriguezvega (approved):
Thank you for catching these mistakes, @siburu!

sangier (approved):
Thank you @siburu for catching this! LGTM!

PR #1123, opened 2024-07-05
ICS20-multiple-versions-folder-structure

sangier:
Change the folder structure to maintain multiple ics20 spec version in separated folders.


sangier:
Git history not preserved in v1/README.md

Issue #1121, opened 2024-07-01
ICS003: Remove client and consensus state validation during the connection handshake


colin-axner:
# Summary

Remove the client and consensus state validation which occurs during the connection handshake. 

## Problem

During the connection handshake, IBC was designed to ensure the correctness of a connection/channel handshake such that when a channel is opened, we know that sending a packet can only be received and processed between two chains which are only connected to each other. 

While on paper, this design approach makes a lot of sense, the abstraction of this logic has lead to difficulties on the implementation front.

Light clients represent other chains. A chain will have many light clients, one for each chain it would like to connect to. This abstraction is intuitive. A chain is only responsible for making sure its light clients are properly setup and functioning. 

The connection handshake went a step further by ensuring not just that a chain's own light client is properly setup, but that the chain it wishes to connect to, has a correctly setup light client for itself. It does this by validating the client state and consensus state. The important information for a tendermint client, being the chain id and validator set hash. 

Note that since we are not proving execution, we are not protecting against a malicious chain. We are only checking correctness of the relayers. 

A few problems have arisen over time with regards to this construction:
* not all chains have consensus state information available https://github.com/cosmos/ibc/pull/839
* not all implementations of a light client share the same client/consensus state structures https://github.com/cosmos/ibc-go/issues/5315
* because client and consensus state keys are proved, additional storage was added to improve iteration https://github.com/cosmos/ibc-go/pull/125, consensus state queries are still inefficient https://github.com/cosmos/ibc-go/issues/1399
* because client and consensus state paths are proved, changing the stored value is very difficult (cannot break client/consensus state structures)
* in ibc-go, historical information is required to be set to lookup past consensus states (leads to additional store writes for infrequent access, also most recently blocked feature req https://github.com/cosmos/cosmos-sdk/issues/20733)

## Proposal

Social consensus is already required to avoid accessing a malicious chain. A user must select a channel which it believes to be correctly setup. Thus, we are not adding much defense with client and consensus state proofs. We should take advantage of the social consensus approach and remove the client and consensus state proof verification in the connection handshake. 

**Pros:**
* reduction in complexity (2 less proofs)
* client and consensus state representation is decoupled from core ibc proofs (allows for breaking changes)
* additional reverse iterator keys in 07-tendermint can be added
* client and consensus state keys can be updated to be iterator efficient
* allows for multiple implementations of a single client with different underlying client state/consensus state structures to be natively supported
* aligns with ibc lite, less work to implement a layer of ibc 

[...]

PR #1128, opened 2024-07-24
ICS-3: Removal of Client Validation from Connection Handshake

AdityaSripal:
Closes: #1121


sangier (approved):
Really great and clear written and visual explanation, thanks @AdityaSripal !!

spec/core/ics-003-connection-semantics/client-validation-removal.md
```diff
@@ -0,0 +1,29 @@
...
+
+Without this check, it is possible in very unlucky circumstances to have two chains that are validly connected to each other and also have misconfigured third-party chains that believe they are connected to a chain that is not connected back to them. For an example of this situation, see the attached [diagram](./client-validation-removal.png). In this case, the validly connected chains will have communication between them that follows IBC's correctness and integrity properties. However, the chains that are misconfigured may misinterpret messages send to other chains as messages intended for them. In this case, the only connection ends that are affected are the misconfigured connection ends that do not correlate to a valid connection end on the intended counterparty. In order for this situation to arise, there would need to not only be relayer error, but also a coincidence in identifiers and handshake message timing. Thus, this is an unlikely situation to arise in practice and the only effect is invalid message processing on misconfigured connections and channel ends.
```
colin-axner, line 13:
```suggestion
Without this check, it is possible in very unlucky circumstances to have two chains that are validly connected to each other and also have misconfigured third-party chains that believe they are connected to a chain that is not connected back to them. For an example of this situation, see the attached [diagram](./client-validation-removal.png). In this case, the validly connected chains will have communication between them that follows IBC's correctness and integrity properties. However, the chains that are misconfigured may misinterpret messages sent to other chains as messages intended for them. In this case, the only connection ends that are affected are the misconfigured connection ends that do not correlate to a valid connection end on the intended counterparty. In order for this situation to arise, there would need to not only be relayer error, but also a coincidence in identifiers and handshake message timing. Thus, this is an unlikely situation to arise in practice and the only effect is invalid message processing on misconfigured connections and channel ends.
```

"messages send" -> "messages sent"

colin-axner (approved):
Thank you!! Beautiful write up :heart_eyes:

crodriguezvega (approved):

PR #1132, opened 2024-08-15
ICS03: delete upgradability protocol

crodriguezvega:
It's out of date and probably incorrect (since it hasn't been updated with the knowledge we've got from channel upgradability) and in a post eureka world, it's unlikely we will ever implement it. Therefore, I think it's better if we just remove it.


AdityaSripal (approved):
Thanks @crodriguezvega !

PR #1142, opened 2024-08-19
Create v2 folder with placeholder document

AdityaSripal:


sangier:
@AdityaSripal please checkout the comments [here](https://github.com/cosmos/ibc/pull/1137). 

For me it's ok even the approach proposed in this PR, however we would need to copy back even the unchanged specs.

AdityaSripal:
Superceded by #1143

Issue #1138, opened 2024-08-19
Create 24-host spec for v2 with vastly limited provablestore surface area


AdityaSripal:
With the new v2 spec, the surface area for what needs to be stored in the provable store is **vastly reduced**

- Only packet level paths are necessary
- We should also explain not just the keys, but also the values that will be stored under the key
- Packet commitment, ack commitment, receipt

We should also explain a little bit about the possiblility that all ICS24 paths will be prefixed and how IBC will handle this.

PR #1144, opened 2024-08-21
24-host: Reduce provable surface area to only packet flow keys

AdityaSripal:
- Removes connection/channel keys
- Client keys are now private since they aren't proven by counterparty
- Keeps key paths and values the same for now
- Add counterparty key/value to private store
- Remove self client validation necessity

Context for Reviewers:

This PR introduces a new high level packet and acknowledgement commitment.
- The packet commitment must now include the destination identifier
- The packet commitment hashes the destination identifier, timeout, and app data all as bytes to create a fixed preimage size and then hashes the result

- The acknowledgement is just a hash of the acknowledgement

Closes: #1138

### DIFF
--- a/spec/core/v2/ics-024-host-requirements/README.md
+++ b/spec/core/v2/ics-024-host-requirements/README.md
@@ -1,0 +1,411 @@
+---
+ics: 24
+title: Host State Machine Requirements
+stage: draft
+category: IBC/TAO
+kind: interface
+requires: 23
+required-by: 2, 3, 4, 5, 18
+version compatibility: ibc-go v7.0.0
+author: Christopher Goes <cwgoes@tendermint.com>
+created: 2019-04-16
+modified: 2022-09-14
+---
+
+## Synopsis
+
+This specification defines the minimal set of interfaces which must be provided and properties which must be fulfilled by a state machine hosting an implementation of the interblockchain communication protocol.
+
+### Motivation
+
+IBC is designed to be a common standard which will be hosted by a variety of blockchains & state machines and must clearly define the requirements of the host.
+
+### Definitions
+
+### Desired Properties
+
+IBC should require as simple an interface from the underlying state machine as possible to maximise the ease of correct implementation.
+
+## Technical Specification
+
+### Module system
+
+The host state machine must support a module system, whereby self-contained, potentially mutually distrusted packages of code can safely execute on the same ledger, control how and when they allow other modules to communicate with them, and be identified and manipulated by a "master module" or execution environment.
+
+The IBC/TAO specifications define the implementations of two modules: the core "IBC handler" module and the "IBC relayer" module. IBC/APP specifications further define other modules for particular packet handling application logic. IBC requires that the "master module" or execution environment can be used to grant other modules on the host state machine access to the IBC handler module and/or the IBC routing module, but otherwise does not impose requirements on the functionality or communication abilities of any other modules which may be co-located on the state machine.
+
+### Paths, identifiers, separators
+
+An `Identifier` is a bytestring used as a key for an object stored in state, such as a connection, channel, or light client.
+
+Identifiers MUST be non-empty (of positive integer length).
+
+Identifiers MUST consist of characters in one of the following categories only:
+
+- Alphanumeric
+- `.`, `_`, `+`, `-`, `#`
+- `[`, `]`, `<`, `>`
+
+A `Path` is a bytestring used as the key for an object stored in state. Paths MUST contain only identifiers, constant strings, and the separator `"/"`.
+
+Identifiers are not intended to be valuable resources — to prevent name squatting, minimum length requirements or pseudorandom generation MAY be implemented, but particular restrictions are not imposed by this specification.
+
+The separator `"/"` is used to separate and concatenate two identifiers or an identifier and a constant bytestring. Identifiers MUST NOT contain the `"/"` character, which prevents ambiguity.
+
+Variable interpolation, denoted by curly braces, is used throughout this specification as shorthand to define path formats, e.g. `client/{clientIdentifier}/consensusState`.
+
+All identifiers, and all strings listed in this specification, must be encoded as ASCII unless otherwise specified.
+
+By default, identifiers have the following minimum and maximum lengths in characters:
+
+| Port identifier | Client identifier | Connection identifier | Channel identifier |
+| --------------- | ----------------- | --------------------- | ------------------ |
+| 2 - 128         | 9 - 64            | 10 - 64               | 8 - 64             |
+
+### Key/value Store
+
+The host state machine MUST provide a key/value store interface 
+with three functions that behave in the standard way:
+
+```typescript
+type get = (path: Path) => Value | void
+```
+
+```typescript
+type set = (path: Path, value: Value) => void
+```
+
+```typescript
+type delete = (path: Path) => void
+```
+
+`Path` is as defined above. `Value` is an arbitrary bytestring encoding of a particular data structure. Encoding details are left to separate ICSs.
+
+These functions MUST be permissioned to the IBC handler module (the implementation of which is described in separate standards) only, so only the IBC handler module can `set` or `delete` the paths that can be read by `get`. This can possibly be implemented as a sub-store (prefixed key-space) of a larger key/value store used by the entire state machine.
+
+Host state machines MUST provide two instances of this interface -
+a `provableStore` for storage read by (i.e. proven to) other chains,
+and a `privateStore` for storage local to the host, upon which `get`
+, `set`, and `delete` can be called, e.g. `provableStore.set('some/path', 'value')`.
+
+The `provableStore`:
+
+- MUST write to a key/value store whose data can be externally proved with a vector commitment as defined in [ICS 23](../ics-023-vector-commitments). 
+- MUST use canonical data structure encodings provided in these specifications as proto3 files
+
+The `privateStore`:
+
+- MAY support external proofs, but is not required to - the IBC handler will never write data to it which needs to be proved.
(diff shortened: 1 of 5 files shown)


spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -311,16 +186,6 @@ This permissioning can be implemented with unique references (object capabilitie
 
 Modules that wish to make use of particular IBC features MAY implement certain handler functions, e.g. to add additional logic to a channel handshake with an associated module on another state machine.
 
-### Datagram submission
```
crodriguezvega, line 314: Just curious: why was this section deleted?

crodriguezvega (approved):
Looks good to me!

sangier (approved):
That's a great simplification!!! If you merge latest change from feat/v2-spec you should be able to get rid of the CI failures.

spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -385,27 +250,10 @@ Key/value store functionality and consensus state type are unlikely to change du
...
-May 11, 2019 - Rename "RootOfTrust" to "ConsensusState"
-
-Jun 25, 2019 - Use "ports" instead of module names
-
-Aug 18, 2019 - Revisions to module system, definitions
-
-Jul 05, 2022 - Lower the minimal allowed length of a channel identifier to 8
-
-Jul 27, 2022 - Move `ClientState` to the `provableStore`, and add "Client state validation" section
+Aug 21, 2024 - Initial draft
```
sangier, line 255:
```suggestion
Aug 21, 2024 - [Initial draft](https://github.com/cosmos/ibc/pull/1144/files)
```

spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -120,15 +120,10 @@ Note that the client-related paths listed below reflect the Tendermint client as
...
-| privateStore   | "ports/{identifier}"                                                           | CapabilityKey     | [ICS 5](../ics-005-port-allocation) |
-| provableStore  | "channelEnds/ports/{identifier}/channels/{identifier}"                         | ChannelEnd        | [ICS 4](../ics-004-packet-semantics) |
-| provableStore  | "nextSequenceSend/ports/{identifier}/channels/{identifier}"                    | uint64            | [ICS 4](../ics-004-packet-semantics) |
-| provableStore  | "nextSequenceRecv/ports/{identifier}/channels/{identifier}"                    | uint64            | [ICS 4](../ics-004-packet-semantics) |
-| provableStore  | "nextSequenceAck/ports/{identifier}/channels/{identifier}"                     | uint64            | [ICS 4](../ics-004-packet-semantics) |
+| privateStore  | "clients/{identifier}/clientState"                                             | ClientState       | [ICS 2](../ics-002-client-semantics) |
+| privateStore  | "clients/{identifier}/consensusStates/{height}"                                | ConsensusState    | [ICS 7](../../client/ics-007-tendermint-client) |
```
sangier, line 124:
```suggestion
| privateStore  | "clients/{identifier}/consensusStates/{height}"                                | ConsensusState    | [ICS 7](../../../client/ics-007-tendermint-client) |
```

spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -120,15 +120,10 @@ Note that the client-related paths listed below reflect the Tendermint client as
...
+| privateStore | "clients/{identifier}/counterparty"                                             | Counterparty      | [ICS 2](../ics-002-client-semantics)
+| privateStore  | "nextSequenceSend/ports/{identifier}/channels/{identifier}"                    | uint64            | [ICS 4](../ics-004-channel-and-packet-semantics) |
```
sangier, line 126:
```suggestion
| privateStore  | "nextSequenceSend/ports/{identifier}/channels/{identifier}"                    | uint64            | [ICS 4](../ics-004-packet-semantics) |
```

AdityaSripal:
Leaving open for now because we may want to remove specification of private keys entirely. it is up to individual implementations to decide how best to execute the logic required to write the provable keys in store.

Also, we should be standardizing the values stored under the provable keys in this specification as well.

Also blocked on discussion around spec format. i.e. ABI for universal data format and moving to IETF standards for specs

spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -103,35 +103,61 @@ The `privateStore`:
...
+
+Since the packet commitment is keyed on the source identifier and sequence, with this key and value together; the receiving chain can prove that the sending chain has sent a packet with the given source and destination identifiers at a given sequence with the timeout and application data.
+
+#### Packet Acknowledgement
+
+The acknowledgement will be provided with the packet to the sending chain. Thus we only need to provably associate the acknowledgement with the original packet. This association is already accomplished by the acknowledgment path which contains the destination identifier and the sequence. Thus on the sending chain, we can prove the acknowledgement was indeed sent for the packet we sent. We prove the packet was sent by us by checking that we stored the packet commitment under the packet commitment path. We can retrieve the client from the source identifier and then prove the counterparty stored under the destination identifier and sequence. Thus, we can associate the acknowledgement stored under this path with the unique packet provided. The acknowledgement commitment can therefore simply consist of a hash of the acknowledgment data sent to the state machine.
```
sangier, line 146: Would be nice to have a visual representation of this

spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -103,30 +103,61 @@ The `privateStore`:
...
 
-Keys used in the provable store MAY safely vary on a per-client-type basis as long as there exists a bipartite mapping between the key formats
-defined herein and the ones actually used in the machine's implementation.
 
-Parts of the private store MAY safely be used for other purposes as long as the IBC handler has exclusive access to the specific keys required.
-Keys used in the private store MAY safely vary as long as there exists a bipartite mapping between the key formats defined herein and the ones
-actually used in the private store implementation.
+| Store          | Path format                                                                | Value type        | Defined in |
+| -------------- | -------------------------------------------------------------------------- | ----------------- | ---------------------- |
+| provableStore  | "commitments/channels/{identifier}/sequences/{bigEndianUint64Sequence}"    | bytes             | [ICS 4](../ics-004-packet-semantics) |
```
sangier, line 115: I'm wondering if there is some space for commitment collision which could led to double usage, but I cannot really think of a scenario that would allow that. At a first glance the provableStore paths look good to me.

spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -103,35 +103,61 @@ The `privateStore`:
...
-| provableStore  | "receipts/ports/{identifier}/channels/{identifier}/sequences/{sequence}"       | bytes             | [ICS 4](../ics-004-packet-semantics) |
-| provableStore  | "acks/ports/{identifier}/channels/{identifier}/sequences/{sequence}"           | bytes             | [ICS 4](../ics-004-packet-semantics) |
+
+| Store          | Path format                                                                | Value type        | Defined in |
+| -------------- | -------------------------------------------------------------------------- | ----------------- | ---------------------- |
+| provableStore  | "commitments/channels/{identifier}/sequences/{bigEndianUint64Sequence}"    | bytes             | [ICS 4](../ics-004-packet-semantics) |
+| provableStore  | "receipts/channels/{identifier}/sequences/{bigEndianUint64Sequence}"       | bytes             | [ICS 4](../ics-004-packet-semantics) |
+| provableStore  | "acks/channels/{identifier}/sequences/{bigEndianUint64Sequence}"           | bytes             | [ICS 4](../ics-004-packet-semantics) |
```
damiannolan, line 117: Looks like we're missing `"nextSequenceSend/ports/{identifier}/channels/{identifier}"`? 

I guess it doesn't need to be provable, but it should also still be defined?? I think we still want the logical separation of sequence/nonce under its own key in v2. And the current impl still relies on port identifier being included

sangier: Since it won't be provable, but the ICS4 still uses it, I have defined the 
```typescript
function nextSequenceSendPath(sourceChannelID: bytes): Path {
    return "nextSequenceSend/{sourceChannelID}"
}
``` in the privateStore. However,  being local and private it could be any path that the chains want to set, thus I guess the implementation should be able to avoid relying on the port, correct?  


```

spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -103,35 +103,61 @@ The `privateStore`:
...
-| provableStore  | "commitments/ports/{identifier}/channels/{identifier}/sequences/{sequence}"    | bytes             | [ICS 4](../ics-004-packet-semantics) |
-| provableStore  | "receipts/ports/{identifier}/channels/{identifier}/sequences/{sequence}"       | bytes             | [ICS 4](../ics-004-packet-semantics) |
-| provableStore  | "acks/ports/{identifier}/channels/{identifier}/sequences/{sequence}"           | bytes             | [ICS 4](../ics-004-packet-semantics) |
+
+| Store          | Path format                                                                | Value type        | Defined in |
+| -------------- | -------------------------------------------------------------------------- | ----------------- | ---------------------- |
+| provableStore  | "commitments/channels/{identifier}/sequences/{bigEndianUint64Sequence}"    | bytes             | [ICS 4](../ics-004-packet-semantics) |
+| provableStore  | "receipts/channels/{identifier}/sequences/{bigEndianUint64Sequence}"       | bytes             | [ICS 4](../ics-004-packet-semantics) |
```
sangier, line 116: seems that sequences will be the main discriminator here, thus would be possible to use a path like `"receipts/{identifier}/sequences/{bigEndianUint64Sequence}"` so deleting the channels reference?

Issue #1136, opened 2024-08-19
Create 02 client spec in v2 core folder


AdityaSripal:
Create the 02 client spec for v2 folder

- Remove delay period passing in
- Specify the register counterparty message and handler

sangier:
In [ICS20](https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer/v1) we decided to: 
- create a v1 folder  
- move the original spec into the v1 folder 
- update the original readme with the V2 version. 

For confirmation, are we going to use the same approach here or, for now and until we have a clear restructure repo plan, we prefer to create v2 folders and place there the new spec versions?

PR #1147, opened 2024-09-09
02-client-TAOv2

sangier:
Closes https://github.com/cosmos/ibc/issues/1136

### DIFF
--- a/spec/core/v2/ics-002-client-semantics/README.md
+++ b/spec/core/v2/ics-002-client-semantics/README.md
@@ -254,17 +254,6 @@ A `ClientMessage` is an opaque data structure defined by a client type which pro
 type ClientMessage = bytes
 ```
 
-#### `Counterparty`
-
-`Counterparty` is the data structure responsible for maintating the counterparty informations.
-
-```typescript
-interface Counterparty {
-    channelId: Identifier
-    keyPrefix: CommitmentPrefix
-}
-```
-
 ### Store paths
 
 Client state paths are stored under a unique client identifier.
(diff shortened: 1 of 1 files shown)


spec/core/v2/ics-002-client-semantics/README.md
```diff
@@ -343,44 +343,44 @@ to intervene to unfreeze a frozen client & provide a new correct ClientMessage w
 It is utilised to verify presence or absence of a particular key/value pair in state
 at a particular finalised height (necessarily associated with a particular commitment root).
 
+#### `Counterparty`
+
+`Counterparty` is the data structure responsible for maintaining the counterparty information.
+
+```typescript
+interface Counterparty {
+    channelId: Identifier
```
sangier, line 352: Shall we change all the channelID references to to clientIDs?

AdityaSripal: I think we should keep this in ICS4 actually, since it is only used for channel level packet processing. In IBC-go we have the channel ID and clientID separate to enable aliasing.

So in the spec I think we should have them separate to note it isn't a strict requirement to have the identifiers the same. In a note, we can say that implementations (like Solidity) can choose to keep them the same.

spec/core/v2/ics-002-client-semantics/README.md
```diff
@@ -527,6 +527,41 @@ type verifyNonMembership = (ClientState, Height, CommitmentProof, Path) => boole
 
 IBC handlers MUST implement the functions defined below.
 
+#### Counterparty Idenfitifcation and Registration
```
AdityaSripal, line 530: Let's move this section to ICS4 as we said.

AdityaSripal (commented):
Great work @sangier ! I think the sections on counterparty are good, but we should have it housed in ICS4 spec

spec/core/v2/ics-002-client-semantics/README.md
sangier: Makes definitely sense to me!

AdityaSripal (approved):
Nice work @sangier !! Looking forward to the ICS4 PR

Issue #1140, opened 2024-08-19
ICS4: v2 handling of packet flow


AdityaSripal:
In v2, the only functions that are necessary for 04-channel are the packet flow handlers. Should we rename this spec so it is no longer called channel?

- Retain only the packet flow handlers and ensure all similar checks are still made
- Specify and explain new packet structure (including multidata)
- Specify and explain new acknowledgment structure (including metadata)
- Specify and explain how the packet flow handler will route verification to the underlying client and route application data to all relevant applications

sangier:
Closed by https://github.com/cosmos/ibc/pull/1148

Issue #1141, opened 2024-08-19
ICS20: Support v2 TAO in ICS20 spec


AdityaSripal:
Make changes to support v2 TAO for ICS20. We need to figure out how to clearly communicate this, as there is ICS20v1 and ICS20 v2. While both versions of ICS20 can be modified to support v2 TAO. Thus we need to find a way to clearly communicate this version compatibility matrix

sangier:
Closed by https://github.com/cosmos/ibc/pull/1157

PR #1148, opened 2024-09-11
04-Packet-Flow-v2

sangier:
This PR introduces the draft specification for the IBC version 2 ICS-04: Channel and Packet Semantics.

It formalizes the semantics for channel creation, counterparty registration, and packet flow management in IBC version 2 taking into account all the v2 spec changes.

Closes https://github.com/cosmos/ibc/issues/1140


spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -190,10 +167,23 @@ The architecture of clients, connections, channels and packets:
 
 #### Store paths
 
+NOTE do we stil need this, or we can retrieve the sequence from the commitment path associated with the clientID? 
+
+- The `nextSequenceSend`, stored separately, tracks the sequence number for the next packet to be sent.
+
+```typescript
+function nextSequenceSendPath(sourceID: bytes, destID: bytes): Path {
```
AdityaSripal: This shouldn't be keyed on the string `clients`.

And there's no need for there to be both source and destId. Since we only need our sides' identifier to uniquely key on the channel.

It can just be

`nextSequenceSend/{sourceId}`

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -190,10 +167,23 @@ The architecture of clients, connections, channels and packets:
 
 #### Store paths
 
+NOTE do we stil need this, or we can retrieve the sequence from the commitment path associated with the clientID?
```
AdityaSripal: We still need this. We want to keep a track of the nonce

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -190,10 +167,23 @@ The architecture of clients, connections, channels and packets:
...
+```
+
+/* Is all of this Unnecessary? 
+
+NOTE Channel paths and capabilities should be maintained for backward compatibility? 
 Channel structures are stored under a store path prefix unique to a combination of a port identifier and channel identifier:
 
 ```typescript
-function channelPath(portIdentifier: Identifier, channelIdentifier: Identifier): Path {
+function channelPath(portIdentifier: bytes, channelIdentifier: bytes): Path {
```
AdityaSripal: Not necessary

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -206,513 +196,111 @@ function channelCapabilityPath(portIdentifier: Identifier, channelIdentifier: Id
...
-}
-```
+*/
 
 Constant-size commitments to packet data fields are stored under the packet sequence number:
 
 ```typescript
-function packetCommitmentPath(portIdentifier: Identifier, channelIdentifier: Identifier, sequence: uint64): Path {
-    return "commitments/ports/{portIdentifier}/channels/{channelIdentifier}/sequences/{sequence}"
+function packetCommitmentPath(sourceId: bytes, sequence: uint64): Path {
```
AdityaSripal, line 166: the sequence needs to be in bigEndian format. See ICS24 for details

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -206,513 +196,111 @@ function channelCapabilityPath(portIdentifier: Identifier, channelIdentifier: Id
...
+    return "receipts/channels/{sourceId}/sequences/{bigEndianUint64Sequence}"
 }
 ```
 
 Packet acknowledgement data are stored under the `packetAcknowledgementPath`:
 
 ```typescript
-function packetAcknowledgementPath(portIdentifier: Identifier, channelIdentifier: Identifier, sequence: uint64): Path {
-    return "acks/ports/{portIdentifier}/channels/{channelIdentifier}/sequences/{sequence}"
+function packetAcknowledgementPath(sourceId: bytes, sequence: uint64): Path {
```
AdityaSripal, line 184: ditto

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -206,513 +196,111 @@ function channelCapabilityPath(portIdentifier: Identifier, channelIdentifier: Id
...
-
-```typescript
-type validateChannelIdentifier = (portIdentifier: Identifier, channelIdentifier: Identifier) => boolean
-```
-
-If not provided, the default `validateChannelIdentifier` function will always return `true`.
+#### Counterparty Idenfitifcation and Registration 
 
-#### Channel lifecycle management
+A client MUST have the ability to idenfity its counterparty. With a client, we can prove any key/value path on the counterparty. However, without knowing which identifier the counterparty uses when it sends messages to us, we cannot differentiate between messages sent from the counterparty to our chain vs messages sent from the counterparty with other chains. Most implementations will not be able to store the ICS-24 paths directly as a key in the global namespace, but will instead write to a reserved, prefixed keyspace so as not to conflict with other application state writes. Thus the counteparty information we must have includes both its identifier for our chain as well as the key prefix under which it will write the provable ICS-24 paths.
```
AdityaSripal, line 203: Again we are not focusing on clients anymore as the operative word in ICS4. I think it makes sense to just use channel and perhaps even channelEnd

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -206,513 +196,111 @@ function channelCapabilityPath(portIdentifier: Identifier, channelIdentifier: Id
...
-}
-```
-
-The `chanOpenTry` function is called by a module to accept the first step of a channel opening handshake initiated by a module on another chain.
+Thus, IBC version 2 introduces a new message `RegisterCounterparty` that will associate the counterparty client of our chain with our client of the counterparty. Thus, if the `RegisterCounterparty` message is submitted to both sides correctly. Then both sides have mirrored <client,client> pairs that can be treated as channel identifiers. Assuming they are correct, the client on each side is unique and provides an authenticated stream of packet data between the two chains. If the `RegisterCounterparty` message submits the wrong clientID, this can lead to invalid behaviour; but this is equivalent to a relayer submitting an invalid client in place of a correct client for the desired chain. In the simplest case, we can rely on out-of-band social consensus to only send on valid <client, client> pairs that represent a connection between the desired chains of the user; just as we currently rely on out-of-band social consensus that a given clientID and channel built on top of it is the valid, canonical identifier of our desired chain.
```
AdityaSripal, line 205: Good explanation!!! Let's update for having channel id and client id separate as well

spec/core/v2/ics-004-packet-semantics/README.md
AdityaSripal, line 203:
```suggestion
Each IBC channel end MUST have the ability to idenfity its counterparty. With a client, we can prove any key/value path on the counterparty. However, without knowing which identifier the counterparty uses when it sends messages to us, we cannot differentiate between messages sent from the counterparty to our chain vs messages sent from the counterparty with other chains. Most implementations will not be able to store the ICS-24 paths directly as a key in the global namespace, but will instead write to a reserved, prefixed keyspace so as not to conflict with other application state writes. Thus the counterparty information we must have includes both its identifier for our chain as well as the key prefix under which it will write the provable ICS-24 paths.
```

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -206,513 +196,111 @@ function channelCapabilityPath(portIdentifier: Identifier, channelIdentifier: Id
...
-        channel.counterpartyPortIdentifier,
-        channel.counterpartyChannelIdentifier,
-        expected
-      ))
-    }
-
-    // write will happen in the handler defined in the ICS26 spec
+// getCounterparty retrieves the stored counterparty identifier
+// given the channelIdentifier on our chain once it is provided
+function getCounterparty(clientId: bytes): Counterparty {
```
AdityaSripal, line 232: This should be mapped to channel ID now

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -206,513 +196,111 @@ function channelCapabilityPath(portIdentifier: Identifier, channelIdentifier: Id
...
-      ))
+function RegisterCounterparty(
+    clientID: bytes, // this will be our own client identifier representing our channel to desired chain
+    counterpartyClientId: bytes, // this is the counterparty's identifier of our chain
+    counterpartyKeyPrefix: CommitmentPrefix,
+    authentication: data, // implementation-specific authentication data
+) {
+    assert(verify(authentication))
+
+    counterparty = Counterparty{
```
AdityaSripal, line 216: We should also just consider this a v2 channel and rename.

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -206,513 +196,111 @@ function channelCapabilityPath(portIdentifier: Identifier, channelIdentifier: Id
...
-    // write may happen asynchronously in the handler defined in the ICS26 spec
-    // if the channel is closing during an upgrade, 
-    // then we can delete all auxiliary upgrade information
-    provableStore.delete(channelUpgradePath(portIdentifier, channelIdentifier))
-    privateStore.delete(counterpartyUpgradePath(portIdentifier, channelIdentifier))
-
-    channel.state = CLOSED
-    provableStore.set(channelPath(portIdentifier, channelIdentifier), channel)
+type IBCRouter struct {
+    versions: portId -> [Version]
```
AdityaSripal, line 251: I don't believe we need this in the router anymore. If the portID doesn't support the version the callback will simply error

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -206,513 +196,111 @@ function channelCapabilityPath(portIdentifier: Identifier, channelIdentifier: Id
...
+    callbacks: portId -> [Callback]
+    clients: clientId -> Client
+    ports: portId -> counterpartyPortId
```
AdityaSripal, line 254: I don't think this is necessary on the router. If the counterpartyPortId is wrong given our portID, it is responsibility of application to error.

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -736,84 +324,78 @@ Represented spatially, packet transit between two machines can be rendered as fo
...
-    abortTransactionUnless(connection !== null)
-
-    // check if the calling module owns the sending port
-    abortTransactionUnless(authenticateCapability(channelCapabilityPath(sourcePort, sourceChannel), capability))
+    sourceClientId: bytes,
+    destClientId: bytes,
+    timeoutTimestamp: uint64,
+    packetData: []byte
+): uint64 {
+    // in this specification, the source channel is the clientId
```
AdityaSripal, line 298: We need to remove this assumption

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -736,84 +324,78 @@ Represented spatially, packet transit between two machines can be rendered as fo
...
-    // check that the channel must be OPEN to send packets;
-    abortTransactionUnless(channel !== null)
-    abortTransactionUnless(channel.state === OPEN)
-    connection = provableStore.get(connectionPath(channel.connectionHops[0]))
-    abortTransactionUnless(connection !== null)
-
-    // check if the calling module owns the sending port
-    abortTransactionUnless(authenticateCapability(channelCapabilityPath(sourcePort, sourceChannel), capability))
+    sourceClientId: bytes,
+    destClientId: bytes,
```
AdityaSripal, line 294: Must use ChannelIds

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -736,84 +324,78 @@ Represented spatially, packet transit between two machines can be rendered as fo
...
+    // NOTE - What is the commit port? Should be the sourcePort? If yes, in the packet we should put destPort and destID?
+    // if the sequence doesn't already exist, this call initializes the sequence to 0
+    //sequence = channelStore.get(nextSequenceSendPath(commitPort, sourceID))
+    sequence = channelStore.get(nextSequenceSendPath(sourceClientID, destClientID))
+    
     // store commitment to the packet data & packet timeout
-    provableStore.set(
-      packetCommitmentPath(sourcePort, sourceChannel, sequence),
-      hash(hash(data), timeoutHeight, timeoutTimestamp)
+    // Note do we need to keep the channelStore? Should this be instead the counterParty store or something similar? Do we keep it for backward compatibility?
```
AdityaSripal: It should just be the provableStore

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -736,84 +324,78 @@ Represented spatially, packet transit between two machines can be rendered as fo
...
+    channelStore.set(
+      packetCommitmentPath(sourceClientId sequence),
+      hash(hash(data), timeoutTimestamp)
```
AdityaSripal, line 319: This is wrong, we need to use the commitment function defined in `PACKET.md`

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -822,166 +404,77 @@ We pass the address of the `relayer` that signed and submitted the packet to ena
...
-
-        abortTransactionUnless(packetReceipt === null)
-        provableStore.set(
-          packetReceiptPath(packet.destPort, packet.destChannel, packet.sequence),
-          SUCCESSFUL_RECEIPT
-        )
-      break;
-    }
+        merklePath,
+        hash(packet.data, packet.timeoutTimestamp)
```
AdityaSripal, line 387: need to use Commitment function defined in PACKET.md

spec/core/v2/ics-004-packet-semantics/README.md
```diff
@@ -822,166 +404,77 @@ We pass the address of the `relayer` that signed and submitted the packet to ena
...
+  
+    // we must set the receipt so it can be verified on the other side
+    // this receipt does not contain any data, since the packet has not yet been processed
+    // it's the sentinel success receipt: []byte{0x01}
+    packetReceipt = channelStore.get(packetReceiptPath(packet.sourceChannelId, packet.sequence))
+    assert(packetReceipt === null)
+
+    channelStore.set(
+        packetReceiptPath(packet.sourceChannelId, packet.sequence),
+        SUCCESSFUL_RECEIPT
```
AdityaSripal, line 405: Must define this

AdityaSripal (changes requested):
Great start!! I really like the explanations you gave for the Counterparty

A couple major tweaks as you continue working on this:

- We need to decide on a new name for Counterparty and make Channel a clear use again in the specification
- We need to not tie client id to channel id (aliasing approach)
- The ICS24 paths and commitments need to follow their new v2 specification as defined in ICS24 README and PACKET.md

AdityaSripal (commented):
Nice start @sangier !

We should also check the counterparty channel id is as expected in the packet handlers

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -0,0 +1,1026 @@
...
+
+Note that a `Packet` is never directly serialised. Rather it is an intermediary structure used in certain function calls that may need to be created or processed by modules interacting with the IBC handler.
+
+When the array of payloads, passed-in the packet, is populated with multiple values, the system will handle the packet as a multi-data packet. The multi-data packet handling logic is out of the scope of the current version of this spec. 
+
+An `OpaquePacket` is a packet, but cloaked in an obscuring data type by the host state machine, such that a module cannot act upon it other than to pass it to the IBC handler. The IBC handler can cast a `Packet` to an `OpaquePacket` and vice versa.
+
+```typescript
+type OpaquePacket = object
+```
```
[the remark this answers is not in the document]
AdityaSripal: Let's remove all references to `OpaquePacket` and just use `Packet`

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -0,0 +1,1028 @@
...
+        proofHeight,
+        proof,
+        merklePath,
+        commit))
+
+    
+    // Executes Application logic ∀ Payload
+    payload=packet.data[0]
+    cbs = router.callbacks[payload.destPort]
+    ack,success = cbs.onReceivePacket(packet.channelDestId,payload,relayer,packet.sequence) // Note that payload includes the version. The application is required to inspect the version to route the data to the proper callback
```
sangier, line 679: Discussing with @chatton about https://github.com/cosmos/ibc/issues/1141 we realised we do need the `packet.sequence` to be passed into the `onRecv` `onAck` `onTimeout` callbacks. 
Indeed, the ics20v2 the `packet.sequence` is used to store the packet in the `packetForwardPath` [e.g](https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer#:~:text=privateStore.set(packetForwardPath(forwarding.hops%5B0%5D.portId%2C%20forwarding.hops%5B0%5D.channelId%2C%20packetSequence)%2C%20packet))

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -0,0 +1,1059 @@
...
+| **Post-Conditions (Success)**  | 1. `onTimeoutPacket` is executed and the application state is modified <br> 2. `packetCommitment` has been cleared out <br> 3. `packetReceipt` is empty <br> 4. Event is Emitted<br> | 1. `onTimeoutPacket(..)==True; app.State(beforeTimeoutPacket)!=app.State(afterTimeoutPacket)`<br> 2. `provableStore.get(packetCommitmentPath(packet.channelSourceId, packet.sequence)) === null`<br> 3. `provableStore.get(packetReceiptPath(packet.channelDestId, packet.sequence))==null`<br> 4. Check Event is Emitted<br> |
+| **Post-Conditions (Error)**    | 1. If `onTimeoutPacket` fails and the application state is unchanged <br> 2. `packetCommitment` is not cleared out <br> 3. No Event Emission<br> | 1. `onTimeoutPacket(..)==True; app.State(beforeTimeoutPacket)!=app.State(afterTimeoutPacket)`<br> 2. `provableStore.get(packetCommitmentPath(packet.channelSourceId, packet.sequence)) === null` <br> 3. Check No Event is Emitted<br>|
```
AdityaSripal, line 934:
```suggestion
| **Post-Conditions (Error)**    | 1. If `onTimeoutPacket` fails and the application state is unchanged <br> 2. `packetCommitment` is not cleared out <br> 3. No Event Emission<br> | 1. `onTimeoutPacket(..)==False; app.State(beforeTimeoutPacket)==app.State(afterTimeoutPacket)`<br> 2. `provableStore.get(packetCommitmentPath(packet.channelSourceId, packet.sequence)) === null` <br> 3. Check No Event is Emitted<br>| 
```

AdityaSripal (approved):
Nice work!! I think this can be made more readable in the future once we decide on the future format of the v2 specs. Fixing the small errors and merging for now.

Thanks for the excellent writeup!

PR #1152, opened 2024-10-11
Packet Spec: Hash App data rather than standardizing encoding

AdityaSripal:
The following is the Packet spec that uses only hashing rather than the cbor encoding in the previous spec. This is meant to allow reviewers to compare the two options.

Benefits:

- No standardized encoding allows implementation by virtually any platform with low maintanence burden
- Canonical by construction

Cons:

- Must specify a unique commitment scheme for the internal packet data and acknowledgement rather than relying on existing encoding
- Multiple hashes might be slow and expensive in certain environments (e.g. EVM, and ZKVM)

### DIFF
--- a/spec/core/v2/ics-004-packet-semantics/PACKET.md
+++ b/spec/core/v2/ics-004-packet-semantics/PACKET.md
@@ -80,14 +80,14 @@ func commitPayload(payload: Payload): bytes {
 }
 
 func commitV2Packet(packet: Packet) {
-    timeoutBytes = LittleEndian(timeout)
+    timeoutBytes = LittleEndian(packet.timeout)
     var appBytes: bytes
     for p in packet.payload {
         appBytes = append(appBytes, commitPayload(p))
     }
-    buffer = sha256.Hash(destIdentifier)
-    buffer = append(buffer, sha256.hash(bigEndian(timeoutBytes)))
-    buffer = append(buffer, sha256.hash(data))
+    buffer = sha256.Hash(packet.destChannel)
+    buffer = append(buffer, sha256.hash(timeoutBytes))
+    buffer = append(buffer, sha256.hash(appBytes))
     return sha256.Hash(buffer)
 }
 ```
--- a/spec/core/v2/ics-004-packet-semantics/PACKET.md
+++ b/spec/core/v2/ics-004-packet-semantics/PACKET.md
@@ -85,6 +85,8 @@ func commitPayload(payload: Payload): bytes {
 // Note: SourceChannel and the sequence are omitted since they will be included in the key
 // Every other field of the packet is committed to in the packet which will be stored in the
 // packet commitment value
+// The final hash will be prepended by the byte 0x02 in order to clearly define the protocol version
+// and allow for future upgradability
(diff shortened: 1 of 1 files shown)


sangier:
Dropping some thoughts about the cons: "Multiple hashes might be slow and expensive in certain environments (e.g. EVM, and ZKVM)"

- Checking the ethereum [yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf) I noticed that a simple transfer cost 21,000 gas, while a keccak256 encoding only costs 30 gas for the first 32 bytes, and 6 gas for each additional 32 bytes. Thus at least for EVM environment multiple hashes are not expensive and should be fast as keccack is the native enconding function.  
- The multiple hashes may affect the zkvm proof generation and verification processes, but I'm not sure how. 
- The native encoding function for Solana is sha-256. 
- Cairo uses natively Poseidon and Pedersen which are zk-snark friendly.   

So, intra-platform (e.g. Cosmos to Cosmos, EVM to EVM, Sol to Sol ...) we could use the native hashing functions being fast and cheap. However inter-platform (e.g EVM to Solana..) we must use one hashing function and use it in both the environments, which may affect the computation of one of the side.

spec/core/v2/ics-004-packet-semantics/PACKET.md
```diff
@@ -0,0 +1,126 @@
...
+    buffer = append(sha256.Hash(payload.appData))
+}
+
+func commitV2Packet(packet: Packet) {
+    timeoutBytes = LittleEndian(timeout)
+    var appBytes: bytes
+    for p in packet.payload {
+        appBytes = append(appBytes, commitPayload(p))
+    }
+    buffer = sha256.Hash(destIdentifier)
```
sangier, line 88:
```suggestion
    buffer = sha256.Hash(packet.destChannel)
```

spec/core/v2/ics-004-packet-semantics/PACKET.md
```diff
@@ -0,0 +1,126 @@
...
+    buffer = append(buffer, sha256.hash(bigEndian(timeoutBytes)))
+    buffer = append(buffer, sha256.hash(data))
```
sangier, line 90:
```suggestion
    buffer = append(buffer, sha256.hash(packet.data))
```

spec/core/v2/ics-004-packet-semantics/PACKET.md
```diff
@@ -0,0 +1,126 @@
...
+func commitPayload(payload: Payload): bytes {
+    buffer = sha256.Hash(payload.sourcePort)
+    buffer = append(sha256.Hash(payload.destPort))
+    buffer = append(sha256.Hash(payload.version))
+    buffer = append(sha256.Hash(payload.encoding))
+    buffer = append(sha256.Hash(payload.appData))
+}
+
+func commitV2Packet(packet: Packet) {
+    timeoutBytes = LittleEndian(timeout)
```
sangier, line 83:
```suggestion
    timeoutBytes = LittleEndian(packet.timeout)
```

sangier (approved):

PR #1155, opened 2024-10-25
ICS24: Restructure provable packet keys

AdityaSripal:
These keys are now less verbose and also keys being proved on the same channel will be closer together in the tree since we key first on channelID. This will improve efficiency when we implement batching


gjermundgaraba (approved):

sangier (approved):
That's great, thanks! Only one thing, shall we specify somewhere that `0x1` is for commitments, `0x2` for receipts and `0x3` for acks?

spec/core/v2/ics-024-host-requirements/README.md
```diff
@@ -110,11 +110,11 @@ IBC/TAO implementations MUST implement the following paths for the `provableStor
...
-| provableStore  | "commitments/channels/{identifier}/sequences/{bigEndianUint64Sequence}"    | bytes             | [ICS 4](../ics-004-packet-semantics) |
-| provableStore  | "receipts/channels/{identifier}/sequences/{bigEndianUint64Sequence}"       | bytes             | [ICS 4](../ics-004-packet-semantics) |
-| provableStore  | "acks/channels/{identifier}/sequences/{bigEndianUint64Sequence}"           | bytes             | [ICS 4](../ics-004-packet-semantics) |
+| Store          | Path format                                       | Value type | Defined in                           |
+| -------------- | ------------------------------------------------- | ---------- | ------------------------------------ |
+| provableStore  | {channelIdentifier}|0x1|{bigEndianUint64Sequence} | bytes      | [ICS 4](../ics-004-packet-semantics) |
+| provableStore  | {channelIdentifier}|0x2|{bigEndianUint64Sequence} | bytes      | [ICS 4](../ics-004-packet-semantics) |
+| provableStore  | {channelIdentifier}|0x3|{bigEndianUint64Sequence} | bytes      | [ICS 4](../ics-004-packet-semantics) |
```
soareschen, line 117: I think you need a length prefix, or it would be vulnerable to ambiguous encoding.

Also, can't we extend the IBC client API with 3 methods with explicit parameters, so that it can verify the commitment in different ways? The IBC client API don't really need a general method to verify all possible blockchain commitments. Also, not all blockchains need to store commitments in a flattened Merkle store. They may be stored in more efficient or more secure ways, e.g. nested Merkle trees or ZK storage proofs.

For backward compatibility, we can just wrap the legacy IBC clients, and use the encoding scheme verified here to verify the IBC commitments.

soareschen: Oh right, it is not ambiguous because u64 is fixed size. Though I think it is tricky to reason about ambiguity depending on whether the subsequent fields are fixed size. It would also become ambiguous in case if we generalize sequence in the future to allow variable-length nonce bytes.

AdityaSripal: Indeed, the one concern I had was that it could be easy to mess up in the future

PR #1165, opened 2024-11-11
ICS4: Event emission Specification

AdityaSripal:
Emit events for relayers to construct the packet messages

- Simple in single payload case
- Support for multipayload

### DIFF
--- a/spec/core/v2/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/v2/ics-004-channel-and-packet-semantics/README.md
@@ -323,7 +323,7 @@ function createChannel(
         nextSequenceSend[channelId]=1
         
         // Event Emission 
-    emitLogEntry("createChannel", {
+    emitEvents("createChannel", {
       channelId: channelId, 
       clientId: clientId, 
       creatorAddress: msg.signer(),
@@ -377,7 +377,7 @@ function registerCounterparty(
 
     // log that a packet can be safely sent
     // Event Emission 
-    emitLogEntry("registerCounterparty", {
+    emitEvents("registerCounterparty", {
       channelId: channelId,
       clientId: channel.clientId
       counterpartyChannelid: counterpartyChannelId,
@@ -597,7 +597,7 @@ function sendPacket(
     nextSequenceSend[sourceChannelId]=sequence+1
     
     // Event Emission for send packet
-    emitLogEntry("send_packet", {
+    emitEvents("send_packet", {
       sourceId: sourceChannelId, 
       destId: channel.counterpartyChannelId,
       sequence: sequence, // value is string in decimal format
@@ -722,7 +722,7 @@ function recvPacket(
         // NOTE: Synchronous ack. 
         writeAcknowledgement(packet.channelDestId,packet.sequence,ack)
         // In case of Synchronous ack we emit the event here as we have all the necessary information, while writeAcknowledgement can only retrieve this in case of asynchronous ack. 
-        emitLogEntry("write_acknowledgement", {
+        emitEvents("write_acknowledgement", {
             sequence: packet.sequence, // value is string in decimal format
             sourceId: packet.channelSourceId,
             destId: packet.channelDestId,
@@ -742,7 +742,7 @@ function recvPacket(
     )
 
     // Event Emission for receive packet
-    emitLogEntry("recv_packet", {
+    emitEvents("recv_packet", {
       sourceId: sourceChannelId, 
       destId: channel.counterpartyChannelId,
       sequence: sequence, // value is string in decimal format
@@ -833,7 +833,7 @@ function writeAcknowledgement(
                 ackString = ackString + "/"
             } 
         }
-        emitLogEntry("write_acknowledgement", {
+        emitEvents("write_acknowledgement", {
         sequence: packet.sequence, // value is string in decimal format
         sourceId: packet.channelSourceId,
         destId: packet.channelDestId,
@@ -841,7 +841,7 @@ function writeAcknowledgement(
         })
 
         // Event Emission for receive packet. emit again so relayer can reconstruct the packet
-        emitLogEntry("recv_packet", {
+        emitEvents("recv_packet", {
         sourceId: sourceChannelId, 
         destId: channel.counterpartyChannelId,
         sequence: sequence, // value is string in decimal format
@@ -952,7 +952,7 @@ function acknowledgePacket(
     }
     
     // Event Emission // Check fields
-    emitLogEntry("acknowledgePacket", {
+    emitEvents("acknowledgePacket", {
       sequence: packet.sequence, // value is string in decimal format
       sourceId: packet.channelSourceId,
       destId: packet.channelDestId,
@@ -1068,7 +1068,7 @@ function timeoutPacket(
     channelStore.delete(packetCommitmentPath(packet.channelSourceId, packet.sequence))
     
     // Event Emission for timeout packet
-    emitLogEntry("timeoutPacket", {
+    emitEvents("timeoutPacket", {
       sequence: packet.sequence, // value is string in decimal format
       sourceId: packet.channelSourceId,
       destId: packet.channelDestId,
--- a/spec/core/v2/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/v2/ics-004-channel-and-packet-semantics/README.md
@@ -378,7 +378,8 @@ function registerCounterparty(
     // log that a packet can be safely sent
     // Event Emission 
     emitLogEntry("registerCounterparty", {
-      channelId: channelId, 
+      channelId: channelId,
+      clientId: channel.clientId
       counterpartyChannelid: counterpartyChannelId,
     })
 }
--- a/spec/core/v2/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/v2/ics-004-channel-and-packet-semantics/README.md
@@ -162,8 +162,8 @@ The ICS-04 use the protocol paths, defined in [ICS-24](../ics-024-host-requireme
 Thus, constant-size commitments to packet data fields are stored under the packet sequence number:
 
 ```typescript
-function packetCommitmentPath(channelSourceId: bytes, sequence: BigEndianUint64): Path {
-    return "{channelSourceId}|0x1|{bigEndianUint64Sequence}"
+function packetCommitmentPath(sourceChannel: bytes, sequence: BigEndianUint64): Path {
+    return "{sourceChannel}|0x1|{bigEndianUint64Sequence}"
 }
 ```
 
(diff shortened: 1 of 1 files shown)


AdityaSripal:
Done! @DimitrisJim

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -596,15 +596,33 @@ function sendPacket(
...
-    // Event Emission 
-    emitLogEntry("sendPacket", {
+    // Event Emission for send packet
+    emitEvents("send_packet", {
       sourceId: sourceChannelId, 
-      destId: channel.counterpartyChannelId, 
-      sequence: sequence,
-      packet: packet,
-      timeoutTimestamp: timeoutTimestamp, 
+      destId: channel.counterpartyChannelId,
```
DimitrisJim, line 602: I see these are defined as `bytes` in spec atm (for eth side of things where these are a contract address I assume?). Should they be hex encoded too in spec?

AdityaSripal: Mmmm this is a good point. In ibc-go this is an alphanumeric string. It feels wasteful to convert that to a hex string only to convert it back. Tho it is possible that other implementations use bytes that do not decode to ASCII characters.

What do you think @gjermundgaraba ?

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -596,15 +596,33 @@ function sendPacket(
     // increment the sequence. Thus there are monotonically increasing sequences for packet flow for a given clientId
     nextSequenceSend[sourceChannelId]=sequence+1
     
-    // log that a packet can be safely sent
-    // Event Emission 
-    emitLogEntry("sendPacket", {
+    // Event Emission for send packet
+    emitEvents("send_packet", {
       sourceId: sourceChannelId,
```
gjermundgaraba, line 601: Is there any particular reason why we use sourceId here rather than sourceChannelId? I don't care too much either way, but why not use the same names as the packet structure?

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -596,15 +596,33 @@ function sendPacket(
...
       sourceId: sourceChannelId, 
-      destId: channel.counterpartyChannelId, 
-      sequence: sequence,
-      packet: packet,
-      timeoutTimestamp: timeoutTimestamp, 
+      destId: channel.counterpartyChannelId,
+      sequence: sequence, // value is string in decimal format
+      timeoutTimestamp: timeoutTimestamp, // value is string in decimal format
+      payloadLength: len(payloads), // value is string in decimal format
+      // include first payload data in events if there is only one payload
```
gjermundgaraba, line 606: Does that mean that if this is a multi-payload packet, these need to be empty strings?

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -596,15 +596,33 @@ function sendPacket(
...
+      encoding: payload[0].encoding,
+      data: toHex(payload[0].appData) // emit app bytes as string in hex format
     })
+
+    // for multi payload cases, we will emit each payload as a separate event
+    // these will include the packet identifier so they can be indexed and
+    // reconstructed by relayers
+    for i, payload in payloads {
+        emitLongEntry("send_payload", {
+            sourceId: sourceChannelId,
```
gjermundgaraba, line 617: Same question as above regarding naming differences

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -690,19 +708,25 @@ function recvPacket(
...
+    }
     abortTransactionUnless(success)
     if ack != nil {
         // NOTE: Synchronous ack. 
         writeAcknowledgement(packet.channelDestId,packet.sequence,ack)
         // In case of Synchronous ack we emit the event here as we have all the necessary information, while writeAcknowledgement can only retrieve this in case of asynchronous ack. 
-        emitLogEntry("writeAcknowledgement", {
-            sequence: packet.sequence,
+        emitEvents("write_acknowledgement", {
+            sequence: packet.sequence, // value is string in decimal format
```
gjermundgaraba, line 726: nit/question: Curious about the use of the word "decimal" here. I immediately think of a number with some kind of precision. Would it be simpler to say integer?

Also, what is the reason for using string-formatted numbers here?

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -784,14 +825,49 @@ function writeAcknowledgement(
     // Note that the event should be emitted by this function only in the asynchrounous ack case. Otherwise the event is emitted during the onReceive 
     packet=getPacket(destChannelId,sequence)
     if(packet!=nil){
-        emitLogEntry("writeAcknowledgement", {
-        sequence: packet.sequence,
+        // construct acknowlegement event by concatenating each app acknowledgment together with a "/" delimiter
```
gjermundgaraba, line 828: Since we emit separate events for multiple payloads, could we not do the same here? Is it because it would be simpler for an event handler to always have to deal with a single writeAcknowledgement event per packet?

gjermundgaraba (approved):
This looks reasonable to me. Is it OK for events to emit extra information, if there are implementation-specific things that are needed?

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
AdityaSripal: I mean that the string is base10 as opposed to hexadecimal or something else.

This is assuming the event system is a <string, string> attribute system.

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
AdityaSripal: We could do this yes. Think I just did this before thinking about the payload case

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
gjermundgaraba: Ah, I see. That makes sense.

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
gjermundgaraba: Right, yes because you would check the length to find out if there should be more to look for.

gjermundgaraba (approved):

srdtrk:
I think this is mostly written under the assumtions that all events must look like key-value pairs where both key and value must be string. Also assumes that there must be a key identifying the event. In solidity it is not necessarily like this, and serializing the packet this way would likely cost a lot of gas. Instead we'd be better off just emitting the full packet (abi encoded) each time rather than emitting each individual field. For example:

```solidity
emit SendPacket(packet)
```

I don't see any issue with this unless we have to implement the events this way in every implementation.

AdityaSripal:
Yes its a bit of a tough needle to thread which is why we faced this problem in the v1 spec. Note the exact view of the events is not strictly necessary to be followed in order to be a compliant implemention (only provable key/value logic is absolutely necessary). If different implementations emit events the same way, then the work on the relayer is simpler since they will use the same parsing and reconstruction logic. We could also just tell different implementations **what** they have to emit without specifying how, then relayers must build custom integrators for each implementation. This has been a consistent complaint and request from relayer teams to standardize and specify the events for the v1 spec. Open to doing things differently but should ensure its ok with potential relayer implementations

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -596,15 +595,33 @@ function sendPacket(
...
-    emitLogEntry("sendPacket", {
-      sourceId: sourceChannelId, 
-      destId: channel.counterpartyChannelId, 
-      sequence: sequence,
-      packet: packet,
-      timeoutTimestamp: timeoutTimestamp, 
+    // Event Emission for send packet
+    emitEvents("send_packet", {
+      sourceChannel: sourceChannelId, 
+      destChannel: channel.counterpartyChannelId,
```
DimitrisJim, line 601: hm, counterparty channel id is equal to source channel on recv no? why not use packet fields here?

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -686,47 +703,65 @@ function recvPacket(
...
-      data: packet.data
-      timeoutTimestamp: packet.timeoutTimestamp,
-      sequence: packet.sequence,
-      sourceId: packet.channelSourceId,
-      destId: packet.channelDestId,
-      relayer: relayer 
+    // Event Emission for receive packet
+    emitEvents("recv_packet", {
+      sourceChannel: sourceChannelId, 
+      destChannel: channel.counterpartyChannelId,
```
DimitrisJim, line 741: meant to drop comment here, counterparty channel id on recv will be equal to source channel no? Should be `packet.DestChannel`?

gjermundgaraba:
@srdtrk @AdityaSripal one other thing we have not looked too closely at is how different way of structuring the events affect gas cost on Ethereum. I would need to dive a bit deeper into it, but the general way it works is `375 + 375 * numberOfIndexedParameters + numberOfUnindexedBits`

Currently, I don't believe we haven't added any extra indexed parameters, so we are mostly concerned with size. So where we can keep size smaller, it will be advantageous. Obviously any marshaling/encoding we have to do first is also very much relevant.

> Yes its a bit of a tough needle to thread

Would at least being able to use the local "type" for different variables such as timestamp and so on be OK in this context (rather than having to encode it to string)?

sangier (approved):

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -784,14 +819,45 @@ function writeAcknowledgement(
...
+            emitEvents("write_acknowledgement", {
+                sourceChannel: packet.sourceChannel,
+                destChannel: packet.destChannel,
+                sequence: packet.sequence, // value is string in decimal format
+                payloadSequence: i, // value is string in decimal format
+                acknowledgement: toHex(ack),
+            })
+        }
+
+        // Event Emission for receive packet. emit again so relayer can reconstruct the packet
```
sangier, line 833: Note that re-emission may be expensive. Can't relayers store somewhere these events instead?

sangier:
working on the solidity events, I can confirm that if we don't emit events like serdar suggested `emit SendPacket(packet)` it increases the cost and we end up in a stack too deep issue. I guess that's fine to emit event like suggested as long as all the relevant fields are emitted. The relayer reconstruction should then be custom for the eth side.

spec/core/v2/ics-004-channel-and-packet-semantics/README.md
```diff
@@ -596,15 +595,33 @@ function sendPacket(
...
+      // include first payload data in events if there is only one payload
+      version: payload[0].version,
+      encoding: payload[0].encoding,
+      data: toHex(payload[0].appData) // emit app bytes as string in hex format
     })
+
+    // for multi payload cases, we will emit each payload as a separate event
+    // these will include the packet identifier so they can be indexed and
+    // reconstructed by relayers
+    for i, payload in payloads {
```
sangier, line 614: In the single payload case, a `send_payload` event would be emitted anyway. Since we emit an event with the first payload related data in the `send_packet` event, this results in a duplicated field emission. 
Do we want to emit the `send_payload` event even in the single payload case? 
If not, shall we discard the first element of the payload array here? Alternatively, shall we remove version,encoding,data from the `send_packet` event?

AdityaSripal:
Superceded by #1172

PR #1173, opened 2024-12-04
Fix outdated description for optimistic send

### LINKED ISSUE #1170: Question: Clarification on support for optimistic packet sending in ICS-004
Hi there,

In the current IBC specifications, is the lack of support for optimistic packet sending intentional?

As described [here](https://github.com/cosmos/ibc/blob/47fea20d4d400e967721396092c6b43398c65d78/spec/core/ics-004-channel-and-packet-semantics/README.md?plain=1#L764), the channel state is required to be OPEN when sending a packet.
Additionally, I know that the ibc-go has decided to remove support for optimistic packet sending in [this issue](https://github.com/cosmos/ibc-go/issues/2670).

However, if optimistic packet sending is not supported in the ICS-004 as well, [the following description about packet timeout](https://github.com/cosmos/ibc/blob/47fea20d4d400e967721396092c6b43398c65d78/spec/core/ics-004-channel-and-packet-semantics/README.md?plain=1#L1157) in the document seems misleading or incorrect:
> Since we allow optimistic sending of packets (i.e. sending a packet before a channel opens), we must also allow optimistic timing out of packets. With optimistic sends, the packet may be sent on a channel that eventually opens or a channel that will never open. If the channel does open after the packet has timed out, then the packet will never be received on the counterparty so we can safely timeout optimistically. If the channel never opens, then we MUST timeout optimistically so that any state changes made during the optimistic send by the application can be safely reverted.

Could you clarify whether this feature is supported and whether the current description aligns with the intended design?

bluele:
fix #1170

### DIFF
--- a/spec/core/ics-004-channel-and-packet-semantics/README.md
+++ b/spec/core/ics-004-channel-and-packet-semantics/README.md
@@ -1537,6 +1537,8 @@ Jan 10, 2022 - Add ORDERED_ALLOW_TIMEOUT channel type and appropriate logic
 
 Mar 28, 2023 - Add `writeChannel` function to write channel end after executing application callback
 
+Dec 4, 2024 - Remove the description for optimistic packet sending
+
 ## Copyright
 
 All content herein is licensed under [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0).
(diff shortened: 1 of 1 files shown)


damiannolan (commented):
Thanks! Would suggest adding an amendment to the "History" section at the bottom of the document.

sangier (approved):
Thanks for catch @bluele !

AdityaSripal (approved):

PR #1213, opened 2025-03-13
feat(ics-033): optimize client updates in multi-hop protocol

VolodymyrBg:
Implemented an optimization in the multi-hop protocol to update clients with specific missing block heights rather than always using the latest block height. This change reduces the likelihood of needing additional client updates on subsequent chains in the multi-hop path, improving overall efficiency.
The implementation:
1. Added a new UpdateClientToHeight function to the Chain struct
2. Modified the client update logic to use the specific consensus height when available
3. Falls back to the existing behavior of updating to the latest height when no specific height is provided
This change addresses the TODO comment in the multi-hop specification while maintaining backward compatibility with existing code.


BoomchainLabs (approved):
Lgtm

PR #1245, opened 2026-02-09
docs: added ics27-gmp (ICS27 v2) specs

srdtrk:


spec/IBC_V2/app/ics-027-gmp/README.md
```diff
@@ -0,0 +1,228 @@
...
+}
+```
+
+Implementations may store additional configuration required to instantiate accounts (e.g. account code hash, beacon address, or router reference).
+
+### Encodings
+
+ICS27-GMP supports multiple serialization encodings for `GMPPacketData` and acknowledgements. Payload encoding follows the per-packet `encoding` field defined by IBC v2 (see `spec/IBC_V2/core/ics-004-packet-semantics/PACKET.md`).
+
+- `application/x-protobuf` (default)
```
mariuszzak: Currently the default one is ABI 

https://github.com/cosmos/ibc-go/blob/main/modules/apps/27-gmp/keeper/msg_server.go#L44-L46

aljo242 (approved):

PR #1252, opened 2026-06-24
feat(link): scaffold the project

swift1337:
- [x] Scaffold the project
- [x] Add YAML config parsing.
- [x] Add support for ENV variables in YAML
- [x] Basic CLI structure
- [x] Add linter

Closes FOU-347
Closes FOU-348


swift1337:
@greptile review

gjermundgaraba (changes requested):
Two things I want us to consider:
1. start from the beginning by putting everything into internal packages - there is zero reason to have public packages at this point and I don't want us to end up in a position where we have to maintain public packages because we just left them in.
2. We don't actually need multiple go modules. The current setup is actually wrong, because it doesn't work without go.work. Lets just make our lives easier have a root go module and put everything that is not main.go in internal packages.

Not exactly critical, but I kinda dislike the double ibc in "ibc/ibc-link". would look nice if it was just "ibc/link" or something.

packages/go/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/cosmos/ibc/ibc-link/packages/go
```
dhfang: hmm shouldn't this be github.com/cosmos/ibc/packages/go

dhfang (commented):
it's typically not recommended to commit the go.work https://go.dev/ref/mod#go-work-file

swift1337:
@greptile review

link/AGENTS.md
```diff
@@ -0,0 +1,13 @@
+# IBC Link Development Guide for AI Agents
+
+1. Use `make lint-fix` to auto-format and lint code before finishing work.
+2. YAML configuration should only contain `camelCase` keywords; never use `snake_case`.
+3. Global CLI flags such as `--home` or `--config` belong in `config.FlagSet`. Add new global flags there, and don't use a separate variable if it can be global.
+
+## CLI development (located in cmd/ibc/)
+
+1. Use a single `init()` function in main.go to wire all subcommands. Never create init() per CLI file.
```
gjermundgaraba: Feel very free to do this in a separate PR, but this should not be in the root AGENTS.md file. If it is needed at all, it should be inside cmd/ibc.

link/AGENTS.md
```diff
@@ -0,0 +1,13 @@
+# IBC Link Development Guide for AI Agents
+
+1. Use `make lint-fix` to auto-format and lint code before finishing work.
+2. YAML configuration should only contain `camelCase` keywords; never use `snake_case`.
```
gjermundgaraba: not necessary right now, but this is something I would prefer to have in a linter, not take up space in every agent's context.

link/AGENTS.md
```diff
@@ -0,0 +1,13 @@
...
+3. Global CLI flags such as `--home` or `--config` belong in `config.FlagSet`. Add new global flags there, and don't use a separate variable if it can be global.
```
gjermundgaraba: this too is only for cli, no?

link/.golangci.yml
gjermundgaraba, line 1: Just a comment for us to remember: I want to discuss what we put in this file, because I think it's important for agent enablement.

link/cmd/ibc/main.go
```diff
@@ -0,0 +1,52 @@
+// Package main is the entrypoint for the ibc-link binary
```
gjermundgaraba:
```suggestion
```

is there a linting reason for this? seems unnecessary 😅

link/internal/config/config.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"os"
+
+	"github.com/goccy/go-yaml"
+	"github.com/pkg/errors"
+
+	"github.com/cosmos/ibc/link/internal/network"
+)
+
+// Config represents a config file
```
gjermundgaraba, line 13: unless there is some linting reason for these comments, I think some comment rules in AGENTS.md would be useful to reduce this. It hurts to see this kind of stuff 😅

link/internal/config/config_test.go
```diff
@@ -0,0 +1,187 @@
...
+				}
+
+				require.NoError(t, err)
+			})
+		}
+	})
+
+	t.Run("LoadFromFile", func(t *testing.T) {
+		t.Run("valid", func(t *testing.T) {
+			// ARRANGE
```
gjermundgaraba, line 52: love this! we should create a very thin skill for writing unit tests!

link/internal/logging/logging.go
gjermundgaraba, line 1: can be done/discussed separately, but I would highly recommend we simplify logging by using the standard library `slog` package.

gjermundgaraba (approved):
LFG. Left some minor comments we can deal with later, no reason to block this PR IMO.

link/internal/config/config_test.go
swift1337: i already have one , will share

dhfang (approved):
🚀

PR #1254, opened 2026-06-29
feat(link): implement `store` package (DB 101)

swift1337:
- [x] Database migrations (`up`, `down`, `status`, `new`)
- [x] Support for Sqlite + sqlite in-memory
- [x] Support for Postgres
- [x] Test cases for migrations and store CRUD
- [x] AI guide for `store` package
- [x] Add `os.Chdir(home)` (`--home`)

---

The gist is that DB callers only interact with the `store.Store` interface, which hides the underlying DB repository.

I hope that [AGENTS.md](https://github.com/cosmos/ibc/blob/89ffb9c5921f63b0be4f7a5b0d8d6d8ab56c44a9/link/internal/store/AGENTS.md) explains the `store` development flow pretty well 

---

Notes: added testcontainers for postgres tests. We can replace the implementtation layer. But it's not slow (<5s for pg tests + they are disabled by default)

```bash
go test -v ./internal/store/... -count=1
=== RUN   TestStore
=== RUN   TestStore/sqlite
    store_test.go:24: filename: /var/folders/98/tp73c8b9341gm5b9l2l82fs00000gn/T/TestStoresqlite1138142896/001/ibc.db
=== RUN   TestStore/sqlite/reopen
=== RUN   TestStore/sqliteInMemory
=== RUN   TestStore/postgres
    store_test.go:82: Postgres container started in 3.068653542s
    store_test.go:86: Created database ibc-link in 47.777958ms
--- PASS: TestStore (3.32s)
    --- PASS: TestStore/sqlite (0.00s)
        --- PASS: TestStore/sqlite/reopen (0.00s)
    --- PASS: TestStore/sqliteInMemory (0.00s)
    --- PASS: TestStore/postgres (3.32s)
PASS
ok      github.com/cosmos/ibc/link/internal/store       3.648s
```

Closes FOU-349


link/cmd/ibc/config.go
```diff
@@ -90,11 +90,42 @@ func printConfigHome(_ *cobra.Command, _ []string) {
...
 		return config.Config{}, errors.Wrap(err, "unable to get config path")
 	}
 
-	return config.LoadFromFile(configPath, true, false)
+	// ensure --home exists
+	if err = config.EnsureDirectory(configPath); err != nil {
+		return config.Config{}, errors.Wrapf(err, "unable to create home directory %s", home)
+	}
+
+	// change process'es working directory to --home
```
gjermundgaraba: again not something we need to deal with here, but this level of comments are more noise than anything else.
Adding some kind of instructions in AGENTS.md that compel the agent to no write comments where the code plainly says the same things as the comment usually helps a bit.

link/cmd/ibc/migrate.go
```diff
@@ -0,0 +1,118 @@
...
+
+	db, err := store.NewStore(context.Background(), cfg)
+	if err != nil {
+		return config.Config{}, nil, err
+	}
+
+	return cfg, db, nil
+}
+
+func printJSON(v any) error {
```
gjermundgaraba: maybe move this so we dont duplicate this later?

link/internal/config/flags.go
```diff
@@ -21,6 +22,7 @@ func DefaultFlagSet() FlagSet {
 	return FlagSet{
 		Home:   "~/.ibc",
 		Config: "ibc.yml",
+		DB:     "",
```
gjermundgaraba, line 25: nit: seems like we're mixing where defaults live right now. smol now, but would be nice if defaults had a clearer single place?

link/internal/store/AGENTS.md
```diff
@@ -0,0 +1,50 @@
+# Store guidelines for AI Agents
```
gjermundgaraba, line 1: *chefs kiss*

link/internal/tests/utils.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func EnvBool(env string) bool {
+	b, _ := strconv.ParseBool(os.Getenv(env))
+
+	return b
+}
+
+// GetFreePorts returns N available ports
+func GetFreePorts(t *testing.T, n int) []int {
```
gjermundgaraba: my clanker told me this is unused?

link/scripts/migratenew.go
```diff
@@ -0,0 +1,57 @@
+//go:build ignore
```
gjermundgaraba: what is this for?

link/.golangci.yml
```diff
@@ -71,5 +71,7 @@ linters:
             # force comments on pubic structs, relax on functions and methods
             - disable-checks-on-functions
             - disable-checks-on-methods
+        - name: blank-imports
```
gjermundgaraba: I don't agree with disabling this wholesale. Better to comment and and ignore on the lines where we _have_ to do this.

link/internal/store/store_postgres.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "parse config")
+	}
+
+	return NewPostgresWithConfig(ctx, poolConfig)
+}
+
+// NewPostgresWithConfig creates a new PostgresDB instance based on pgxpool.Config
+// Allows to modify the config before creation.
+func NewPostgresWithConfig(ctx context.Context, config *pgxpool.Config) (*PostgresDB, error) {
```
gjermundgaraba: One pattern we've used that in the platform backend is having a quick ping after pool creation. It doesn’t have to be in this function necessarily, but it's useful to fail quickly on a bad connection.
https://github.com/cosmos/platform/blob/ab99ca8d540b14d7201bf8f437786d499f4cb122/libs/go/db/db.go#L31-L52

link/internal/store/sqlc.yml
```diff
@@ -0,0 +1,22 @@
+# https://docs.sqlc.dev/en/latest/reference/config.html
```
gjermundgaraba, line 1: We have some potentially nice configs from the platform we can consider (now or later): https://github.com/cosmos/platform/blob/ab99ca8d540b14d7201bf8f437786d499f4cb122/apps/backend/sqlc.yaml#L1-L19

gjermundgaraba (changes requested):
Left some comments.

Another thing I noticed which I think there might be a better solution for is the double-implementation of the same logic in store_sqlite.go and store_postgres.go.
I have a local branch with some suggestions for that that I'll open as a PR against this branch and we can discuss.

Not sure if we've talked about this yet, and it could be a separate issue/pr, but having auto-migration is nice UX: https://github.com/cosmos/platform/blob/ab99ca8d540b14d7201bf8f437786d499f4cb122/apps/backend/internal/assembly/assembly.go#L40-L55

link/internal/store/sqlc.yml
swift1337: I think we don't need anything from there except `emit_pointers_for_null_types: true`

link/internal/store/store_postgres.go
swift1337: I'll make it optional, yes

link/.golangci.yml
swift1337: sg

link/scripts/migratenew.go
swift1337: It was an experiment, I'll remove that

link/internal/tests/utils.go
swift1337: Yeah, it was an experiment

link/internal/config/flags.go
swift1337: I think the default config and the default CLI args/flags are different things. I'd keep `config.DefaultFlagSet()` and `config.DefaultConfig()` as is for now.

link/cmd/ibc/config.go
swift1337: I wrote this myself 😅 — sometimes I add these obvious things so I can glance at the code quickly without diving into every line.

I'll cut the number of comments and add your request to agents.md

swift1337:
@gjermundgaraba 

> Not sure if we've talked about this yet, and it could be a separate issue/pr, but having auto-migration is nice UX


We'll have auto-migrations for sure, but as part of CLI. It will be either `ibc relayer run ... --migrate` or the opposite `ibc relayer run --no-migrate` (we need to agree on the default behaviour)

link/internal/store/store_postgres.go
```diff
@@ -0,0 +1,140 @@
...
+	db := &PostgresDB{
+		pool:       pool,
+		sqlWrapper: stdlib.OpenDBFromPool(pool),
+		repo:       postgres.New(pool),
+		logger:     slog.With("module", "database"),
+	}
+
+	if ping {
+		if err := db.Ping(ctx); err != nil {
+			return nil, err
```
Eric-Warehime, line 68: This will leak the `pgxpool` when `Ping` fails.

link/internal/store/store.go
```diff
@@ -0,0 +1,89 @@
...
+// ValidateConfigLive assumes the config.Config is valid,
+// and checks if the database is reachable.
+func ValidateConfigLive(cfg config.Config) error {
+	// noop, don't create an empty sqlite db
+	if cfg.DB.Type == config.DBTypeSQLite {
+		return nil
+	}
+
+	// contains Ping()
+	_, err := NewStore(context.Background(), cfg)
```
Eric-Warehime: This leaks the `store.DB` without ever calling `Close`. Fine for now, but if this is ever called in a place where it doesn't trigger process to exit it will leak resources.

link/cmd/ibc/config.go
```diff
@@ -63,22 +63,26 @@ func configNew(_ *cobra.Command, _ []string) error {
...
 	}
 
-	_, err = config.LoadFromFile(configPath, true, flagConfigValidateStrict)
+	cfg, err := config.LoadFromFile(configPath, true, flagConfigValidateStrict)
 	if err != nil {
 		return errors.Wrap(err, "config load")
 	}
 
+	if flagConfigValidateLive {
+		if err := store.ValidateConfigLive(cfg); err != nil {
```
Eric-Warehime, line 72: Seems like this doesn't apply `--db` overrides since we're not calling `setupHomeWithConfig` I think for people that are permanently using that flag to invoke they will want to validate against the same parameters without having to serialize to a file.
For example, if your CLI uses variable expansion to inject a password or something then you'd want to validate against that no?

link/internal/config/config.go
```diff
@@ -93,3 +119,74 @@ func (c GRPCConfig) Validate() error {
...
+
+// DBConfigFromURL infers DB type from a CLI database URL override.
+func DBConfigFromURL(raw string) (DBConfig, error) {
+	if raw == "" {
+		return DBConfig{}, errors.New("empty db url")
+	}
+
+	if isSQLitePath(raw) {
+		return DBConfig{
+			Type: DBTypeSQLite,
```
dhfang: We shouldn't need to infer type from the path formatting since it is a config param

link/internal/config/config.go
```diff
@@ -93,3 +119,74 @@ func (c GRPCConfig) Validate() error {
...
+
+// Label returns a human-readable label for the DB config.
+func (c DBConfig) Label() string {
+	if c.Type != DBTypeSQLite {
+		return c.Type
+	}
+
+	// sqlite case
+	path := c.URL
+	if abs, err := filepath.Abs(path); err == nil {
```
dhfang, line 146: In the in-memory case where path = :memory: won't this return some meaningless path like /a/b/:memory:

link/internal/config/config.go
```diff
@@ -93,3 +119,74 @@ func (c GRPCConfig) Validate() error {
...
+
+	fmt.Println(string(bz))
+
+	return nil
+}
+
+// isSQLitePath reports whether raw should be treated as a sqlite file path
+// rather than a database connection URL.
+//
+// Anything without a "scheme://" is considered a local file path, e.g.
```
dhfang: Is this convention from a different project? I've seen some use `sqlite://` as scheme

link/cmd/ibc/migrate.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/ibc/link/internal/config"
+	"github.com/cosmos/ibc/link/internal/store"
+)
+
+var (
```
dhfang, line 12: Running the migrate command using in-memory sqlite mode creates a new in memory database that is tossed when the cli exits. So it's kind of a pointless operation that we should consider just rejecting with an error.

link/internal/store/store_sqlite.go
```diff
@@ -0,0 +1,176 @@
...
+		logger: logger,
+	}, nil
+}
+
+func (db *SqliteDB) Close() error {
+	return db.db.Close()
+}
+
+func (db *SqliteDB) Ping(_ context.Context) error {
+	return db.db.Ping()
```
dhfang: Do they not offer a ping call that takes a context?

swift1337: Ping is most likely a no-op call for SQLite since it's an in-process DB

link/internal/config/config.go
swift1337: I'll drop `:memory:` from the config. It should be used mostly for unit tests

link/internal/config/config.go
swift1337: This is what we get from the `--db` override, not from the config.

dhfang (approved):
lets gooo

link/internal/store/queries/relayer_submissions.sql
```diff
@@ -0,0 +1,9 @@
+-- name: GetRelaySubmission :one
+SELECT * FROM relay_submissions
+WHERE source_chain_id = sqlc.arg(chain_id)
+AND source_tx_hash = sqlc.arg(tx_hash);
+
+-- name: UpsertRelaySubmission :exec
+INSERT INTO relay_submissions (source_chain_id, source_tx_hash)
+VALUES (sqlc.arg(chain_id), sqlc.arg(tx_hash))
+ON CONFLICT (source_chain_id, source_tx_hash) DO NOTHING;
```
Eric-Warehime, line 9: maybe a nit but this isn't a true upsert. An upsert would replace existing entries and this could cause confusion about the behaviour if you're just reading the name of the function.

gjermundgaraba (approved):

Eric-Warehime (approved):
I don't see anything about cleaning up db entries yet--maybe that's intentional, but probably we want some delete or ttl or something.

link/internal/store/queries/relayer_submissions.sql
swift1337: Yeah, I copied this from go-relayer. The goal of this PR is to add the boilerplate; actual feature-related CRUDs will be in the follow-up PRs

PR #1257, opened 2026-07-06
feat(link): server wiring

swift1337:
- [x] Add protos 
- [x] Add proto codegen with ConnectRPC; Wire GRPC server
- [x] Add GRPC reflection
- [x] Add a single bootstrap entry point
- [x] Auto relayer migrations + `--no-migrate`
- [x] Graceful shutdown
- [x] Scaffold remote/local attestors
- [x] Basic test cases
- [x] AGENTS.md

**Notes**

1.  GRPC handlers have NO business logic and only proxy data to underlying services (`relayer.Service`, `attestor.Service`)
2. `attestor.Attestor` would also hide its implementation under `&LocalAttestor{}` or `&RemoteAttestor{}` 
3. ⛳ Adds the ability to run both relayer and attestor in-process!

Closes FOU-421


link/internal/pkg/graceful/graceful.go
gjermundgaraba, line 1: Seems worthwhile to have some tests for something like this?

link/internal/server/server.go
```diff
@@ -0,0 +1,110 @@
...
+	Name() string
+}
+
+var errInternal = connect.NewError(connect.CodeInternal, errors.New("internal server error"))
+
+// New Server constructor.
+func New(addr string, useReflection bool) *Server {
+	protocols := new(http.Protocols)
+	protocols.SetHTTP1(true)
+	protocols.SetUnencryptedHTTP2(true)
```
gjermundgaraba, line 41: I think this at the very least deserves a comment

gjermundgaraba (approved):
Just some minor stuff. Would love to get this merged soon. I have some work that I want to do on top of this

link/cmd/ibc/relayer.go
```diff
@@ -19,13 +21,43 @@ var (
...
 	}
 
-	fmt.Printf("Running relayer on addr %q...\n", cfg.GRPC.ListenAddress)
+	app, err := bootstrap.BuildRelayer(cfg)
+	if err != nil {
+		return err
+	}
+
+	if flagRelayerNoMigrate {
+		app.Logger.Info("--no-migrate flag passed, skipping migrations")
```
dhfang, line 38: Nit: We should fail if flagRelayerNoMigrate is supplied when using in-memory sqlite

link/internal/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,109 @@
...
+		Server:          srv,
+		Store:           db,
+		RelayerService:  relayerService,
+		AttestorService: nil,
+	}
+
+	// Dual mode: if .attestor config is provided, then we can run both relayer and attestor in the same process.
+	// This might be useful for PoC/testing environments or when an operator wants to run the relayer
+	// and one of attestors in the same process
+	if len(cfg.Attestor.Attestations) > 0 {
```
dhfang, line 59: Nit: I think we'll want to construct the attestor before the relayer when doing in-process. I'm assuming this BuildRelayer block will change meaningfully as we start the relayer implementation so no need to address in this PR.

link/internal/pkg/graceful/graceful.go
```diff
@@ -0,0 +1,119 @@
...
+
+var globalHandler *shutdownHandler
+
+func init() {
+	setupGlobalHandler()
+}
+
+func setupGlobalHandler() {
+	notify := make(chan os.Signal, 1)
+	signal.Notify(notify, syscall.SIGINT, syscall.SIGTERM)
```
dhfang, line 44: This seems like it would change the sigint sigterm behavior of other commands that don't listen to this notify channel like the migrate command. I think those commands will just ignore ctrl-c, which seems undesriable.

dhfang (approved):
Nice

link/cmd/ibc/relayer.go
swift1337: We already enforce that during config/flags parsing:

```bash
./bin/ibc relayer run --db ":memory:" 
Error: invalid --db: .url must not be :memory: for sqlite
```

link/internal/bootstrap/bootstrap.go
swift1337: I agree it might change. The goal is to have a single bootstrap.go with a super explicit wiring of the app

link/internal/pkg/graceful/graceful.go
swift1337: As long as we control the codebase, this will be the single OS signal handler for the entire process. I don't think third-party libraries rely on global signal handling, so everything should be fine.

dhfang: What I mean is that the migrate command will ignore ctrl-c now since it does not read off the notify channel created here unlike the attestor and relayer

link/cmd/ibc/relayer.go
dhfang: Hmm why did we decide to reject in memory sqlite as a run mode https://github.com/cosmos/ibc/blob/981953e8253b57fa71930c2ee7648470f3313486/link/internal/config/config.go#L155-L156?

link/internal/pkg/graceful/graceful.go
swift1337: ooh, that's a good catch

PR #1259, opened 2026-07-09
feat(link): integrate cosmos/kms

swift1337:
This PR adds local and remote signing keys for IBC Link

- [x] local ECDSA, EDDSA signers
- [x] remote KMS grpc signers
- [x] CLI: `ibc keys [new|show]`
- [x] Cross config attestors and signers validation
- [x]  Test cases, KMS grpc mocks

---

`Signer` is basically an interface that would be used by attestors (LC attestations) and relayers (relay tx). The caller doesn't know the exact implementation.

---

Example usage:

```bash
ibc keys new ecdsa foobar
{
  "keyPath": "/Users/dmitry/.ibc/keys/foobar.json",
  "keyType": "ecdsa",
  "publicKey": "0x022323efe4285529956bb504ab56ac2f3613f4354cffbf90f8aa9fcafb9bea3004"
}

ibc keys show foobar
{
  "keyPath": "/Users/dmitry/.ibc/keys/foobar.json",
  "keyType": "ecdsa",
  "publicKey": "0x022323efe4285529956bb504ab56ac2f3613f4354cffbf90f8aa9fcafb9bea3004"
}
```

```bash
cat ~/.ibc/ibc.yml
grpc:
  listenAddr: 0.0.0.0:3000

db:
  type: sqlite
  url: ibc.db

attestor:
  attestations:
    - chainId: 1
      name: eth-enjoyer
      signer: foobar

signers:
  - type: local
    alias: foobar
    file: foobar # auto resolves to keys/foorbar.json -> abs ibc path
```

Config validation:

```diff
- alias: foobar
+ alias: foobar2
```

```bash
ibc config validate
Using home: ~/.ibc
Error: setup home with config: validation failed: .attestor.attestations[0].signer references unknown signer: "foobar"
```

Closes FOU-384


link/cmd/ibc/keys.go
```diff
@@ -0,0 +1,116 @@
...
+		Use:   "new [type] [name]",
+		Short: "Create a new signing key",
+		Long:  "Saves key into <ibc-home>/keys/<name> or prints to stdout if name is not provided",
+		Args:  cobra.RangeArgs(1, 2),
+		RunE:  keysNew,
+	}
+
+	cmdKeysShow = &cobra.Command{
+		Use:   "show [name]",
+		Short: "Show key details",
```
dhfang, line 30: Should add a long description similar to the above that identifies the path the keys are assumed to be in.

link/internal/service/signer/signer.go
```diff
@@ -0,0 +1,78 @@
...
+
+// KeyType key type supported by local Signer
+type KeyType string
+
+// Signer represents signer that can either sign digests or messages.
+type Signer interface {
+	Type() KeyType
+	IsLocal() bool
+
+	Sign(ctx context.Context, message []byte) ([]byte, error)
```
dhfang, line 19: I'm not a huge fan of how there are different expectations on how `message` and the return bytes are formed depending on the underlying signer but I dont see a good way around it. Even if we simplify the ambiguity by making `message` always raw with the underlying implementations deriving any digests, the way the return bytes are formed is also implicit in the signer implementation.

Have we at least verified the local and remote secp signer are consistent in their expectations for whether message is a digest or raw?

link/internal/service/signer/remote.go
```diff
@@ -0,0 +1,129 @@
...
+	case signerservice.SignatureScheme_ED25519:
+		return EDDSA, nil
+	case signerservice.SignatureScheme_ECDSA_SECP256K1ETH:
+		return ECDSA, nil
+	default:
+		return "", errors.Errorf("unsupported remote key scheme: %s", scheme)
+	}
+}
+
+// todo: revisit security if needed. we can convert `signer.grpc string` to `signer.grpc{<options>}`
```
dhfang, line 118: lets track this in a ticket

dhfang (approved):
LGTM

link/cmd/ibc/keys.go
```diff
@@ -0,0 +1,116 @@
...
+	keyPath, err := signer.KeyFilePath(globalFlags.Home, args[1])
+	if err != nil {
+		return err
+	}
+
+	if err := key.StoreToFile(keyPath); err != nil {
+		return err
+	}
+
+	// note that we don't print the private key here to avoid leaking it to the user
```
gjermundgaraba, line 76: very nit, take it or ignore it as you prefer :)

```suggestion
	// note that we don't print the private key here to avoid it ending up in logs or terminal history
```

link/internal/config/config.go
```diff
@@ -188,24 +242,117 @@ func DBConfigFromURL(url string) (DBConfig, error) {
...
+		if err := fileExistsInAny(fallbacks...); err != nil {
+			return errors.Wrapf(err, ".file %s", path)
 		}
-		set[attestation.Name] = struct{}{}
 	}
 
 	return nil
 }
 
+func KeyFileFallbacks(keyPath string) []string {
```
gjermundgaraba, line 331: do we need this?

link/internal/service/signer/signer.go
swift1337: >Have we at least verified the local and remote secp signer are consistent in their expectations for whether message is a digest or raw?

They use the same KMS logic, yes.

> I'm not a huge fan of how there are different expectations on how message and the return bytes are formed depending on the underlying signer but I dont see a good way around it.

Agreed. I also raised this. @Eric-Warehime mentioned this is due to HSM internals (many can't do eth keccak)

@Eric-Warehime maybe we can update KMS and perform smth like `if isDigest(message) ? sign(digest) : sign(keccak(message))`?

link/internal/service/signer/remote.go
swift1337: FOU-549

link/internal/config/config.go
swift1337: Yes. `file:` path is resolved based on IBC home directory (defaults to `~/.ibc/`), but keys are resolved at either an absolute path (`/path/to/key.json`), or a relative path (`~/my-custom-keys/key.json`).

But when a user creates a key (`ibc key new ecdsa foobar`) it's not obvious that you have to specify the signer like  `file: keys/foobar.json` thus I added a fallback. 

`foobar` or `foobar.json` both resolve to `~/.ibc/keys/foobar.json`  => `/home/user/.ibc/keys/foobar.json`

gjermundgaraba: is this logic from the signer package or from us? just seems like it would be better to have one simple way to do this :)

PR #1267, opened 2026-07-10
[E2E harness 1/8] Restrict config environment expansion

gjermundgaraba:
Restricts environment expansion to explicit `${NAME}` references in `db.url`. Other fields and dollar-sign forms remain literal, and unset references now return an error.

Stack: 1/8.

Validation: `go test ./internal/config`.

### DIFF
--- a/link/internal/config/config.go
+++ b/link/internal/config/config.go
@@ -106,8 +106,8 @@ func DefaultConfig() Config {
 	}
 }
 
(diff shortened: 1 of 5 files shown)


gjermundgaraba:
Closing this layer: the existing `--db` whole-value override is sufficient. Environment-reference handling needed by the E2E harness will live in the stub layer instead.

PR #1268, opened 2026-07-10
add relayer config

dhfang:
Adds configs required for the Relayer. Defines a chains config at the top level of the config with configs shared by the relayer and attestor, and defines a relayer block with relayer-specific configs. 

Closes FOU-644

### DIFF
--- a/link/internal/config/config.go
+++ b/link/internal/config/config.go
@@ -132,7 +132,7 @@ func DefaultConfig() Config {
 		},
 		Chains: []ChainConfig{},
 		Relayer: RelayerConfig{
-			ChainOverrides: []ChainOverride{},
+			ChainOverrides: []RelayerChainOverride{},
 		},
 		Attestor: AttestorConfig{
 			Attestations: []AttestationConfig{},
--- a/link/internal/config/config.go
+++ b/link/internal/config/config.go
@@ -102,13 +102,21 @@ const (
 // ChainConfig chain information shared by the attestor and relayer.
 type ChainConfig struct {
 	ChainID string          `yaml:"chainId"`
-	Type    ChainType       `yaml:"type"`
-	RPC     string          `yaml:"rpc"`
 	EVM     *EVMChainConfig `yaml:"evm,omitempty"`
+}
+
+// Type returns the chain type implied by the configured settings.
+func (c ChainConfig) Type() ChainType {
+	if c.EVM != nil {
+		return ChainTypeEVM
+	}
+
+	return ""
 }
 
 // EVMChainConfig EVM-specific chain details.
 type EVMChainConfig struct {
+	RPC         string `yaml:"rpc"`
 	ICS26Router string `yaml:"ics26Router"`
 }
 
@@ -259,19 +267,12 @@ func (c ChainConfig) Validate() error {
 	switch {
 	case c.ChainID == "":
 		return errors.New(".chainId required")
-	case c.Type != ChainTypeEVM:
-		return errors.Errorf(".type unknown chain type: %q", c.Type)
-	case c.RPC == "":
-		return errors.New(".rpc required")
-	}
-
-	if c.Type == ChainTypeEVM {
-		switch {
-		case c.EVM == nil:
-			return errors.Errorf(".evm required for %s chains", ChainTypeEVM)
-		case c.EVM.ICS26Router == "":
-			return errors.New(".evm.ics26Router required")
-		}
+	case c.Type() != ChainTypeEVM:
+		return errors.Errorf("unknown chain type (only %s chains are supported)", ChainTypeEVM)
+	case c.EVM.RPC == "":
+		return errors.New(".evm.rpc required")
+	case c.EVM.ICS26Router == "":
+		return errors.New(".evm.ics26Router required")
 	}
 
 	return nil
--- a/link/internal/config/config.go
+++ b/link/internal/config/config.go
@@ -33,6 +33,8 @@ const sqliteInMemory = ":memory:"
 type Config struct {
 	Server   ServerConfig   `yaml:"server"`
 	DB       DBConfig       `yaml:"db"`
+	Chains   []ChainConfig  `yaml:"chains"`
+	Relayer  RelayerConfig  `yaml:"relayer"`
 	Attestor AttestorConfig `yaml:"attestor"`
 	Signers  Signers        `yaml:"signers"`
 }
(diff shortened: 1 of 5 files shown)


link/internal/config/config.go
```diff
@@ -89,6 +91,27 @@ type AttestationConfig struct {
...
+type ChainConfig struct {
+	ChainID string          `yaml:"chainId"`
+	Type    ChainType       `yaml:"type"`
+	RPC     string          `yaml:"rpc"`
+	EVM     *EVMChainConfig `yaml:"evm,omitempty"`
+}
+
+// EVMChainConfig EVM-specific chain details.
+type EVMChainConfig struct {
+	ICS26Router string `yaml:"ics26Router"`
```
dhfang, line 120: I moved ICS26 router to the top level since it is used by both the attestor and relayer

link/internal/config/ibc.yml
```diff
@@ -7,7 +7,67 @@ db:
...
+      txSubmissionDelay: 5s
+      packetBatchSize: 50
+
+  # attestors the relayer queries for packet and state attestations; they
+  # point to a configured client
+  attestors:
+    - name: "attestor-base"
+      type: remote
+      grpc: attestor.example.com:3000
+      client: "eth-to-base"
```
dhfang: This is the alias of a client configured in the clients block below

link/internal/config/config.go
```diff
@@ -177,6 +224,59 @@ func (c Config) crossValidate() error {
...
+		}
+	}
+
+	for _, client := range c.Relayer.Clients {
+		if _, ok := c.Chain(client.ChainID); client.ChainID != "" && !ok {
+			return errors.Errorf(
+				".clients[%s] chainId %q not declared in top-level chains",
+				client.ClientID, client.ChainID,
+			)
+		}
```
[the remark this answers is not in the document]
dhfang: you're wrong greptile, it passes client.ClientID, client.ChainID

dhfang:
@greptile re-review

link/internal/config/config.go
```diff
@@ -89,6 +91,27 @@ type AttestationConfig struct {
...
+	ChainTypeEVM ChainType = "evm"
+)
+
+// ChainConfig chain information shared by the attestor and relayer.
+type ChainConfig struct {
+	ChainID string          `yaml:"chainId"`
+	Type    ChainType       `yaml:"type"`
+	RPC     string          `yaml:"rpc"`
+	EVM     *EVMChainConfig `yaml:"evm,omitempty"`
+}
```
swift1337, line 106: Can we change the config to this structure and determine chain's type by checking `if cc.EVM != nil or cc.EVM != &EVMChainConfig{}`?

```suggestion
type ChainConfig struct {
	ChainID string          `yaml:"chainId"`
	EVM     *EVMChainConfig `yaml:"evm,omitempty"`
}
```

This results in a nice yml + we're can apply chain-specific/additional fields in EVMChain instead of assuming RPC is a string

before
```yaml
chains:
  - chainId: 1
    type: evm
    rpc: rpc.site.com
    evm:
      ics26Router: 0xABC
      ....
```


after
```yaml
chains:
  - chainId: 1
    evm:
      rpc: rpc.site.com
      ics26Router: 0xABC
      ....
```

link/internal/config/config.go
```diff
@@ -147,6 +174,26 @@ func (c Config) Validate() error {
...
+			return errors.Wrapf(err, ".chains[%s]", chain.ChainID)
+		}
+
+		if _, ok := chainIDs[chain.ChainID]; ok {
+			return errors.Wrapf(errors.Errorf("duplicate chainId: %q", chain.ChainID), ".chains")
+		}
+		chainIDs[chain.ChainID] = struct{}{}
+	}
+
+	if err := c.validateChainReferences(); err != nil {
```
swift1337: this should be called inside `c.crossValidate()`

link/internal/config/ibc.yml
```diff
@@ -7,7 +7,67 @@ db:
...
+      ics26Router: "0x0000000000000000000000000000000000000000"
+  - chainId: "8453"
+    type: evm
+    rpc: https://base-rpc.example.com
+    evm:
+      ics26Router: "0x0000000000000000000000000000000000000000"
+
+relayer:
+  # relay specific chain settings
+  chains:
```
swift1337: `$.relayer.chains` vs `$.chains` is confusing (i assume this is done before feedback on ADR-011)

link/internal/config/relayer.go
```diff
@@ -0,0 +1,390 @@
...
+const (
+	AttestorTypeRemote AttestorType = "remote"
+	AttestorTypeLocal  AttestorType = "local"
+)
+
+// RelayerConfig the relayer block of the config.
+type RelayerConfig struct {
+	Chains    []RelayerChainConfig `yaml:"chains"`
+	Clients   []ClientConfig       `yaml:"clients"`
+	Attestors []AttestorEntry      `yaml:"attestors"`
```
swift1337: Attestors are unique for each light client and chain; they can't be reused. I think it's confusing to reference them one-to-one

link/internal/config/relayer_test.go
```diff
@@ -0,0 +1,474 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const fullRelayerConfig = `
```
swift1337: let's move that to `./testdata/sample.yml`

link/internal/config/relayer.go
gjermundgaraba: But not all client or all chains will have attestors.

link/internal/config/config.go
gjermundgaraba: what if you have three chain types? nil checks would get frustrating

link/internal/config/ibc.yml
dhfang: hmm well there are still chain level configurations that are relayer specific. if you do a route first configuration where each entry is a one way route, you get duplicate chain level configs since you need two route entries per connection

dhfang: I think renaming this to `chainOverrides` resolves the confusion - we'll set sensible defaults people can override using the `chainOverrides` section

dhfang:
@greptile re-review

link/internal/config/relayer.go
dhfang: It's kind of a wash for me, there's downsides to both. I updated this to move the attestor entries back under the `attestorSet` configuration, mainly because I think the extra level of indirection the `attestors` block creates and also having an `attestors` block at the relayer config level in addition to the top level block is a bit more confusing then one additional level of nesting

link/internal/config/config.go
dhfang: I removed the `type` field and added a `Type()` helper which infers the type based on the config

PR #1277, opened 2026-07-17
Add Relaying Logic

dhfang:
Ports the packet relaying logic from cosmos/ibc-relayer into link

### Structure

internal/relay/
--  processors/   each processor handles one step of the relaying pipeline
--  pipeline/       defines a pipeline struct which chains the processors together to complete a relay
--  dispatch/     reads unfinished transfers from the db and pushes them onto the pipeline for relaying
--  proofapi/     connect client for the proof api (vendored proto)
internal/txmgr/   defines TxManager which handles landing transactions on chain
internal/chains/  added chain queries

### Reviewer Considerations
- The logic added in internal/relay and internal/chains is copied mostly verbatim from cosmos/relayer with some removals and reorganization of the relay files into more coherent packages.
- On further consideration, the pipeline abstraction is a bit too complex and can be replaced with something simpler. The pipeline processes relays in a series of sequential stages but the packet lifecycle actually does not map perfectly onto a linear set of stages. Alot of the complex logic in internal/relay/pipeline exists to encode rules for when a packet should skip steps in the pipeline. I'd like to land it as is now so we can get something working but we should be able to create a simpler relaying abstraction that consolidates the stages defined by the processors into a much smaller number of asynchronous workers (probably just one for recv, one for ack, and one for timeout) and then replaces the pipeline with some worker that just triages what stage a packet is at and forwards it to the appropriate worker.
- The TxManager abstraction is new - it consolidates tx submission logic that was spread across a different places in the old relayer. Changes we should make to it in follow-up work:
  - Nonce selection is based on whatever nonce the node returns. When submitting multiple txs within a single block, the relayer does not have the ability to calculate the sequential nonces for the transactions in order to land them simultaneously. We currently insert an artificial delay between tx submission attempts to avoid multiple transactions getting assigned the same nonce and failing on submission. We should add logic to choose the nonces more intelligently and remove the delay.
  - It contains some mechanisms for landing transactions reliably but currently relies on the caller to handle tx resubmission if required. Other similar abstractions like those in the chainlink and op relayer handle bumping the gas price and resubmission, which we should extend ours to do as well.


dhfang:
@greptile re-review

dhfang:
@greptile re-review

dhfang:
@greptile re-review

link/internal/bootstrap/bootstrap.go
```diff
@@ -59,13 +66,32 @@ func BuildRelayer(cfg config.Config) (*Services, error) {
...
+	}
+
 	services := &Services{
 		Context:         ctx,
 		Logger:          logger,
 		Server:          srv,
 		Store:           db,
 		Signers:         signers,
 		RelayerService:  relayerService,
+		Dispatcher:      dispatcher,
```
gjermundgaraba: not crucial in this pr, but one architecutural improvement i want use to make is trying to avoid separate services for things that are directly coupled like the dispatcher and the relayer service. but we'll discuss this later :)

link/internal/chains/evm/client.go
```diff
@@ -11,38 +14,50 @@ import (
...
 	"github.com/pkg/errors"
 
 	"github.com/cosmos/ibc/link/internal/chains/evm/contracts/ics26router"
 
 	v2 "github.com/cosmos/ibc/link/internal/types/v2"
+	ethereum "github.com/ethereum/go-ethereum"
 )
 
-const sendPacketEvent = "SendPacket"
+// ics26 router events consumed by the client
```
gjermundgaraba, line 29: just leaving as a note for follow-up when https://github.com/cosmos/ibc/pull/1281 is merged that we use ibc-go canonicals instead

link/internal/relay/dispatch/manager.go
```diff
@@ -0,0 +1,109 @@
...
+	logger *slog.Logger
+	cfg    config.Config
+	deps   pipeline.Deps
+
+	mu        sync.Mutex
+	pipelines map[processors.Route]pipeline.TransferPipeline
+}
+
+// RouteManager identifies the pipeline a transfer is relayed through.
+type RouteManager interface {
```
gjermundgaraba: i see that we do this quite a few places in the code, but we should avoid creating these interfaces for everything next to the concrete types. this is a consumer-side concern. not critical to fix before merging, but there is some cleanup we should be doing around this code later on.

link/internal/relay/dispatch/manager.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/pkg/errors"
+
+	"github.com/cosmos/ibc/link/internal/config"
+	"github.com/cosmos/ibc/link/internal/relay/pipeline"
+	"github.com/cosmos/ibc/link/internal/relay/processors"
+)
+
+// Manager creates and caches one pipeline per route.
+type Manager struct {
```
gjermundgaraba: we need to ban the name "Manager". it doesnt mean anything or give me any clue what is going on here. this is dispatch.Manager, but is is managing dispatches? or routes? or pipelines? it doesnt help answer why this is separate from the dispatcher itself either.

link/internal/relay/pipeline/batch_processor.go
gjermundgaraba, line 1: nit, but i dont see why this is called _conditionally_ batch processor. why not just batch processor? the fact that is batches "shouldProcess" doesn't seem to warrant the verbose name. as far as i can tell, there is no unconditional batch processor :P

link/internal/relay/pipeline/pipeline.go
```diff
@@ -0,0 +1,240 @@
+package pipeline
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/deliveryhero/pipeline/v2"
```
gjermundgaraba, line 7: not for now, but not sure we want to use this dep? it was last updated 2 years ago and isn't widely used.

gjermundgaraba (approved):
nothing blocking, but i do think we should try to get to a cleaner architecture to make this code more maintainable and easier to understand.

link/cmd/ibc/relayer.go
```diff
@@ -56,6 +58,22 @@ func relayerRun(_ *cobra.Command, _ []string) error {
 	}
 
 	graceful.AddCallback(app.Store.Close)
+
+	if app.Dispatcher != nil {
+		dispatcherCtx, cancelDispatcher := context.WithCancel(context.Background())
+
+		go func() {
+			if err := app.Dispatcher.Run(dispatcherCtx); err != nil && !errors.Is(err, context.Canceled) {
```
swift1337: It should have same Start() and Close() semantics as grpc.server

https://github.com/cosmos/ibc/blob/70272b86c3687bc4a70774c231e0fc1601390718/link/internal/server/server.go#L58

link/internal/bootstrap/bootstrap.go
swift1337: I agree that the dispatcher is confusing; it does what the relayer service is supposed to do.

link/internal/bootstrap/bootstrap.go
```diff
@@ -59,13 +66,32 @@ func BuildRelayer(cfg config.Config) (*Services, error) {
 	srv := server.New(cfg.Server.ListenAddress, true)
 	srv.Register(relayerHandler)
 
+	// Relaying dispatcher
+	var dispatcher *dispatch.RelayDispatcher
+	if len(cfg.Relayer.Routes) > 0 {
```
swift1337: We should drop this check and fail the process if `len(routes ) == 0`

link/internal/chains/client.go
```diff
@@ -14,7 +15,56 @@ import (
...
+
+	// IsTimestampFinalized reports whether the chain has finalized a block at
+	// or after the timestamp. A nil finalityOffset compares against the
+	// finalized head; otherwise against the block finalityOffset blocks behind
+	// the latest.
+	IsTimestampFinalized(ctx context.Context, timestamp time.Time, finalityOffset *uint64) (bool, error)
+
+	// WaitForChain blocks until the chain's latest block time catches up to
+	// the current time.
+	WaitForChain(ctx context.Context) error
```
swift1337, line 60: Why do we need this?

proto/proofapi/proofapi.proto
```diff
@@ -0,0 +1,120 @@
...
+// vendored from skip-mev/relayer proto/eurekarelayer/relayer.proto (renamed
+// to proofapi.proto so the registered filename does not collide with
+// link/relayer.proto); the
+// go_package is rewritten to generate into this repo
+option go_package = "github.com/cosmos/ibc/link/internal/types/proofapi";
+
+// The proof API service definition.
+// Each proof API service is defined between two chains.
+// One chain is the source chain, and the other chain is the target chain.
+service ProofApiService {
```
swift1337: ~~Can you explain why we have proof api protos in the first place?~~

link/internal/store/store.go
```diff
@@ -36,6 +36,43 @@ type Repository interface {
...
+
+	UpdatePacketAckTx(ctx context.Context, key PacketKey, tx PacketTx) error
+	ClearPacketAckTx(ctx context.Context, key PacketKey) error
+
+	UpdatePacketTimeoutTx(ctx context.Context, key PacketKey, tx PacketTx) error
+	ClearPacketTimeoutTx(ctx context.Context, key PacketKey) error
+}
+
+// PacketKey uniquely identifies a packet.
+type PacketKey struct {
```
swift1337, line 58: nice

link/internal/relay/proofapi/client.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/cosmos/ibc/link/internal/config"
+
+	proto "github.com/cosmos/ibc/link/internal/types/proofapi"
+)
+
+// relay transactions carry proofs and can exceed default message limits
+const maxResponseBytes = 10 * 1024 * 1024
+
+// NewClient creates a proof api client speaking gRPC.
+func NewClient(cfg config.ProofAPIConfig) proto.ProofApiServiceClient {
```
swift1337: Same Q regarding the separate proof API

swift1337 (changes requested):
Reviewed two-thirds. Requesting changes to make sure we don't merge proof API protos (I assume it here for platform e2e tests)

link/internal/relay/pipeline/batch_processor.go
dhfang: The context here is the library has a batch processor but it didn't skip processing in certain conditions where we wanted it to. I've renamed it though

link/internal/relay/dispatch/manager.go
dhfang: I banished all managers to the shadow realm

link/internal/relay/dispatch/manager.go
dhfang: will create a ticket for this

link/internal/chains/client.go
dhfang: For certain light clients, proofs contain a timestamp and the light client compares the proof's timestamp against the chain's. If we don't wait for the chain to catch up to the timestamp in the proof, the tx fails during simulation.

proto/proofapi/proofapi.proto
dhfang: Just an intermediate step before the proof api port.

link/internal/bootstrap/bootstrap.go
```diff
@@ -31,6 +38,10 @@ func BuildRelayer(cfg config.Config) (*Services, error) {
 	ctx := context.Background()
 	logger := slog.With("module", "bootstrap")
 
+	if len(cfg.Relayer.Routes) == 0 {
```
swift1337: this should be in config.Validate()

swift1337 (approved):

PR #1280, opened 2026-07-22
feat(link): implement attestor

swift1337:
- [x] In-process attestor
- [x] Remote attestor
- [x] Port encoding/cryptography from Solidity / ibc-attestor (Rust)
- [x] Unit tests  

The PR ensures that all existing components are seamlessly integrated (Signer x Attestor x Client) while maintaining domain responsibilities.

Closes FOU-502


link/internal/chains/evm/client.go
```diff
@@ -24,19 +25,24 @@ const sendPacketEvent = "SendPacket"
...
-	eth           ETHClient
-	router        *ics26router.ContractFilterer
-	routerABI     *abi.ABI
-	logger        *slog.Logger
+	chainID string
+
+	eth ETHClient
+
+	routerAddress  common.Address
+	routerFilterer *ics26router.ContractFilterer
```
dhfang: Should be able to get rid of this and just use the router field

dhfang (commented):
LGTM!

gjermundgaraba (approved):
as long as we use ibc-go canonicals in this follow-up pr, i think we're good: https://github.com/cosmos/ibc/pull/1281

PR #1282, opened 2026-07-24
Feat/relayer attestation proofgen

dhfang:
Removes the relayer's dependency on the external Rust proof-api service for building relay transactions, replacing it with in-process proof generation against this repo's own attestor.

### Changes

- Removed anything related to the proof api
- `internal/relay/proofgen`: Adds a `ProofGenerator` abstraction that takes packet events and returns state and packet proofs. There will be a different `ProofGenerator` implementation for each light client implementation the relayer supports. The PR adds an implementation for the attestation light client under `internal/relay/proofgen/attestation`
- `internal/relay/txbuilder`: Adds a TxBuilder abstraction that takes packet and state proofs and forms relaying transactions. There will be a different `TxBuilder` implementation for each chain type. The PR adds an EVM implementation that packs a client update and a batch of packet operations into a single `ICS26Router.multicall` transaction.
- Updated the three processors that submit relay transactions (`batch_ack_packet.go`, `batch_recv_packet.go`, `timeout_packet.go`) onto the new proof generation / tx creation flow. They no longer call the proof api to form relay transactions and leverage `ProofGenerator` and `TxBuilder` instead. The shared logic is in `batch_relay.go`.


link/internal/relay/processors/batch_relay.go
```diff
@@ -0,0 +1,120 @@
...
+
+	submission, err := txSubmitter.Submit(ctx, v2.TxIntent{
+		To:   common.BytesToAddress(relayTx.To).Hex(),
+		Data: relayTx.Data,
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "submitting relay tx")
+	}
+
+	return submission, nil
```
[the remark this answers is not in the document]
dhfang: This is not relevant right now since solana is not supported. It will break in obvious ways when adding solana support.

link/internal/relay/proofgen/proofgen.go
```diff
@@ -0,0 +1,137 @@
...
+	localAttestors *attestor.Service,
+) (attestor.Attestor, error) {
+	switch entry.Type {
+	case config.AttestorTypeRemote:
+		if entry.GRPC == "" {
+			return nil, errors.New("no grpc address configured")
+		}
+
+		// TODO: Support TLS
+		return attestor.NewRemoteFromURL(watchedChainID, entry.Name, entry.Name, "http://"+entry.GRPC), nil
```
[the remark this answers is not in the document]
dhfang: Future work.

link/internal/relay/proofgen/proofgen.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+
+			counterpartyChain, ok := clientSet.Get(clientCfg.CounterpartyChainID)
+			if !ok {
+				return nil, errors.Errorf(
+					"client %q: no configured chain client for counterparty chain %q", clientCfg.Alias, clientCfg.CounterpartyChainID,
+				)
+			}
+
+			generators[Key(clientCfg.ChainID, clientCfg.ClientID)] = attestation.New(attestors, clientCfg.AttestorSet.Threshold, counterpartyChain)
```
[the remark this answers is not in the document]
dhfang: There is a step in teh relaying pipeline that uses this config value to determine how long to wait for finality. Moreover, we always check first what the latest provable height is which queries the attestors directly for what they consider a finalized height.

dhfang:
@greptile re-review

link/internal/relay/txbuilder/evm/evm.go
```diff
@@ -0,0 +1,189 @@
...
+		if len(item.Acks) == 0 {
+			return nil, errors.Errorf("no acknowledgement recorded for sequence %d", item.Packet.Sequence)
+		}
+
+		return packAckPacket(packet, item.Acks[0], item.Proof, item.ProofHeight)
+	case v2.RelayKindTimeout:
+		return packTimeoutPacket(packet, item.Proof, item.ProofHeight)
+	default:
+		return nil, errors.Errorf("unsupported relay kind %v", item.Kind)
+	}
```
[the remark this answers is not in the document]
dhfang: the solidity implementation does not currently support multi-payload

link/internal/relay/proofgen/attestation/signing.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// Domain separation tags, matching the attestor's own (see
+// service/attestor/evm.SignABI).
+const (
+	TagStateAttestation  byte = 0x01
```
swift1337: let's dedup and use attestor.evm package directly

swift1337: UPD: Let's move all of this to link/internal/service/attestor/evm/signing.go

link/internal/relay/proofgen/attestation/quorum.go
```diff
@@ -0,0 +1,251 @@
...
+		buckets[string(resp.data)] = append(buckets[string(resp.data)], resp)
+	}
+
+	for _, bucket := range buckets {
+		seenSigners := make(map[common.Address]struct{})
+
+		var signatures [][]byte
+
+		for _, resp := range bucket {
+			if _, dup := seenSigners[resp.signer]; dup {
```
swift1337, line 143: It should not be possible to have responses from duplicate signers. Validation logic should fail on this even before making the call

link/internal/relay/proofgen/attestation/quorum.go
```diff
@@ -0,0 +1,251 @@
...
+		return quorumResponse{name: a.Name(), err: errors.Wrapf(err, "attestor %q", a.Name())}
+	}
+
+	return quorumResponse{name: a.Name(), signer: signer, data: data, sig: sig}
+}
+
+// reduceQuorum groups responses by their exact attestationData value
+// and returns the first value whose distinct signers reach threshold.
+func reduceQuorum(responses []quorumResponse, threshold int) (quorumResult, error) {
+	buckets := make(map[string][]quorumResponse)
```
swift1337, line 127: Can you explain what a bucket is here and why we need it?

link/internal/relay/proofgen/attestation/quorum.go
```diff
@@ -0,0 +1,251 @@
...
+		for _, resp := range bucket {
+			if _, dup := seenSigners[resp.signer]; dup {
+				continue
+			}
+
+			seenSigners[resp.signer] = struct{}{}
+			signatures = append(signatures, resp.sig)
+		}
+
+		if len(signatures) >= threshold {
```
swift1337, line 151: Is that how the existing ibc-attestor works? Please help me to clarify:

This seems to drop real signatures from live attestors after a threshold. Let's say we have a quorum of 5/8; this means that three attestors would not be present in the on-chain verification. I think it's good to have all attestors for liveness, i.e., third-party observers can verify that all attestors participate instead of only a subset. (I know that Sol LC also does a fast exit, and we don't spend gas verifying signatures after a threshold).

link/internal/relay/proofgen/proofgen.go
```diff
@@ -0,0 +1,153 @@
...
+	localAttestors *attestor.Service,
+) (attestor.Attestor, error) {
+	switch entry.Type {
+	case config.AttestorTypeRemote:
+		if entry.GRPC == "" {
+			return nil, errors.New("no grpc address configured")
+		}
+
+		// TODO: Support TLS
+		return attestor.NewRemoteFromURL(watchedChainID, entry.Name, entry.Name, "http://"+entry.GRPC), nil
```
swift1337, line 138: (not blocking) Please add `todo` to merge this into the attestor service; it's a bit confusing that proofgen creates an attestor instance. We could take the route config, resolve attestors, then use the local attestors and call something like:
```go
routeAttestorService := attestorService.Merge(routeAttestors...)

proofGen := NewFromConfig(cfg, attestorService, clientSet)
```

link/internal/relay/proofgen/attestation/generator.go
```diff
@@ -0,0 +1,142 @@
...
+
+func New(attestors []attestor.Attestor, threshold int, counterpartyChain chains.Client) *Generator {
+	return &Generator{attestors: attestors, threshold: threshold, counterpartyChain: counterpartyChain}
+}
+
+func (g *Generator) LatestProvableHeight(ctx context.Context) (uint64, time.Time, error) {
+	return latestProvableHeight(ctx, g.attestors, g.threshold, g.counterpartyChain)
+}
+
+func (g *Generator) StateProof(ctx context.Context, height uint64) ([]byte, error) {
```
swift1337, line 34: This seems identical to Attestor, we could move this to attestorService in the future, i'd place a `todo`

dhfang: I don't think so, this is an interface defined in proofgen.go that all proof generators will implement. This is just the implementation for the attestation light client and it calls attestor aggregation logic that doesn't belong in the attestor service.

link/internal/relay/proofgen/attestation/quorum.go
dhfang: This is a recovered signer from the attestor response. The config validation doesn't actually tell you anything about what signing key an attestor is actually configured with.

link/internal/relay/proofgen/attestation/quorum.go
dhfang: I don't think I'm going to change this:
- Relaying is permissionless so you don't actually have guarantees around what inclusion rules a relayer uses to decide which attestor's signatures to submit
- The solidity light client verifies ALL attestor signatures
- I'd actually want to go in the opposite direction and stop listening for a response from any remaining attestors after a quorum of responses has been received.

link/internal/relay/proofgen/attestation/quorum.go
dhfang: The attestors might not return the same attested data i.e. if you have a misbehaving attestor for example. This just buckets the responses by the attested data

link/internal/relay/proofgen/proofgen.go
dhfang: will create a ticket

swift1337 (approved):

PR #1284, opened 2026-07-28
refactor(link): publish the relayer readiness contract in api/v2

gjermundgaraba:
Moves the relayer's stdout readiness event from `cmd/relayercmd` to `api/v2/relayer`, beside the attestor's readiness type, as a plain data type validated by its consumer. The e2e harness parses the shared type directly; the readiness-only `relayercmd` test moves with the contract. `relayercmd` keeps only the acceptance-test packet vocabulary, which the next PR in the stack removes.

Second PR of a 3-PR stack (base: #1283):
1. harness simplification (#1283)
2. **this** — readiness contract → `api/v2`
3. packet outcome states end to end

🤖 Generated with [Claude Code](https://claude.com/claude-code)


link/cmd/ibc/relayer.go
```diff
@@ -6,9 +6,10 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
-	"github.com/cosmos/ibc/link/cmd/relayercmd"
 	"github.com/cosmos/ibc/link/internal/bootstrap"
 	"github.com/cosmos/ibc/link/internal/pkg/graceful"
+
+	relayerv2 "github.com/cosmos/ibc/link/api/v2/relayer"
```
swift1337, line 12:
```suggestion
	"github.com/cosmos/ibc/link/api/v2/relayer"
```

e2e/internal/harness/ibclink/daemon.go
```diff
@@ -45,7 +45,7 @@ type RelayerOptions struct {
 }
 
 type Relayer struct {
-	readiness    relayercmd.Readiness
+	readiness    relayerv2.ProcessReadiness
```
swift1337, line 48:
```suggestion
	readiness    relayer.ProcessReadiness
```

swift1337 (approved):

dhfang:
Could you give some context on the purpose of this readiness related logic and why we don't add and use a health check endpoint instead?

gjermundgaraba:
> Could you give some context on the purpose of this readiness related logic and why we don't add and use a health check endpoint instead?

it is also about address discovery and enabling better probing of what is running.

link/cmd/ibc/relayer.go
gjermundgaraba: we could, but this currently is consistent with how attestorv2 is used. if you want this changed, i'd be happy to open a separate pr for that

e2e/internal/harness/ibclink/daemon.go
gjermundgaraba: see previous comment

dhfang (approved):

dianab-cl (approved):

PR #1285, opened 2026-07-28
feat(proto): report packet outcome states end to end

gjermundgaraba:
Replaces the collapsed `TransferState` with `PacketState` outcome values: `SUCCEEDED`, `TIMED_OUT`, `REJECTED`, and `RELAY_FAILED` travel the wire instead of being folded into `COMPLETE` with a timeout-tx pointer and a `write_ack_error` side channel. The service maps the store's terminal relay statuses onto the enum directly, and request fields are renamed to `source_chain_id`.

On the e2e side this deletes the shadow packet vocabulary (the remaining `relayercmd` types): tests assert on the wire enum, which removes the lossy re-derivation of outcomes. Verified end to end by the IFT-timeout and GMP-error-ack anvil tests.

Third PR of a 3-PR stack (base: #1284):
1. harness simplification (#1283)
2. readiness contract → `api/v2` (#1284)
3. **this** — packet outcome states end to end

🤖 Generated with [Claude Code](https://claude.com/claude-code)


proto/link/relayer.proto
```diff
@@ -41,14 +51,19 @@ message TransactionInfo {
...
-  TransferState state = 1;
+  PacketState state = 1;
+  // Together with source_client_id, identifies the packet.
   uint64 sequence_number = 2;
+  // Together with sequence_number, identifies the packet.
   string source_client_id = 3;
+  // The source-chain SendPacket transaction. Always present.
   TransactionInfo send_tx = 4;
+  // The destination-chain receive transaction, once submitted or discovered.
   TransactionInfo recv_tx = 5;
```
swift1337, line 62:
```suggestion
  TransactionInfo destination_tx = 5;
```

proto/link/relayer.proto
```diff
@@ -5,34 +5,44 @@ package ibc.v2.relayer;
...
+  repeated PacketStatus packet_statuses = 1;
+}
 
-enum TransferState {
-  TRANSFER_STATE_UNKNOWN = 0;
-  TRANSFER_STATE_PENDING = 1;
-  TRANSFER_STATE_COMPLETE = 2;
-  TRANSFER_STATE_FAILED = 3;
+enum PacketState {
+  PACKET_STATE_UNSPECIFIED = 0;
```
swift1337, line 34:
```suggestion
  PACKET_STATE_INVALID = 0;
```

proto/link/relayer.proto
gjermundgaraba: hmm, im not sure this would be more accurate. it's not incorrect, but recv is the name of the ibc lifecycle transaction, similar to send and the others in this list.

proto/link/relayer.proto
gjermundgaraba: this changes the meaning. `UNSPECIFIED` is the official protobuf style guide name and what we use all over cosmos sdk already too.

link/internal/service/relayer/service.go
```diff
@@ -261,12 +262,14 @@ func (s *Service) validateRelayArgs(chainID, txHash string) (string, error) {
...
 	case store.RelayStatusCompleteWithAck,
-		store.RelayStatusCompleteWithTimeout,
-		store.RelayStatusCompleteWithWriteAckSuccess,
-		store.RelayStatusCompleteWithWriteAckError:
-		return StateComplete
+		store.RelayStatusCompleteWithWriteAckSuccess:
+		return StateSucceeded
+	case store.RelayStatusCompleteWithTimeout:
+		return StateTimedOut
+	case store.RelayStatusCompleteWithWriteAckError:
```
dhfang, line 268: I think this state is actually dead code now. `RelayStatusCompleteWithWriteAckError` and `RelayStatusCompleteWithWriteAckSuccess` actually represent a terminal state that used to be possible where the relayer didn't relay the ack. We had config values that set whether success and/or error acks were relayed to save gas when relaying from ethereum. When the acknowledgement is actually relayed, the relayer's internal states dont distinguish between error or success right now, it just has `RelayStatusCompleteWithAck`. We'd have to also check the `write_ack_status` to determine if it's a success or error ack.

gjermundgaraba: CompleteWithWriteAckSuccess was ded. CompleteWithWriteAckError is still in use. so fixed what could be fixed there.

dhfang: Ahh I didn't realize you'd wired up CompleteWithWRiteAckError like that in a previous pr. Thanks for cleaning up CompleteWIthWriteAckSuccess as well.

dhfang (approved):

PR #1288, opened 2026-07-29
feat(link): add docker & artifact builds

swift1337:
- [x] build & publish binaries as CI artifacts (mac arm, linux amd64, linux arm64)
- [x] build & publush docker images (linux amd64, arm64) 
- [x] triggers: manual, tag push, or PR merge with `build-artifacts` label
- [x] add `make docker-build`

### DIFF
--- a/.github/workflows/ibc-link-build.yml
+++ b/.github/workflows/ibc-link-build.yml
@@ -1,0 +1,167 @@
+name: Link Build Artifacts
+
+on:
+  workflow_dispatch:
+  push:
+    tags: ["*"]
+  pull_request_target:
+    # PR must contain `build-artifacts` label
+    types: [closed]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.base.ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  IMAGE_NAME: ghcr.io/${{ github.repository }}
+  BUILD_SHA: ${{ github.event.pull_request.merge_commit_sha || github.sha }}
+  GO_BUILD_FLAGS: -trimpath -ldflags="-s -w"
+
+jobs:
+  # Build binaries
+  build:
+    name: Build ${{ matrix.goos }}-${{ matrix.goarch }} binary
+    if: >-
+      github.event_name == 'workflow_dispatch' || github.event_name == 'push' ||
+      (github.event.pull_request.merged == true &&
+      contains(github.event.pull_request.labels.*.name, 'build-artifacts'))
+    runs-on: depot-ubuntu-24.04-4
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - goos: darwin
+            goarch: arm64
+            artifact: ibc-darwin-arm64
+          - goos: linux
+            goarch: amd64
+            artifact: ibc-linux-amd64
+          - goos: linux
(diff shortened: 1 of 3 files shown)


.github/workflows/ibc-link-build.yml
```diff
@@ -0,0 +1,167 @@
...
+        run: chmod +x image/ibc
+      - uses: docker/setup-buildx-action@v3
+      - name: Log in to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build & Push
+        uses: docker/build-push-action@v6
```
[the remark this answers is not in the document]
swift1337: we want to maintain `ibc:development` image that is always up to date.

.github/workflows/ibc-link-build.yml
```diff
@@ -0,0 +1,167 @@
...
+
+jobs:
+  # Build binaries
+  build:
+    name: Build ${{ matrix.goos }}-${{ matrix.goarch }} binary
+    if: >-
+      github.event_name == 'workflow_dispatch' || github.event_name == 'push' ||
+      (github.event.pull_request.merged == true &&
+      contains(github.event.pull_request.labels.*.name, 'build-artifacts'))
+    runs-on: depot-ubuntu-24.04-4
```
dianab-cl, line 32: Why is depot needed ?

dianab-cl: no need, see example https://github.com/cosmos/ibc-attestor/blob/main/.github/workflows/docker-publish.yml

swift1337: Binary builds happen on Depot to speed up the CI, so I placed them everywhere for cache locality.

dianab-cl: I'm not sure if we want to go that route. 
The idea was to decouple all releases and docker builds from depot. 
Especially for public repositories where github runner minutes are free.
No depot runner minute is free.

gjermundgaraba: it depends on how long it takes to run. we def want to keep CI times down, but it depends on where the bottlenecks are of course. the longest job is always the bottleneck.

.github/workflows/ibc-link-build.yml
gjermundgaraba, line 1: I think this is more complex than it needs to be. Take a look at the platform workflow in `.github/workflows/backend-docker.yml`. I think it will be simpler and faster to run too.

link/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+FROM golang:1.26.4-alpine AS builder
+
+ARG LDFLAGS="-s -w"
+
+WORKDIR /src
+
+COPY go.mod go.sum ./
+RUN --mount=type=cache,target=/go/pkg/mod go mod download
+
+COPY . .
```
gjermundgaraba, line 13: we need a `link/.dockerignore` if we do this

dianab-cl (approved):

.github/workflows/ibc-link-build.yml
swift1337: It doesn't build and publish the binaries as artifacts, whereas this CI action does both binaries and images

link/Dockerfile
swift1337: What would you include in this file? `.git` is already in the parent directory

PR #1289, opened 2026-07-29
feat(link): QoL improvements

swift1337:
- [x] add option for json logging
- [x] handle `err` error log field
- [x] `ibc keys import [type] [name] --private-key ...`
- [x] keys: show evm address, show base64 private key


link/internal/service/signer/signer.go
```diff
@@ -70,3 +73,16 @@ func NewSignerFromConfig(ctx context.Context, cfg config.SignerConfig) (signer S
 func ParseKeyType(raw string) (keyfile.Type, error) {
 	return keyfile.ParseType(raw)
 }
+
+func PublicKeyToEVMAddress(pk []byte) string {
+	publicKey, err := crypto.DecompressPubkey(pk)
+	if err != nil {
+		return ""
```
dhfang: Why do we fail silently here?

dhfang (approved):
LGTM

PR #1292, opened 2026-07-31
feat(e2e): support relayer config overrides

gjermundgaraba:
## Summary

This gives downstream attestor, quorum, batching, and remote-signer E2E tests a supported configuration path instead of per-test YAML workarounds.

- expose per-client local and remote attestor sets with configurable quorum
- expose per-chain packet batch size and timeout overrides
- support local and remote relayer signers while preserving the default generated config
- surface the relayer's startup log tail in `StartRelayer` errors, so link's own config validation explains misconfigurations in the test failure instead of dying silently with the TempDir; the harness deliberately does not duplicate link's validation
- expose `Endpoint()`, `LatestHeight()`, `Stop()`, and `Restart()` on `environment.Attestor`; restart keeps the attestor name, signer key, and endpoint while using a fresh work directory (FOU-1075)

## Tests

- `make test-harness` (includes the attestor endpoint/lifecycle integration checks and the startup-log surfacing check)
- `make test-unit`
- `make test-e2e`
- `make lint-e2e`

## Linear

Closes FOU-1122, FOU-1075


dhfang (approved):
Curious about the`TestTransfer_RemoteAttestor` test, but the changes in the diff LGTM

PR #1293, opened 2026-08-03
test(e2e): add IFT error-ack and multi-transfer relay coverage

dhfang:
## Summary
Adds three new tests:
- `TestIFTTransfer_ErrorAck_Refund` — zero-address receiver forces the destination mint to revert; asserts error ack, zero mint, full refund, and IFT tranfer pending record cleared.
- `TestIFTTransfer_ErrorAck_UnregisteredBridge` — unregistered destination bridge hits `IFTBridgeNotFound`; asserts error ack, zero mint, full refund, and IFT tranfer pending record cleared.
- `TestIFTTransfer_MultiPacketPending` — sends 3 transfers before any are relayed, asserts one pending record per sequence with correct sender/amount, then relays and asserts each record is cleared by its own ack with exact per-receiver amounts.


e2e/ift_test.go
```diff
@@ -72,4 +79,105 @@ func TestIFTTimeout_Refund(t *testing.T) {
...
+	relayer = e2etest.StartRelayer(t, driver, env)
+	destination, err := env.Chain(route.Destination)
+	require.NoError(t, err)
+	for _, transfer := range transfers {
+		err := e2etest.AwaitState(ctx, relayer, transfer.Packet(),
+			relayerv2.PacketState_PACKET_STATE_SUCCEEDED, destination.Timing())
+		require.NoError(t, err)
+		require.NoError(t, transfer.VerifyDelivered(ctx))
+		require.NoError(t, transfer.VerifyPendingCleared(ctx))
+	}
```
gjermundgaraba, line 183: not critical, but we could add a final check to make sure the amount is still burned after the successful acks here with something like:
```
require.NoError(t, transfers[len(transfers)-1].VerifyBurned(ctx))
```

e2e/ift_test.go
```diff
@@ -28,13 +32,15 @@ func TestIFTTransfer_AutoRelay(t *testing.T) {
...
 	require.NoError(t, transfer.VerifyBurned(ctx))
+	require.NoError(t, transfer.VerifyPending(ctx))
 
 	destination, err := env.Chain(route.Destination)
 	require.NoError(t, err)
 	err = e2etest.AwaitState(ctx, relayer, transfer.Packet(),
 		relayerv2.PacketState_PACKET_STATE_SUCCEEDED, destination.Timing())
 	require.NoError(t, err)
 	require.NoError(t, transfer.VerifyDelivered(ctx))
+	require.NoError(t, transfer.VerifyPendingCleared(ctx))
```
gjermundgaraba, line 43: same as the other, we could add a final check to make sure the amount is still burned after the successful acks here with something like:
```go
require.NoError(t, transfer.VerifyBurned(ctx))
```

gjermundgaraba (changes requested):
Great stuff!

My clanker found a smol issue which is only partially from this pr, but worth doing here imo:
🤖:
I rechecked the multi-packet flow and the per-sequence cleanup is already covered: `AwaitState` submits one source transaction at a time, and `VerifyPendingCleared` checks that same sequence.

I see two remaining FOU-1094 coverage gaps:

1. Successful ACKs are not checked for an accidental source refund.

In `e2e/ift_test.go`, after the single-transfer success assertions around line 43:

```go
require.NoError(t, transfer.VerifyBurned(ctx))
```

After the multi-transfer ACK loop around line 182:

```go
require.NoError(t, transfers[len(transfers)-1].VerifyBurned(ctx))
```

The last packet’s snapshot represents the aggregate balance after all three burns.

2. `VerifyNotMinted` only checks the requested receiver. FOU-1094 says the destination must mint nothing, so it should also verify that destination `totalSupply` remains unchanged.

In `e2e/e2etest/traffic_ift.go`, add a snapshot to `IFTPacket`:

```go
destinationSupplyBefore *big.Int
```

Capture it in `Send` before submitting the transfer:

```go
destinationSupplyBefore, err := i.totalSupply(ctx, i.destination.evm, i.destIFT)
if err != nil {
	return nil, err
}
```

Store it in the returned packet:

```go
destinationSupplyBefore: destinationSupplyBefore,
```

Then extend `VerifyNotMinted`:

```go
supply, err := p.app.totalSupply(ctx, p.app.destination.evm, p.app.destIFT)
if err != nil {
	return err
}
if supply.Cmp(p.destinationSupplyBefore) != 0 {
	return fmt.Errorf(
		"e2etest: IFT packet %s destination total supply: got %s, want %s",
		p.packet.reference(),
		supply,
		p.destinationSupplyBefore,
	)
}
```

The generated IFT binding already provides `TotalSupply`, so this can use a helper matching the existing `balance` helper:

```go
func (i *IFT) totalSupply(
	ctx context.Context,
	client *environment.EVM,
	contract common.Address,
) (*big.Int, error) {
	var supply *big.Int
	err := client.UseContractCaller(func(caller bind.ContractCaller) error {
		bound, err := ift.NewContractCaller(contract, caller)
		if err != nil {
			return fmt.Errorf("e2etest: bind IFT %s: %w", contract, err)
		}
		supply, err = bound.TotalSupply(&bind.CallOpts{Context: ctx})
		return err
	})
	if err != nil {
		return nil, fmt.Errorf("e2etest: query IFT total supply: %w", err)
	}
	return supply, nil
}
```

gjermundgaraba (approved):

PR #1294, opened 2026-08-03
test(e2e): ift multi packet single tx

dhfang:
## Summary
- Add `IFTBatchTransferShim.sol`, a small contract that loops `iftTransfer` over an array of transfers in one cal
- Add `TestIFTTransfer_MultiPacketSingleTx`: emits 10 packets from one source transaction, relays them with a single `Relay` call, and asserts the relayer reports 10 distinct SUCCEEDED packets


gjermundgaraba (commented):
It looks clean enough to me. My clanker did find one issue that looks legit to me:
🤖:
The shim is needed: without it, looping over `IFT.Send` creates separate source transactions and does not satisfy FOU-1086.

There is one API mismatch in the batch result. `SendBatch` returns ordinary `IFTPacket` values and records the shim balance at `e2e/e2etest/traffic_ift.go:186`, but the inherited `IFTPacket.VerifyBurned` and `VerifyRefunded` methods query the application EOA at lines 228 and 286.

Those methods cannot be made correct merely by storing the shim address: an atomic batch has no observable intermediate per-packet source balances. Source burn and refund accounting should remain batch-level.

I suggest exposing only packet identities from `IFTBatch` and moving delivery verification onto the batch.

In `e2e/e2etest/traffic_ift.go`, remove the unusable source snapshot from batch packets:

```diff
 			receiver:          common.HexToAddress(transfers[k].Receiver),
 			amount:            transfers[k].Amount,
-			sourceBefore:      batcherBefore,
 			destinationBefore: destinationsBefore[k],
```

Replace `IFTBatch.Packets` around line 202:

```go
func (b *IFTBatch) Packets() []Packet {
	packets := make([]Packet, len(b.packets))
	for i, packet := range b.packets {
		packets[i] = packet.Packet()
	}
	return packets
}

func (b *IFTBatch) VerifyDelivered(ctx context.Context) error {
	for _, packet := range b.packets {
		if err := packet.VerifyDelivered(ctx); err != nil {
			return err
		}
	}
	return nil
}
```

Then update `TestIFTTransfer_MultiPacketSingleTx`:

```diff
 	batch, err := iftApp.SendBatch(ctx, requests)
 	require.NoError(t, err)
-	require.Len(t, batch.Packets(), packetCount)
+	packets := batch.Packets()
+	require.Len(t, packets, packetCount)

 	wantSequences := make(map[uint64]struct{}, packetCount)
-	for _, packet := range batch.Packets() {
-		wantSequences[packet.Packet().Sequence] = struct{}{}
+	for _, packet := range packets {
+		wantSequences[packet.Sequence] = struct{}{}
 	}
 	require.Len(t, wantSequences, packetCount, "packets must have distinct sequences")

-	require.NoError(t, e2etest.Relay(ctx, relayer, batch.Packets()[0].Packet()))
+	require.NoError(t, e2etest.Relay(ctx, relayer, packets[0]))

 	destination, err := env.Chain(route.Destination)
 	require.NoError(t, err)
-	for _, packet := range batch.Packets() {
-		err = e2etest.AwaitState(ctx, relayer, packet.Packet(),
+	for _, packet := range packets {
+		err = e2etest.AwaitState(ctx, relayer, packet,
 			relayerv2.PacketState_PACKET_STATE_SUCCEEDED, destination.Timing())
 		require.NoError(t, err)
-		require.NoError(t, packet.VerifyDelivered(ctx))
 	}
+	require.NoError(t, batch.VerifyDelivered(ctx))
 	require.NoError(t, batch.VerifyBurned(ctx))

-	statuses, err := relayer.PacketStatuses(ctx, string(route.Source), batch.Packets()[0].Packet().SourceTxHash)
+	statuses, err := relayer.PacketStatuses(ctx, string(route.Source), packets[0].SourceTxHash)
```

This keeps the correct aggregate burn assertion and prevents batch callers from accidentally using source-verification methods that only apply to ordinary EOA-originated transfers.

gjermundgaraba (approved):

PR #1297, opened 2026-08-03
test(e2e): add attested IFT auto-relay coverage

gjermundgaraba:
## Summary

- add an attested IFT auto-relay acceptance test with distinct remote attestors
- verify burn/mint delivery and anchor both receive and acknowledgement proofs to on-chain client heights
- share the attestation-state decoder for follow-up attested tests
- clean up leaked attestor processes, including custom `IBC_BIN` basenames

## Why

FOU-1072 adds the first end-to-end IFT path whose packet proofs are verified by attestation clients rather than the permissive dummy client.

## Checks

- `make build-link`
- `make lint-e2e`
- `make test-harness`
- full default E2E suite
- focused test on Anvil, interval Anvil, and Besu
- `shellcheck e2e/scripts/clean.sh`

Linear: FOU-1072

### DIFF
--- a/e2e/AGENTS.md
+++ b/e2e/AGENTS.md
@@ -2,7 +2,8 @@
 
 This module contains one root repository-level acceptance package: linear Go tests over the
 `internal/harness` surface. Tests relay real IBC packets through the real Link relayer and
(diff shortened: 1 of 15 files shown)


dhfang (approved):
🚀🚀🚀 

What do we plan to do with all the dummy client usages, replace them with the attestations spec?

PR #1298, opened 2026-08-03
test(e2e): cover multi-attestor quorum recovery

gjermundgaraba:
## Summary

- add a 2-of-3 destination-attestor quorum acceptance test
- prove one inactive attestor still permits delivery, losing a second leaves the packet pending without minting, and restoring quorum completes the same transfer
- verify the configured signer set, threshold, client heights, and both live attestors’ latest attestable heights

## Why

FOU-1076 covers Link quorum aggregation and recovery through the black-box E2E surface.

## Checks

- `make build-link`
- `make lint-e2e`
- `make test-e2e`
- focused test on Anvil, interval Anvil, and Besu

Linear: FOU-1076

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/e2e/ift_test.go
+++ b/e2e/ift_test.go
@@ -25,8 +25,12 @@ const (
 	attestedClientBID  environment.ClientID    = "attested-client-b"
 	attestorAID        environment.AttestorID  = "attestor-a"
 	attestorBID        environment.AttestorID  = "attestor-b"
+	attestorCID        environment.AttestorID  = "attestor-c"
+	attestorDID        environment.AttestorID  = "attestor-d"
 	attestorAAuthority environment.AuthorityID = "attestor-a-authority"
 	attestorBAuthority environment.AuthorityID = "attestor-b-authority"
+	attestorCAuthority environment.AuthorityID = "attestor-c-authority"
+	attestorDAuthority environment.AuthorityID = "attestor-d-authority"
 )
 
 func TestAttestedIFTTransfer_AutoRelay(t *testing.T) {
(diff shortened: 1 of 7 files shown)


e2e/ift_test.go
```diff
@@ -95,19 +98,132 @@ func TestAttestedIFTTransfer_AutoRelay(t *testing.T) {
...
-	require.GreaterOrEqual(t, destinationState.LatestHeight, sendReceipt.BlockNumber.Uint64())
-	// The acknowledgement proof advanced the source client through the receive block.
-	require.GreaterOrEqual(t, sourceState.LatestHeight, receiveReceipt.BlockNumber.Uint64())
+	// One live attestor cannot satisfy the destination client's quorum.
+	require.NoError(t, destinationAttestorB.Stop(ctx))
+	pending, err := iftApp.Send(ctx, e2etest.IFTRequest{Amount: big.NewInt(2_345_000)})
+	require.NoError(t, err)
+	require.NoError(t, pending.VerifyBurned(ctx))
+	require.NoError(t, e2etest.Relay(ctx, relayer, pending.Packet()))
+	require.NoError(t, e2etest.AwaitStable(ctx, relayer, pending.Packet(),
```
dhfang, line 221: Just a general note on `AwaitStable`, it wasn't really obvious what this does to me so I looked at the [implementation](https://github.com/cosmos/ibc/blob/6e96cf6fe2e6eeb2af908c778889d7ff5961ba19/e2e/e2etest/traffic_relayer.go#L63-L105) and it seems like the purpose is to assert the relayer stalls at some state. However, this is based solely on a test configuration `environment.Timing` and doesn't take into account the relayer's relay interval or finality offset or the attestor's finality offset. This is a bit confusing to me as the developer as I need to know that AwaitStable's assertion is valid only if the relayer is relaying packets at an interval quicker than the settlement window. Perhaps build some assertions into AwaitStable on the relayer's relay interval relative to the settlement window?

e2e/ift_test.go
```diff
@@ -18,12 +17,16 @@ import (
...
-	attestorAID        environment.AttestorID  = "attestor-a"
-	attestorBID        environment.AttestorID  = "attestor-b"
-	attestorAAuthority environment.AuthorityID = "attestor-a-authority"
-	attestorBAuthority environment.AuthorityID = "attestor-b-authority"
+	attestedClientAID        environment.ClientID    = "attested-client-a"
+	attestedClientBID        environment.ClientID    = "attested-client-b"
+	attestorAID              environment.AttestorID  = "attestor-a"
+	attestorBID              environment.AttestorID  = "attestor-b"
+	attestorQuorumBID        environment.AttestorID  = "attestor-quorum-b"
+	attestorQuorumCID        environment.AttestorID  = "attestor-quorum-c"
```
dhfang: This was a bit confusing for me to parse, why are some of these attestors prefixed `attestor` and others `attestorQuorum`? Why not just attestorAID, attestorBID, attestorCID, attestorDID?

dhfang (approved):

e2e/ift_test.go
gjermundgaraba: Good point. AwaitStable uses an end-to-end harness stability window, although accepting environment.Timing made that contract look chain-only. The managed harness fixes the Link dispatch interval at 100 ms, batch size at 1, and finality offset at 1; its settle windows cover multiple dispatch and block-production opportunities in every lane. This test also calls Relay before AwaitStable. A generic assertion would require exposing the effective relayer and attestor timing through the API, which AwaitStable cannot derive today. I clarified SettleWindow as end-to-end here; the broader timing assertion belongs with the lane/timing refactor.

e2e/ift_test.go
gjermundgaraba: Agreed. A is the source attestor, and B/C/D are the destination attestors. The Quorum infix was intended to distinguish the added fixtures, but it made the set harder to parse. I renamed the IDs, authorities, and local variables consistently.

e2e/ift_test.go
gjermundgaraba: sorry, my clanker decided to respond for me. but yes, agreed, and will roll this into a couple of other refactors im lining up

PR #1304, opened 2026-08-04
fix(link): avoid cross-chain timeout height comparison

gjermundgaraba:
## Summary

- separates packet-event lookup from the same-chain event-height guard used by recv and acknowledgement relays
- keeps the guard where the event and proof heights belong to the same chain
- avoids comparing the source SendPacket height with the destination timeout-proof height
- adds a regression case with source event height 150 and destination proof height 100

## Root cause

The shared packet-event helper assumed every event height and proof height came from the same chain. Timeout relay is different: it reconstructs packet data from the source SendPacket event but proves non-receipt on the destination chain. Comparing those chain-local heights rejected valid timeouts when chain heights diverged.

## Impact

Valid timeout refunds can proceed even when the source and destination chains have substantially different block heights.

## Validation

- `make check` (builds, all Link and harness tests, both linters, generated bindings, and the complete Anvil E2E suite)

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/link/internal/relay/pipeline/pipeline_test.go
+++ b/link/internal/relay/pipeline/pipeline_test.go
@@ -328,7 +328,9 @@ func TestPipelineLifecycle(t *testing.T) {
 		env.dstClient.EXPECT().IsTimestampFinalized(mock.Anything, mock.Anything, (*uint64)(nil)).Return(true, nil).Once()
 
 		// timeout delivery on the source chain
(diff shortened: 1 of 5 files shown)


dhfang (approved):
Good catch

PR #1307, opened 2026-08-05
Initial deploy wiring

Eric-Warehime:
Creates `ibc deploy` commands with the following expected interface:

```
ibc deploy core --chain <chainId>
    Provision the core routing stack on one chain (all creation happens
    in-process from artifacts embedded in the binary; relaying is opened to
    any EOA as part of the same provisioning). Writes/updates the chain's
    manifest.

ibc deploy client --chain <chainId> --counterparty-chain <chainId>
    Provision a light client on --chain tracking --counterparty-chain, then
    register it on the router. Records clientId + pairing in the manifest.
    A connection is two mirrored invocations, one per chain. Rerunning with
    the same id continues (skips) a matching deployment; a rerun whose
    attestors/threshold conflict with the recorded deployment fails with the
    differences listed, as does a client registered on-chain but missing
    from the manifest — both suggest a new --client-id.
    Flags: --type <clientType>       default "attestation"; validated against
                                     the driver's supported types
           --attestors <a,…> --threshold <n>
                                     attestation params; values are
                                     attestation names or signer aliases
                                     (resolved through local keys; remote
                                     signers error), and anything matching
                                     no alias passes through verbatim as an
                                     address for the chain driver to
                                     validate; default derives from
                                     attestor.attestations[] for the
                                     tracked chain
           --client-id / --counterparty-client-id
                                     both default to one shared id,
                                     link-<a>-<b> with the chain ids
                                     sorted, so mirrored invocations pair
                                     up without id flags
           --height/--timestamp      initial trusted state; default fetched
                                     from the counterparty chain head
                                     (clamped to height 1 on fresh chains)

ibc deploy status [--chain <chainId>]
    Read-only health report: manifest vs live chain — connected chain id
    matches, router bytecode present, clients registered with matching
    addresses and counterparties. Without --chain, sweeps the union of
    configured chains and existing manifests, reporting per-chain errors
    without aborting; exits non-zero on any failure. Works without a
    configured deployer.

ibc deploy render-config <chainA> <chainB>
    Load the two chains' manifests and emit relayer.clients[] entries for
    every mutual client pair (each side's counterparty fields referencing
    the other back), printed to stdout for the user to merge — closing the
    loop to `relayer run`. Errors if no mutual pair exists. Attestor set
    names are left for the operator; only the threshold is carried over.
    Read-only: no chain access, works without a configured deployer.
```

Note that I've intentionally left out app registration for a future PR. So once you've used this to deploy the core and clients you'll still be left without an app to use to send packets back and forth--likely that will take the form of `ibc deploy app --type ics20` or something similar.


link/cmd/ibc/deploy.go
```diff
@@ -0,0 +1,540 @@
...
+			timestamp = ts
+		}
+	}
+	spec := deploy.ClientSpec{
+		ClientID:             clientID,
+		Type:                 flagDeployClientType,
+		CounterpartyChainID:  counterpartyChainID,
+		CounterpartyClientID: counterpartyClientID,
+	}
+	if flagDeployClientType == deploy.ClientTypeAttestation {
```
dhfang: Error on unsupported type

link/cmd/ibc/deploy.go
```diff
@@ -0,0 +1,540 @@
...
+		return discovered
+	}
+	merged := *discovered
+	merged.Clients = nil
+	for _, c := range discovered.Clients {
+		if old, ok := existing.Client(c.ClientID); ok {
+			if c.CounterpartyChainID == "" {
+				c.CounterpartyChainID = old.CounterpartyChainID
+			}
+			if c.Params == nil {
```
dhfang: Can c.Params and c.Type be reconstructed from on-chain information

link/cmd/ibc/deploy.go
```diff
@@ -0,0 +1,540 @@
...
+				CounterpartyChainID:  c.CounterpartyChainID,
+				CounterpartyClientID: c.CounterpartyClientID,
+				Type:                 config.ClientType(c.Type),
+			}
+			if c.Type == deploy.ClientTypeAttestation {
+				set := &config.AttestorSetConfig{}
+				if threshold, ok := c.Params["threshold"].(float64); ok {
+					set.Threshold = int(threshold)
+				}
+				// attestor names cannot be derived from addresses; user fills them in
```
dhfang: I think we can get rid of the attestorSet config altogether hopefully https://linear.app/cosmoslabs/issue/FOU-1221/investigate-whether-we-can-remove-the-attestorset-config-in-the

link/cmd/ibc/deploy.go
```diff
@@ -0,0 +1,540 @@
...
+				client.AttestorSet = set
+			}
+			out.Relayer.Clients = append(out.Relayer.Clients, client)
```
dhfang: I think we should only append client configs where there also exists a client config for the counterparty, it will fail config validation otherwise. Not necessary to change now, I am working on restructuring the configs and likely will get rid of the top level clients list in the relayer configs.

link/cmd/ibc/main.go
```diff
@@ -75,4 +81,32 @@ func init() {
...
+
+	// Deploy commands
+	cmdDeploy.AddCommand(
+		cmdDeployCore, cmdDeployClient, cmdDeployConnect,
+		cmdDeployImport, cmdDeployStatus, cmdDeployRenderConfig,
+	)
+	dpf := cmdDeploy.PersistentFlags()
+	dpf.StringVar(&flagDeployManifestDir, "manifest-dir", "deployments", "manifest directory relative to home")
+	dpf.StringVar(&flagDeployDeployer, "deployer", "", "signer alias override for deployment transactions")
+	dpf.StringVar(&flagDeployChain, "chain", "", "chain id")
```
dhfang: Better description please

link/cmd/ibc/main.go
```diff
@@ -75,4 +81,32 @@ func init() {
...
+	dpf.BoolVar(&flagDeployDryRun, "dry-run", false, "print the step plan without submitting transactions")
+	dpf.BoolVar(&flagDeployYes, "yes", false, "skip confirmation prompts")
+
+	cmdDeployClient.Flags().
+		StringVar(&flagDeployCounterparty, "counterparty", "", "counterparty chain id the client tracks")
```
dhfang: Kinda ambiguous - prefer `counterparty-chain`

link/cmd/ibc/main.go
```diff
@@ -75,4 +81,32 @@ func init() {
...
+	for _, c := range []*cobra.Command{cmdDeployClient, cmdDeployConnect} {
+		c.Flags().StringVar(&flagDeployClientType, "type", deploy.ClientTypeAttestation, "light client type")
+		c.Flags().StringSliceVar(&flagDeployAttestors, "attestors", nil, "attestor addresses (attestation clients)")
```
dhfang: Can you describe the expected format in the description? i.e. comma separated list of addresses or attestor aliases. If it's addresses I think we can consider taking attestor aliases to save the operator from having to derive addresses.

link/internal/deploy/evm/wiring.go
```diff
@@ -0,0 +1,273 @@
...
+		if !registered {
+			continue
+		}
+		cp, err := contract.GetCounterparty(&bind.CallOpts{Context: ctx}, ev.ClientId)
+		if err != nil {
+			return nil, fmt.Errorf("query counterparty for %q: %w", ev.ClientId, err)
+		}
+		m.UpsertClient(manifest.Client{
+			ClientID:             ev.ClientId,
+			Type:                 deploy.ClientTypeAttestation,
```
dhfang: Is it possible to determine the type from onchain information rather than hardcoding here. Similarly, is it possible to query the light client params and populate that here.

link/cmd/ibc/deploy.go
```diff
@@ -0,0 +1,540 @@
...
+// wherever discovered left them empty.
+func mergeManifests(existing, discovered *manifest.Manifest) *manifest.Manifest {
+	if existing == nil {
+		return discovered
+	}
+	merged := *discovered
+	merged.Clients = nil
+	for _, c := range discovered.Clients {
+		if old, ok := existing.Client(c.ClientID); ok {
+			if c.CounterpartyChainID == "" {
```
dhfang: If a client is imported via discovery and there is no prior entry in the manifests to pull counterpartyChainID and params from, I believe the render step just renders those fields using zero values. Some of those zero values will pass config validation, which could be pretty confusing for operators.

link/internal/deploy/evm/driver.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return abi.ABI{}, nil, fmt.Errorf("parse access manager abi: %w", err)
+	}
+	return parsed, common.FromHex(artifact.Bytecode.Object), nil
+}
+
+// ProvisionCore deploys AccessManager + ICS26Router (implementation behind
+// an initialized ERC1967 proxy), then binds the relaying selectors to
+// PUBLIC_ROLE so any relayer EOA can submit packets. The deployer key stays
+// the AccessManager admin; role hardening is a follow-up.
```
dhfang, line 123: Can you make a ticket tracking this followup.

link/internal/deploy/evm/wiring.go
```diff
@@ -0,0 +1,273 @@
...
+
+	"github.com/cosmos/ibc/link/internal/deploy"
+	"github.com/cosmos/ibc/link/internal/deploy/manifest"
+)
+
+// newTestDriver builds a Driver on an injected backend, without RPC or forge.
+// Only exercised from _test.go files, which golangci-lint's unused check
+// (run.tests: false) can't see.
+//
+//nolint:unused
```
dhfang: This constructor seems pretty unnecessary

link/internal/deploy/evm/driver.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	attestors, err := attestationArgs(params)
+	if err != nil {
+		return deploy.ClientRef{}, fmt.Errorf("client %q: %w", spec.ClientID, err)
+	}
+	opts, err := d.transactOpts(ctx)
+	if err != nil {
+		return deploy.ClientRef{}, err
+	}
+	addr, tx, _, err := attestation.DeployContract(
```
dhfang, line 212: Right now the last roleManger arg is set to the zero address resulting in the `PROOF_SUBMITTER` role being set to the zero address. This means anyone can submit client updates to the client contract directly. While this is not a security issue, I typically set the router as the proof submitter because events get emitted when submitting client updates through the router that aren't when submitted directly to the client contract.

link/README.md
```diff
@@ -22,6 +22,12 @@ make build
 # run the relayer and/or attestor after populating config
 ./bin/ibc relayer run
 ./bin/ibc attestor run
+
+# deploy IBC between two configured chains (idempotent; rerun to repair)
+./bin/ibc deploy connect <chainA> <chainB> --attestors 0x... --yes
```
dhfang: We shouldn't be reusing attestor sets for multiple chains :)

link/cmd/ibc/deploy.go
```diff
@@ -0,0 +1,540 @@
...
+	specA, err := clientSpec(cmd.Context(), targetB, chainA, chainB) // on A, tracking B
+	if err != nil {
+		return err
+	}
+	specB, err := clientSpec(cmd.Context(), targetA, chainB, chainA) // on B, tracking A
+	if err != nil {
+		return err
+	}
+	// connect derives both IDs; per-direction overrides only apply to `deploy client`
+	specA.ClientID, specA.CounterpartyClientID = deploy.DefaultClientID(chainB), deploy.DefaultClientID(chainA)
```
dhfang: `connect` always uses the default client ids, but I think it will actually be a fairly common to want to deploy a new client pair after having ran `connect` once. IIUC the idempotency check here https://github.com/cosmos/ibc/pull/1307/changes#diff-afbbb595d032b82f1638dbfb0c73f815a973b11443fa0721243d18a6ee5ca849R111-R128 actually creates an issue in that deployment will be skipped but if it is invoked with different client params, the manifest will be updated with the new client params. i.e. a user could call `connect` once which deploys the client contracts and then calls it again with a different attestor set, which will not deploy new contracts but will still update the manifests with the new attestor set. If new configurations are then generated using the out-of-sync client params, relaying will fail.

dhfang (changes requested):
I'm mostly aligned with the cli interface and think there's alot of neat and useful functionality built into it 🚀🚀🚀 

The `import` command doesn't seem that useful in it's current state though for a couple reasons:
- It relies on there being an existing manifest entry for any discovered clients in order to populate counterparty chain id, client params, and type. This premise doesn't seem real to me - it seems like import is likely to be used when you have no prior manifest data for the chain. An operator that uses this will end up with a list of clients with counterparty client ids identified but I think it's unlikely they'll know from these ids alone what the counterparty chain id, client params and type are. 
- Import is for a single chain. The router may have clients for counterparty chains that the relayer is not interested in relaying for. The manifest will be populated with clients but not necessarily their counterparties. In this case, the render step will generate a list of client configs that fail config validation due to the lack of matching counterparty configs.

I think we can consider `import-connections` instead where it takes two chains as arguments and it only identify clients with a counterparty on the other chain. This lets us at least populate counterparty chain id for the clients and also avoids the issue of populating the manifests with clients the relayer is not interested in. I realize that clients are not registered in the contracts with any identifiers for the counterparty chain, so this is vulnerable to an edge case where the client ids match based on the registered counterparty client id but they are not true counterparties with each other. I think this is an unlikely scenario if operators are deploying the clients using link given the client ids the deploy script uses.

link/internal/deploy/evm/driver.go
Eric-Warehime: https://linear.app/cosmoslabs/issue/FOU-1235/integrate-access-management-to-ibc-link-deploy

link/cmd/ibc/deploy.go
Eric-Warehime: Ended up changing this behaviour to just fail if we attempt to register a client that is already registered with differing fields/params.

link/cmd/ibc/deploy.go
```diff
@@ -216,23 +203,44 @@ func deployCore(cmd *cobra.Command, _ []string) error {
...
-// clientSpec assembles the ClientSpec for --chain tracking --counterparty,
+// clientSpec assembles the ClientSpec for --chain tracking --counterparty-chain,
 // defaulting trusted state from the counterparty chain head.
 func clientSpec(
 	ctx context.Context,
+	cfg config.Config,
 	counterpartyTarget deploy.Target,
 	chainID, counterpartyChainID string,
 ) (deploy.ClientSpec, error) {
+	// both ids default to one shared name: client ids are per-chain
```
dhfang, line 214: Nice I like this more

link/cmd/ibc/deploy.go
```diff
@@ -470,71 +400,82 @@ func deployStatus(cmd *cobra.Command, _ []string) error {
...
 
-// renderedDeployment is the subset of config.Config that `deploy
-// render-config` actually populates. config.Config itself isn't used as the
-// output type: printed as YAML it would also emit every zero-valued field
-// (server:, db:, signers: [], etc.), which would corrupt a working config
-// if merged in as printed.
+// renderedDeployment is the subset of the config schema render-config emits,
+// so the output can be merged into ibc.yml without zero-valued sections.
 type renderedDeployment struct {
-	Chains  []config.ChainConfig `yaml:"chains"`
```
dhfang: Seems like it would still be useful for surfacing router addresses

dhfang (approved):
LGTM

PR #1317, opened 2026-08-09
feat(link): support sequence-filtered relaying

gjermundgaraba:
Closes FOU-1091

## What

Adds explicit packet selection to Link's Relay API so callers must choose either every relayable packet in a source transaction or an exact set of source-client/sequence selectors.

- Adds public `PACKET_STATE_NOT_SELECTED` status and persists every configured and routed packet before conditionally admitting selected packets as `PENDING`.
- Keeps packet admission monotonic, atomic, and idempotent across duplicate, in-flight, and terminal selections.
- Rejects absent selectors and packets outside the configured source/destination client pair before any mutation.
- Adds explicit relay-all/relay-selected E2E clients and aggregate IFT batch verification for balances, supply, and pending-transfer records.
- Covers the specified 10-packet filtered receive flow (sequences 2 and 6) and five-packet real-clock timeout flow (sequences 2 and 3), including stable unselected packets and later admission of the remainder.
- Orders packet persistence deterministically to avoid conflicting concurrent lock order.

## Validation

- `go -C link test ./internal/service/relayer ./internal/store`
- `make -C link lint`
- `go -C e2e test ./... -run '^$'`
- `E2E_MODE=fast go -C e2e test . -timeout 60m -parallel 2 -run '^Test(IFTTimeout_Refund|Relay_FilteredSequences|Relay_FilteredTimeoutSequences)$' -count=1 -v`
- `make check-e2e-matrix`
- `git diff --check`


proto/link/relayer.proto
```diff
@@ -19,6 +19,28 @@ service RelayerApiService {
 message RelayRequest {
   string tx_hash = 1;
   string source_chain_id = 2;
+  // Selection is required and controls only this relayer instance; IBC
+  // relaying remains permissionless.
+  oneof selection {
+    AllPackets all_packets = 3;
+    SelectedPackets selected_packets = 4;
+  }
+}
```
dhfang, line 28:
```suggestion
message RelayRequest {
string tx_hash = 1;
string source_chain_id = 2;
repeated PacketSelector filter_packets = 3;
}
```

How about just this - unspecified `filter_packets` means relay all packets which should be the common case

link/internal/service/relayer/service.go
```diff
@@ -181,9 +292,19 @@ func (s *Service) packetsFromEvents(chainID, txHash string, events []v2.PacketEv
 
 			continue
 		}
+		if _, routed := s.cfg.Relayer.RouteBySourceClient(client.Alias); !routed {
```
dhfang, line 275: Thanks for catching this.

dhfang (approved):
I dont really think there are any product requirements for this right now, did we do this so we can port over certain e2e tests?

proto/link/relayer.proto
gjermundgaraba: Yeah, I did consider that the only reason I didn't do this was because it makes it impossible to know if someone accidentally sent in an empty list that they meant to and that they meant to filter something or if they meant to relay all I don't think the failure scenario is too likely but it does introduce some ambiguity of what the relay request is meant to do. Relay all or relay filtered packets that said I am not strong strongly opinionated about this and I'm very open to changing this. We could potentially also add a separate message for this

dhfang: Yeah I think the error case is unlikely and something like filter_packets is more concise but still self-documenting. Not strongly opinionated either, I'm happy for you to go with your preference here.

gjermundgaraba: in that case i think i'll keep it as is, even if unlikely, if you accidentally do this, it would instead relay all - which would be quite surprising since you are trying to filter something. this feels more explicit at least. we can also revisit before final release ofc

PR #1321, opened 2026-08-10
legal: normalize license and add SPDX headers

gjermundgaraba:
## Summary
- replace the modified root license with canonical Apache 2.0 text
- preserve the existing attribution in `NOTICE`
- add `Apache-2.0` SPDX headers to 261 hand-authored code and configuration files
- refresh generated Proto/Connect bindings so they inherit source headers, and refresh generated SQL after adding its query-source header
- emit SPDX headers from the E2E matrix and SQL migration generators
- enforce future headers across supported file types with pinned License-Eye checks in Make and CI

## Deferred spec scope
The spec documents still declare Apache 2.0. Their `Community-Spec-1.0` headers must land with the CSL conversion tracked by FOU-1152 through FOU-1155; adding them here would create contradictory declarations. The License-Eye exclusion records this boundary for removal with that conversion.

## Validation
- exact comparison with `https://www.apache.org/licenses/LICENSE-2.0.txt`
- `actionlint`, YAML parsing, shell syntax, and `gofmt`
- `make check-license-headers`, including a negative smoke test with a headerless Go file
- `make -C link codegen-sql` and `make -C link codegen-proto`
- migration-generator smoke check
- `TEST_POSTGRES=true make -C link check`
- `make check` (build, linters, Link tests, harness tests, generated artifacts, and fast E2E suite)

Linear: FOU-1165  
Stack: 4/4; depends on #1320


NOTICE
dianab-cl, line 1: Expected to mention Interchain?

dianab-cl (approved):

NOTICE
gjermundgaraba: for now, creating a follow-up issue to check with legal

PR #1328, opened 2026-08-10
refactor(relayer): config simplification

dhfang:
### Summary

- Reworks the relayer config schema to be connection-first: one entry per IBC connection with two client ends (clientA/clientB).
- Previously remote attestors were defined in the relayer's `attestorSet` config and local attestors were defined in the top level `attestor` configuration block. Now remote and local attestor definition is consolidated in a top level `attestors` config. 
- Removes the relayer's `attestorSet` config(threshold + attestor list + finality offset, per client end) and instead associates the attestors configured in the top level `attestors` block with their corresponding light client by associating their self-reported addresses with the addresses queried from the on-chain attestation light client. A misconfigured or stale quorum now fails at startup instead of silently drifting from what the chain actually enforces.
- Adds live client counterparty validation to ibc config validate --live: confirms each connection's on-chain registered counterparty actually matches what's configured, in both directions.
- The relayer used to rely on the finality offset configured in the `attestorSet` to determine when a packet commitment is finalized. This PR removes that as a concept and instead just waits till all the attestors report they can attest to the height of a commitment. This happens in the Relayer's CheckSendFinality/CheckTimeoutFinality/CheckWriteAckFinality pipeline stages where they now determine finality by calling LatestProvableHeight() on the attestor light client implementation of the ProofGenerator abstraction.

The refactors required to support the new configuration format are quite large - I added some comments in the diff to help reviewers parse the changes, but let me know if anyone prefers me to split into smaller PRs.

Closes FOU-1219, FOU-1221


link/internal/bootstrap/bootstrap.go
```diff
@@ -49,29 +49,35 @@ func BuildRelayer(cfg config.Config) (*Services, error) {
...
-	signers, err := signerSet(ctx, cfg)
+	signers, err := signer.NewSetFromConfig(ctx, cfg.Signers)
 	if err != nil {
 		return nil, err
 	}
 
-	// Dual mode: if .attestor config is provided, then we can run both relayer and attestor in the same process.
-	// This might be useful for PoC/testing environments or when an operator wants to run the relayer
+	// Attestors
+	local, remote, err := attestor.ResolveFromConfig(ctx, cfg.Attestors, clientSet, signers)
```
dhfang, line 60: We resolve both local and remote attestors up front now as opposed to the previous flow which was confusing. We previously initialized the attestor grpc service handlers, which internally resolved the local attestors. We'd pipe that attestor service into the ProofGenerators which retrieved local attestors from it and did their own remote attestor resolution as needed.

link/internal/relay/pipeline/pipeline.go
```diff
@@ -111,21 +111,25 @@ func NewPipeline(
 	output = pipeline.ProcessConcurrently(ctx, stageConcurrency,
 		NewProcessorMW(deps.Storage, processors.NewCheckPacketCommitment(deps.Chains, deps.Storage)), output)
 
-	// wait for the send tx to finalize on the source chain
```
dhfang, line 116: We pipe the ProofGenerators directly into the processors that handle waiting for tx finality so that they can query the LatestProvableHeight method to determine when a packet commitment can be relayed. This replaces static finality offsets configs.

link/internal/relay/proofgen/attestation/resolve.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, errors.Errorf("no configured chain client for counterparty chain %q", counterparty.ChainID)
+	}
+
+	return New(matched, int(minRequiredSigs), counterpartyChain), nil
+}
+
+// MatchAttestors resolves self's on-chain attestation set and returns the
+// attestors that watch counterparty's chain and whose self-reported address
+// appears in it, erroring if too few match to meet the on-chain threshold.
+func MatchAttestors(
```
dhfang, line 42: We no longer specify in the configs which attestors belong to which light client in the relayer configs. It is replaced with this logic that associates the attestors with their light clients on startup. This lets us get rid of the `attestorSet` config block entirely. The `attestorSet` config block was confusing to dog fooders because you configure it under the client for one chain but it actually should be attesting to the counterparty client's chain.

proto/link/attestor.proto
```diff
@@ -14,6 +14,9 @@ service AttestationService {
 
   // Returns the latest height the attestor will generate attestations for.
   rpc LatestHeight(LatestHeightRequest) returns (LatestHeightResponse);
+
+  // Returns identity information about a configured attestor.
+  rpc Info(InfoRequest) returns (InfoResponse);
```
dhfang, line 21: Added an info endpoint to the attestor so that the relayer can discover the attestor's chain id and address on startup.

link/internal/relay/proofgen/attestation/resolve.go
```diff
@@ -0,0 +1,89 @@
...
+			continue
+		}
+
+		configured = append(configured, fmt.Sprintf("%s=%s", a.Name(), a.Address()))
+
+		if _, inOnChainSet := onChainSet[strings.ToLower(a.Address())]; !inOnChainSet {
+			continue
+		}
+
+		matched = append(matched, a)
```
gjermundgaraba, line 89: might want to dedup here. with two configured attestors with the same attestor key this could "reach quorum" according to this code (misconfiguration probably, but it would create some confusing errors later).

link/cmd/ibc/attestors.go
```diff
@@ -7,14 +7,13 @@ import (
...
-// only the target driver can judge.
+// resolveAttestorToken resolves one attestor token: an attestors[].name or a
+// signers[] alias resolves through its key; anything else is passed through
+// verbatim as an address, whose format only the target driver can judge.
 func resolveAttestorToken(cfg config.Config, token string) (string, error) {
 	alias := token
-	if attestation, ok := cfg.AttestationByName(token); ok {
-		alias = attestation.Signer
+	if attestor, ok := cfg.AttestorByName(token); ok {
+		alias = attestor.Signer
```
gjermundgaraba, line 24: Given:

attestors:
  - name: attestor-1
    type: remote
    grpc: attestor.example.com:3000

And:

ibc deploy client \
  --chain 1 \
  --counterparty-chain 8453 \
  --attestors attestor-1

The failure is:

1. The CLI finds attestor-1 in attestors[].
2. link/cmd/ibc/attestors.go:13 tries to resolve its signer.
3. The resolver fails to find a signer and returns the original string "attestor-1".
4. The EVM driver receives that string and rejects it as invalid attestor address "attestor-1".

link/internal/relay/pipeline/pipeline.go
```diff
@@ -111,21 +111,25 @@ func NewPipeline(
...
+	// wait for the send tx's packet event to be at or before what the
+	// destination client can currently prove
+	checkSendFinality, err := processors.NewCheckSendFinality(deps.Chains, deps.ProofGenerators, route)
```
gjermundgaraba, line 118: 🤖: These per-transfer finality checks duplicate the proof-height/timestamp gates already enforced by `BatchRecvPacket`, `BatchTimeoutPacket`, and `BatchAckPacket`. This adds redundant RPCs and leaves `TxHeight` plus the finalized-time fields with no independent runtime purpose.

link/internal/config/relayer.go
```diff
@@ -47,41 +45,26 @@ type RelayerEVMConfig struct {
...
-type RouteConfig struct {
-	SourceClient string `yaml:"sourceClient"`
-	// SourceSignerAlias and DestSignerAlias are the signer aliases used to submit relay
-	// transactions on the route's source and destination chains.
-	SourceSignerAlias string          `yaml:"sourceSignerAlias"`
-	DestSignerAlias   string          `yaml:"destSignerAlias"`
-	AutoRelay         AutoRelayConfig `yaml:"autoRelay,omitempty"`
+	// AutoRelay configures auto-relay for packets flowing FROM this end's
+	// chain TOWARD the counterparty end.
+	AutoRelay AutoRelayConfig `yaml:"autoRelay,omitempty"`
```
gjermundgaraba, line 69: 🤖: `autoRelay.enabled` and `lookback` have no runtime readers. In particular, `enabled: false` does not disable anything, so the config accepts and documents values that silently have no effect.

link/internal/config/relayer.go
```diff
@@ -47,41 +45,26 @@ type RelayerEVMConfig struct {
...
-	ChainID              string             `yaml:"chainId"`
-	CounterpartyChainID  string             `yaml:"counterpartyChainId"`
-	CounterpartyClientID string             `yaml:"counterpartyClientId"`
-	Type                 ClientType         `yaml:"type"`
-	AttestorSet          *AttestorSetConfig `yaml:"attestorSet,omitempty"`
+// ConnectionConfig one bidirectional IBC connection the relayer actively
+// relays, in both directions. ClientA's counterparty is simply ClientB (and
+// vice versa).
+type ConnectionConfig struct {
+	Alias   string    `yaml:"alias"`
```
gjermundgaraba, line 54: the connection alias is required and uniqueness-validated, but `ConnectionByAlias` has no production callers and is otherwise only used in diagnostics atm. should we remove it?

proto/link/attestor.proto
```diff
@@ -51,6 +54,16 @@ message LatestHeightRequest { string attestor = 1; }
...
 message LatestHeightResponse { uint64 height = 1; }
 
+message InfoRequest { string attestor = 1; }
+
+message InfoResponse {
+  // The chain this attestor watches.
+  string chain_id = 1;
+  // The attestor's signing address.
+  string address = 2;
+  uint64 finality_offset = 3;
```
gjermundgaraba: `finality_offset` is transported and cached but never read by proof or quorum logic. `LatestHeight` already returns the attestor's offset-adjusted height, so this duplicates wire metadata across the proto, handler, remote state, and mocks.

link/internal/service/attestor/service.go
```diff
@@ -82,45 +90,6 @@ var (
...
-	for _, spec := range cfg.Attestor.Attestations {
-		if err := add(spec); err != nil {
-			return nil, fmt.Errorf("attestor %s: %w", spec.Name, err)
-		}
-	}
-
-	return New(attestorsSpecs)
-}
-
 // New Service constructor. Attestors should have unique aliases
```
gjermundgaraba: code smell? `Alias()` is identical to `Name()` for both implementations and only used to key `Service`. can we drop it?

e2e/internal/harness/ibclink/relayer_config.go
```diff
@@ -161,142 +148,99 @@ func buildRelayerFileConfig(cfg RelayerConfig) (fileConfig, error) {
...
-		addLocalAttestor(
-			&file, processSigner, cfg.FinalityOffset,
-			chain.ChainID, localAttestorName(chain.ChainID), "",
-		)
 	}
 
-	clients := make(map[string]string, 2*len(cfg.Connections))
-	type localAttestor struct{ chainID, keyFile string }
-	locals := make(map[string]localAttestor)
+	sourceClients := make(map[string]struct{}, 2*len(cfg.Connections))
```
gjermundgaraba: 🤖: `sourceClients` exists only to validate `cfg.Routes`; those routes never reach the generated YAML, while `Connections` already configure both directions. The harness shape therefore suggests route-level control that the relayer config no longer has.

link/internal/service/attestor/resolve.go
```diff
@@ -0,0 +1,73 @@
...
+			if errRemote != nil {
+				slog.Warn(
+					"Skipping unresolvable configured attestor",
+					"name", entry.Name, "type", entry.Type, "err", errRemote,
+				)
+				continue
+			}
+
+			remote = append(remote, a)
+		default:
```
gjermundgaraba, line 46: 🤖: The unsupported-type fallback here and the missing-gRPC check in `resolveRemote` duplicate authoritative config validation. Production callers cannot reach either state, so maintaining the same schema policy in this resolver risks drift.

link/internal/config/config.go
```diff
@@ -443,44 +454,75 @@ func DBConfigFromURL(url string) (DBConfig, error) {
...
+		case c.FinalityOffset != 0:
+			return errors.New(".finalityOffset must not be set for remote attestors")
+		}
 	}
 
 	return nil
 }
 
+// Locals returns the subset of attestors this process runs itself.
+func (a Attestors) Locals() []AttestorConfig {
```
gjermundgaraba: looking at how this is used, it's unclear if we need this?

e2e/internal/e2etest/traffic_setup.go
```diff
@@ -364,21 +364,19 @@ func buildConfig(
 			ICS26Router: apps.ICS26Router.Hex(),
 		})
 	}
-	attestorIDs := env.Attestors()
-	attestorSets := make(map[string]*ibclink.RelayerAttestorSet, len(attestorIDs))
-	for _, id := range attestorIDs {
+	declaredAttestors := map[string]bool{}
```
gjermundgaraba: 🤖: `Environment.Attestors()` enumerates a map whose IDs were already uniqueness-validated, so duplicates cannot reach this loop. `declaredAttestors` only rechecks an invariant already guaranteed structurally.

link/internal/relay/proofgen/attestation/resolve.go
dhfang: good catch

link/internal/config/relayer.go
dhfang: We'll be implementing it this week! Lmk if you'd still want this removed though

link/internal/service/attestor/resolve.go
dhfang: Hmm I don't think it's an issue to duplicate validation if that's what this comment is referring to

link/internal/relay/pipeline/pipeline.go
dhfang: Yeah the way that the batch processors work is that they will consider candidate packets in batches and if any of those packets fail, it will trim the batches but not try and re-batch the smaller batches into larger batches. So without this check packets that aren't finalized will constantly reach the batch processor and get tossed from the batch affecting the ability of the relayer to form batches. It seems like a cheap call that we can keep for now.

link/cmd/ibc/deploy.go
```diff
@@ -454,23 +454,12 @@ func renderedChain(cfg config.Config, m *manifest.Manifest) config.ChainConfig {
...
-		// params; the operator fills them in to match attestor config.
-		threshold, _ := c.Params["threshold"].(float64)
-		client.AttestorSet = &config.AttestorSetConfig{Threshold: int(threshold)}
-	}
-	return client
+func renderedClientEnd(m *manifest.Manifest, c manifest.Client) config.ClientEnd {
+	return config.ClientEnd{
+		ChainID:  m.ChainID,
+		ClientID: c.ClientID,
+		Type:     config.ClientType(c.Type),
```
gjermundgaraba, line 462: is it a problem that renderClientEnd never sets Signer?

link/internal/config/config.go
```diff
@@ -445,39 +456,58 @@ func DBConfigFromURL(url string) (DBConfig, error) {
...
-func (c AttestorConfig) Validate() error {
-	nameSet := make(map[string]struct{})
-	signerSet := make(map[string]struct{})
-
-	for i, attestation := range c.Attestations {
-		if err := attestation.Validate(); err != nil {
-			return errors.Wrapf(err, ".attestations[%d]", i)
+// Validate validates the attestors list. Allows empty.
+func (a Attestors) Validate() error {
+	localNames := make(map[string]struct{})
```
gjermundgaraba, line 456: might be ideal to just outright reject duplicate local signers here?

link/cmd/ibc/deploy.go
```diff
@@ -486,9 +475,13 @@ func renderRelayConfig(cfg config.Config, a, b *manifest.Manifest) (renderedDepl
 		if !ok || cb.CounterpartyChainID != a.ChainID || cb.CounterpartyClientID != ca.ClientID {
 			continue
 		}
-		out.Relayer.Clients = append(out.Relayer.Clients, renderedClient(a, ca), renderedClient(b, cb))
+		out.Relayer.Connections = append(out.Relayer.Connections, config.ConnectionConfig{
+			Alias:   ca.ClientID,
```
gjermundgaraba: this could be problematic? aliases must be globally unique, but client ids might not be.

link/internal/livevalidate/quorum.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return errors.Wrap(err, "attestors")
+	}
+
+	attestors := make([]attestor.Attestor, 0, len(local)+len(remote))
+	attestors = append(attestors, local...)
+	attestors = append(attestors, remote...)
+
+	for _, conn := range cfg.Relayer.Connections {
+		for _, end := range []struct {
```
gjermundgaraba: 🤖: `checkAttestorQuorum` duplicates part of relayer startup: it resolves the same attestors, loops over the same connection ends, and checks the same quorum. The two paths can drift—for example, live validation currently skips client types that startup rejects. Could live validation reuse `proofgen.NewSetFromConfig` so it checks the same setup the relayer actually uses?

link/internal/service/attestor/resolve_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+		local, remote, err := ResolveFromConfig(ctx, entries, clients, signers)
+
+		require.NoError(t, err)
+		require.Len(t, local, 1)
+		require.Equal(t, "alice", local[0].Name())
+		require.Empty(t, remote, "an unreachable remote is skipped, not fatal")
+	})
+
+	t.Run("remoteMissingGRPCSkipped", func(t *testing.T) {
```
gjermundgaraba: This is just a deslop comment from my 🤖, take it or leave it :)

🤖: `remoteMissingGRPCSkipped` and `unsupportedTypeSkipped` exercise states that `AttestorConfig.Validate` rejects before any production caller reaches `ResolveFromConfig`. These tests lock in duplicate fallback policy that can drift from config validation. Could we drop the impossible-state cases while retaining the reachable behavior: unavailable remotes are skipped and missing local chain clients fail?

link/internal/config/config.go
```diff
@@ -243,51 +248,38 @@ func (c Config) crossValidate() error {
...
 	ChainID     string
 	SignerAlias string
 }
 
-// RelayerChainSignerPairs resolves the unique (chain, signer) pairs across all
-// configured routes. A chain may appear with multiple signers when different
-// clients on it are relayed by different routes.
+// RelayerChainSignerPairs resolves the unique (chain, signer) pairs across
+// every configured connection's two client ends.
 func RelayerChainSignerPairs(c Config) ([]ChainSignerPair, error) {
```
gjermundgaraba: does this still need the error?

e2e/internal/harness/ibclink/relayer_config.go
```diff
@@ -163,142 +140,88 @@ func buildRelayerFileConfig(cfg RelayerConfig) (fileConfig, error) {
...
-		}
-		file.Relayer.Routes = append(file.Relayer.Routes, routeFileConfig{
-			SourceClient:      alias,
-			SourceSignerAlias: cfg.SignerAlias,
-			DestSignerAlias:   cfg.SignerAlias,
+		file.Relayer.Connections = append(file.Relayer.Connections, connectionFileConfig{
+			Alias:   connection.ClientA + "-" + connection.ClientB,
+			ClientA: ends[connection.ChainA+"/"+connection.ClientA],
+			ClientB: ends[connection.ChainB+"/"+connection.ClientB],
 		})
```
gjermundgaraba: `ends` contains exactly two values whose chain and client IDs are already in hand, and both are read once immediately below. Constructing each end directly is equivalent for valid inputs and removes the compound-key collision behavior:

```suggestion
		file.Relayer.Connections = append(file.Relayer.Connections, connectionFileConfig{
			Alias: connection.ClientA + "-" + connection.ClientB,
			ClientA: clientEndFileConfig{
				ChainID:  connection.ChainA,
				Signer:   cfg.SignerAlias,
				ClientID: connection.ClientA,
				Type:     "attestation",
			},
			ClientB: clientEndFileConfig{
				ChainID:  connection.ChainB,
				Signer:   cfg.SignerAlias,
				ClientID: connection.ClientB,
				Type:     "attestation",
			},
		})
```

link/internal/config/relayer.go
gjermundgaraba: thats fine!

link/cmd/ibc/deploy.go
dhfang: Yeah so the flow right now involves manually filling those out at the end, but to make this more clear, I modified render-config to output a comment next to the blank field.

link/internal/config/config.go
dhfang: removed

link/cmd/ibc/deploy.go
dhfang: https://github.com/cosmos/ibc/pull/1328/changes/28999b7e713e775f234a0bb982c2127ee336174f

gjermundgaraba (approved):

PR #1337, opened 2026-08-10
feat!: IFT Deploy Commands

Eric-Warehime:
This implements `ibc deploy` commands for GMP and IFT.

```
 deploy gmp
deploy ift --name --symbol [--owner]
 deploy ift-bridge --symbol --client-id --counterparty-ift [--send-call-constructor]
```

The `ift-bridge` command will also deploy a send call constructor contract if an address is not provided, or it is not already present in the manifests. This contract is required for the cross chain call to be properly formatted. This command also requires the counterparty IFT to have been deployed already.

These calls generally require core and a client to have been deployed for the counterparty chain already.


link/internal/deploy/steps.go
```diff
@@ -179,6 +179,257 @@ func ClientSteps(t Target, dir, chainID string, spec ClientSpec) []Step {
...
+			Name: fmt.Sprintf("send call constructor on chain %s", chainID),
+			Done: func(_ context.Context) (bool, error) {
+				if ctorOverride != "" {
+					return true, nil
+				}
+				m, err := manifest.Load(dir, chainID)
+				if err != nil || m == nil {
+					return false, err
+				}
+				return m.SendCallConstructor != "", nil
```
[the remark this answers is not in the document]
Eric-Warehime: code-less will revert on chain.

link/internal/deploy/manifest/manifest.go
```diff
@@ -14,12 +14,15 @@ const schemaVersion = 1
...
-	Clients       []Client          `json:"clients"`
-	TargetData    map[string]string `json:"targetData,omitempty"`
+	SchemaVersion       int               `json:"schemaVersion"`
+	ChainID             string            `json:"chainId"`
+	Target              string            `json:"target"`
+	Core                Core              `json:"core"`
+	Clients             []Client          `json:"clients"`
+	GMP                 *GMP              `json:"gmp,omitempty"`
+	Tokens              []Token           `json:"tokens,omitempty"`
+	SendCallConstructor string            `json:"sendCallConstructor,omitempty"`
```
dhfang: Can we restructure this to have multiple send constructors keyed by chain type? So we can easily add support for other chain types in the future without breaking changes

link/cmd/ibc/deploy.go
```diff
@@ -496,3 +521,82 @@ func deployRenderConfig(_ *cobra.Command, args []string) error {
...
+	}
+	target, err := newTarget(cmd.Context(), cfg, flagDeployChain, flagDeployDeployer, true)
+	if err != nil {
+		return err
+	}
+	spec := deploy.IFTSpec{Owner: owner, Name: flagDeployIFTName, Symbol: flagDeployIFTSymbol}
+	return planThenRun(cmd.Context(), deploy.IFTSteps(target, flagDeployManifestDir, flagDeployChain, spec))
+}
+
+func deployIFTBridge(cmd *cobra.Command, _ []string) error {
```
dhfang, line 589: Did we consider registering both sides of the connection? Deploying the ift contract itself for a single chain makes sense as one ift may be cross-registered with mutiple others, but the registration operation should always be registered in both directions.

link/cmd/ibc/deploy.go
```diff
@@ -496,3 +521,82 @@ func deployRenderConfig(_ *cobra.Command, args []string) error {
...
+	cfg, err := setupHomeWithConfig()
+	if err != nil {
+		return err
+	}
+	if flagDeployChain == "" {
+		return errors.New("--chain is required")
+	}
+	if flagDeployIFTSymbol == "" || flagDeployBridgeClientID == "" || flagDeployBridgeCounterparty == "" {
+		return errors.New("--symbol, --client-id and --counterparty-ift are required")
```
dhfang: It's a bit strange to me that we take symbol on the source chain and the address on the counterparty. They will already have to resolve the symbol that they use in one direction to an address when they register it as a counterparty in the opposite direction. This doesn't feel like a meaningful simplification to me over just using the address.

link/internal/deploy/steps.go
```diff
@@ -179,6 +179,266 @@ func ClientSteps(t Target, dir, chainID string, spec ClientSpec) []Step {
...
+		Done: func(ctx context.Context) (bool, error) {
+			m, err := manifest.Load(dir, chainID)
+			if err != nil || m == nil {
+				return false, err
+			}
+			tok, ok := m.Token(spec.Symbol)
+			if !ok {
+				return false, nil
+			}
+			// token symbol identifies the deployment; a rerun that disagrees on
```
dhfang: Why do we enforce a uniqueness constraint on the symbol?

link/internal/deploy/manifest/manifest.go
Eric-Warehime: Yeah we can restructure it. 
The complicated bit here is that the `CosmosIFTSendCallConstructor` takes token-specific args like `denom` so you need to deploy a new one per IFT whereas the EVM one is stateless so every IFT on a chain that's connecting to an EVM counterparty can re-use the same contract.

So we can just name this field EVMSendCallConstructor since the Bridge entry already has a specific address.

link/cmd/ibc/deploy.go
Eric-Warehime: Yeah using the address on the counterparty chain was meant to avoid having to have the counterparty chain as part of the deployment manifests in case you just want to operate on a single chain, but I guess in practice you always have to deploy on both sides so it doesn't really make sense to just register in one place.

Using the token as opposed to addresses _is_ meant to be a convenience thing. Potentially if we just unify all of this into a single cross chain call you can use a single token name and never need to think about addresses.

link/internal/deploy/steps.go
```diff
@@ -179,6 +179,238 @@ func ClientSteps(t Target, dir, chainID string, spec ClientSpec) []Step {
...
+							"register a new client pair",
+						spec.ClientID,
+						iftAddr,
+						chainID,
+						cp,
+						spec.CounterpartyIFT,
+					)
+				}
+				// a rerun requesting a different send-call constructor must re-register
+				ctor, err := resolveConstructor(m, ctorOverride)
```
gjermundgaraba: should this precheck avoid requiring state that the preceding step has only planned? `planThenRun` evaluates every `Done` before any `Run`. after registering once with an explicit constructor, `EVMSendCallConstructor` remains empty, so a later run without the override fails here before step 1 can populate it.

a useful regression case would be an already-registered bridge with no chain-level constructor recorded and no override: the dry-run should return a plan rather than fail its precheck.

link/cmd/ibc/deploy.go
```diff
@@ -496,3 +525,89 @@ func deployRenderConfig(_ *cobra.Command, args []string) error {
...
+	}
+	targetA, err := newTarget(cmd.Context(), cfg, flagDeployBridgeChainA, flagDeployDeployer, true)
+	if err != nil {
+		return errors.Wrapf(err, "chain %s", flagDeployBridgeChainA)
+	}
+	targetB, err := newTarget(cmd.Context(), cfg, flagDeployBridgeChainB, flagDeployDeployer, true)
+	if err != nil {
+		return errors.Wrapf(err, "chain %s", flagDeployBridgeChainB)
+	}
+	clientID := flagDeployBridgeClientID
```
gjermundgaraba: could we account for each chain’s local client id here? `deploy client` supports asymmetric pairs through `--client-id` and `--counterparty-client-id`, but this applies one id to both sides. if it exists on chain A while chain B uses another id, A’s registration is mined and saved before B fails, leaving a one-sided bridge.

this should either support separate local client ids for A and B or reject the pair before submitting either registration; a test should verify that a missing client on B cannot leave A committed.

link/cmd/ibc/deploy.go
```diff
@@ -496,3 +525,89 @@ func deployRenderConfig(_ *cobra.Command, args []string) error {
...
+	}
+	spec := deploy.IFTSpec{Owner: owner, Name: flagDeployIFTName, Symbol: flagDeployIFTSymbol}
+	return planThenRun(cmd.Context(), deploy.IFTSteps(target, flagDeployManifestDir, flagDeployChain, spec))
+}
+
+// deployIFTBridge registers both sides of an IFT bridge in one invocation:
+// contract --ift-a on --chain-a and contract --ift-b on --chain-b each register
+// a bridge over the shared client, pointing at the other. Both chains must be
+// EVM chains this tool can deploy to.
+func deployIFTBridge(cmd *cobra.Command, _ []string) error {
```
gjermundgaraba, line 589: shoulw we validate the complete two-chain request before either side can submit transactions? both targets are EVM today, but the address flags are not consistently validated or canonicalized, the client id bypasses `ValidClientID`, and equal chain ids are accepted.

link/internal/deploy/evm/wiring.go
```diff
@@ -237,5 +232,159 @@ func (d *Driver) Verify(ctx context.Context, m *manifest.Manifest) (deploy.Repor
...
+
+// IFTBridge queries the token for a bridge registered under clientID.
+func (d *Driver) IFTBridge(ctx context.Context, iftAddr, clientID string) (string, bool, error) {
+	contract, err := ift.NewContract(common.HexToAddress(iftAddr), d.backend)
+	if err != nil {
+		return "", false, err
+	}
+	bridge, err := contract.GetIFTBridge(&bind.CallOpts{Context: ctx}, clientID)
+	if err == nil {
+		return bridge.CounterpartyIFTAddress, true, nil
```
gjermundgaraba: should `IFTBridge` retain the constructor returned by `getIFTBridge` too? dropping it makes `Done` depend on a manifest bridge record that doesnt exist for external tokens and may be missing after a successful transaction followed by a failed save. those cases currently fall through.

link/internal/deploy/steps.go
Eric-Warehime: Removed.

link/cmd/ibc/deploy.go
Eric-Warehime: Changed to this :D

link/cmd/ibc/deploy.go
Eric-Warehime: Adding test 👍

link/internal/deploy/manifest/manifest.go
dhfang: > The complicated bit here is that the CosmosIFTSendCallConstructor takes token-specific args like denom so you need to deploy a new one per IFT


painge

dhfang (approved):
🚀

PR #1358, opened 2026-08-11
ci: Add 'DCO Check'

dianab-cl:


.github/workflows/dco.yml
```diff
@@ -0,0 +1,128 @@
...
+            echo "**Fix all commits in the branch**"
+            echo '```sh'
+            echo "git rebase --signoff origin/${BASE_REF:-main} && git push --force-with-lease"
+            echo '```'
+            echo ""
+            echo "_Tip: run \`git config alias.cs 'commit -s'\` so \`git cs\` always signs off._"
+          }
+
+          # Best-effort sticky comment. Never fails the run (fork PRs get a
+          # read-only token and simply skip this; the check status still blocks).
```
dianab-cl:
```suggestion

```

gjermundgaraba (commented):
im not sure if we need to manually implement this?
maybe better to use cncf's dco2? https://github.com/cncf/dco2?

PR #1362, opened 2026-08-12
refactor: centralize generated Solidity bindings

srdtrk:
## Summary

- establish `gen/go/solidity-abi` as the shared Go module for generated Solidity bindings
- move six reusable E2E bindings into that module without changing generated code
- wire the module into `e2e` with `require`/`replace` directives, alongside `link`
- update generation paths, CI path filters, caches, linting, and test targets

## Why

Generated contract code was owned by an implementation-specific E2E module even when it was reusable elsewhere. This aligns ownership with ADR-003: shared generated artifacts live under `gen/`, while implementation-specific tests remain beside their implementation.

Sibling modules are resolved with `replace` directives, as before. `go.work`/`go.work.sum` are gitignored at the repo root, mirroring `link/.gitignore`, so local workspaces stay possible without being committed.

## Stack

1. #1361 — use typed bindings for calldata
2. **#1362 (this PR)** — centralize shared generated bindings
3. #1363 — use the canonical ICS26 router binding
4. #1364 — canonicalize artifact generation and revision checks

## Validation

- generated binding module tests and lint
- full `link` test suite
- non-Docker E2E and harness tests
- E2E lint
- license-header checks
- `make check-test-apps` — regeneration reproduces the moved bindings byte for byte

Linear: FOU-1241


go.work
gjermundgaraba, line 1: no go.work committed plz

gen/go/solidity-abi/go.mod
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+
+module github.com/cosmos/ibc/gen/go/solidity-abi
+
+go 1.26.4
+
+require github.com/ethereum/go-ethereum v1.17.4
```
gjermundgaraba: nit: bump to the same verison as the rest of the repo

scripts/generate-contract-bindings.sh
gjermundgaraba, line 1: does this script belong here still?

e2e/internal/e2etest/traffic_setup.go
```diff
@@ -21,11 +21,11 @@ import (
...
 	"github.com/cosmos/ibc/e2e/internal/harness/environment"
-	"github.com/cosmos/ibc/e2e/internal/harness/environment/solidityibc/counter"
-	"github.com/cosmos/ibc/e2e/internal/harness/environment/solidityibc/iftbatchtransfershim"
-	"github.com/cosmos/ibc/e2e/internal/harness/environment/solidityibc/iftsendcallconstructor"
-	"github.com/cosmos/ibc/e2e/internal/harness/environment/solidityibc/testerc20"
 	"github.com/cosmos/ibc/e2e/internal/harness/ibclink"
+	"github.com/cosmos/ibc/gen/go/solidity-abi/counter"
+	"github.com/cosmos/ibc/gen/go/solidity-abi/iftbatchtransfershim"
+	"github.com/cosmos/ibc/gen/go/solidity-abi/iftsendcallconstructor"
+	"github.com/cosmos/ibc/gen/go/solidity-abi/testerc20"
```
gjermundgaraba, line 28: not critical, but when i see this, i'm thinking it would've looked a lot nicer if it was just "github.com/cosmos/ibc/solidity-abi/*". I'm not sure it's super critical that the package is called "gen".

.github/workflows/ibc-link-ci.yml
```diff
@@ -102,7 +108,9 @@ jobs:
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
         with:
           go-version-file: link/go.mod
-          cache-dependency-path: link/go.sum
+          cache-dependency-path: |
```
gjermundgaraba, line 110: maybe, but it's not actually used in link yet, is it?

gjermundgaraba (changes requested):

.github/workflows/ibc-link-ci.yml
gjermundgaraba: nevermind, i see it is used later in the stack

go.work
srdtrk: Done — `go.work`/`go.work.sum` are removed and now gitignored at the repo root, mirroring `link/.gitignore`, so a local workspace is still possible but never committed. `e2e/go.mod` gets its `require`/`replace` wiring back for both `link` and the new `solidity-abi` module, which is what was resolving the graph before this PR. Verified with `GOWORK=off`.

gen/go/solidity-abi/go.mod
srdtrk: Bumped `github.com/ethereum/go-ethereum` to `v1.17.5`, matching `e2e` and `link`. (The `go` directive was already `1.26.4`, same as the rest of the repo.)

solidity-abi/scripts/generate-contract-bindings.sh
srdtrk: Good catch — it does not. Moved to `solidity-abi/scripts/generate-contract-bindings.sh`, next to the module it writes into. `Makefile` was the only caller; `make test-apps` / `make check-test-apps` are updated.

e2e/internal/e2etest/traffic_setup.go
srdtrk: Agreed. The module now lives at the repo root as `solidity-abi/` with module path `github.com/cosmos/ibc/solidity-abi`, so the imports read `github.com/cosmos/ibc/solidity-abi/counter` etc. Dropping `gen/` also keeps the module path matching its repo-relative directory, and it is consistent with protobuf output already landing directly in `link/`.

srdtrk: Correction to my earlier reply on this thread: the `gen/go/solidity-abi` location is fixed by one of our internal ADRs, so I have reverted the move. Imports stay as `github.com/cosmos/ibc/gen/go/solidity-abi/*`. Happy to raise the naming separately if you want to revisit the ADR.

gen/go/solidity-abi/scripts/generate-contract-bindings.sh
srdtrk: Small correction: the script landed at `gen/go/solidity-abi/scripts/generate-contract-bindings.sh` (I briefly had it under a top-level `solidity-abi/`, now reverted to match the ADR-mandated location). Still colocated with the module it writes into, and still out of `e2e/`.

scripts/generate-contract-bindings.sh
srdtrk: Final location: `scripts/generate-contract-bindings.sh` at the repo root. `gen/` holds generated code only, so the handwritten generator does not belong beside its output — and a root `scripts/` dir suits a script whose inputs are in `e2e/` and outputs in `gen/`. Added `scripts/**` to the e2e `code` path filter so editing it still triggers the staleness check.

gjermundgaraba (approved):

PR #1368, opened 2026-08-12
feat(cli): attestor/relayer grpc query commands + IFT mint/send

dhfang:
Adds commands for interacting with the grpc api and ifts
```
ibc attestor info <name> [--host host:port]
ibc attestor latest-height <name> [--host host:port]
ibc attestor state-attestation <name> --height N [--host host:port]

ibc relayer relay --tx-hash <hash> --chain-id <id> [--host host:port]
ibc relayer status --tx-hash <hash> --chain-id <id> [--host host:port]

ibc tx ift mint --chain <id> --ift <address> --to <recipient|signer-alias> --amount <base-units> --from <signer-alias>
ibc tx ift send --chain <id> --ift <address> --client-id <id> --to <receiver|signer-alias> --amount <base-units> --from <signer-alias> [--timeout 15m]

ibc deploy show <chain-id> [--manifest-dir deployments]
```

`packet-attestation` was left out for now — constructing a valid request means hand-assembling ABI-encoded IBC packet bytes, which isn't something anyone would do by hand and it's not something people will use for the walkthrough demo

### DIFF
--- a/link/cmd/ibc/attestor.go
+++ b/link/cmd/ibc/attestor.go
@@ -3,14 +3,22 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
+	"net"
+	"net/http"
 
+	"connectrpc.com/connect"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	attestorv2 "github.com/cosmos/ibc/link/api/v2/attestor"
 	"github.com/cosmos/ibc/link/internal/bootstrap"
+	"github.com/cosmos/ibc/link/internal/config"
 	"github.com/cosmos/ibc/link/internal/pkg/graceful"
 )
+
+var flagAttestorHost string
 
 var (
 	cmdAttestor = &cobra.Command{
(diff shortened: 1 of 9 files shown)


link/cmd/ibc/attestor.go
```diff
@@ -23,8 +31,31 @@ var (
...
+
+	cmdAttestorStateAttestation = &cobra.Command{
+		Use:   "state-attestation [name]",
+		Short: "Query a local attestor for a state attestation at --height",
+		Args:  cobra.ExactArgs(1),
+		RunE:  attestorStateAttestation,
+	}
 )
 
+var flagAttestorHeight uint64
```
Eric-Warehime: nit: I like to put all cmd vars at the top in a single block, or at least above the command that's binding to them.

link/cmd/ibc/relayer.go
```diff
@@ -89,3 +110,51 @@ func relayerRun(cmd *cobra.Command, _ []string) error {
...
+	address := flagRelayerHost
+	if address == "" {
+		if cfg.Server.ListenAddress == "" {
+			return errors.New("server.listenAddr is not configured; pass --host to target a server directly")
+		}
+		address = cfg.Server.ListenAddress
+	}
+
+	client := relayerv2.NewRelayerApiServiceClient(
+		newGRPCHTTPClient(), "http://"+dialableAddress(address), connect.WithGRPC(),
```
Eric-Warehime, line 149: https?

Eric-Warehime (approved):

link/cmd/ibc/relayer.go
dhfang: Link doesn't support tls yet 😬

PR #1378, opened 2026-08-14
refactor(link): expose config package

gjermundgaraba:
## Summary

- move Link config types, loading, validation, and YAML helpers into public `link/config`
- keep CLI flags/output and filesystem helpers outside the public config package
- replace duplicated E2E config types and serialization with `link/config`

## Testing

- `make build-link`
- `make lint`
- `go test ./...` in `link/`
- `make test-harness`
- `make test-e2e`
- `git diff --check`


link/config/config.go
```diff
@@ -0,0 +1,293 @@
...
+
+	return db, db.Validate()
+}
+
+const finalityOffsetTODO = `TODO: set appropriately. 0 defaults to chain finality`
+
+// CollectComments builds TODO comments for every field in cfg that's left
+// for the operator to fill in by hand, keyed by YAML path for
+// MarshalYAMLWithComments.
+func CollectComments(cfg Config) map[string]string {
```
dhfang, line 251: I wouldn't want to make this public yet, not sure it's something we want to keep

dhfang (commented):
Is this causing alot of friction for you in E2E development? I'm not super opposed to making this public - my main concern is that breaking changes seem likely as we continue development due to multiple reasons
- `AutoRelayConfig` is already changing as the design is being fleshed out.
- We want to simplify the relayer architecture and some configs currently reflect the way the relayer is designed. I could see `DispatchPollInterval` changing. We also want to expose parts of the relayer and attestors as libraries and make it possible to define custom implementations of some of the interfaces like `ProofGenerator` for example - I think the configs will need to change to support that as well.
- We've made a best guess at choosing configs that can be extended without breaking changes to support multiple light client types and multiple chain types. As we actually extend beyond just attestations and just EVM support we may find that we require breaking changes to support additional light client and chain types or there is a more ideal config structure.

Unless it's a huge inconvenience right now, I would consider waiting till we add QBFT LC support, auto-relay and finish defining the public interface for the relayer + attestor. Lmk what you think!

PR #1394, opened 2026-08-19
feat(make): improve Makefile, deduplicate commands

swift1337:
- [x] Streamline and simplify e2e's `make` commands 
- [x] Move e2e's `make` commands to e2e/Makefile
- [x] Fix CI `make` references
- [x] Rename some `make` commands (e.g. `make check` → `make run-all-checks`)

Closes FOU-1411

### DIFF
--- a/link/Makefile
+++ b/link/Makefile
@@ -35,7 +35,7 @@ lint-fix: ## Lint the code and fix errors
 	golangci-lint run --fix
 
 test: ## Run tests
-	go test ./... 
+	go test ./... -v
 
 codegen: codegen-sql codegen-proto codegen-abi codegen-mocks ## Generate code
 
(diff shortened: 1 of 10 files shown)


link/Makefile
```diff
@@ -72,9 +64,30 @@ codegen-abi: ## Generate Go bindings for EVM contract ABIs
...
+
+run-all-checks: ## Run "all-in-one" code validation step.
+	@echo "==== 1. Linting ===="
+	$(MAKE) lint
+
+	@echo "==== 2. Checking for stale code ===="
+	$(MAKE) check-stale
+
+	@echo "==== 3. Building ===="
+	$(MAKE) build
```
gjermundgaraba, line 86: super nit: wouldn't it be more natural to just write:
```
check-all: lint check-stale build
```

gjermundgaraba (approved):

link/Makefile
swift1337: I did that initially, but it's not clear what step is being executed right now and what is slow

PR #1395, opened 2026-08-19
feat: Make signers required for ibc deploy render-config

Eric-Warehime:
Add required arguments to render config in order to fill in relay signer per chain connection being output.


gjermundgaraba (approved):
fix the lint and failed sign check and it should be good to go

PR #1396, opened 2026-08-20
feat(link): relayer packets endpoint

dhfang:
- Adds a `Packets` RPC listing packets the relayer knows about filterable by chain, client, state, source transaction, and sequence. Paginated.
- Adds `ibc relayer packets`
- Removes `Status` along with `ibc relayer status`

closes FOU-1218


link/internal/store/store_postgres.go
```diff
@@ -400,3 +400,33 @@ func (db *PostgresDB) ClearPacketTimeoutTx(ctx context.Context, key PacketKey) e
...
+	rows, err := db.repo.ListPackets(ctx, postgres.ListPacketsParams{
+		Statuses:            filter.statusList(),
+		SourceChainID:       filter.SourceChainID,
+		DestinationChainID:  filter.DestinationChainID,
+		SourceClientID:      filter.SourceClientID,
+		DestinationClientID: filter.DestinationClientID,
+		SourceTxHash:        filter.SourceTxHash,
+		SequenceNumber:      filter.sequenceFilter(),
+		Before:              page.before(),
+		RowLimit:            int32(page.Limit), //nolint:gosec // bounded by the caller's page cap
```
gjermundgaraba, line 424: the service caps this today, but `Page` is unvalidated at the store boundary. negative limits fail on postgres and are unbounded on sqlite, while values outside `int32` can wrap here. validate the page before either backend queries?

link/internal/server/relayer_handler_test.go
```diff
@@ -27,6 +28,18 @@ func (s *relayerServiceStub) Status(context.Context, string, string) ([]relayers
 	return s.status, nil
```
gjermundgaraba: `RelayerService` no longer has `Status`, so this method is dead. delete it?

e2e/internal/harness/ibclink/daemon.go
```diff
@@ -297,19 +309,24 @@ func (d *Relayer) wireChainID(harnessChainID string) (string, error) {
...
+// without a chain mapping have nothing to probe.
+func (d *Relayer) probePacketsEndpoint(ctx context.Context) error {
 	for _, chainID := range d.chainIDs {
-		_, err := d.client.Status(ctx, connect.NewRequest(&relayerv2.StatusRequest{
-			SourceChainId: chainID,
-			TxHash:        zeroTxHash,
+		response, err := d.client.Packets(ctx, connect.NewRequest(&relayerv2.PacketsRequest{
+			Filter: &relayerv2.PacketFilter{
+				SourceChainId: &chainID,
+				SourceTxHash:  &[]string{zeroTxHash}[0],
```
gjermundgaraba: this builds a one-element slice just to get a pointer. use an addressable local instead? it matches the pointer pattern elsewhere and reads more clearly.

link/internal/service/relayer/service.go
```diff
@@ -43,6 +43,7 @@ type ChainClients interface {
 type Store interface {
 	GetRelayRequest(ctx context.Context, chainID string, txHash string) (*store.RelayRequest, error)
```
gjermundgaraba: these were only used by the removed `Status` service path. drop them from this interface and regenerate `service.mock.go`? the repository methods are still used elsewhere.

link/internal/service/relayer/packets_test.go
```diff
@@ -0,0 +1,273 @@
...
+			"status %q expands under %v but reports as %v", status, state, mapPacketState(status))
+	}
+
+	require.Len(t, covered, len(store.AllRelayStatuses()),
+		"every relay status must be covered exactly once")
+}
+
+// A packet mid-pipeline is not in the literal PENDING status, and must still be
+// listed by a PENDING filter.
+func TestDBStatusesForStatePendingCoversIntermediates(t *testing.T) {
```
gjermundgaraba: `TestMapPacketState` already covers the mapping, and the `Packets` test covers expansion. delete this duplicate?

link/internal/store/list_packets_test.go
```diff
@@ -0,0 +1,228 @@
...
+			hashes[i] = packet.SourceTxHash
+		}
+
+		return hashes
+	}
+
+	str := func(v string) *string { return &v }
+
+	t.Run("listPackets", func(t *testing.T) {
+		t.Run("bySourceChain", func(t *testing.T) {
```
gjermundgaraba: these filter cases all have the same setup and assertion. the two unknown-hash cases are duplicates too. table-drive the filters and keep one no-match case?

link/internal/service/relayer/packets_test.go
```diff
@@ -0,0 +1,273 @@
...
+			}
+
+			// The cursor must name the last returned row, not the probe row,
+			// or the next page would skip it.
+			require.Equal(t, encodeCursor(rows[tt.wantPackets-1].ID), page.NextCursor)
+		})
+	}
+}
+
+func TestPacketsCapsTheLimit(t *testing.T) {
```
gjermundgaraba: this repeats the setup in the paging table. fold the cap case into `TestPacketsProbesForFurtherPages` and add the expected store limit?

link/internal/server/relayer_handler_test.go
```diff
@@ -15,8 +15,9 @@ import (
 )
 
 type relayerServiceStub struct {
-	relay  func(relayerservice.RelayRequest) error
-	status []relayerservice.PacketStatus
+	relay   func(relayerservice.RelayRequest) error
+	status  []relayerservice.PacketStatus
+	packets func(relayerservice.PacketFilter, relayerservice.PacketQuery) (relayerservice.PacketPage, error)
```
gjermundgaraba: `packets` is never set here, so the callback and branch are dead. have `Packets` return the `status` fixture directly?

link/internal/service/relayer/service.go
```diff
@@ -87,6 +88,48 @@ type PacketStatus struct {
...
+
+// PacketPage is one page of a Packets listing.
+type PacketPage struct {
+	Packets    []PacketStatus
+	HasMore    bool
+	NextCursor string
+}
+
+// normalizeLimit applies the default and cap so every listing is bounded.
+func normalizeLimit(limit int64) int64 {
```
gjermundgaraba: `normalizeLimit` has one caller, and `Packets` already tests both branches. inline it there and use `min`?

link/internal/store/list_packets_test.go
```diff
@@ -0,0 +1,228 @@
...
+		{
+			// Different source client and chain, to prove filters discriminate.
+			Status: RelayStatusAwaitingSendFinality, SourceChainID: chainTwo, DestinationChainID: chainOne,
+			SourceTxHash: "0xlist3", SourceTxTime: base, PacketSequenceNumber: 3,
+			PacketSourceClientID: "src-b", PacketDestinationClientID: "dst-b",
+			PacketTimeoutTimestamp: base.Add(time.Hour),
+		},
+	}
+
+	finalStatus := []RelayStatus{
```
gjermundgaraba: `finalStatus` repeats `packet.Status` and can get out of sync. use `packet.Status` for the transition?

link/cmd/ibc/relayer.go
```diff
@@ -47,8 +53,38 @@ var (
...
+var packetStates = map[string]relayerv2.PacketState{
+	"not-selected": relayerv2.PacketState_PACKET_STATE_NOT_SELECTED,
+	"pending":      relayerv2.PacketState_PACKET_STATE_PENDING,
+	"succeeded":    relayerv2.PacketState_PACKET_STATE_SUCCEEDED,
+	"timed-out":    relayerv2.PacketState_PACKET_STATE_TIMED_OUT,
+	"rejected":     relayerv2.PacketState_PACKET_STATE_REJECTED,
+	"relay-failed": relayerv2.PacketState_PACKET_STATE_RELAY_FAILED,
+}
+
+func packetStateNames() []string {
```
gjermundgaraba, line 78: go 1.26 already has `slices.Sorted(maps.Keys(packetStates))`. use that?

gjermundgaraba (changes requested):
Just a bunch of small stuff for the most part.

link/internal/store/migrations/sqlite/005-packet-source-tx-index.sql
```diff
@@ -0,0 +1,10 @@
+-- SPDX-License-Identifier: Apache-2.0
+
+-- +migrate Up
+
+create index if not exists index_packets_source_tx
```
swift1337: Before the v1 release, it's safe to amend the existing migration.

link/internal/store/queries/relayer.sql
```diff
@@ -139,3 +139,16 @@ WHERE status NOT IN (
     'FAILED'
 )
 ORDER BY id;
+
+-- name: ListPackets :many
+SELECT * FROM packets
+WHERE ',' || sqlc.arg(statuses) || ',' LIKE '%,' || status || ',%'
```
swift1337, line 145: can you explain what's that?

link/internal/service/relayer/service.go
```diff
@@ -87,6 +88,48 @@ type PacketStatus struct {
...
+	case limit > MaxPacketPageLimit:
+		return MaxPacketPageLimit
+	}
+
+	return limit
+}
+
+// PacketFilter narrows a Packets listing
+type PacketFilter struct {
+	SourceChainID       *string
```
swift1337: let's remove pointers

link/internal/service/relayer/pagination.go
```diff
@@ -0,0 +1,32 @@
...
+package relayer
+
+import (
+	"encoding/base64"
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+func encodeCursor(id int64) string {
```
swift1337: this file can be merged into service.go

link/internal/store/queries/relayer.sql
dhfang: Couldn't design a query that used the native sqlc list primitive and worked for both sqlite and postgres. It fails for sqlite due to this bug https://github.com/sqlc-dev/sqlc/issues/2452.  On SQLite, sqlc gives sqlc.slice() a single numbered placeholder at codegen time but expands it to N placeholders at runtime, so every parameter positioned after it binds to the wrong argument. So this takes the list of statuses from the filter adds a comma to the beginning and end i.e. `PENDING,COMPLETE` -> `,PENDING,COMPLETE,` and does the same for the status of the entry `PENDING` -> `,PENDING,` and then does a substring match. It is definitely not performant, but I think we can just address this in a followup pr if this becomes an issue. Would probably have to manually edit the generated bindings

gjermundgaraba (approved):

PR #1397, opened 2026-08-20
feat(link): relay response detail

dhfang:
- `Relay` returned `{}` on success, so callers couldn't tell a full relay from one that selected nothing; it now reports every send packet in the transaction with its selection
(`SELECTED`, `NOT_SELECTED`, or `UNCONFIGURED`). 
- Removes the `relay_requests` table. This was previously used by the /status endpoint to distinguish to users between a tx that hand't been submitted to the api and one that has no packets. We replaced /status with /packets which no longer relies on this table. I just deleted the migration, but let me know if there are any users like the sandbox this would be annoying for and I can rewrite the deletion as a new migration.

closes FOU-1344


link/internal/service/relayer/service.go
```diff
@@ -41,7 +41,6 @@ type ChainClients interface {
 
 // Store queries used by the relayer gRPC handlers.
 type Store interface {
-	GetRelayRequest(ctx context.Context, chainID string, txHash string) (*store.RelayRequest, error)
 	ListPacketsBySourceTx(ctx context.Context, chainID string, txHash string) ([]store.Packet, error)
```
gjermundgaraba: `ListPacketsBySourceTx` isn't used through this interface

link/internal/server/relayer_handler_test.go
```diff
@@ -15,13 +15,17 @@ import (
 )
 
 type relayerServiceStub struct {
-	relay   func(relayerservice.RelayRequest) error
-	status  []relayerservice.PacketStatus
-	packets func(relayerservice.PacketFilter, relayerservice.PacketQuery) (relayerservice.PacketPage, error)
+	relay       func(relayerservice.RelayRequest) error
+	relayResult relayerservice.RelayResult
+	status      []relayerservice.PacketStatus
+	packets     func(relayerservice.PacketFilter, relayerservice.PacketQuery) (relayerservice.PacketPage, error)
```
gjermundgaraba: `Status` isn't part of `RelayerService` anymore, and `packets` is never set. remove both dead stub paths and have `Packets` return the fixture directly?

link/internal/service/relayer/service.go
```diff
@@ -136,6 +135,29 @@ type PacketSelector struct {
...
+
+// ObservedPacket is a send packet in the transaction
+type ObservedPacket struct {
+	Selector  PacketSelector
+	Selection PacketSelection
+}
+
+// RelayResult reports what a relay request did with each packet in the
+// transaction
+type RelayResult struct {
```
gjermundgaraba: `RelayResult` only wraps `[]ObservedPacket`, and every caller immediately unwraps it. return the slice directly?

gjermundgaraba (changes requested):
just some smol cleanup items

gjermundgaraba (approved):

PR #1412, opened 2026-08-20
feat(relayer): Add send packet ws subscription for auto relayed routes

mattac21:
Adds a basic websocket `watcher` for clients within a connection configured with `autoRelay: true`. When a config specifics that auto relay should be enabled for a chain, a websocket url is now required. The `watcher` simply listens to sent packet events for clients it is configured on, and adds `PENDING` packets into the `packets` table. 

It handles no logic around reorgs or the "clearing" pass (i.e if the relayer goes down, we will currently miss packets sent during the downtime). That will be implemented in upcoming prs.


e2e/internal/e2etest/traffic_setup.go
mattac21, line 1: e2e updates are mostly just wiring for the websocket field. not yet adding in e2e tests doing full auto relay since many of them stop the relayer, send packets, and then start it again and expect it to relay which is not yet implemented

link/internal/relay/watcher/watcher.go
mattac21, line 1: main logic implemented here

link/internal/service/relayer/service.go
```diff
@@ -314,18 +330,24 @@ func (s *Service) Status(ctx context.Context, chainID, txHash string) ([]PacketS
 		return nil, err
 	}
 
-	switch _, errGet := s.store.GetRelayRequest(ctx, chainID, txHash); {
```
mattac21: I think this is all about to change in dennis recent pr to change /status to /packets so this probably doesn't matter to review. but currently status expects a relay request to be in the db in order to return a status, which auto relayed packets dont have, so I removed this

link/internal/relay/watcher/watcher.go
```diff
@@ -0,0 +1,199 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+
+		case event := <-events:
+			if err := w.HandleEvent(ctx, event); err != nil {
+				w.logger.Error("Recording send packet", "err", err)
+			}
```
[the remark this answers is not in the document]
mattac21: @greptile this is a no fix for now, we are going to implement a separate "clearing pass" in a later pr that will retry on these packets that the websocker watcher failed to relay. so there is no need to build in this retrying for now.

link/internal/relay/watcher/watcher.go
```diff
@@ -0,0 +1,199 @@
...
+	// canceling the per-subscription context is what releases the
+	// subscription's goroutine; unsubscribing alone leaves it running
+	subCtx, subCancel := context.WithCancel(ctx)
+	defer subCancel()
+
+	sub, err := w.subscriber.SubscribeSendPackets(subCtx, w.clientIDs, events)
+	if err != nil {
+		w.logger.Error("Subscribing to send packets failed", "err", err)
+		return
+	}
```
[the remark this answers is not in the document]
mattac21: fixed

link/internal/relay/watcher/watcher.go
```diff
@@ -0,0 +1,199 @@
...
+
+	w.logger.Debug(
+		"Send packet observed",
+		"clientID", event.Packet.SourceClient,
+		"sequence", event.Packet.Sequence,
+		"txHash", event.TxHash,
+	)
+
+	row := packetRow(w.chainID, w.destChains[event.Packet.SourceClient], event)
```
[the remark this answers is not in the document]
mattac21: fixed

link/internal/bootstrap/bootstrap.go
```diff
@@ -110,8 +111,14 @@ func BuildRelayer(cfg config.Config) (*Services, error) {
 	}
 	dispatcher := dispatch.NewRelayDispatcher(db, pipelines, pollInterval, logger)
 
+	// Packet discovery
```
swift1337:
```suggestion
	// Packet discovery (auto-relaying)
```

link/internal/relay/watcher/watcher.go
```diff
@@ -0,0 +1,239 @@
...
+
+	w.cancel()
+	<-w.stopped
+
+	return nil
+}
+
+func (w *Watcher) run(ctx context.Context, stream stream) {
+	defer close(w.stopped)
+	defer stream.close()
```
swift1337, line 147: let's add panic recovery here

swift1337 (approved):

link/internal/relay/watcher/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+const eventBuffer = 128
+
+// Subscriber the chain-side event stream.
+type Subscriber interface {
+	SubscribeSendPackets(ctx context.Context, clientIDs []string, out chan<- v2.PacketEvent) (v2.Subscription, error)
+}
+
+// PacketStore the persistence discovered packets are written to.
+type PacketStore interface {
+	UpsertPacket(ctx context.Context, input store.UpsertPacket) error
```
dhfang, line 27: The upsertpacket query was written in a very specific way for the relay endpoint. The insert on non-existence always happens but I believe the update only happens when the packet row is in the NOT_SELECTED state. Can we double check it's appropriate for the watcher's purpose?

link/internal/relay/watcher/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case event := <-stream.events:
+			if err := w.HandleEvent(ctx, event); err != nil {
+				w.logger.Error("Recording send packet", "err", err)
+			}
+		case err := <-stream.sub.Err():
```
dhfang, line 165: Do we plan to have any logic that attempts to recover from a failure here? Right now it seems like errors that happen in the subscription loop get bubbled up and then cause the watcher to stop without any recovery.

PR #1415, opened 2026-08-21
feat(cli): remote prover

dhfang:
## Summary

- Adds support for a remote gRPC prover.
- Renames `ProofGenerator` to `Prover` throughout.

## The contract

`proto/link/prover.proto` mirrors the relayer's internal prover interface
method for method, with the client named on every request so one service can
serve many clients across many chains:
```
LatestProvableHeight(Client)                -> height, timestamp
StateProof(Client, height)                  -> proof
PacketProofs(Client, height, kind, packets) -> proofs[]
```

Configuration:

```yaml
clientA:
  chainId: "1"
  clientId: link-1-2
  signer: relayer
  type: remote
  params:
    url: http://prover:9090

### CODE REVIEW

File: cli/internal/config/relayer.go [hunk 1]
```diff
@@ -64,11 +68,71 @@ type ClientEnd struct {
 	ClientID string     `yaml:"clientId"`
 	Type     ClientType `yaml:"type"`
 
+	// Params is this client type's settings.
+	Params yaml.RawMessage `yaml:"params,omitempty"`
```
dhfang, line 71: The format of this is dependent on the type of light client, so we don't need to fill the ClientEnd config with tons of fields as we add support for more light clients.

cli/internal/config/relayer.go
```diff
@@ -64,11 +68,71 @@ type ClientEnd struct {
...
+
+// AttestationParams is empty
+type AttestationParams struct{}
+
+func (*AttestationParams) isClientParams() {}
+
+func (*AttestationParams) Validate() error { return nil }
+
+// RemoteParams is the params block a remote client declares.
+type RemoteParams struct {
```
dhfang, line 92: Defines how `ClientEnd.Params` is formed. I'm defining these in config.go for now while we only have two light clients, but could move these definitions to be besides the light client implementations eventually and have the `type` and `params` registered on startup similar to the other PoC.

proto/link/prover.proto
```diff
@@ -0,0 +1,89 @@
...
+}
+
+message LatestProvableHeightRequest {
+  Client client = 1;
+}
+
+message LatestProvableHeightResponse {
+  uint64 height = 1;
+  // Unix seconds, the counterparty-chain timestamp of height.
+  int64 timestamp = 2;
```
swift1337: maybe also uint64 for consistency?

proto/cli/prover.proto
```diff
@@ -0,0 +1,89 @@
+// SPDX-License-Identifier: Apache-2.0
```
dhfang, line 1: This is pretty much identical to the internal interface except it specifies the chain and client in each request

cli/internal/testutil/proverservice/cmd/main.go
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: Apache-2.0
```
dhfang, line 1: Exposing a remote prover for e2es via a binary - relies on internal attestations provers

cli/internal/config/relayer.go
swift1337: What parameters are available for configuration when using a remote prover? (Potentially).

cli/internal/config/relayer.go
```diff
@@ -14,6 +15,8 @@ type ClientType string
 // Client types
 const (
 	ClientTypeAttestation ClientType = "attestation"
+	// ClientTypeRemote delegates proof generation to a remote service
+	ClientTypeRemote ClientType = "remote"
```
swift1337, line 19: Let's update the example ibc.yml to showcase this scenario

swift1337 (approved):

cli/internal/config/relayer.go
dhfang: They are defined here
https://github.com/cosmos/ibc/pull/1415/changes#diff-c2870bd694c64d502f6618374a7edcd10c9cbf6e726ce307bb75627b354f5498R92-R95

PR #1431, opened 2026-08-29
feat(cli): render config improvements

dhfang:
- new `--populate-config` flag writes the result to the config file. Without `--yes` it prints the config it would write, lists what that changes, and asks before writing.
- Default output is now the whole config, the existing one with the settings for relaying between the two chains added, rather than a fragment to copy-paste.

closes fou-1473


cli/internal/config/patch.go
```diff
@@ -0,0 +1,122 @@
...
+
+package config
+
+import (
+	"reflect"
+	"slices"
+)
+
+// Patch is the set of config sections projected out of deployment manifests.
+type Patch struct {
```
swift1337, line 11: This struct seems redundant as it's a subset of `Config{}`

cli/internal/config/patch.go
```diff
@@ -0,0 +1,122 @@
...
+	Kind string
+	ID   string
+}
+
+func (c Conflict) String() string {
+	return c.Kind + " " + c.ID
+}
+
+// WithPatch returns c with p merged in, alongside the entries it overwrites.
+func (c Config) WithPatch(p Patch) (Config, []Conflict) {
```
swift1337, line 28: Consider refactoring to 
```suggestion
func (c Config) Merge(override Config) (Config, []Conflict) {
```

swift1337: `merge.go`

cli/internal/config/patch.go
dhfang: This patching functionality doesn't support all config fields yet, so we just define the subset that we do support in this Patch struct.

cli/internal/config/patch.go
dhfang: Hmm I'd make this change if I didn't think there was still a reason for keeping the Patch struct as I mentioned above

swift1337 (approved):

PR #1440, opened 2026-09-02
feat(cli)!: config hardening

swift1337:
- [x] streamline config package
- [x] restrict chain config with `evm == nil`
- [x] restrict yaml unknown fields by default (no more `--strict`)
- [x] add yaml path to errors for better visibility
- [x] add more cross-validation checks
- [x] add `RelayerSufficiency()` and `AttestorSufficiency()` which simplify checking that config is actually "runnable" e.g. there's at least one chain or connection configured 

```
ibc config validate  --help
Validates config's correctness.
Optionally, if [target] is provided ["relayer", "attestor"], it performs additional checks
to ensure the config is sufficient to run:
- ibc relayer run
- ibc attestor run

Usage:
  ibc config validate [target: relayer|attestor] [flags]

Flags:
  -h, --help   help for validate
      --live   extra validation checks

Global Flags:
      --config string   Config file relative to home (default "ibc.yml")
      --db string       Database URL override
      --home string     IBC home directory (default "~/.ibc")
      --log-json        Enable JSON logging
  -q, --quiet           Quiet mode
```

 

Closes FOU-1413
Closes FOU-1259

PR #1442, opened 2026-09-04
docs: update the reference generation script

evanorti:
Follow-up to #1440. This fixes the docs generator. The generator reads validation rules to decide required-ness, enum values, and the attestor/signer tables — when it can't read how errors are built it harvests nothing.

- Refuse when `Validate()` builds errors the scraper doesn't read
- Recognise any receiver name, not just `c`
- Test that any struct with `Validate()` yields a rule, or is listed as ruleless
- Re-anchor two mutation tests that had silently stopped running
- Regenerate `6-cli-commands.md`, stale on main
- Repoint four citations whose line numbers moved
- `AGENTS.md` notes in `cli/` and `proto/`; ignore `__pycache__`


dhfang (approved):
I'm going to approve this for now, but I think the doc gen strategy is incredibly brittle as it relies on parsing specific keywords out of the config. They depend on implementation details like helper function names that are likely to change frequently and at any time, which creates additional burden on developers to update this doc gen script. If this becomes a burden we'll need to revisit this strategy or consider removing these doc gen scripts entirely.

PR #1446, opened 2026-09-04
feat: Log additions to relay

Eric-Warehime:
Adds some logging to the attestor/relayer.

Most of this PR is just threading the logger through, with a couple of additions of info/warn and a few debug logs.

### DIFF
--- a/cli/internal/relay/prover/attestation/generator.go
+++ b/cli/internal/relay/prover/attestation/generator.go
@@ -4,6 +4,7 @@ package attestation
 
 import (
 	"context"
+	"log/slog"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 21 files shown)


cli/internal/relay/prover/attestation/generator.go
```diff
@@ -23,18 +24,29 @@ type Generator struct {
...
+		attestors:         attestors,
+		threshold:         threshold,
+		counterpartyChain: counterpartyChain,
+		logger:            logger,
+	}
 }
 
 func (g *Generator) LatestProvableHeight(ctx context.Context) (uint64, time.Time, error) {
-	return latestProvableHeight(ctx, g.attestors, g.threshold, g.counterpartyChain)
+	return latestProvableHeight(ctx, g.logger, g.attestors, g.threshold, g.counterpartyChain)
```
swift1337, line 45: For such cases I'd do someling like this:
// prev code sets logging.Context(ctx, g.logger) - attaches slog to ctx, 
// log.FromContext() returns attached logger or slog.Default()
```suggestion
	return latestProvableHeight(ctx, g.attestors, g.threshold, g.counterpartyChain)
```

swift1337 (approved):

cli/internal/relay/prover/attestation/generator.go
Eric-Warehime: Mmmm, I tried it out and it's basically the same level of complexity just in a different place. Yes, you keep the logger out of the function signature in these cases, but it's inconsistent with the way the logger is passed in other places and you have to maintain the custom logger context stuff. I think we should just keep it as is.

PR #1448, opened 2026-09-07
feat(relayer): add metrics

swift1337:
Adds OpenTelemetry-based metrics to the IBC CLI relayer and attestor.

- [README](https://github.com/cosmos/ibc/blob/feat/metrics-101/cli/internal/otel/README.md) for metrics
- [README](https://github.com/cosmos/ibc/blob/feat/metrics-101/cli/scripts/otel/README.md) for sample infra setups

---

- [x] `observability.type: simple` exposes Prometheus metrics at `http://127.0.0.1:9090/metrics`.
- [x] `observability.type: otel` load OTPL config from a file/env variable
- [x] Includes a local Grafana LGTM stack for the `simple` endpoint.

| Category   | Metrics                                        |
| ---------- | ---------------------------------------------- |
| Go runtime | Memory, GC, goroutines, processor limits       |
| ConnectRPC | Request duration, size, messages, status       |
| Attestor   | Operation duration/results, latest height      |
| Signer     | Signing duration/results by signer and type    |
| Prover     | Operation duration/results, height, batch size |
| Dispatch   | Pending packets by route                       |
| Pipeline   | State transitions, batch results and sizes     |
| Processors | Completed relays, duration, retries            |
| Watcher    | Packet-send and acknowledgement events         |

**Notes**
- tx submissions and gas metrics will be added in follow-up RPs

Closes FOU-1214


cli/internal/otel/otel.go
```diff
@@ -0,0 +1,185 @@
...
+	if !cfg.Metrics {
+		logger.Info("Observability is disabled")
+		return
+	}
+
+	provider, err := New(ctx, cfg, logger)
+	if err != nil {
+		logger.Warn("Failed to setup observability. Skipping", "err", err)
+		return
+	}
```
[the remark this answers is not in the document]
swift1337: This is intentional when the port is already allocated. It's better to log a warning and disable the metrics rather than crash the startup.

cli/internal/relay/pipeline/metrics.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return &instrumentation{
+		PacketsTotal: packetsTotal,
+		BatchesTotal: batchesTotal,
+		BatchSize:    batchSize,
+	}, nil
+}
+
+func (m *instrumentation) packetTransition(ctx context.Context, tr *processors.Transfer, state store.RelayStatus) {
+	m.PacketsTotal.Add(ctx, 1, otel.WithAttributes(
```
gjermundgaraba, line 48: on a normal recv, `BatchRecvPacket` and `RetryRecvPacket` both report `RelayStatusDeliverRecvPacket`, so `packets_total` counts one transition twice even without a retry. ack and timeout do the same.

```suggestion
	if tr.Status == state {
		return
	}

	m.PacketsTotal.Add(ctx, 1, otel.WithAttributes(
```

cli/internal/relay/processors/retry_recv_packet.go
```diff
@@ -47,6 +47,8 @@ func (p RetryRecvPacket) Process(ctx context.Context, tr *Transfer) (*Transfer,
 		return nil, errors.Wrapf(err, "clearing recv tx %s", *tr.RecvTxHash)
 	}
 
+	metrics.transactionRetry(ctx, tr, relayTypeSendToRecv)
```
gjermundgaraba: when a batch transaction fails or expires, each packet sharing its `RecvTxHash` increments `transaction_retries_total`, so the retry count scales with batch size. ack and timeout do the same.

cli/internal/relay/watcher/metrics.go
```diff
@@ -0,0 +1,50 @@
...
+		otel.AttrType.String(eventType),
+	))
+}
+
+func eventType(kind v2.EventKind) (string, bool) {
+	switch kind {
+	case v2.KindSendPacket:
+		return "send_packet", true
+	case v2.KindWriteAck:
+		return "write_ack", true
```
gjermundgaraba: `Watcher.subscribe` only calls `SubscribeSendPackets`, whose EVM implementation filters for send packets and emits only `KindSendPacket`, so this branch is unreachable and `watcher_events_total{type="write_ack"}` is never emitted even when acknowledgements are written.

cli/internal/otel/otel.go
```diff
@@ -0,0 +1,221 @@
...
+	case len(bytes.TrimSpace(bz)) == 0:
+		return nil, nil, fmt.Errorf("OTEL config file %q is empty", path)
+	}
+
+	otelConfig, err := otelconf.ParseYAML(bz)
+	if err != nil {
+		return nil, nil, fmt.Errorf("parse OTEL config file %q: %w", path, err)
+	}
+
+	sdk, err := otelconf.NewSDK(otelconf.WithOpenTelemetryConfiguration(*otelConfig))
```
gjermundgaraba, line 202: with `OTEL_CONFIG_FILE=''`, `ConfigFile()` accepts a valid `otelFile`, but `NewSDK` reads the environment again and tries to open an empty filename, disabling metrics.

cli/internal/config/config.go
```diff
@@ -73,6 +87,15 @@ type DBConfig struct {
 	URL  string `yaml:"url"`
 }
 
+// Observability config for metrics and tracing.
+// note: in future we'll add `tracing: true` if needed.
```
dhfang, line 102: I think it's likely we add tracing the future and we can structure this config block now to let us do that without breaking changes. I'm not aware of any standard tracing setup where traces are scraped from the service, so I don't think ListenAddress or observability.type=simple are relevant to both tracing and metrics. We could consider moving metrics configs under an observability.metrics block or rename existing fields to make it more clear which ones are metrics specific.

cli/internal/relay/dispatch/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// update the map for the next call
+	m.routesFromPrevCall = make(map[routeKey]struct{}, len(counts))
+	for key := range counts {
+		m.routesFromPrevCall[key] = struct{}{}
+	}
+}
+
+func (m *instrumentation) recordPending(ctx context.Context, key routeKey, count int64) {
+	m.PacketsPending.Record(ctx, count, otel.WithAttributes(
```
dhfang, line 108: nit: The way this metric is currently collected, we don't register a pending number for a route until this helper sees a packet on that route. It would be nice if it recorded 0 for a route that it hasn't encountered a packet for yet.

cli/internal/relay/dispatch/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+		otel.AttrChainID.String(key.chainID),
```
dhfang: nit: it's a bit strange to pair the labels `chain_id` and `dest_chain_id` imo, would prefer identifying source explicitly as `source_chain_id`

cli/internal/relay/pipeline/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/cosmos/ibc/cli/internal/otel"
+	"github.com/cosmos/ibc/cli/internal/relay/processors"
+	"github.com/cosmos/ibc/cli/internal/store"
+)
+
+type instrumentation struct {
+	PacketsTotal metric.Int64Counter
+	BatchesTotal metric.Int64Counter
+	BatchSize    metric.Int64Histogram
```
dhfang, line 17: I believe the histogram already results in a `batch_size_count` metric, so I think we can consolidate BatchesTotal and BatchSize

cli/internal/relay/processors/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (m *instrumentation) txSubmitted(ctx context.Context, chainID, clientID, txHash string) {
+	key := txKey(chainID, txHash)
+	m.submittedTransactions.Store(key, struct{}{})
+	m.TransactionsSubmitted.Add(ctx, 1, txMetricAttributes(chainID, clientID))
+}
+
+func (m *instrumentation) txConfirmed(ctx context.Context, chainID, clientID, txHash string) {
+	key := txKey(chainID, txHash)
+	if _, submitted := m.submittedTransactions.LoadAndDelete(key); !submitted {
```
dhfang, line 101: It seems like we're using the submittedTransactions map to dedupe when multiple packets are carried by the same tx, but this tactic doesn't work if the service restarts between `txSubmitted` and `txRetry`/`txConfirmed`.

dhfang (commented):
Feel free to skip the nits

cli/internal/relay/processors/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+
+type relayType string
+
+const (
+	relayTypeSendToRecv    relayType = "send_to_recv"
+	relayTypeSendToTimeout relayType = "send_to_timeout"
+	relayTypeRecvToAck     relayType = "recv_to_ack"
+)
+
+type instrumentation struct {
```
dhfang, line 25: We had an excessive relay latency metric that triggered whenever a transfer had been pending for some duration. While the new metrics do include latency metrics we don't have anything we can use to identify when a transfer has failed to relay for some time. This was a pretty useful metric and one I'd like to carry forward

cli/internal/txsubmitter/evm/metrics.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+type instrumentation struct {
+	// EVMGasSpent is the total amount of gas spent on EVM chains in token (wei/10^18).
+	// Note that this value is set per {chain_id, wallet} pair via OTEL observer.
+	// This helps to convert cumulativeGasCostWei's bigInt to float64 only when requested by telemetry.
+	EVMGasSpent metric.Float64ObservableCounter
+	GasBalance  metric.Float64ObservableGauge
+
+	txOwners             map[transactionKey]string
```
dhfang: Similar to the in memory maps used in the processor metrics, this data gets lost on restarts. How does that impact metrics collection?

cli/internal/txsubmitter/evm/metrics.go
```diff
@@ -0,0 +1,206 @@
...
+	instruments := &instrumentation{
+		EVMGasSpent:          evmGasSpent,
+		GasBalance:           gasBalance,
+		txOwners:             make(map[transactionKey]string),
+		cumulativeGasCostWei: make(map[chainWallet]*big.Int),
+		clients:              make(map[chainWallet]balanceClient),
+		lastObservation:      make(map[chainWallet]time.Time),
+	}
+
+	_, err = m.RegisterCallback(instruments.observeEVMGasSpent, evmGasSpent)
```
dhfang, line 81: Did not know this was a thing, nice.

dhfang (commented):
Another useful place to collect metrics would be around attestor aggregation in the relayer. It would be useful to identify if some misbehaving attestor disagreed with the quorum for example. Happy to make a ticket for this and address this later as you've already created a bunch of metrics.

cli/internal/relay/dispatch/metrics.go
swift1337: This is done on purpose to avoid three chain labels (chai_id, src_chain_id, dest_chain_id). I'll add a note that the convention is that chain_id and client_id also mean source_chain_id, souce_client_id

cli/internal/relay/processors/metrics.go
swift1337: >  but this tactic doesn't work if the service restarts between

yeah, that's why there is a guard with `m.submittedTransactions.LoadAndDelete(key)`. 

There's no clear way to tackle restarts; all metrics are dropped on restart (common practice). Technically, if we notice the need to make it more reliable in the future, we could offload in-progress metrics to a WAL or DB; for now, I'd keep it as is.

cli/internal/relay/processors/metrics.go
swift1337: I initially ported it as well, but then dropped it because it was redundant. I'll bring it back

cli/internal/txsubmitter/evm/metrics.go
swift1337: https://github.com/cosmos/ibc/pull/1448#discussion_r3979337252

cli/internal/relay/processors/metrics.go
dhfang: Yeah the prometheus metrics are cleared on restart since they are in-memory but I don't see why that has to be true for the underlying data used to derive those metric values. I don't think it's at all unlikely for restarts to occur while there are one or more transactions pending verification, and in those cases `txConfirmed`/`txRetry` just no-ops and misses valid confirmations since the `submittedTransactions` map has been cleared. I'll create a ticket in the backlog to address this.

cli/internal/txsubmitter/evm/metrics.go
dhfang: https://github.com/cosmos/ibc/pull/1448#discussion_r3985119812

I think for the same reasons mentioned in this comment the current implementation is not ideal. IIUC the gas costs of transactions are calculated at the scrape interval and if the restart happens while gas costs are pending calculation, it just completely drops the pending gas calculation. An unexpected spike in gas costs is something I would want to create alerts on as an operator, but observations can easily be dropped around restarts in the current implementation. Doesn't need to be addressed now, I'll create a ticket.

dhfang (approved):

cli/internal/txsubmitter/evm/metrics.go
```diff
@@ -0,0 +1,206 @@
...
+
+	for key, client := range clients {
+		addr := common.HexToAddress(key.wallet)
+		attr := metric.WithAttributes(
+			otel.AttrChainID.String(key.chainID),
+			otel.AttrWallet.String(key.wallet),
+		)
+
+		eg.Go(func() error {
+			balance, err := client.BalanceAt(ctx, addr, nil)
```
gjermundgaraba: in `simple` mode, this RPC has neither a context deadline nor a client timeout. A stalled node blocks collection indefinitely, causing this and subsequent Prometheus scrapes to hang, including unrelated metrics.

cli/internal/relay/dispatch/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+	switch p.Status {
+	case store.RelayStatusCompleteWithAck,
+		store.RelayStatusCompleteWithWriteAckError,
+		store.RelayStatusCompleteWithTimeout,
+		store.RelayStatusFailed:
+		return "", false
+	}
+
+	if p.WriteAckTxHash == nil {
+		if now.After(p.PacketTimeoutTimestamp) {
```
gjermundgaraba: when a received packet's write-ack lookup fails, `WriteAckTxHash` remains nil, so this reports excessive `send_to_timeout` latency five minutes after expiry. The packet cannot time out, and neither 60-minute relay threshold may have elapsed.

gjermundgaraba (changes requested):

cli/internal/txsubmitter/evm/metrics.go
```diff
@@ -0,0 +1,206 @@
...
+	var (
+		now     = time.Now()
+		clients = make(map[chainWallet]balanceClient)
+		eg      errgroup.Group
+	)
+
+	m.mu.Lock()
+	for key, client := range m.clients {
+		// To prevent live frequent RPC calls, skip if observed within the threshold.
+		if now.Sub(m.lastObservation[key]) < observationThreshold {
```
gjermundgaraba: a scrape within ten seconds of the last balance query omits that wallet's `evm_gas_balance`, causing Prometheus to mark it stale. Another scraper can trigger this even when the main scraper polls less frequently.

gjermundgaraba (changes requested):

cli/internal/txsubmitter/evm/metrics.go
```diff
@@ -0,0 +1,223 @@
...
+
+	return nil
+}
+
+func (m *instrumentation) observeGasBalance(ctx context.Context, observer metric.Observer) error {
+	now := time.Now()
+	wg := sync.WaitGroup{}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
gjermundgaraba: this holds the shared metrics mutex until all balance RPCs finish, while `Submit` and `ShouldRetry` need the same mutex for transaction tracking. A stalled balance endpoint can therefore delay both operations across every EVM chain for three seconds per scrape, repeating on subsequent scrapes because failed queries don't update the cache timestamp.

gjermundgaraba (changes requested):
one last one sry

cli/internal/txsubmitter/evm/metrics.go
swift1337: updated

gjermundgaraba (approved):

PR #1461, opened 2026-09-14
feat(relayer): auto-relay

swift1337:
This PR adds support for auto-relay based on WebSocket connections and a retrospective, ticker-based clearing function. **It follows ADR-012**

**E2E tests**

- `TestTransfer_AutoRelay`
- `TestIFTTransfer_AutoRelay`
- `TestGMPCall_AutoRelay`
- `TestAutoRelay_PacketIsClearedOnRestart`
- `TestAutoRelay_AllPacketsAreClearedOnStart`
- `TestAutoRelay_SubscriptionReconnect`

---

**How to review this PR**

1. View config changes (autoRelay, clearOnStart, clearInterval)
2. View SQL updates (new tables, new queries)
3. View E2E tests
4. View `watcher.go`, `clearer.go`

---

**Original stack**:

1. #1452
5. #1453 
6. #1454
7. #1455
8. #1456
9. #1457

- [x] Address Greptile findings from the above

Closes FOU-385
Closes FOU-1590


cli/internal/relay/watcher/clearer.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+}
+
+// Clear writes a packet row for every outstanding commitment on clientID we
+// hold no row for, probing the sequences above the watermark plus whatever the
+// last pass could not resolve. Sequences we already know about are skipped
+// whatever state they are in, terminal and deselected ones included: clearing
+// discovers packets, it does not reopen decisions the pipeline or an operator
+// has already made.
+func (c *Clearer) Clear(ctx context.Context, clientID string) (Result, error) {
```
swift1337, line 90: Could you take a look one more time? @mattac21

e2e/internal/harness/chain/evm/wsproxy.go
```diff
@@ -0,0 +1,162 @@
...
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// WebSocketProxy forwards TCP between a local listener and an upstream ws:// host:port.
+// Kill() closes active connections and rejects new ones until Revive() is called.
+type WebSocketProxy struct {
```
swift1337, line 17: This is needed to exercise WebSocket connection disruption. 

I tried to use container (un)pause and Docker network disconnect/connect, but on my machine the container runtime keeps TCP connections. And this test-driven proxy would work on all architectures.

cli/internal/chains/evm/client.go
```diff
@@ -337,6 +340,196 @@ func (c *Client) commitmentExists(ctx context.Context, clientID string, sequence
 	return commitment != [32]byte{}, nil
 }
 
+// prevSequenceSends is the second field of the ibc.storage.IBCStore ERC-7201
+// namespace, whose base is
+// keccak256(uint256(keccak256("ibc.storage.IBCStore")) - 1) & ~0xff.
+var prevSequenceSendsSlot = common.BigToHash(new(big.Int).Add(
```
swift1337, line 346: we'll move to evm getter in followups

cli/internal/relay/watcher/clearer.go
```diff
@@ -0,0 +1,422 @@
...
+// ClearStore the persistence a clearing pass reads and writes.
+type ClearStore interface {
+	PacketStore
+
+	MaxPacketSequence(ctx context.Context, chainID string, clientID string) (uint64, error)
+	ListPacketSequencesFrom(ctx context.Context, chainID string, clientID string, fromSequence uint64) ([]uint64, error)
+	GetClearingState(ctx context.Context, chainID string, clientID string) (store.ClearingState, error)
+
+	// Transact carries the pass's rows and its watermark, so a watermark is never
+	// persisted on the strength of rows that did not land.
```
dhfang, line 37: This one's a bit too claude-y for me

```suggestion
	// Transact saves the clearing pass's recovered packets and watermark atomically.
```

cli/internal/relay/watcher/clearer.go
```diff
@@ -0,0 +1,422 @@
...
+		c.logger.Warn(
+			"The chain reports fewer sends than we hold packets for. Either the rpc endpoint is "+
+				"behind the one the subscription reads, the client was reset, or the IBCStore "+
+				"storage layout has moved",
+			"clientID", clientID,
+			"chainSequence", latest,
+			"recordedSequence", recorded,
+		)
+	}
+
```
dhfang, line 127: Since the websocket subscription runs asynchronously to this clearing pass, this case seems easy to hit in normal operation. The weboscket subscription can discover new packets after the header query above that is used to pin subsequent requests. We can also get rid of the `MaxPacketSequence` query

```suggestion
```

cli/internal/relay/watcher/clearer.go
```diff
@@ -0,0 +1,422 @@
...
+		return result, errors.Wrapf(err, "reading the clearing state for client %s", clientID)
+	}
+
+	from := state.LastProbed + 1
+
+	// abandoned sequences are carried rather than probed: dropping them from the
+	// probe is what bounds the pass, and keeping them on record is what lets an
+	// operator who later has an archive endpoint recover them by turning the
+	// setting back off.
+	// ponytail: the abandoned set is read every pass and thrown away; make the
```
dhfang, line 139: What is ponytail?

cli/internal/relay/watcher/clearer.go
```diff
@@ -0,0 +1,422 @@
...
+// The height it returns is the one the carried set was read at, which is what
+// the store weighs a resolve against: resolving a sequence read absent at a
+// stale height on the strength of a height a later chunk raised would forget a
+// packet that is still live.
+func (c *Clearer) calcOutstanding(
+	ctx context.Context,
+	clientID string,
+	unresolved []uint64,
+	from, latest, height uint64,
+) ([]uint64, int, uint64, error) {
```
dhfang, line 221: Naming these return values would make this a bit more self documenting and easy to read

cli/internal/relay/watcher/clearer.go
```diff
@@ -0,0 +1,422 @@
...
+			)
+		}
+		return nil
+	})
+}
+
+// unresolvedDelta is the change a pass makes to the unresolved set: what it
+// found outstanding with no send log, and what it probed and no longer needs to
+// remember. Sequences absent from both are left alone.
+func unresolvedDelta(probed, unresolved []uint64, height uint64) store.UnresolvedDelta {
```
dhfang, line 388: Could we rename the arguments to something like `previousUnresolved`, `currentUnresolved`? `probed` and `unresolved` made this pretty difficult to understand.

dhfang (commented):
Alot of these comments are pretty difficult to read. They use language I would never expect a human to use in programming comments. If it's possible to easily do a cleanup pass, can we do so? Like maybe just have an agent do a deslop pass at the comments.

cli/internal/relay/watcher/watcher.go
```diff
@@ -232,7 +218,211 @@ func (w *Watcher) HandleEvent(ctx context.Context, event v2.PacketEvent) error {
...
+		// whatever the dropped one had already buffered
+		eventsChan = eventStream.events
+	)
+
+	clearTick := time.NewTicker(w.cfg.ClearInterval)
+	defer clearTick.Stop()
+
+	// combines live events with periodically triggered clearing
+	// handles errors, re-subscriptions, and context cancellation
+	for {
```
dhfang, line 302: This run function and this loop are doing alot and I don't look forward to debugging it in the future. I wouldn't be against splitting this into a thread processing subscription events and a thread running clearing loops periodically. I'd be fine with dropping the ability to trigger a clearing loop on resubscription and just reducing the default clearing interval instead.

cli/internal/relay/watcher/clearer.go
```diff
@@ -0,0 +1,422 @@
...
+	// operator who later has an archive endpoint recover them by turning the
+	// setting back off.
+	// ponytail: the abandoned set is read every pass and thrown away; make the
+	// query aware of the setting if anyone parks enough of them to notice.
+	reprobe := state.Unresolved
+	if c.abandon {
+		reprobe = nil
+	}
+
+	outstanding, probed, probeHeight, err := c.calcOutstanding(ctx, clientID, reprobe, from, latest, head.Height)
```
dhfang, line 146: After the clearing loop determines the range of sequence numbers between the last watermark and the latest on chain sequence number it can query the database first before doing more expensive on-chain queries. Under normal operation, the packets will typically already be in the database from the subscription loop.

e2e/relayer_lifecycle_test.go
```diff
@@ -4,14 +4,174 @@ package e2e_test
...
+	// Given packet1 sent and auto-relayed
+	transfer1 := mustSend(t, transferApp, big.NewInt(100_000))
+	mustDeliver(t, relayer, transfer1)
+
+	// ACT #1: break the relayer's websocket subscription through the proxy.
+	chainWSProxy.Kill()
+	time.Sleep(time.Second)
+
+	// ACT #2: restore the proxy; the relayer resubscribes and triggers ws reconnection + clearing pass.
+	chainWSProxy.Revive()
```
dhfang, line 167: We can also send a packet during the websocket outage before revive and verify it gets relayed.

cli/internal/relay/watcher/clearer.go
swift1337: It's inside calcUnrecorded

cli/internal/relay/watcher/watcher.go
```diff
@@ -232,126 +239,100 @@ func (w *Watcher) subscribe(ctx context.Context, events chan v2.PacketEvent) (st
 	return stream{sub, events, cancel}, nil
 }
 
-func (w *Watcher) run(ctx context.Context, eventStream stream) {
+// connects to live events and recovers broken subscriptions.
+func (w *Watcher) runSubscription(ctx context.Context, eventStream stream) {
 	defer func() {
```
dhfang, line 244: The current panic recovery behavior does not involve reconnecting to the sub. Even if it was intermittent, now the relayer will keep running since the panic is intercepted but it will not identify new packets. I think it's probably better not to suppress the panic so that:
- the service can restart and recover from intermittent failures
- operator configured alerts on service health get triggered

Or we can add logic to resubscribe on panic and also increment some metric that captures the panic.

cli/internal/relay/watcher/watcher.go
```diff
@@ -232,126 +239,100 @@ func (w *Watcher) subscribe(ctx context.Context, events chan v2.PacketEvent) (st
...
 			w.logger.Info("Resubscribed to send packets", "clientIDs", w.clientIDs)
+
+			// trigger clearer
+			w.clearRequest()
 		}
 	}
 }
 
+func (w *Watcher) runClearer(ctx context.Context) {
+	defer func() {
```
dhfang, line 302: Same as above

cli/internal/relay/watcher/clearer.go
```diff
@@ -81,112 +79,118 @@ func NewClearer(
...
-	// query aware of the setting if anyone parks enough of them to notice.
+	// query unresolved packages one more time. This might succeed ONLY
+	// if an operator changes RPC to an archival one
 	reprobe := state.Unresolved
 	if c.abandon {
 		reprobe = nil
 	}
 
-	outstanding, probed, probeHeight, err := c.calcOutstanding(ctx, clientID, reprobe, from, latest, head.Height)
+	seqsOutstanding, seqsQueried, err := c.queryLiveSequences(ctx, seqFrom, seqLatest, reprobe)
```
dhfang, line 143: Regarding https://github.com/cosmos/ibc/pull/1461#discussion_r4063497560, but using new helper names:

The order is queryLiveSequences then buildUnrecorded. My point is that since the goal of the clearer is to identify packets the relayer is not already tracking, we can skip the on-chain queries in queryLiveSequences with a less expensive db lookup first to see if the relayer already tracks the sequence number in its packets table.

cli/internal/relay/watcher/watcher.go
```diff
@@ -135,28 +135,35 @@ func (w *Watcher) Start() error {
 	}
 
 	w.cancel = cancel
-	w.stopped = make(chan struct{})
+	w.wg.Add(2)
 
-	go w.run(ctx, eventStream)
+	go func() {
```
dhfang, line 140: Nice the refactored workers look much cleaner.

dhfang (commented):
Thanks for landing this Dmitry, I left a few more comments but none of these need to be addressed in the current PR. You can create tickets for someone to address later if you agree on the points.

One thing we identified as in scope was addressing reorgs. Right now if auto-relay ingests a new packet and then the tx gets reorged out, we don't have a mechanism currently to identify the new packet using the same sequence number that replaces it. I think we wanted to add a processor to the relay pipeline that handles this edge case by checking whether the source tx still exists and updating accordingly if it no longer does.

dhfang (approved):

cli/internal/relay/watcher/watcher.go
swift1337: Yeah, the intention of this recovery is to avoid panic in a goroutine to prevent the process from crashing if something goes wrong with the auto relayer, so we can keep running a manual relayer, for example. 

I agree that it's not obvious, and we'd better explicitly crash the process rather than keep it running semi-functional in this case.

swift1337:
> Thanks for landing this Dmitry, I left a few more comments but none of these need to be addressed in the current PR. You can create tickets for someone to address later if you agree on the points.
> 
> One thing we identified as in scope was addressing reorgs. Right now if auto-relay ingests a new packet and then the tx gets reorged out, we don't have a mechanism currently to identify the new packet using the same sequence number that replaces it. I think we wanted to add a processor to the relay pipeline that handles this edge case by checking whether the source tx still exists and updating accordingly if it no longer does.

FOU-1668

cli/internal/relay/watcher/watcher.go
swift1337: We'll address this in a follow-up (created a ticket)

swift1337: FOU-1669

cli/internal/relay/watcher/clearer.go
swift1337: FOU-1670

PR #1462, opened 2026-09-15
fix(link): attestation commitment validation

dhfang:
- Add validation in attestor aggregation that the attested data matches the expected commitment
- Add E2E test which enforces relayer does not accept corrupt attestations


cli/internal/relay/prover/attestation/generator.go
```diff
@@ -136,6 +135,41 @@ func (g *Generator) PacketProofs(
...
+			Path:       crypto.Keccak256Hash(hostv2.PacketCommitmentKey(packet.SourceClient, packet.Sequence)),
+			Commitment: [32]byte(channeltypesv2.CommitPacket(packet)),
+		}, nil
+	case attestor.CommitmentTypeReceipt:
+		return attestorevm.PacketCompact{
+			Path: crypto.Keccak256Hash(hostv2.PacketReceiptKey(packet.DestinationClient, packet.Sequence)),
+		}, nil
+	case attestor.CommitmentTypeAck:
+		path := crypto.Keccak256Hash(hostv2.PacketAcknowledgementKey(packet.DestinationClient, packet.Sequence))
+		commitment, err := g.counterpartyChain.GetCommitment(ctx, height, path)
```
gjermundgaraba: this makes acknowledgement proofs depend on the relayer RPC serving historical contract state even when a valid attestor quorum is available. The relay already has the acknowledgement bytes in `event.Acks`, so this adds a historical-state availability requirement and a serial RPC per packet just to establish the expected commitment.

cli/internal/relay/prover/attestation/quorum.go
```diff
@@ -137,53 +146,38 @@ func queryOne(
...
+	seen := make(map[common.Address]bool)
 	for _, resp := range responses {
-		if resp.err != nil {
+		if resp.err != nil || seen[resp.signer] {
 			continue
 		}
-
-		buckets[string(resp.data)] = append(buckets[string(resp.data)], resp)
+		seen[resp.signer] = true
+		result.AttestationData = resp.data
```
gjermundgaraba: every successful response already matches `expectedData`, so carrying each response's data into the reducer and repeatedly assigning it retains unnecessary claim-selection machinery. The reducer's comment also still describes grouping competing claims, obscuring its actual reliance on upstream filtering.

swift1337 (approved):

cli/internal/relay/prover/attestation/generator.go
dhfang: > even when a valid attestor quorum is available

Hmm I'm not sure I fully understand the issue raised here since the purpose of this PR is to address the possibility of misbehaving attestors returning unexpected attested data.

> The relay already has the acknowledgement bytes in event.Acks

This is true. I initially avoided optimizing this query out by passing acks through, since I didn't want to extend the Prover interface for something that I think is only needed by the attestations prover. I made the change though and I don't think it's made the interface any more difficult to undersatnd, lmk what you think.

gjermundgaraba (approved):

PR #1465, opened 2026-09-16
fix(cli): improve EVM metrics and observability documentation

gjermundgaraba:
## Summary
Split out of the Besu QBFT work. Two leftover correctness/cleanup items, unrelated to that feature: count gas on reverted txs (`evm_gas_spent` was success-only, so retries under-reported spend), and replace the hand-rolled EVM HTTP histogram with `otelhttp` so we stop maintaining a custom transport.

- Replace HTTP transport instrumentation with a metered EVM client shared by chain clients and transaction submitters.
- Report consistent success/error results and protocol error codes; treat not-found responses as successful RPC calls.
- Simplify metric helpers and batch error handling, expand regression tests, and document logging, observability, and gas accounting.

## Validation
- Passed targeted tests for EVM clients, relay packages, EVM transaction submission, OTEL, config, logging, bootstrap, and attestors.
- Passed `git diff --check`.
- Passed `GOTOOLCHAIN=go1.26.4 make lint-fix` (0 issues). Go 1.27.1 is incompatible with the transitive `cockroachdb/swiss` dependency.


dianab-cl (approved):

cli/internal/bootstrap/bootstrap.go
```diff
@@ -130,7 +130,7 @@ func BuildRelayer(cfg config.Config) (*Services, error) {
 	relayerHandler := server.NewRelayerHandler(relayerService)
 
 	// Server
-	srv := server.New(cfg.Server.ListenAddress, rpcEnableReflection, cfg.Observability.Enabled())
+	srv := server.New(cfg.Server.ListenAddress, rpcEnableReflection, cfg.Observability.Metrics)
```
swift1337, line 133: I'd keep as it was because it hides the decision logic e.g. if we add tracing then `enabled = tracing || metrics`.

cli/internal/chains/evm/metrics.go
```diff
@@ -32,92 +33,165 @@ func newInstrumentation(m metric.Meter) (*instrumentation, error) {
...
+	var (
+		httpErr rpc.HTTPError
+		rpcErr  rpc.Error
+	)
+
+	switch {
+	case err == nil:
+		return ""
+	case errors.As(err, &httpErr):
+		return "http_" + strconv.Itoa(httpErr.StatusCode)
```
swift1337, line 75: `"http_200"` looks noisy. Consider simplifying to `"200"`

cli/internal/otel/metrics.go
```diff
@@ -18,6 +18,8 @@ const (
 	AttrOp     attribute.Key = "operation"
 	AttrType   attribute.Key = "type"
 	AttrResult attribute.Key = "result"
+	// AttrCode qualifies an error result with a protocol code where one exists.
```
swift1337: Let's remove this comment, as code might mean a different thing in different metrics

```suggestion

```

cli/internal/otel/otel.go
```diff
@@ -44,7 +44,7 @@ const simpleMetricsPath = "/metrics"
 func GlobalSetup(ctx context.Context, cfg config.Observability, logger *slog.Logger) {
 	logger = logger.With("module", "otel")
 
-	if !cfg.Enabled() {
+	if !cfg.Metrics {
```
swift1337, line 47: same comment

cli/internal/txsubmitter/evm/evm.go
```diff
@@ -69,9 +69,9 @@ type ChainOptions struct {
 
 // NewFromRPC dials the chain's RPC and builds its tx submitter.
 func NewFromRPC(chainID, rpcURL string, chainSigner signer.Signer, opts ChainOptions) (*TxSubmitter, error) {
-	eth, err := ethclient.Dial(rpcURL)
```
swift1337, line 72: This isn't for the current PR, but txsubmitter creates its own client instead of using an existing one. That would affect future RPC caching. I would create a ticket for that

cli/internal/bootstrap/bootstrap.go
gjermundgaraba: if we add tracing, this will need more changes in any case. this feels simpler, and i dont think the decision logic for this needs to show up inline in this code right here, no?

cli/internal/chains/evm/metrics.go
gjermundgaraba: it's to distinguish between http_XXX and jsonrpc_XXX (also, it wouldnt be 200, since these are for errors)

cli/internal/otel/metrics.go
gjermundgaraba: good call! done.

cli/internal/txsubmitter/evm/evm.go
gjermundgaraba: created FOU-1610

swift1337 (approved):

PR #1468, opened 2026-09-16
fix(cli): reconcile deployment config without losing custom settings

gjermundgaraba:
## Why
`ibc deploy render-config --populate` merged by alias/name, so rerunning after a customize-and-redeploy cycle could overwrite aliases, signers, and operational settings, or write duplicate identities that then failed validation. Incomplete drafts also could not be reopened to fill in signers later.

## Summary
- Match deployment connections by endpoint identity and preserve aliases, remote provers, and operational settings.
- Make relay signer overrides optional and support repairing incomplete configuration drafts.
- Allocate unused connection aliases and reject ambiguous identities, including signer aliases, before rendering.
- Simplify reconciliation and validation boundaries; add regression coverage and documentation.

## Validation
- `cd cli && go test ./internal/config ./cmd/ibc` (passed after rebasing onto latest main)
- `git diff --check origin/main...HEAD`


cli/cmd/ibc/deploy.go
```diff
@@ -612,22 +595,24 @@ func renderRelayConfig(
...
 		return err
 	}
+	if err = cfg.ValidateIdentities(); err != nil {
+		return err
+	}
 	signers := []string{flagDeployRenderSignerA, flagDeployRenderSignerB}
-	for i, alias := range signers {
+	for _, alias := range signers {
 		if alias == "" {
-			return errors.Errorf("--signer-%c is required: the signers[] alias submitting relay txs on %s",
```
dhfang, line 629: Why did we decide to make this optional?

cli/internal/config/patch.go
```diff
@@ -61,62 +82,87 @@ func mergeChains(existing, incoming []ChainConfig, conflicts *[]Conflict) []Chai
...
-			continue
+		merged := out[idx]
+		changed := false
+		for _, incomingEnd := range []ClientEnd{conn.ClientA, conn.ClientB} {
+			end := &merged.ClientA
+			if end.identity() != incomingEnd.identity() {
+				end = &merged.ClientB
+			}
+			// A remote prover is an operational choice, not an on-chain client type.
+			if end.Type != ClientTypeRemote && end.Type != incomingEnd.Type {
```
dhfang, line 115: When redeploying to a new router on the same chains, the default client IDs remain unchanged. Once deployment tooling supports additional client types, those IDs could refer to a different type in the new deployment. i.e. first deployment rlies on attestations client and the redeploy relies on besu. Wouldn't this check reject that legitimate redeployment because reconciliation identifies clients only by chain ID and client ID? Or is that intended behavior

cli/internal/config/patch.go
```diff
@@ -61,62 +82,87 @@ func mergeChains(existing, incoming []ChainConfig, conflicts *[]Conflict) []Chai
...
 }
 
-func mergeConnections(existing, incoming []ConnectionConfig, conflicts *[]Conflict) []ConnectionConfig {
-	out := append([]ConnectionConfig(nil), existing...)
+func sameConnection(a, b ConnectionConfig) bool {
+	return (a.ClientA.identity() == b.ClientA.identity() && a.ClientB.identity() == b.ClientB.identity()) ||
+		(a.ClientA.identity() == b.ClientB.identity() && a.ClientB.identity() == b.ClientA.identity())
+}
 
+// mergeConnections matches connections by their client pair regardless of alias
```
dhfang, line 90: I like the matching by client pair part.

The field-level conflict reconciliation is something I intentionally tried to stay away from though. Replacing the connections block entirely I thought was simpler and easier to maintain. I think we are likely to add more fields to the connections block that should be overwritten like the signer is currently and when developers add those fields they have to know to modify this reconciliation logic.

dhfang (commented):
Thanks for taking the time to improve on this

cli/cmd/ibc/deploy.go
gjermundgaraba: the issue is that if you dont, you have to set the signers, even if they are already in the config.
if you have:
```yaml
  clientA:
    signer: alice
  clientB:
    signer: bob
```

you would have to do this:
```bash
ibc deploy render-config 1 2 --signer-a alice --signer-b bob
```

happy to consider other approaches!

cli/internal/config/patch.go
gjermundgaraba: yeah, it would be simpler, but it creates other problems where you could end up erasing other setting.

cli/internal/config/patch.go
gjermundgaraba: yes, this rejects a type change even when the router changes. deployment
tooling only supports attestation clients today, so i've kept the guard and added
a clearer error, documentation, and a regression test for router replacement. might need a re-look with the besu changes. added a TODO for that.

cli/internal/config/patch.go
dhfang: Okay well it is an improvement although I don't know how common the redeploy case is. I'm fine with just going with the field level reconciliation for now and deal with it later if it becomes a maintainability issue.

dhfang (approved):
LGTM

PR #1469, opened 2026-09-16
feat(e2e): add e2e dump

swift1337:
Adds an opt-in dump mechanism that copies attestor configs, relayer logs, diagnostics, and other workspace files before cleanup removes them - useful for debugging

`E2E_DUMP=true`


e2e/internal/e2etest/traffic_setup.go
dhfang, line 183: Seems like it may be useful to register the dump before relayer start so we can capture the dump if relayer startup fails part of the way through, but still produces useful artifacts.

dhfang (approved):
nice

PR #1481, opened 2026-09-18
feat: add Besu QBFT light client support

gjermundgaraba:
## Summary
- Add Besu QBFT header validation, consensus checks, and membership proof encoding.
- Support Besu client deployment, bootstrap configuration, and native relayer proof generation.
- Integrate generated contract bindings, metered proof RPCs, and capability-based prover wiring.
- Add fixtures, unit/e2e coverage, and configuration and CLI documentation.

closes FOU-1404

## Validation
- Relevant Besu, chain, and prover package tests passed before rebase.
- `GOTOOLCHAIN=go1.26.8 make -C cli lint-fix` passed before rebase.
- Fixture sync target checks passed; full CI pending.

## Follow-up
- FOU-1406: move fixture generation into this repository’s e2e flow.


gjermundgaraba:
@greptile review

gjermundgaraba:
FYI: this needs https://github.com/cosmos/ibc-contracts/pull/1138 and an update to the go.mod before it can be merged

cli/api/v2/prover/prover.connect.go
```diff
@@ -37,22 +37,22 @@ const (
...
 type ProverServiceClient interface {
-	// LatestProvableHeight returns the highest height a subsequent StateProof
+	// LatestProvableHeight returns the highest height a subsequent ClientUpdatePayload
 	// and PacketProofs call sharing that height can currently succeed at, with
 	// that height's counterparty-chain timestamp.
 	LatestProvableHeight(context.Context, *connect.Request[LatestProvableHeightRequest]) (*connect.Response[LatestProvableHeightResponse], error)
-	// StateProof proves the light client's counterparty state at a height.
-	StateProof(context.Context, *connect.Request[StateProofRequest]) (*connect.Response[StateProofResponse], error)
+	// ClientUpdatePayload returns the encoded light-client update payload at a height.
+	ClientUpdatePayload(context.Context, *connect.Request[ClientUpdatePayloadRequest]) (*connect.Response[ClientUpdatePayloadResponse], error)
```
srdtrk, line 55: Seems to be an unrelated naming change?

cli/besu/besutest/encoding.go
```diff
@@ -0,0 +1,20 @@
...
+var messageBindings = besumsgs.NewBindings()
+
+// EncodeClientState produces getClientState() output, for tests and fakes.
+func EncodeClientState(state besumsgs.IBesuLightClientMsgsClientState) ([]byte, error) {
+	data, err := messageBindings.TryPackClientState(state)
+	if err != nil {
+		return nil, fmt.Errorf("encode client state: %w", err)
+	}
+	return data[4:], nil
+}
```
srdtrk: What is this function doing? Why are we dropping the first 4 bytes?

cli/besu/besutest/encoding.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"fmt"
+
+	"github.com/cosmos/solidity-ibc-eureka/packages/go-abigen/besumsgs"
+)
+
+var messageBindings = besumsgs.NewBindings()
+
+// EncodeClientState produces getClientState() output, for tests and fakes.
+func EncodeClientState(state besumsgs.IBesuLightClientMsgsClientState) ([]byte, error) {
```
srdtrk, line 14: It's a shame we need these helpers. For example, if we were okay with the non-try version of the functions, we could do this inline with

```go
encoded := messageBindings.PackClientState(state)[4:]
```

cli/besu/besutest/builder.go
srdtrk, line 1: I wonder why this must all live under cli. But I haven't worked with link much so I guess this is the standard in this repo

cli/besu/besutest/builder.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/cosmos/ibc/cli/besu"
+	"github.com/cosmos/ibc/cli/besu/internal/headerlayout"
+)
+
+// Builder mutates a real Besu QBFT header and re-seals it with test keys, so
+// tests can exercise validator turnover, heights and timestamps the fixture
+// does not contain.
+type Builder struct {
+	items      []rlp.RawValue
+	extraItems []rlp.RawValue
```
srdtrk: Is there no header struct in go-ethereum which you could just decode the rlp values into rather than keeping items, and extra items

cli/besu/messages_abi.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// EncodeMembershipProof builds the proof bytes for verifyMembership and
+// verifyNonMembership: the preimage of the consensus state at the proof height,
+// the state trie nodes proving the router account and the storage trie nodes
+// eth_getProof returned for the commitment slot. accountProofNodes may be
+// empty when an earlier call in the same transaction proved the same height.
+func EncodeMembershipProof(
+	preimage besumsgs.IBesuLightClientMsgsConsensusState,
+	accountProofNodes, proofNodes [][]byte,
```
srdtrk, line 40: nit: If this is an encoding function, then it should simply take `IBesuLightClientMsgsMembershipProof` input imo

cli/besu/messages_abi.go
```diff
@@ -0,0 +1,52 @@
...
+
+var messageBindings = besumsgs.NewBindings()
+
+// EncodeUpdateClient builds the updateClient payload: the raw header, the
+// trusted height (revision 0) and the preimage of the consensus state trusted
+// at that height.
+func EncodeUpdateClient(
+	headerRLP []byte,
+	trustedHeight uint64,
+	preimage besumsgs.IBesuLightClientMsgsConsensusState,
```
srdtrk, line 19: nit: If this is an encoding function, then it should simply take `IBesuLightClientMsgsMsgUpdateClient` input imo

cli/besu/storage.go
```diff
@@ -0,0 +1,20 @@
...
+package besu
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// IBCStoreStorageSlot is the ERC-7201 namespace slot of IBCStoreUpgradeable's
+// commitment mapping, keccak256(abi.encode(uint256(keccak256("ibc.storage.IBCStore")) - 1)) & ~0xff.
+var IBCStoreStorageSlot = common.HexToHash("0x1260944489272988d9df285149b5aa1b0f48f2136d6f416159f840a3e0747600")
```
srdtrk, line 12: This is actually availible in go-abigen package. [See](https://github.com/cosmos/ibc-contracts/blob/1718df93b9a72adddfca64bb552bd5e6ae1e5866/packages/go-abigen/ics26router/store.go?plain=1#L6)

cli/besu/thresholds.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"fmt"
+
+	"github.com/cosmos/solidity-ibc-eureka/packages/go-abigen/besumsgs"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// OverlapRequired is how many commit seal signers of a new header must belong
+// to the trusted validator set: strictly more than one third of it.
+func OverlapRequired(trusted int) int { return trusted/3 + 1 }
```
srdtrk: This is actually changed upstream now. And this is the problem with this code. Why do we have all this written manually. What use does this serve in link. I'd want all this removed. We are not writing the besu light client in golang. If we did, it should go to ibc-go

cli/besu/thresholds.go
```diff
@@ -0,0 +1,64 @@
...
+
+// QuorumRequired is how many commit seal signers must belong to the new
+// header's own validator set: Besu's ceil(2n/3), written n - floor(n/3).
+func QuorumRequired(validators int) int { return validators - validators/3 }
```
srdtrk: ugly. Improved in solidity, should be here too

cli/besu/thresholds.go
```diff
@@ -0,0 +1,64 @@
...
+	count := 0
+
+	for _, signer := range signers {
+		if _, ok := members[signer]; ok {
+			count++
+		}
+	}
+
+	return count
+}
```
srdtrk: Why does link need to do this?

cli/cmd/ibc/deploy.go
```diff
@@ -278,27 +289,66 @@ func clientSpec(
...
+		CounterpartyClientID: counterpartyClientID,
+	}
+	switch flagDeployClientType {
+	case deploy.ClientTypeAttestation:
+		params, err := attestationParams(ctx, cfg, counterpartyTarget, counterpartyChainID)
+		if err != nil {
+			return deploy.ClientSpec{}, err
+		}
+		spec.Params = params
+	case deploy.ClientTypeBesuQBFT:
```
srdtrk, line 305: no IBFT? I'd prefer having this if possible

cli/besu/header.go
```diff
@@ -0,0 +1,109 @@
...
+	"slices"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+// Header is a parsed Besu QBFT block header. RLP is the exact encoding the
+// light client receives; the other fields are decoded from it.
+type Header struct {
```
srdtrk, line 9: Kind of duplicate of builder

cli/besu/thresholds.go
srdtrk, line 1: just relay bro (simulate if not sure)

cli/besu/besutest/encoding.go
srdtrk, line 1: maybe move to `clientstate.go` in internal

srdtrk (commented):
initial (limited scope) review, would like consensus light client logic to be removed

cli/internal/chains/evm/client.go
```diff
@@ -288,6 +320,213 @@ func (c *Client) GetAttestationSet(ctx context.Context, clientID string) ([]stri
 	return addresses, set.MinRequiredSigs, nil
 }
 
+// SealedHeader returns the Besu QBFT header at height, parsed from the node's
+// sealed RLP. Validators come from extraData.
+func (c *Client) SealedHeader(ctx context.Context, height uint64) (*besu.Header, error) {
```
swift1337: I think Besu-specific clients should live under `chains/besu` or `chains/evm/besu.go`

```go
type BesuClient struct {
  *Client
}
```

cli/besu/besutest/builder.go
gjermundgaraba: it made slightly more sense when it was called "link" :P
but i do agree that a restructure should be done before we release!

cli/api/v2/prover/prover.connect.go
gjermundgaraba: No, it is very much related. This naming for the prover was very attestation-geared, so refactored to a more general name

cli/besu/besutest/encoding.go
gjermundgaraba: since it's test-only, we are! removed it.

cli/besu/messages_abi.go
gjermundgaraba: fixed

cli/cmd/ibc/deploy.go
gjermundgaraba: Made a follow-up issue for IBFT that i'll take after this: FOU-1664

cli/besu/header.go
gjermundgaraba: yeah, trimmed it a bit

cli/besu/storage.go
gjermundgaraba: oh, true, fixed

cli/besu/thresholds.go
gjermundgaraba: fixed

cli/besu/header.go
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package besu
+
+import "github.com/ethereum/go-ethereum/common"
+
+// Header contains the fields needed to construct light-client payloads.
+// RLP is preserved exactly; consensus validity is checked by the contract.
+type Header struct {
```
srdtrk, line 9: This is not the full header, (the actual header would have way more fields, so, maybe a better name is needed

cli/besu/storage.go
srdtrk, line 1: super nit: better name for this file could be `path.go`. Storage can be somewhat ambiguous

cli/cmd/ibc/deploy.go
```diff
@@ -278,27 +289,66 @@ func clientSpec(
...
 		counterpartyClientID = defaultClientID(chainID, counterpartyChainID)
 	}
+	spec := deploy.ClientSpec{
+		ClientID:             clientID,
+		Type:                 flagDeployClientType,
+		CounterpartyChainID:  counterpartyChainID,
+		CounterpartyClientID: counterpartyClientID,
+	}
+	switch flagDeployClientType {
+	case deploy.ClientTypeAttestation:
```
srdtrk: I guess I understand the architecture. For each chain we support, we add all the supported client types. Not sure if I like this because this means that we will duplicate this attestation specs stuff many times right?

Anyway, this is not related to this PR

cli/cmd/ibc/deploy.go
```diff
@@ -309,27 +359,122 @@ func clientSpec(
...
+// values that truncation would silently change.
+func wholeSeconds(d time.Duration, flag string) (uint64, error) {
+	if d < 0 {
+		return 0, errors.Errorf("--%s must not be negative, got %s", flag, d)
+	}
+	if d%time.Second != 0 {
+		return 0, errors.Errorf("--%s must be whole seconds, got %s", flag, d)
+	}
+	return uint64(d / time.Second), nil
 }
```
srdtrk, line 478: There is gotta be a better way to convert durations to seconds using std library functions...
Also, flag is passed here only to be returned in the errors... If this is needed, flag can be used to wrap the error in one layer above

cli/cmd/ibc/deploy.go
```diff
@@ -278,27 +289,66 @@ func clientSpec(
 	if counterpartyClientID == "" {
 		counterpartyClientID = defaultClientID(chainID, counterpartyChainID)
 	}
+	spec := deploy.ClientSpec{
+		ClientID:             clientID,
+		Type:                 flagDeployClientType,
+		CounterpartyChainID:  counterpartyChainID,
+		CounterpartyClientID: counterpartyClientID,
+	}
+	switch flagDeployClientType {
```
srdtrk, line 298: this is ugly. Ideally, the top level `cli/cmd/ibc/deploy.go` shouldn't care about implementation details and it should interact with chain/client specific packages through interfaces.

Look at how much code is added to this file for supporting a single light client type

cli/cmd/ibc/deploy.go
```diff
@@ -309,27 +359,122 @@ func clientSpec(
...
+			"counterparty chain %s needs a valid nonzero evm.ics26Router in config", counterpartyChainID,
+		)
+	}
+	height := flagDeployHeight
+	if height == 0 {
+		head, _, headErr := counterpartyTarget.Head(ctx)
+		if headErr != nil {
+			return deploy.BesuQBFTParams{}, errors.Wrap(headErr, "fetch counterparty head for initial trusted state")
+		}
+		height = max(head, 1)
```
srdtrk, line 430: why are we doing this, if height is still zero, there is clearly something wrong and we should just return an error

cli/cmd/ibc/deploy.go
```diff
@@ -309,27 +359,122 @@ func clientSpec(
...
+		if flags.Changed(flagNameMaxClockDrift) {
+			params.MaxClockDrift = maxClockDrift
+		}
+		return params, nil
+	}
+	if !flags.Changed(flagNameTrustingPeriod) {
+		return deploy.BesuQBFTParams{}, errors.New(
+			"--trusting-period is required for a new besu-qbft client: choose a positive duration based on validator governance",
 		)
 	}
```
srdtrk, line 412: unlike what is happening in `ClientTypeAttestation` case, you are not checking `flags.Changed(flagNameTrustingPeriod) && trustingPeriod == 0`

cli/cmd/ibc/deploy.go
```diff
@@ -309,27 +359,122 @@ func clientSpec(
...
-	return spec, nil
+	source, ok := counterpartyTarget.(deploy.BesuQBFTSource)
+	if !ok {
+		return deploy.BesuQBFTParams{}, errors.Errorf(
+			"counterparty chain %s cannot serve a besu-qbft trusted state", counterpartyChainID,
+		)
+	}
+	if !common.IsHexAddress(counterpartyRouter) || common.HexToAddress(counterpartyRouter) == (common.Address{}) {
```
srdtrk, line 419: I believe there is an IsZero function rather than having to check equality.

cli/cmd/ibc/deploy.go
```diff
@@ -473,6 +625,9 @@ func deployShow(_ *cobra.Command, args []string) error {
 // attestorsFromClient projects one client's on-chain attestor addresses into
 // attestors
 func attestorsFromClient(cfg config.Config, c manifest.Client, watchedChainID string) config.Attestors {
+	if c.Type != deploy.ClientTypeAttestation {
+		return nil
+	}
```
srdtrk, line 630: it is weird that this function can get called when type is not attestation. And that if this happens, this returns nil rather than some error or panic

cli/cmd/ibc/main.go
```diff
@@ -152,19 +152,27 @@ func init() {
 	cmdDeployClient.Flags().
 		StringVar(&flagDeployCounterparty, "counterparty-chain", "", "counterparty chain id the client tracks")
 	_ = cmdDeployClient.MarkFlagRequired("counterparty-chain")
-	cmdDeployClient.Flags().StringVar(&flagDeployClientType, "type", deploy.ClientTypeAttestation, "light client type")
+	cmdDeployClient.Flags().
+		StringVar(&flagDeployClientType, "type", deploy.ClientTypeAttestation, "light client type: attestation or besu-qbft")
```
srdtrk, line 156: hard coded strings: `attestation or besu-qbft`. Use imported constants. Ditto elsewhere in this file

cli/internal/chains/evm/besu.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	sealed, err := besu.ParseSealedHeader(header)
+	if err != nil {
+		return nil, errors.Wrapf(err, "header %d on chain %s is not a Besu QBFT header", height, c.chainID)
+	}
+
+	if height != v2.LatestBlock && sealed.Height != height {
+		return nil, errors.Errorf("chain %s returned header %d for height %d", c.chainID, sealed.Height, height)
+	}
```
srdtrk, line 53: I'm not sure what this check is doing?

cli/internal/chains/evm/besu.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/cosmos/ibc/cli/besu"
+	v2 "github.com/cosmos/ibc/cli/internal/types/v2"
+)
+
+// AccountProof is an eth_getProof result for one account at one height: the
+// account proof nodes against the block's state root and one storage proof
+// per requested slot, in request order.
+type AccountProof struct {
```
srdtrk, line 25: Calling this account proof is a bit misleading. It is true that the first field is the account proof, but the second field is not. This contains more than the proof of the account, so I think it would be good to rename it

cli/internal/chains/evm/besu.go
```diff
@@ -0,0 +1,243 @@
...
+
+	return proof, nil
+}
+
+// accountProofFromResult converts an eth_getProof result, requiring a
+// well-formed storage proof for every requested slot regardless of response
+// order.
+func accountProofFromResult(result *gethclient.AccountResult, slots [][32]byte) (AccountProof, error) {
+	byKey := make(map[[32]byte]StorageProof, len(result.StorageProof))
+
```
srdtrk, line 84: I don't like that we are de-duping using this map. Can we just de-dup the slots before calling `get_proof` so that we don't need to do this map. Then we would just validate that the proof response is returned in the same order (no maps)

cli/internal/chains/evm/besu.go
```diff
@@ -0,0 +1,243 @@
...
+			)
+		}
+
+		return [32]byte{}, errors.Wrapf(
+			err, "querying consensus state hash at height %d for client %q on chain %s", height, clientID, c.chainID,
+		)
+	}
+
+	return hash, nil
+}
```
srdtrk, line 201: I actually don't think you ever need to use this function. You don't need to check whether or not the trusted consensus state you produce will match this hash, you can see that by simulating the transaction anyway. You can consider removing this to reduce the maintenance surface

cli/internal/chains/evm/besu.go
```diff
@@ -0,0 +1,243 @@
...
+		return false
+	}
+
+	decoded, decodeErr := besuErrorBindings.UnpackError(revert)
+	if decodeErr != nil {
+		return false
+	}
+	_, ok = decoded.(*besuerrors.BindingsConsensusStateNotFound)
+	return ok
+}
```
srdtrk, line 243: I guess if you did what I said, you could also remove this function too. Also, it is weird to me that you have function returning a bool for checking an error...

cli/internal/chains/evm/client.go
```diff
@@ -61,14 +68,18 @@ type Client struct {
 	logger        *slog.Logger
 }
 
+// ErrConsensusStateNotFound reports that a light client stores nothing at the
+// requested height.
+var ErrConsensusStateNotFound = errors.New("consensus state not found")
+
```
srdtrk, line 74: This error should not be here!

cli/internal/chains/evm/rpc.go
srdtrk, line 1: It seems this is not a necessary change to support besu. Why is this included in this PR? If this is an improvement we want, we can judge it on it's own merits in another PR.

If the comments are true, you can just do this wrapping elsewhere in runtime, or fully replace ethclient with gethclient in `cli/internal/chains/evm/client.go`. Since it seems gethclient can do everything ethclient can do

cli/internal/deploy/evm/driver.go
```diff
@@ -211,6 +227,108 @@ func (d *Driver) ProvisionClient(ctx context.Context, router string, spec deploy
...
+		d.backend,
+		args.router,
+		params.InitialHeight,
+		params.InitialTimestamp,
+		args.stateRoot,
+		args.validators,
+		params.TrustingPeriod,
+		params.MaxClockDrift,
+		common.HexToAddress(router),
+	)
```
srdtrk, line 260: I kinda wanna change the constructor args, we should just deploy with client state and consensus state. WDYT?

cli/internal/deploy/evm/driver.go
```diff
@@ -211,6 +227,108 @@ func (d *Driver) ProvisionClient(ctx context.Context, router string, spec deploy
...
+type besuQBFTConstructorArgs struct {
+	router     common.Address
+	stateRoot  [32]byte
+	validators []common.Address
+}
+
+// besuQBFTArgs validates besu-qbft params and converts them for the contract
+// constructor.
+func besuQBFTArgs(p deploy.BesuQBFTParams) (besuQBFTConstructorArgs, error) {
+	if !common.IsHexAddress(p.IBCRouter) {
```
srdtrk, line 279: check if it is zero. `IsZero()`?

cli/internal/deploy/evm/driver.go
```diff
@@ -211,6 +227,108 @@ func (d *Driver) ProvisionClient(ctx context.Context, router string, spec deploy
...
+	if p.InitialHeight == 0 || p.InitialTimestamp == 0 {
+		return besuQBFTConstructorArgs{}, fmt.Errorf("initial height and timestamp required")
+	}
+	root, err := hexutil.Decode(p.InitialStateRoot)
+	if err != nil || len(root) != common.HashLength {
+		return besuQBFTConstructorArgs{}, fmt.Errorf(
+			"initial state root %q must be 32 hex bytes",
+			p.InitialStateRoot,
+		)
+	}
```
srdtrk, line 291: check that that it is non-zero (`bytes32(0)`)

cli/internal/deploy/evm/driver.go
```diff
@@ -211,6 +227,108 @@ func (d *Driver) ProvisionClient(ctx context.Context, router string, spec deploy
...
+	validators := make([]common.Address, len(p.InitialValidators))
```
srdtrk, line 292: make sure the list is non-zero

cli/internal/deploy/evm/driver.go
```diff
@@ -211,6 +227,108 @@ func (d *Driver) ProvisionClient(ctx context.Context, router string, spec deploy
...
+	for i, v := range p.InitialValidators {
+		if !common.IsHexAddress(v) {
```
srdtrk, line 294: check non-zero address. I think our code could use re-usable helpers for address validation

cli/internal/deploy/evm/driver.go
```diff
@@ -211,6 +227,108 @@ func (d *Driver) ProvisionClient(ctx context.Context, router string, spec deploy
...
+	}
+	parsed, err := besu.ParseSealedHeader(header)
+	if err != nil {
+		return deploy.BesuQBFTTrustedState{}, fmt.Errorf("header %d is not a Besu QBFT header: %w", height, err)
+	}
+	validators := make([]string, len(parsed.Validators))
+	for i, v := range parsed.Validators {
+		validators[i] = v.Hex()
+	}
+	return deploy.BesuQBFTTrustedState{
```
srdtrk, line 324: This seems to be. a relayer construct. It'd better to use a trusted consensus state (At a given height). So that we need not maintain these structs

cli/internal/deploy/deploy.go
```diff
@@ -31,8 +34,37 @@ type AttestationParams struct {
...
+	StateRoot  string
+	Validators []string
+}
+
+// BesuQBFTSource is implemented by targets whose chain runs Besu QBFT and can
+// serve the trusted state a client tracking it is initialized with.
+type BesuQBFTSource interface {
+	BesuQBFTTrustedState(ctx context.Context, height uint64) (BesuQBFTTrustedState, error)
+}
+
```
srdtrk, line 64: ditto: I don't like this construct since it is not needed when we can use Consensus State

cli/internal/deploy/deploy.go
```diff
@@ -31,8 +34,37 @@ type AttestationParams struct {
...
+	MaxClockDrift     uint64   `json:"maxClockDrift"`
+}
+
+// BesuQBFTTrustedState is the counterparty state a Besu QBFT client starts
+// trusting, read from the counterparty chain at Height.
+type BesuQBFTTrustedState struct {
+	Height     uint64
+	Timestamp  uint64
+	StateRoot  string
+	Validators []string
```
srdtrk, line 56: at the very least, these 3 items can. be replaced by the auto-generated consensus state

cli/internal/deploy/steps.go
```diff
@@ -421,26 +421,67 @@ func specToClient(spec ClientSpec, address string) (manifest.Client, error) {
...
 			"threshold":        p.Threshold,
 			"initialHeight":    p.InitialHeight,
 			"initialTimestamp": p.InitialTimestamp,
 		}
+	case ClientTypeBesuQBFT:
+		p, ok := spec.Params.(BesuQBFTParams)
+		if !ok {
+			return manifest.Client{}, paramsMismatch(spec)
+		}
+		client.Params = map[string]any{
```
srdtrk, line 441: Not related to this PR. Why do we pass info like this??

cli/internal/livevalidate/quorum.go
```diff
@@ -15,11 +15,12 @@ import (
...
-// and confirms every attestation-type client end of every configured
-// connection can currently satisfy its attestor quorum against on-chain
-// state.
-func checkAttestorQuorum(ctx context.Context, cfg config.Config, clientSet *chains.ClientSet) error {
+// checkProvers resolves every configured attestor (local and remote) and
+// builds a prover for every client end of every configured connection, which
+// confirms against on-chain state that attestation ends can satisfy their
+// attestor quorum and that besu-qbft ends track the configured counterparty
+// router with a trusted consensus state the relayer can rebuild.
+func checkProvers(ctx context.Context, cfg config.Config, clientSet *chains.ClientSet) error {
```
srdtrk, line 23: why is this function renamed if we aren't doing besu consensus stuff anymore

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func (g *Generator) clientState(ctx context.Context) (besumsgs.IBesuLightClientMsgsClientState, error) {
+	state, err := g.host.GetBesuQBFTClientState(ctx, g.clientID)
+	if err != nil {
+		return state, err
+	}
+	if state.TrustingPeriod == 0 {
+		return state, errors.New("besu qbft trusting period must be nonzero")
+	}
```
srdtrk, line 107: this is already deployed client state right? So why do we do these checks

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+	return state, nil
+}
+
+func unixSeconds(timestamp time.Time) (uint64, error) {
+	seconds := timestamp.Unix()
+	if seconds < 0 {
+		return 0, errors.New("negative chain timestamp")
+	}
+	return uint64(seconds), nil
+}
```
srdtrk, line 117: Should be a generic helper in the repo if we are doing this.

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+
+func exceedsClockDrift(target, host, drift uint64) bool {
+	return target > host && target-host > drift
+}
```
srdtrk, line 121: There is no way we need this

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+			g.clientID, state.IbcRouter, g.counterparty.ChainID(), counterpartyRouter,
+		)
+	}
+
+	if _, err := g.preimage(ctx, state.LatestHeight.RevisionHeight); err != nil {
+		return fmt.Errorf("client %q: verifying trusted consensus state: %w", g.clientID, err)
+	}
+
+	return nil
+}
```
srdtrk, line 145: what is this function doing?

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+			}
+			if exceedsClockDrift(seconds, hostSeconds, state.MaxClockDrift) {
+				high = mid - 1
+			} else {
+				low = mid
+				timestamp = header.Timestamp
+			}
+		}
+		height = low
+	}
```
srdtrk, line 204: Wow... This is never going to happen. And if it does, it is okay. The relayer need not concern itself with this. You should pick maxClockDrift so that this cannot happen

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+
+	if _, err := g.preimage(ctx, state.LatestHeight.RevisionHeight); err != nil {
+		return fmt.Errorf("client %q: verifying trusted consensus state: %w", g.clientID, err)
+	}
+
+	return nil
+}
+
+// LatestProvableHeight selects the newest clock-admissible height. Consensus
+// validity is left to contract simulation; intermediate updates are not supported.
```
srdtrk, line 148: In besu, this is always the latest height. The only time this is not the case is if the validator set has changed more than trusting level (2/3) in which case you need to find an earlier height. But this function is concerned with every scenario EXCEPT that one. I'd say adjust this function to return the latest height, and it's up to you to implement the validator scenario.

Checking `checkTrustingPeriod` kinda makes sense, so you could keep that

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+			}
+			if hash != common.Hash(stored) {
+				return nil, fmt.Errorf("%w: height %d stores %s, counterparty state hashes to %s",
+					ErrConflictingConsensusState, target, common.Hash(stored), hash)
+			}
+			return nil, nil
+		case !errors.Is(storedErr, evm.ErrConsensusStateNotFound):
+			return nil, storedErr
+		}
+	}
```
srdtrk, line 268: This doesn't make sense to me. Just because `target <= state.LatestHeight.RevisionHeight`, doesn't mean that target should be stored in the client.

If `target <= state.LatestHeight.RevisionHeight`, we probably shouldn't even be calling this function with that target height...

It seems to me that our relayer likely cannot handle multiple relayers relaying at the same time, which is sad

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+	// Equivalent to Solidity's widened timestamp + period > block.timestamp.
+	if host >= timestamp && host-timestamp >= period {
+		return fmt.Errorf("%w: timestamp %d, trusting period %ds", ErrClientExpired, timestamp, period)
+	}
+	return nil
+}
+
+// ClientUpdatePayload returns an encoded updateMsg from the client's trusted state to target,
+// or nil when the client already stores an unexpired target. Intermediate updates are not supported.
+func (g *Generator) ClientUpdatePayload(ctx context.Context, target uint64) ([]byte, error) {
```
srdtrk, line 228: It's interesting choice to not have a `trustedHeight` parameter in this function, which is what leads to all the funny checks you need to perform below. I'd say it's an architectural issue which might be out of scope for this PR

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+) error {
+	if trustErr := checkTrustingPeriod(state.TrustingPeriod, target.Timestamp, hostSeconds); trustErr != nil {
+		return fmt.Errorf("proof height %d: %w", target.Height, trustErr)
+	}
+	if exceedsClockDrift(target.Timestamp, hostSeconds, state.MaxClockDrift) {
+		return fmt.Errorf("target height %d exceeds host clock drift", target.Height)
+	}
+
+	return nil
+}
```
srdtrk, line 304: I say remove all these time checks

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+		return besumsgs.IBesuLightClientMsgsConsensusState{}, err
+	}
+
+	if hash != common.Hash(stored) {
+		return besumsgs.IBesuLightClientMsgsConsensusState{}, fmt.Errorf(
+			"consensus state rebuilt for height %d hashes to %s but the light client stores %s",
+			height, hash, common.Hash(stored),
+		)
+	}
+
```
srdtrk, line 472: I don't think you need to do this but I guess it doesn't cost much, so I'm okay with it

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	proof, err := g.counterparty.GetRouterProof(ctx, height, slots)
+	if err != nil {
+		return nil, fmt.Errorf("proving router at height %d: %w (%s)", height, err, historyHint)
+	}
+
+	return &snapshot{header: header, proof: proof, consensus: consensusOf(header)}, nil
```
srdtrk, line 440: I think this snapshot stuff is not needed. I thought I removed it from the proof API as well. The reason why snapshot was useful was because account proofs and update client were coupled. But it is no longer useful once we decoupled them

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+	case v2.ProofKindPacketCommitment:
+		return hostv2.PacketCommitmentKey(packet.SourceClient, packet.Sequence), nil
+	case v2.ProofKindAcknowledgement:
+		return hostv2.PacketAcknowledgementKey(packet.DestinationClient, packet.Sequence), nil
+	case v2.ProofKindReceiptAbsence:
+		return hostv2.PacketReceiptKey(packet.DestinationClient, packet.Sequence), nil
+	default:
+		return nil, fmt.Errorf("unsupported proof kind %v", kind)
+	}
+}
```
srdtrk, line 388: not besu/qbft specific (bad location)

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+	switch kind {
+	case v2.ProofKindPacketCommitment:
+		expected := common.BytesToHash(channeltypesv2.CommitPacket(packet))
+		if value != expected {
+			return fmt.Errorf("%w: stored %s, packet commits to %s", ErrCommitmentMismatch, value, expected)
+		}
+	case v2.ProofKindAcknowledgement:
+		if value == (common.Hash{}) {
+			return ErrAckMissing
+		}
```
srdtrk, line 401: You can add a proper equality check here rather than enforcing non-zero if you pass the acks, similar to (ProofKindPacketCommitment)

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+		return hostv2.PacketAcknowledgementKey(packet.DestinationClient, packet.Sequence), nil
+	case v2.ProofKindReceiptAbsence:
+		return hostv2.PacketReceiptKey(packet.DestinationClient, packet.Sequence), nil
+	default:
+		return nil, fmt.Errorf("unsupported proof kind %v", kind)
+	}
+}
+
+// checkValue verifies the slot value matches the claim before it is proven.
+func checkValue(kind v2.ProofKind, packet channeltypesv2.Packet, value common.Hash) error {
```
srdtrk, line 391: Having one function to check all three cases feels weird for reasons mentioned

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+			return fmt.Errorf("%w: stored %s, packet commits to %s", ErrCommitmentMismatch, value, expected)
+		}
+	case v2.ProofKindAcknowledgement:
+		if value == (common.Hash{}) {
+			return ErrAckMissing
+		}
+	case v2.ProofKindReceiptAbsence:
+		if value != (common.Hash{}) {
+			return ErrReceiptExists
+		}
```
srdtrk, line 405: it is weird that we are checking a hash when we shouldn't even have it (so function signature issues)

cli/internal/relay/prover/besuqbft/generator.go
srdtrk, line 391: ditto: not besu/qbft specific (bad location)

cli/internal/relay/prover/besuqbft/generator.go
```diff
@@ -0,0 +1,474 @@
...
+func (g *Generator) header(ctx context.Context, height uint64) (*besu.Header, error) {
+	header, err := g.counterparty.SealedHeader(ctx, height)
+	if err != nil {
+		return nil, fmt.Errorf("reading counterparty header %d: %w", height, err)
+	}
+
+	if header.Height != height {
+		return nil, fmt.Errorf("counterparty returned header %d for height %d", header.Height, height)
+	}
+
```
srdtrk, line 423: This validation should be done in `g.counterparty.SealedHeader(ctx, height)`

srdtrk (changes requested):
did my second pass

cli/internal/relay/txbuilder/evm/evm.go
```diff
@@ -53,27 +53,27 @@ func New(router common.Address) *TxBuilder {
...
+		}
 
-	calls = append(calls, updateCall)
+		calls = append(calls, updateCall)
+	}
 
 	for _, item := range packetRelayItems {
-		var call []byte
-
-		call, err = packRelayItem(item)
```
srdtrk, line 76: Not related to this PR fully, but seems like packet aggregation support is problematic? Which we have in besu light client

cli/internal/relay/prover/besuqbft/generator.go
gjermundgaraba: Done for the first part: `LatestProvableHeight` now returns the counterparty head and only keeps the trusting-period check. There's no clock-drift search any more; the contract rejects a header that is too far ahead at simulation.

Validator turnover beyond the trust level is tracked in [FOU-1612](https://linear.app/cosmoslabs/issue/FOU-1612): catching up through intermediate headers, anchoring on an older stored consensus state, and a clear error when a transition can't be crossed. Until then, an update that fails the overlap check fails simulation, and the docs say catch-up isn't supported.

cli/internal/relay/prover/besuqbft/generator.go
gjermundgaraba: The checks this caused are gone now: `ClientUpdatePayload` anchors on the client's latest height, returns nothing when the target is that height, and leaves the rest to simulation.

Agreed that the parameter itself is an architectural question. I've written it up as a consideration in [FOU-1699](https://linear.app/cosmoslabs/issue/FOU-1699), to decide together with validator catch-up ([FOU-1612](https://linear.app/cosmoslabs/issue/FOU-1612)), which is where the caller would need to pick trusted heights.

cli/internal/deploy/steps.go
gjermundgaraba: Agreed. This predates the PR; the Besu case just follows the existing attestation pattern. Tracked in [FOU-1700](https://linear.app/cosmoslabs/issue/FOU-1700) to store typed params in the manifest.

